PR #26, opened 2018-10-16
Parameterize cache on whether or not is has a cache

Keno:
Otherwise, in order to get rid of all the ObjectId related code,
SROA would have to see the allocation of the context and be able
to eliminate it. It's much easier on the compiler if we don't
generate it in the first place when we don't need it.

### DIFF
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -1,9 +1,10 @@
-mutable struct Context
-  cache::Union{IdDict{Any,Any},Nothing}
+mutable struct Context{T<:Union{IdDict{Any,Any},Nothing}}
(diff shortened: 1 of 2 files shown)


MikeInnes:
I'm fine with this in principle, but as I recall from trying something similar, making `Context` a parameterised type causes issues in the parts where we abuse type inference (and indeed the `usetyped` tests are now failing, though possibly for a separate reason). I think it's simple to fix that by feeding the actual context type through inference though.

Keno:
Shall we revisit this? I'd like to cut down the diff between master and the TPU branch.

CarloLucibello:
closing as outdated

Issue #29, opened 2018-10-25
Automatic differenciation for complex numbers


GiggleLiu:
I find `Zygote` supports complex number AD, I like it more now! e.g.
```julia console
julia> gradient(imag, x)
((re = nothing, im = 1),)
```
But I have several questions,
## Question 1:
Why it returns a `NamedTuple` rather than a complex number.

## Question 2:
Why the following expression complaints not implemented error
```
gradient(x-> log(x) |> abs, x)
```
`log` is holomophic and complex gradient for `abs` is already defined in Zygote, there is no reason to break here. I suggest to create some traits for functions (is it possible?!) to help program know `log` is holomophic.

## Question 3:
About convention, is the way of `Zygote` defining gradients for complex numbers proper?
We have two conventions here
* zygote: gradient
* partial derivative: https://giggleliu.github.io/2018/02/01/complex_bp.html

I think they are conjugate to each other.

I have implementation of the latter one here
https://github.com/GiggleLiu/poorman_nn/blob/44b54d4a86f689f528a301e3a4db6b05210bb16a/poornn/functions.py#L1034

I choose the latter convention because we want `gradient(log, x) = log'(x) = 1/x` rather than `gradient(log, x) = log'(x)' = 1/x^*`, it is more convenient for holomophic functions, right?

MikeInnes:
So this is actually working entirely by accident; from Zygote's perspective a complex number is just a pair of reals `(re, im)`, and the gradient value reflects that.

I'd be happy to help you start to add real support for this though. The first step would be to add gradient definitions for `real` and `imag` that return complex numbers rather than structs like this. Then we just need gradient definitions for things like `log`. Many definitions – like `abs2` – might work automatically if they are fairly simple.

MikeInnes:
I just added those definitions in https://github.com/FluxML/Zygote.jl/commit/f2a85600981b672d150f01dffa8b0163052c6353, so a couple of simple things work now.

```julia
julia> gradient(x -> abs2(log(x)), 1+2im)
(0.21478062759933578 - 0.29549955871733735im,)
```

Let me know if I can help with any more support for this stuff.

GiggleLiu:
Zygote is like a magic! I sincerely hope it can be stablized and documented as soon as possible, so that we can try cool stuffs on it.

Since I am not familiar with this fresh new framework. Could you please explain a bit why it is able to treat complex as a tuple by accident? Is it decoding the underlying structure of an instance?

[...]

PR #35, opened 2018-11-05
Fix complex gradients

antoine-levitt:
Ref https://github.com/FluxML/Zygote.jl/issues/29

### DIFF
--- a/src/lib/complex.jl
+++ b/src/lib/complex.jl
@@ -1,4 +1,10 @@
 @grad real(x::Complex) = real(x), r̄ -> (r̄ + zero(r̄)*im,)
 @grad imag(x::Complex) = imag(x), ī -> (zero(ī) + ī*im,)
 
-@grad log(x::Complex) = log(x), ȳ -> (ȳ/x,)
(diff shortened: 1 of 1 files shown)


MikeInnes:
Thanks!

Do you think we could generally re-use the rules in [DiffRules](https://github.com/JuliaDiff/DiffRules.jl) for this? (see e.g. the real.jl file for example usage).

antoine-levitt:
I think it's in general true that any function defined in standard libs that is C -> C is holomorphic, and therefore can be differentiated using the same rules as in this PR. The single exception I can see is `conj`.

Then there are the C->R functions (abs, abs2, hypot, real, imag) that have to be special-cased, and the comparison-based (max, mod, etc.) that aren't defined for complex arguments.

If there's a reliable way of predicting whether a function is C->C, then that could be used. Otherwise it'll have to be manual.

MikeInnes:
I suppose that at the very least, we could just make a list of valid functions and look them up in DiffRules, rather than writing out the gradients again. Then at least adding new definitions would be very easy. Not urgent but just a thought for now.

PR #45, opened 2018-12-17
WIP: Replace Zygote's Cassette-style transform with actual Cassette

jrevels:
Requires https://github.com/jrevels/Cassette.jl/pull/97, which can be updated alongside this PR if need be.

This PR is literally in its infancy right now, but the plan is roughly:

1. Get Zygote's AD pass running as a Cassette pass. We're halfway there; see `zygote_pass` in `emit_new.jl`. Right now, that function properly accepts Cassette-provided pass input and (as far as we have checked) produces the correct output w.r.t. Zygote's version of overdub. The next step is to tweak Zygote's pass machinery to produce correct output w.r.t. Cassette's overdub (see TODO comments there).

2. Refactor Zygote to utilize e.g. an actual Cassette context instead of Zygote's Cassette-style context, the Cassette-based AD pass from 1, etc.

3. Delete all of the Zygote overdub machinery that should now be handled by Cassette. This step may or may not be totally completed as part of this PR; we'll have to see. I'm not a good judge yet of where Zygote's actual AD parts are intertwined with it's implementation of Cassette-style overdub - if they are easy to decouple we can just do it here, if it's a bigger project it'd probably be better to do as a follow-up.

cc @Keno

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -15,6 +15,14 @@ git-tree-sha1 = "055eb2690182ebc31087859c3dd8598371d3ef9e"
 uuid = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
 version = "0.5.3"
 
+[[Cassette]]
+deps = ["InteractiveUtils", "LinearAlgebra", "Test"]
+git-tree-sha1 = "93bd4a23e2f3de79672279abda7ea1404b988649"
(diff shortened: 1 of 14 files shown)


CarloLucibello:
closing as outdated

PR #58, opened 2019-01-16
Update `Manifest.toml` to fix missing `IRTools`

staticfloat:
Note that the Julia 1.0.X tests will continue to fail with this, because `IRTools v0.1.1` is not compatible with Julia 1.0.X, and backing it down to `v0.1.0` causes the `Zygote` test suite to segfault Julia 1.0.3.  Our best option is likely to wait until Julia 1.1 is released, then use that for our testing.


staticfloat:
Using `IRTools#master` gets the tests actually running, so this seems to be working.

PR #56, opened 2019-01-15
Distances rules

willtebbutt:
Some rules for euclidean distances. @MikeInnes if this stuff looks okay to you I'll tack on some `ToeplitzMatrices.jl` stuff before we merge.

Also fixes some indentation that I messed up in a previous PR.

PR #60, opened 2019-01-24
Euclidean distance rules

willtebbutt:
I screwed up and deleted the fork that #56 was associated with as a I forgot that I had this PR open. This is a fresh fork with the same changed :)


willtebbutt:
@MikeInnes I'm not sure what could be causing (what appears to be) the infinite recursion that causing this branch to fail. Any chance that you could take a look?

MikeInnes:
I ended up just throwing in a `Zygote.refresh()`. Typed mode exacerbates 265-y issues so it probably just couldn't see one of the gradient definitions. Thanks for the patch!

PR #63, opened 2019-01-25
vcat and hcat for matrices and vectors

willtebbutt:
What is says on the tin. Also sneakily removes `rand` from the `@nograd` list.


MikeInnes:
This is great. Just for reference: In the future would it make sense to support `cat(; dims = ...)` and forward `hcat` and `vcat` to them?

Issue #49, opened 2018-12-20
`@nograd rand` a bit overzealous


willtebbutt:
`rand` is currently flagged as a `@nograd`. While this is probably appropriate for the methods of `rand` one finds in `Base`, it's certainly not for things in `Distributions`. For example:
```julia
Zygote.gradient(x->rand(Normal(x, 1)), x)
```
Here the gradient w.r.t. `x` definitely shouldn't be zero in the general case.

I would suggest that `@nograd` should be extended slightly to allow it to be applied only to specific methods of function.

MikeInnes:
`@nograd` is more-or-less a hack that I added back when everything was segfaulting, so simple examples would work. As you point out, it's basically never useful in its current form since it's almost always overly general.

Feel free to delete the `@nograd` usages entirely where they don't break the tests, or otherwise to just add more sensible definitions.

PR #64, opened 2019-01-27
one, ones, zero, zeros

willtebbutt:
Zygote was having some trouble with the above functions before. I propose to `@nograd` them as I can't imagine a scenario in which gradient information needs to be propagated through them.


MikeInnes:
Sounds good to me. We should eventually give them more specific definitions as in #49, but this is fine for now.

PR #70, opened 2019-02-11
Fix up multiplication and scalar cholesky

willtebbutt:
- Multiplication: there was an issue whereby multiplication returns stuff of the wrong size. This PR simply reshapes stuff to be the appropriate size. e.g. if you don't reshape, you might get an `N`-vector rather than an `N x 1`-matrix. This can cause issues if you're expecting a matrix rather than a vector.
- `cholesky(::Real)`: pretty self-explanatory.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -108,8 +108,11 @@ end
 
 # LinAlg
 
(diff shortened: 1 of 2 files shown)


MikeInnes:
Perfect, thanks a lot!

PR #74, opened 2019-02-23
Correctly set slotflags for generated CodeInfo

Keno:
Setting it wrong causes memory corruption and crashes


MikeInnes:
:+1: seems like a smart move to me. I guess this may as well be part of `argnames!`?

Issue #37, opened 2018-11-12
Zygote DiffEq Integration (Pun Intended)


ChrisRackauckas:
MWE:

```julia
using OrdinaryDiffEq, ParameterizedFunctions, ForwardDiff

f = @ode_def begin
  dx = a*x - b*x*y
  dy = -c*y + x*y
end a b c

p = [1.5,1.0,3.0]
prob = ODEProblem(f,[1.0;1.0],(0.0,10.0),p)
t = 0.0:0.5:10.0

function G(p)
  tmp_prob = remake(prob,u0=convert.(eltype(p),prob.u0),p=p)
  sol = solve(tmp_prob,Vern9(),abstol=1e-14,reltol=1e-14,saveat=t)
  A = convert(Array,sol)
  sum(((1 .- A).^2)./2)
end
G([1.5,1.0,3.0])
res2 = ForwardDiff.gradient(G,[1.5,1.0,3.0])

using Zygote
Zygote.derivative(G,[1.5,1.0,3.0])
```

Krastanov:
Does this still work?

I deleted the `ForwardDiff` stuff for this test.

On a brand new install of julia 1.0.3 with `add DifferentialEquations Zygote` I simply get a crash when `Zygote.derivative` is called (no error messages, the notebook kernel simply crashes, with no explanation or a segfault on the terminal).

With `add DifferentialEquations#master Zygote#master` I get some error about dual numbers:

```
MethodError: no method matching Float64(::ForwardDiff.Dual{Nothing,Float64,2})
  ...
Stacktrace:
 [1] convert(::Type{Float64}, ::ForwardDiff.Dual{Nothing,Float64,2}) at ./number.jl:7
 [2] (::getfield(Zygote, Symbol("##819#822")){typeof(convert)})(::Type, ::Float64) at /home/stefan/.julia/packages/Zygote/Ohw1K/src/lib/broadcast.jl:113
```

I even tried a much simplified piece of code, which also fails:

```julia
f(x,p,t) = p*x
t = 0:0.1:1
function sol(a)
    pr = ODEProblem(f,1.,(0.,1.),a)
    s = solve(pr,Euler(),dt=0.01, saveat=t)
    s.u
end
Zygote.gradient(a->sum(sol(a)),1.)
```
I get with `add DifferentialEquations#master Zygote#master`:
```
[...]

PR #75, opened 2019-02-23
Support Mutation

MikeInnes:
Specifically, mutation of arrays/values, as opposed to mutation of data structures, which we already support well.

This introduces some internal complexity and makes performance a little trickier, so it remains open whether we'll actually want to merge it in. The main goal right now is that people can play with this and test it, and in particular I'd like to get some nice benchmarks with differential equations (#37).

It is of course not ideal to maintain this separately. One option might be to make mutation optional and compile different code if it's enabled, though again this is a significant additional complexity in the system.


andreasnoack:
Tried this out with `TransformVariables`. The examples below transforms back and forth from `Vector` to `NamedTuple`, i.e. the identity, and the sums the result.
```julia
julia> using TransformVariables

julia> t = as((μ = asℝ, σ = asℝ₊))
TransformVariables.TransformNamedTuple{(:μ, :σ),Tuple{TransformVariables.Identity,TransformVariables.ShiftedExp{true,Float64}}}((TransformVariables.Identity(), TransformVariables.ShiftedExp{true,Float64}(0.0)), 2)

julia> sum(inverse(t, transform(t, ones(2))))
2.0

julia> gradient(s -> sum(inverse(t, transform(t, s))), ones(2))
ERROR: Compiling Tuple{typeof(inverse!),Array{Float64,1},TransformVariables.TransformNamedTuple{(:μ, :σ),Tuple{TransformVariables.Identity,TransformVariables.ShiftedExp{true,Float64}}},NamedTuple{(:μ, :σ),Tuple{Float64,Float64}}}: MethodError: no method matching exprtype(::Core.Compiler.IRCode, ::ArgCheck.ArgCheckFlavor)
Closest candidates are:
  exprtype(::Core.Compiler.IRCode, ::Expr) at /Users/andreasnoack/.julia/dev/Zygote/src/tools/ir.jl:64
  exprtype(::Core.Compiler.IRCode, ::QuoteNode) at /Users/andreasnoack/.julia/dev/Zygote/src/tools/ir.jl:61
  exprtype(::Core.Compiler.IRCode, ::GlobalRef) at /Users/andreasnoack/.julia/dev/Zygote/src/tools/ir.jl:60
  ...
Stacktrace:
 [1] _broadcast_getindex_evalf at ./broadcast.jl:625 [inlined]
 [2] _broadcast_getindex at ./broadcast.jl:598 [inlined]
 [3] getindex at ./broadcast.jl:558 [inlined]
 [4] copyto_nonleaf!(::Array{DataType,1}, ::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1},Tuple{Base.OneTo{Int64}},typeof(Zygote.exprtype),Tuple{Base.RefValue{Core.Compiler.IRCode},Base.Broadcast.Extruded{Array{Any,1},Tuple{Bool},Tuple{Int64}}}}, ::Base.OneTo{Int64}, ::Int64, ::Int64) at ./broadcast.jl:982
 [5] copy at ./broadcast.jl:836 [inlined]
```

andreasnoack:
Oh. I'm now reading the actual error more carefully and can see that it's related to `ArgCheck.jl`. I guess handling macros could be complicated. Any ideas to how to proceed?

MikeInnes:
Small update. Maintaining this separately is kind of a pain. I think I'd like to merge this and keep it under a feature flag like `ZYGOTE_TYPED`, and then it'll get CI and so on. Still no immediate plan to make it default behaviour, though.

Roger-luo:
remember to add this (or something equivalent) later to make tuple vector conversion work.

```julia
Zygote.@adjoint! function copyto!(xs::AbstractVector, ys::Tuple)
    xs_ = copy(xs)
    copyto!(xs, ys), function (dxs)
        copyto!(xs_, xs)
        return (nothing, Tuple(dxs))
    end
end
```

cossio:
Is there a way to support mutation of arrays whose gradient is dropped (`Zygote.dropgrad`)? That is, suppose I have a function:

```
A = zeros(2,2)
f(x) = (Zygote.dropgrad(A .= x); return x .+ A)
f'(x) # doesn't work
```

So here I want to treat `A` as a constant and mutate it.

MikeInnes:
If you don't care about gradient at all you can define something like

```julia
ignore(f) = f()
@nograd f
```

and you can do the mutation inside a `do` block.

maartenvd:
So what is the plan for supporting mutation?

CarloLucibello:
closing as outdated

PR #94, opened 2019-03-07
Wct/more rules

willtebbutt:
Some rules from StatsFuns. `softmax` appears to be implemented in NNlib also. Is there a simple way to avoid the conflict between the two?


MikeInnes:
LGTM; you'll have to explicitly `import StatsFuns: ...` for the things you need to avoid the name conflict (or the same for NNlib).

willtebbutt:
New commit resolves the above issue. I'm content with this PR if tests pass.

Issue #76, opened 2019-02-23
Deprecation warning in tests


MikeInnes:
```
┌ Warning: implicit `dims=2` argument now has to be passed explicitly to specify that distances between columns should be computed
│   caller = ip:0x0
└ @ Core :-1
```

Bit unfortunate that there's no line info, but I'm guessing this is down to Distances.jl

PR #103, opened 2019-03-10
Fix #102

### LINKED ISSUE #102: Distances & keyword dims
`Distances.pairwise` now wants a keyword `dims=2` (due to https://github.com/JuliaStats/Distances.jl/pull/121) but this is not taken into account by gradients from #60 here: 
```julia
using Zygote, Distances

cols = rand(2,3)

pairwise(SqEuclidean(), cols) # Warning: implicit `dims=2` argument now has to be passed explicitly to specify that distances between columns should be computed

pairwise(SqEuclidean(), cols; dims=2)


Zygote.gradient(x -> pairwise(SqEuclidean(), x)[1,2], cols) # ok!

Zygote.gradient(x -> pairwise(SqEuclidean(), x; dims=2)[1,2], cols) # MethodError: no method matching adjoint(::Zygote.Context, ::typeof(pairwise), ::SqEuclidean, ::Array{Float64,2}; dims=2)
```

mcabbott:
Fixes #102.

`Distances.pairwise(SqEuclidean(), cols; dims=1)` just inserts lots of transposes, so so did I.

### DIFF
--- a/src/lib/distances.jl
+++ b/src/lib/distances.jl
@@ -15,18 +15,32 @@ end
   end
 end
 
-@adjoint function pairwise(s::SqEuclidean, x::AbstractMatrix, y::AbstractMatrix)
-  return pairwise(s, x, y), function(Δ)
-    x̄ = 2 .* (x * Diagonal(vec(sum(Δ; dims=2))) .- y * Δ')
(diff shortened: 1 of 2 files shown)


mcabbott:
Thanks for the original version! 

BTW I just found a tab I had open with #76, which this should also fix.

PR #106, opened 2019-03-11
Setup coverage using Codecov.

tpapp:
Using a project-based setup (current best practice), running coverage
on Julia 1.1.

Assumes that the associated tooling is enabled (Github app, Coverage).

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,1 +1,1 @@
-comment: false
+comment: off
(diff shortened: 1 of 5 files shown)


MikeInnes:
Cool. My general concern with this kind of thing is not getting coverage of the actual compiler machinery, but it seems like it's working here.

Can we turn off the obnoxious comments it leaves on PRs?

tpapp:
I think it was already intended to be disabled in the repo, but using the wrong syntax. I tried to disable it based on the [codecov docs](https://docs.codecov.io/docs/pull-request-comments).

CarloLucibello:
we do have coverage now

Issue #109, opened 2019-03-11
Inexact Error with Zygote gradient on TaylorSeries derivative


jessebett:
```
using Zygote
using Flux
using TaylorSeries # master branch

F = Chain(Dense(1,10,σ),Dense(10,10))

shift_taylor(a,n) = a + Taylor1(typeof(a),n)
t = shift_taylor(1.,5)

# Try to take the gradient of of the first derivative:
Zygote.gradient(()->TaylorSeries.derivative(1,sum(F([t]))),params(F))
```

This gives the error:
```
ERROR: InexactError: UInt128(0.07210570448898418)
Stacktrace:
 [1] Type at ./float.jl:682 [inlined]
 [2] convert at ./number.jl:7 [inlined]
 [3] setindex! at ./array.jl:767 [inlined]
 [4] setindex! at ./multidimensional.jl:460 [inlined]
 [5] #765 at /Users/jesse/.julia/packages/Zygote/8pKck/src/lib/array.jl:17 [inlined]
 [6] (::getfield(Zygote, Symbol("##1974#back#769")){getfield(Zygote, Symbol("##765#767")){Array{UInt128,1},Tuple{Int128}}})(::Float64) at /Users/jesse/.julia/packages/Zygote/8pKck/src/lib/grad.jl:41
 [7] factorial_lookup at ./combinatorics.jl:21 [inlined]
 [8] (::typeof(∂(Base.factorial_lookup)))(::Float64) at /Users/jesse/.julia/packages/Zygote/8pKck/src/compiler/interface2.jl:0
 [9] factorial at ./combinatorics.jl:25 [inlined]
 [10] (::typeof(∂(factorial)))(::Float64) at /Users/jesse/.julia/packages/Zygote/8pKck/src/compiler/interface2.jl:0
 [11] derivative at /Users/jesse/.julia/packages/TaylorSeries/cbisT/src/calculus.jl:92 [inlined]
 [12] (::typeof(∂(TaylorSeries.derivative)))(::Int8) at /Users/jesse/.julia/packages/Zygote/8pKck/src/compiler/interface2.jl:0
 [13] #7 at ./REPL[9]:1 [inlined]
 [14] (::typeof(∂(getfield(Main, Symbol("##7#8"))())))(::Int8) at /Users/jesse/.julia/packages/Zygote/8pKck/src/compiler/interface2.jl:0
[...]

PR #110, opened 2019-03-11
Nograd Factorial

jessebett:
Discussed in #109 @willtebbutt suggested `@nograd` for `factorial`

### DIFF
--- a/src/lib/real.jl
+++ b/src/lib/real.jl
@@ -32,4 +32,4 @@ end
 
 @adjoint a // b = (a // b, c̄ -> (c̄ * 1//b, - c̄ * a // b // b))
 
-@nograd floor, ceil, trunc, round, hash
+@nograd floor, ceil, trunc, round, hash, factorial


willtebbutt:
Thanks for doing this @jessebett . Could you please add a quick test to `test/gradcheck.jl` to ensure that we don't accidentally revert this at some point?

Keno:
Would it make sense to instead use the derivative of the Gamma function here?

willtebbutt:
Hmm... is the Gamma function the unique extension of factorial to the reals, or are there other choices?

devmotion:
One needs at least the three properties stated in the [Bohr-Mollerup theorem](https://en.wikipedia.org/wiki/Bohr%E2%80%93Mollerup_theorem): `f(1) = 1`, `f(x + 1) = x f(x)` for `x > 0`, and `f` being logarithmically convex.

CarloLucibello:
we have 
```
@non_differentiable factorial(n::Integer)
```
in ChainRules.jl now

PR #117, opened 2019-03-18
Fix getindex

willtebbutt:
Related to [#139](https://github.com/invenia/Nabla.jl/issues/139). I'm surprised that this is only coming to light now if I'm honest.


willtebbutt:
@MikeInnes what's the feeling on this? Not sure what we decided to do about the duplication problem. Do we wait for a backported fix, or go ahead and merge this?

baggepinnen:
I ran into a bug where gradients were dropped silently from a code looking like this
```julia
function linearityloss()
    Z = map(input) do x
        some code ...
    end
    Z = reduce(hcat, Z)'  # Using custom adjoint from SlicedMap.jl
    Y = Z[2:end,:]
    A = Z[1:end-1,:]
    w = A\Y
    lr = mean(abs2, Y-A*w)
end
```
modifying it to the following solved the problem, indicating that `getindex` might have been the problem.
```julia
function linearityloss()
    Z = map(input) do x
        some code ...
    end
    Z = reduce(hcat, Z)' 
    Y = view(Z, 2:size(Z,1),:) # Using custom adjoint from #219 
    A = view(Z, 1:size(Z,1)-1,:)
    w = A\Y
    lr = mean(abs2, Y-A*w) #+ 0.1sum(abs2,f.cell.b)
end
```

MikeInnes:
Yes, we should go ahead and merge this if it fixes that bug.

willtebbutt:
@MikeInnes any idea why CI failed / should we be worried about this, or just go ahead and merge?

MikeInnes:
It might be a conflict with master; worth merging master just to see if it goes away.

CarloLucibello:
what's the status of this? Do we still need this fix?

willtebbutt:
Closing as is now redundant.

PR #123, opened 2019-03-25
Implements `cholesky` for `Diagonal` matrices

willtebbutt:
Also handles `Diagonal` matrices. `gradtest` is currently producing a weird error though (that I don't think it should be / I don't understand). Will try to resolve in the morning.


willtebbutt:
See #124

willtebbutt:
Merged #127 to resolve issues around #124 . Also took the opportunity to refine some functionality with `Symmetric` as there was a hole there.

willtebbutt:
@MikeInnes have fixed up this PR. Should be good to go now.

PR #137, opened 2019-04-09
Add adjoint for LinearAlgebra.tr

Roger-luo:
This implement the backward pass for `tr`.

So it is working generally with the original definition, but the performance is poor.

```julia
julia> using Zygote, LinearAlgebra, BenchmarkTools

julia> @benchmark gradient(tr, $(rand(30, 30)))
BenchmarkTools.Trial:
  memory estimate:  433.33 KiB
  allocs estimate:  303
  --------------
  minimum time:     61.731 μs (0.00% GC)
  median time:      87.743 μs (0.00% GC)
  mean time:        128.345 μs (34.01% GC)
  maximum time:     53.973 ms (99.78% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

I'm not sure if Zygote should be able to generate the fast version directly, but it seems it records a lot `getindex` in the stack (since `LinearAlgebra.tr` is implemented as a for loop.

But anyway, this will make it 4x faster than Tracker

```julia
julia> @benchmark gradient(tr, $(rand(30, 30)))
BenchmarkTools.Trial:
  memory estimate:  160 bytes
  allocs estimate:  6
  --------------
  minimum time:     321.061 ns (0.00% GC)
  median time:      349.290 ns (0.00% GC)
  mean time:        404.831 ns (8.65% GC)
  maximum time:     211.392 μs (99.77% GC)
  --------------
  samples:          10000
  evals/sample:     214
```

I'm not sure if I should replace the original `FillArray` in tools with what's in `FillArrays`? since I added it as dependency to use `Eye` here.


Roger-luo:
Tests somehow fails on nightly, but seems not related to this PR...

MikeInnes:
:+1: Thanks!

PR #138, opened 2019-04-09
Propagate gradients of mutable structs

### LINKED ISSUE #93: Gradients for mutable structs
```julia
using Zygote

mutable struct Bar
  x
end

gradient(b -> b.x,  Bar(1)) # nothing
```

Zygote calculates correct gradients here but doesn't show them, because they are stored in an internal cache rather than propagating via the usual data flow. We should carry on propagating by data flow and use the cache to retrieve the adjoint structs where needed.

How should we represent mutable adjoint structs? We currently use a named-tuple-of-refs, but I think a ref-of-named-tuples might be nicer if this is going to be user exposed.

MikeInnes:
fixes #93. Also implements the change mentioned there; gradients of mut structs are now refs-of-named-tuples.

### DIFF
--- a/src/lib/lib.jl
+++ b/src/lib/lib.jl
@@ -1,3 +1,5 @@
+using Base: RefValue
+
 # Interfaces
 
 accum() = nothing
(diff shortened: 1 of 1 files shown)


jekbradbury:
I would naively have thought named-tuples-of-refs?

MikeInnes:
That's actually what we had before but I figured this was simpler if we're going to expose it. Do you think that would be better?

jekbradbury:
I can't think of any fundamental reason to prefer nt-of-refs, so I'd just go with your judgment

MikeInnes:
It's sort-of less convenient right now, but hopefully once we get better support for mutating immutables it'll be the closest thing to an anonymous mutable struct.

PR #141, opened 2019-04-10
More complex tests, and some words of explanation

antoine-levitt:
Ref https://github.com/FluxML/Zygote.jl/commit/e1b00f78665486343d397a82590c27aa736e8bce#commitcomment-33119240, and https://github.com/FluxML/Zygote.jl/issues/29. Taking `z.re` and `z.im` doesn't seem to work, so I commented it out for now.


test/complex.jl
```diff
@@ -1,3 +1,72 @@
...
+# For holomorphic functions, ∇f (as defined above) = du/dx + i du/dy = du/dx - i dv/dx = conj(f'), with f' the complex derivative
+
+fs_C_to_R = (real,
+             imag,
+             abs,
+             abs2,
+             z -> abs(z)*cos(im*angle(z)),
+             # z->z.re,
+             # z->z.im,
+             # z->2z.re + 3z.im,
```
MikeInnes: You could just use the `real` and `imag` functions here

antoine-levitt: Sure, the point was to test these as well

MikeInnes: If we add them they will trivially be the same as `real`/`imag`, so it's probably not too important to test them independently.

MikeInnes:
Heads up, I just went ahead and removed the implicit `real`, so it'd be best to add that back in explicitly for the `->C` tests.

test/complex.jl
antoine-levitt: That's not clear to me why it needs a specific implementation (it should just be a struct access?), but I can remove them, sure

MikeInnes: The default rule for struct access creates an "adjoint struct" for `Complex` like `(re = grad, im = nothing)`; we have to define it explicitly so that it produces a new complex number instead (and also so that it handles complex `grad` by dropping the imaginary part).

chriscoey:
more complex tests is great!

Issue #130, opened 2019-03-28
Efficient Higher-Order Forward Mode


jessebett:
I've already opened a few disparate issues related to this, but I'd like to centralize these efforts.

The goal is to compute higher-order derivatives of (scalar-input) functions. The input v. output dimensionality of the problems we're working with favours forward mode autodiff. But naively nesting calls to ForwardDiff causes an exponential blow-up. 

However, we believe that the N-th order derivative of a scalar-input function should be computable on the order of N-times the cost of evaluating the function, so not exponential.

Experiments with https://github.com/JuliaDiff/TaylorSeries.jl suggests that their Taylor Series implementation of automatic differentiation is able to achieve this scaling. Likely this is done by sharing computation of intermediate values required to compute lower order derivatives with the higher orders. This sharing is kinda hand designed by writing the rules for the computation of Taylor coefficients. 

We've also mentioned this to the guys working on jax and they're considering a CSE approach to make this more general. You can see their progress on the issue here: google/jax#520 

To clarify what the problem is here. TaylorSeries is able to give us efficient higher order derivatives of scalar functions with their Taylor coefficient method for automatic differentiation. However, this is currently not composable with Zygote (or Flux). We will want (reverse-mode) gradients through the higher-order derivative computations wrt parameters of the function.

i.e. quick pseudocode for the goal
```
f(t, z, p) # vector-output function with scalar input, t, other vector input we'll want grads wrt, and large number of parameters p, also want grads wrt
dNt = TaylorSeries.derivative(N,f(t,z,p)) # Use Taylor series to get efficient higher order derivatives of f wrt t
dNtdp = Zygote.gradient(()->dNt,p) # Use Zygote to get gradients of the N-th order time derivative wrt the parameters p
```

The issues I've already run into are that TaylorSeries are custom types, so `getproperty` needs adjoints defined, which is trickier than I'd like: #111
Also `factorial` is used when converting taylor coefficients into the respective order derivative, so adjoint needs to be defined for that: #109 (got distracted, we couldn't decide if this should be a nograd or if this is relaxed to gamma #110)
Also unearthed some problems with TaylorSeries like their eltype JuliaDiff/TaylorSeries.jl#198 and rounding JuliaDiff/TaylorSeries.jl#199 which I ran up against while trying to use TaylorSeries through Flux Dense layers, which force rounding to Float32, for example.

[...]

PR #144, opened 2019-04-10
Taylor Series

MikeInnes:
Addressing #130.

```julia
using Zygote, TaylorSeries
using Zygote: taylor, derive

gradient(5, 3) do x, n
  x, n = taylor(x), taylor(n)
  x = x + Taylor1(10)
  y = x*n
  derive(y, 1)
end # (0, 1)
```

`derive(y, n)` gets the derivative `d^n y / dx^n`, and then we take the gradient of `x` and `n` with respect to that. It's a bit clunky right now; in particular, anything you want gradients for, you have to call `taylor` on it before using it. So it's not totally automated when it comes to e.g. running this through a Flux model, though this can still be done with `destructure` and `restructure` as in DiffEqFlux.

We can also differentiate against many orders at once:

```julia
gradient(3) do n
  n = taylor(n)
  x = 1 + Taylor1(10)
  y = sin(x+n)
  sum(derive(y, i) for i = 1:10)
end # 1.41
```

TaylorSeries is accurate and instantaneous even up to about 100th order.

Hoping @jessebett can play around with this and get something more realistic working.


jessebett:
`derive` means take nth order derivatives with TaylorSeries method? I think that should be clearly named something like `taylordiff`

jessebett:
Also, not the biggest fan of `n=taylor(n)`.  This confuses where the taylor expansion actually is. For example, in your above code you have 
```
n = taylor(n)
x = 1 + Taylor1(10)
```

So it should be clear that the higher order taylor terms are for x. So it's not clear what `taylor(n)` is doing, right? This should be very clear because otherwise what variable is `derive(y,1)` wrt? It's the variable in the TaylorSeries expansion for x, but having `taylor(n)` makes this ambiguous.

MikeInnes:
Yeah, the interface is not great right now. The right way to do this would clearly be `taylordiff(f, x)`, and then we could even do things like automatically handling the vector `x` case.

The technical reason for this is that the adjoint of `derive` is not a normal value but a `TaylorAdjoint` that stores the derivative at multiple orders. At some point we have to reduce that `TaylorAdjoint` back to a normal value; currently there's no obvious place to do that unless you ask for it via `taylor`.

jessebett:
Possibly my issue is that I'm not sure where `taylor` is supposed to go. But gradients through basic operations with `Taylor1` are still failing

```
θ = taylor(5.)
f(x) = x .* θ
x = Taylor1(1.,3)
f(x)
gradient(()->f(x)[0],Params([θ]))
```

This currently returns `ERROR: ArgumentError: broadcasting over dictionaries and `NamedTuple`s is reserved`

jessebett:
```
θ = taylor(5.)
f(t,z) = θ .* t.^5 + z
x = taylor(1.)
t = shift_taylor(0,10)
f(t,x)
derive(f(t,x),5)
gradient(()->derive(f(t,x),5),Params([θ]))
```

This returns `ERROR: MethodError: no method matching length(::Zygote.TaylorAdjoint{Int8})`

MikeInnes:
`taylor` needs to be part of the forward pass, since it does its work during the backwards pass. Here's a working version:

```julia
using Zygote, TaylorSeries
using Zygote: taylor, derive

shift_taylor(x,n) = x + Taylor1(typeof(x),n)

θ = [5.]

f(t,z) = taylor(θ[]) .* t.*5.0 + z

x = 1.
t = shift_taylor(0,10)
f(t,x)
derive(f(t,x),5)

gs = gradient(()->derive(f(t,x),5),Params([θ]))
gs[θ]
```

Unfortunately this fails with `t^5` due to us trying to call `log` of a `0` taylor series in the backwards pass. If you need that particular operation we'll have to work around that some how.

MikeInnes:
Closing this for now as I think we want something more general for this to be worthwhile.

PR #145, opened 2019-04-10
Update for NNlib overhaul

staticfloat:
With these changes, I can actually use Flux-on-Zygote to run some convolutional models, which is fun.


MikeInnes:
Cool! I guess it just needs a manifest update?

staticfloat:
Looks like I needed to copy-paste some tests from my recent `Tracker.jl` PR.  All green now!

Issue #154, opened 2019-04-14
Loop call in _forward function of collect


GiggleLiu:
I got stackoverflowerror when calling a function that contains `intersect`

```julia
julia> gradient(x->(intersect((1,2), (2,3)); sum(x)), a)
ERROR: StackOverflowError:
Stacktrace:
 [1] _forward(::Zygote.Context, ::typeof(collect), ::Base.Generator{Base.Iterators.Filter{getfield(Base, Symbol("##83#84")){typeof(in),typeof(pop!),Set{Int64}},Tuple{Int64,Int64}},getfield(Base, Symbol("##85#86"))}) at /home/leo/.julia/dev/Zygote/src/lib/array.jl:92
 [2] map at ./abstractarray.jl:2044 [inlined]
 [3] _forward(::Zygote.Context, ::typeof(map), ::getfield(Base, Symbol("##85#86")), ::Base.Iterators.Filter{getfield(Base, Symbol("##83#84")){typeof(in),typeof(pop!),Set{Int64}},Tuple{Int64,Int64}}) at /home/leo/.julia/dev/Zygote/src/compiler/interface2.jl:0
 ... (the last 3 lines are repeated 32442 more times)
 [97330] _forward(::Zygote.Context, ::typeof(collect), ::Base.Generator{Base.Iterators.Filter{getfield(Base, Symbol("##83#84")){typeof(in),typeof(pop!),Set{Int64}},Tuple{Int64,Int64}},getfield(Base, Symbol("##85#86"))}) at /home/leo/.julia/dev/Zygote/src/lib/array.jl:92
 [97331] _forward(::Zygote.Context, ::typeof(Base.vectorfilter), ::getfield(Base, Symbol("##83#84")){typeof(in),typeof(pop!),Set{Int64}}, ::Tuple{Int64,Int64}) at ./array.jl:2393
 [97332] _shrink at ./array.jl:2397 [inlined]
 [97333] _forward(::Zygote.Context, ::typeof(Base._shrink), ::typeof(intersect!), ::Tuple{Int64,Int64}, ::Tuple{Tuple{Int64,Int64}}) at /home/leo/.julia/dev/Zygote/src/compiler/interface2.jl:0
 [97334] intersect at ./array.jl:2400 [inlined]
[...]

PR #156, opened 2019-04-14
[WIP] Fix issue #152, #154

### LINKED ISSUE #152: Broadcasting causes incorrect gradient for complex numbers
Sample code

```julia
using Zygote
julia> h1(a) = real((a*conj(a)));

julia> h2(a) = real((a.*conj(a)));

julia> h3(a) = real(([a].*conj([a])))[];

julia> h1'(0.3im)
0.0 + 0.6im

julia> h2'(0.3im)
ERROR: MethodError: no method matching mapfoldl_impl(::typeof(identity), ::typeof(Zygote.accum), ::NamedTuple{(:dims,),Tuple{Colon}}, ::Complex{Float64})
Closest candidates are:
  mapfoldl_impl(::Any, ::Any, ::NamedTuple{(:init,),T} where T<:Tuple, ::Any, ::Any...) at reduce.jl:40
  mapfoldl_impl(::Any, ::Any, ::NamedTuple{(),T} where T<:Tuple, ::Any) at reduce.jl:55
Stacktrace:
 [1] #mapfoldl#187(::Base.Iterators.Pairs{Symbol,Colon,Tuple{Symbol},NamedTuple{(:dims,),Tuple{Colon}}}, ::Function, ::Function, ::Function, ::Complex{Float64}) at ./reduce.jl:72
 [2] (::getfield(Base, Symbol("#kw##mapfoldl")))(::NamedTuple{(:dims,),Tuple{Colon}}, ::typeof(mapfoldl), ::Function, ::Function, ::Complex{Float64}) at ./none:0
 [3] #mapreduce#191(::Base.Iterators.Pairs{Symbol,Colon,Tuple{Symbol},NamedTuple{(:dims,),Tuple{Colon}}}, ::Function, ::Function, ::Function, ::Complex{Float64}) at ./reduce.jl:205
 [4] (::getfield(Base, Symbol("#kw##mapreduce")))(::NamedTuple{(:dims,),Tuple{Colon}}, ::typeof(mapreduce), ::Function, ::Function, ::Complex{Float64}) at ./none:0
 [5] #reduce#192(::Base.Iterators.Pairs{Symbol,Colon,Tuple{Symbol},NamedTuple{(:dims,),Tuple{Colon}}}, ::Function, ::Function, ::Complex{Float64}) at ./reduce.jl:361
 [6] (::getfield(Base, Symbol("#kw##reduce")))(::NamedTuple{(:dims,),Tuple{Colon}}, ::typeof(reduce), ::Function, ::Complex{Float64}) at ./none:0
 [7] #accum_sum#1040(::Function, ::Function, ::Complex{Float64}) at /home/leo/.julia/dev/Zygote/src/lib/broadcast.jl:33
 [8] accum_sum(::Complex{Float64}) at /home/leo/.julia/dev/Zygote/src/lib/broadcast.jl:33
 [9] unbroadcast(::Complex{Float64}, ::Complex{Float64}) at /home/leo/.julia/dev/Zygote/src/lib/broadcast.jl:47
 [10] (::getfield(Zygote, Symbol("##1054#1055")){Complex{Float64},Complex{Float64}})(::Float64) at /home/leo/.julia/dev/Zygote/src/lib/broadcast.jl:60
 [11] (::getfield(Zygote, Symbol("##2667#back#1056")){getfield(Zygote, Symbol("##1054#1055")){Complex{Float64},Complex{Float64}}})(::Float64) at /home/leo/.julia/dev/Zygote/src/lib/grad.jl:46
 [12] h2 at ./REPL[29]:1 [inlined]
 [13] (::typeof(∂(h2)))(::Float64) at /home/leo/.julia/dev/Zygote/src/compiler/interface2.jl:0
 [14] (::getfield(Zygote, Symbol("##71#72")){typeof(∂(h2))})(::Float64) at /home/leo/.julia/dev/Zygote/src/compiler/interface.jl:38
 [15] gradient(::Function, ::Complex{Float64}) at /home/leo/.julia/dev/Zygote/src/compiler/interface.jl:47
 [16] (::getfield(Zygote, Symbol("##73#74")){typeof(h2)})(::Complex{Float64}) at /home/leo/.julia/dev/Zygote/src/compiler/interface.jl:50
 [17] top-level scope at none:0

julia> h3'(0.3im)
0.0 - 0.6im
```

An array wrapper changes to behavior to the wrong direction. This is really wield... Which part of code should I look into?

GiggleLiu:
fix #152, #154

As well as the slow `conj` backward problem for array.

### DIFF
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -34,8 +34,8 @@ accum_sum(xs; dims = :) = reduce(accum, xs, dims = dims)
 
 # Work around reducedim_init issue
 accum_sum(xs::AbstractArray{Nothing}; dims = :) = nothing
-accum_sum(xs::AbstractArray{<:Real}; dims = :) = sum(xs, dims = dims)
-accum_sum(xs::Real; dims = :) = xs
+accum_sum(xs::AbstractArray{<:Number}; dims = :) = sum(xs, dims = dims)
+accum_sum(xs::Number; dims = :) = xs
 
 trim(x, Δ) = reshape(Δ, ntuple(i -> size(Δ, i), Val(ndims(x))))
 
@@ -58,6 +58,11 @@ unbroadcast(x::Union{Number,Ref}, x̄) = accum_sum(x̄)
 
 @adjoint broadcasted(::typeof(*), x, y) = x.*y,
   z̄ -> (nothing, unbroadcast(x, z̄ .* y), unbroadcast(y, z̄ .* x))
+
+@adjoint broadcasted(::typeof(*), x::Array{<:Number}, y::Array{<:Number}) = x.*y,
+  z̄ -> (nothing, unbroadcast(x, z̄ .* conj(y)), unbroadcast(y, z̄ .* conj(x)))
+@adjoint broadcasted(::typeof(*), x::Number, y::Number) = x.*y,
+  z̄ -> (nothing, unbroadcast(x, z̄ .* conj(y)), unbroadcast(y, z̄ .* conj(x)))
 
 # General Fallback
 # ================
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -61,12 +61,14 @@ Numeric{T<:Number} = Union{T,AbstractArray{<:T}}
 @adjoint broadcasted(::typeof(*), x::Numeric, y::Numeric) = x.*y,
   z̄ -> (nothing, unbroadcast(x, z̄ .* conj.(y)), unbroadcast(y, z̄ .* conj.(x)))
 
-@adjoint broadcasted(::typeof(conj), x::Numeric) = conj.(x),
-    z̄ -> (nothing, unbroadcast(x, conj.(z̄)))
-@adjoint broadcasted(::typeof(real), x::Numeric) = real.(x),
-    z̄ -> (nothing, unbroadcast(x, real.(z̄)))
-@adjoint broadcasted(::typeof(imag), x::Numeric) = imag.(x),
-    z̄ -> (nothing, unbroadcast(x, im .* real.(z̄)))
+@adjoint broadcasted(::typeof(conj), x::Numeric) =
+  conj.(x), z̄ -> (nothing, conj.(z̄))
+
+@adjoint broadcasted(::typeof(real), x::Numeric) =
+  real.(x), z̄ -> (nothing, real.(z̄))
+
+@adjoint broadcasted(::typeof(imag), x::Numeric) =
+  imag.(x), z̄ -> (nothing, im .* real.(z̄))
 
 # General Fallback
 # ================
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -64,6 +64,13 @@ unbroadcast(x::Union{Number,Ref}, x̄) = accum_sum(x̄)
 @adjoint broadcasted(::typeof(*), x::Number, y::Number) = x.*y,
   z̄ -> (nothing, unbroadcast(x, z̄ .* conj(y)), unbroadcast(y, z̄ .* conj(x)))
 
+@adjoint broadcasted(::typeof(conj), x) = conj.(x),
+    z̄ -> (nothing, unbroadcast(x, conj.(z̄)))
+@adjoint broadcasted(::typeof(real), x) = real.(x),
+    z̄ -> (nothing, unbroadcast(x, real.(z̄)))
+@adjoint broadcasted(::typeof(imag), x) = imag.(x),
+    z̄ -> (nothing, unbroadcast(x, im .* real.(z̄)))
+
 # General Fallback
 # ================
 
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -53,22 +53,19 @@ unbroadcast(x::Union{Number,Ref}, x̄) = accum_sum(x̄)
 # to do CSE, then broadcast-ify the expression so that the closure captures the
 # right arrays.
 
-@adjoint broadcasted(::typeof(+), xs...) =
+Numeric{T<:Number} = Union{T,AbstractArray{<:T}}
+
+@adjoint broadcasted(::typeof(+), xs::Numeric...) =
   broadcast(+, xs...), ȳ -> (nothing, map(x -> unbroadcast(x, ȳ), xs)...)
 
-@adjoint broadcasted(::typeof(*), x, y) = x.*y,
-  z̄ -> (nothing, unbroadcast(x, z̄ .* y), unbroadcast(y, z̄ .* x))
+@adjoint broadcasted(::typeof(*), x::Numeric, y::Numeric) = x.*y,
+  z̄ -> (nothing, unbroadcast(x, z̄ .* conj.(y)), unbroadcast(y, z̄ .* conj.(x)))
 
-@adjoint broadcasted(::typeof(*), x::Array{<:Number}, y::Array{<:Number}) = x.*y,
-  z̄ -> (nothing, unbroadcast(x, z̄ .* conj(y)), unbroadcast(y, z̄ .* conj(x)))
-@adjoint broadcasted(::typeof(*), x::Number, y::Number) = x.*y,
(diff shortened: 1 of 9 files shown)


src/lib/forward.jl
```diff
@@ -42,9 +42,9 @@ function forward_jacobian(f, x)
 end
 
 vec_scalar(x) = vec(x)
-vec_scalar(x::Real) = [x]
+vec_scalar(x::Number) = [x]
 reshape_scalar(x, y) = reshape(y, size(x))
-reshape_scalar(x::Real, y) = y[]
+reshape_scalar(x::Number, y) = y[]
```
MikeInnes: Where is this needed/used? `forwarddiff` only works with real numbers.

src/lib/broadcast.jl
```diff
@@ -59,6 +59,11 @@ unbroadcast(x::Union{Number,Ref}, x̄) = accum_sum(x̄)
 @adjoint broadcasted(::typeof(*), x, y) = x.*y,
   z̄ -> (nothing, unbroadcast(x, z̄ .* y), unbroadcast(y, z̄ .* x))
 
+@adjoint broadcasted(::typeof(*), x::Array{<:Number}, y::Array{<:Number}) = x.*y,
+  z̄ -> (nothing, unbroadcast(x, z̄ .* conj(y)), unbroadcast(y, z̄ .* conj(x)))
+@adjoint broadcasted(::typeof(*), x::Number, y::Number) = x.*y,
+  z̄ -> (nothing, unbroadcast(x, z̄ .* conj(y)), unbroadcast(y, z̄ .* conj(x)))
```
MikeInnes: You may as well fold these definitions into the `Any` definition above and add the `conj` there. Also, best if it's broadcasted, so that it can be fused where possible.

src/lib/number.jl
```diff
@@ -54,4 +54,9 @@ end
 @adjoint conj(x::Number) = conj(x), r̄ -> (conj(r̄),)
 @adjoint imag(x::Number) = imag(x), ī -> (real(ī)*im,)
 
+# Improve Array `conj`, `real` and `imag` Performance
+@adjoint real(x::AbstractArray) = real(x), r̄ -> (real(r̄),)
+@adjoint conj(x::AbstractArray) = conj(x), r̄ -> (conj(r̄),)
+@adjoint imag(x::AbstractArray) = imag(x), ī -> (real(ī)*im,)
```
MikeInnes: I would prefer this was fixed by adding the underlying rules for `broadcasted`, so that it also works when one does `conj.(x)` etc.

src/lib/forward.jl
GiggleLiu: It is not needed, I will roll back this change. Thanks for reading carefully.

src/lib/broadcast.jl
GiggleLiu: I thinkg for user defined `*` operations, using `conj` here may cause trouble? Calling the backward function of `*` with respect to specific types is desired.

src/lib/number.jl
GiggleLiu: Do you mean adding backward for the following interface?
```julia
@adjoint broadcasted(::typeof(conj), x) 
```

GiggleLiu: Thanks, after adding this interface, `conj.` also works. The only problem is, here the rule is duplicated. Is it possible to reuse the backward rule defined for element-wise `conj`?

MikeInnes: Not yet unless we share things via DiffRules or add a macro for this. This is fine for now, thanks.

src/lib/broadcast.jl
MikeInnes: The only way to do that is to just delete the special case entirely and have it fall back.

It shouldn't matter though; if someone has defined a `*` operation it should behave mathematically like `*`, and if it doesn't have the same gradient I see that as their issue.

GiggleLiu: > if someone has defined a * operation it should behave mathematically like *

This is often not true, maybe an error is better?

I tried to construct a counter example, but the error is not as expected
```julia
julia> using Zygote

julia> struct A
       data::Float64
       end
julia> Base.:*(a::A, b::A) = A(a.data+b.data)

julia> gradient(x->(A(x)*A(x)).data, 0.2)
(2.0,)

julia> gradient(x->([A(x)].*[A(x)])[].data, 0.2)
ERROR: MethodError: no method matching *(::NamedTuple{(:data,),Tuple{Float64}}, ::A)
Closest candidates are:
  *(::Any, ::Any, ::Any, ::Any...) at operators.jl:502
  *(::A, ::A) at REPL[6]:1
Stacktrace:
 [1] _broadcast_getindex_evalf at ./broadcast.jl:578 [inlined]
 [2] _broadcast_getindex at ./broadcast.jl:551 [inlined]
 [3] getindex at ./broadcast.jl:511 [inlined]
 [4] macro expansion at ./broadcast.jl:843 [inlined]
 [5] macro expansion at ./simdloop.jl:73 [inlined]
 [6] copyto! at ./broadcast.jl:842 [inlined]
 [7] copyto! at ./broadcast.jl:797 [inlined]
 [8] copy at ./broadcast.jl:773 [inlined]
 [9] materialize at ./broadcast.jl:753 [inlined]
 [10] (::getfield(Zygote, Symbol("##1054#1055")){Array{A,1},Array{A,1}})(::Array{Any,1}) at /home/leo/.julia/dev/Zygote/src/lib/broadcast.jl:60
 [11] #5 at /home/leo/.julia/dev/Zygote/src/lib/grad.jl:46 [inlined]
 [12] (::typeof(∂(getfield(Main, Symbol("##5#6"))())))(::Float64) at /home/leo/.julia/dev/Zygote/src/compiler/interface2.jl:0
 [13] (::getfield(Zygote, Symbol("##71#72")){typeof(∂(getfield(Main, Symbol("##5#6"))()))})(::Float64) at /home/leo/.julia/dev/Zygote/src/compiler/interface.jl:38
 [14] gradient(::Function, ::Float64) at /home/leo/.julia/dev/Zygote/src/compiler/interface.jl:47
 [15] top-level scope at none:0
```

If I comment the following two lines in `broadcast.jl`
```julia
#@adjoint broadcasted(::typeof(*), x, y) = x.*y,
#  z̄ -> (nothing, unbroadcast(x, z̄ .* y), unbroadcast(y, z̄ .* x))
```
I got
```julia
julia> gradient(x->([A(x)].*[A(x)])[].data, 0.2)
ERROR: ArgumentError: tuple must be non-empty
Stacktrace:
 [1] first(::Tuple{}) at ./tuple.jl:77
 [2] unzip(::Tuple{}) at /home/leo/.julia/dev/Zygote/src/lib/array.jl:76
 [3] adjoint at /home/leo/.julia/dev/Zygote/src/lib/array.jl:83 [inlined]
 [4] _forward at /home/leo/.julia/dev/Zygote/src/lib/grad.jl:44 [inlined]
 [5] broadcasted at ./broadcast.jl:1166 [inlined]
 [6] #17 at ./none:1 [inlined]
 [7] _forward(::Zygote.Context, ::getfield(Main, Symbol("##17#18")), ::Float64) at /home/leo/.julia/dev/Zygote/src/compiler/interface2.jl:0
 [8] _forward(::Function, ::Float64) at /home/leo/.julia/dev/Zygote/src/compiler/interface.jl:31
 [9] forward(::Function, ::Float64) at /home/leo/.julia/dev/Zygote/src/compiler/interface.jl:37
 [10] gradient(::Function, ::Float64) at /home/leo/.julia/dev/Zygote/src/compiler/interface.jl:46
 [11] top-level scope at none:0
```

I believe fixing this is beyond my ability.

MikeInnes: I think it'd be reasonable to define `Numeric{T<:Number} = Union{T,AbstractArray{T}}` and restrict the broadcast definition to that, rather than the current `Any`, to avoid this kind of problem.

MikeInnes: As another example could also define `map(A, B) = A*B` but at some point we just have to say ok, you're on your own here. We haven't nailed down exactly what that point is yet, but we can at least rule out certain cases that we're definitely not going to support.

src/lib/number.jl
MikeInnes: So with the `broadcasted` definitions added, can we remove these without losing performance?

MikeInnes:
Thanks @GiggleLiu!

PR #165, opened 2019-04-23
custom adjoints for lyap and matrix exponential

sdewaele:


MikeInnes:
Neat, thanks!

sdewaele:
I realize now it is better to limit the custom adjoint for the matrix exponential to `AbstractMatrix`, so line 281 in array.jl should read
```
@adjoint exp(A::AbstractMatrix) = exp(A), function(F̄)
```
@MikeInnes Would you like another PR for this or is this something you could add to one of your commits?

MikeInnes:
Ah, yes, that would be a good idea. A follow up PR would be great.

PR #166, opened 2019-04-24
Add more worked examples, including linear regression and an MLP MNIST

staticfloat:
I think it is worthwhile to have more examples showing users what usage of Zygote can look like, as it is just so incredibly clean and beautiful.

### DIFF
--- a/examples/linear_regression.jl
+++ b/examples/linear_regression.jl
@@ -1,0 +1,99 @@
+# Initialize environment in current directory
+@info("Ensuring example environment instantiated...")
+import Pkg
+Pkg.activate(@__DIR__)
+Pkg.instantiate()
+
+@info("Loading Zygote...")
+using Zygote, LinearAlgebra
+
+# This example will showcase how we do a simple linear fit with Zygote, making
+# use of complex datastructures, a home-grown stochastic gradient descent
+# optimizer, and some good old-fashioned math.  We start with the problem
+# statement:  We wish to learn the mapping `f(X) -> Y`, where `X` is a matrix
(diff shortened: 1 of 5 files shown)


oxinabox:
Do we have something that autobakes this into the docs?

staticfloat:
> Do we have something that autobakes this into the docs?

I'm not sure what this means.  But we should have some kind of CI for this, because I just discovered that the `profiler.jl` example has bitrotted, and I'm not sure how to fix it.  :P

oxinabox:
It would be nice if the code examples,
from this file

Showed up inside the documention website (with syntax highlighting etc)

examples/linear_regression.jl
```diff
@@ -0,0 +1,99 @@
...
+# of our model a tad bit according to the gradients
+function sgd_update!(model::LinearRegression, grads, η = 0.001)
+    model.weights .-= η .* grads.weights
+    model.bias -= η * grads.bias
+end
+
+# Now let's do that for each example in our training set:
+@info("Running train loop for $(size(X,2)) iterations")
+for idx in 1:size(X, 2)
+    grads = Zygote.gradient(m -> loss(m, X[:, idx], Y[idx]), model)[1][]
```
JeffBezanson: I think the `[]` is not needed here?

staticfloat: The `[]` is because `mutable struct` objects get wrapped in a `Ref` when passed back from Zygote.

JeffBezanson: Oops, that's only if you make it immutable.

MikeInnes:
Really nice examples, thanks!

PR #188, opened 2019-05-15
WIP: try to fix static array

Roger-luo:
I'm not sure if this is correct. but it looks like working. But I also tried to define

```julia
@adjoint function StaticArrays.SArray{S, T, N, L}(x::AbstractArray) where {S, T, N, L}
  return SArray{S, T, N, L}(x), function (Δ)
    return (Δ, )
  end
end
```

which is never called, I guess is because no matter what you feed into `SArray` it will be converted to a tuple first?

And maybe it is better to use `Requires`?


MikeInnes:
I think adding StaticArrays would be ok. We do need to add a gradient check; doing that in the normal way should work.

CarloLucibello:
closing as outdated

Issue #190, opened 2019-05-16
relation between builtin FillArray and FillArrays


Roger-luo:
I'm trying to get rid of the mutation happens here

https://github.com/QuantumBFS/LuxurySparse.jl/blob/09795174112120a542fe15b786c8379e7e4b28dc/src/arraymath.jl#L29

with `Fill` (it doesn't make sense to add Zygote as deps for `FillArrays`) but I'm wondering if Zygote could use `FillArrays` directly, so there're no duplicated things.

And we could include some `adjoints` for `FillArrays` directly here since it is a dependency already.

MikeInnes:
Closed by #191.

PR #191, opened 2019-05-16
use FillArrays instead

Roger-luo:
FYI: #190


Roger-luo:
Any comments? @MikeInnes

MikeInnes:
I guess we might as well, at this point.

PR #200, opened 2019-05-21
Add a few methods to `Buffer` to make it behave like an `Array`

YingboMa:
This PR will fix the test failures in https://github.com/JuliaDiffEq/DiffEqSensitivity.jl/pull/71

### DIFF
--- a/src/lib/buffer.jl
+++ b/src/lib/buffer.jl
@@ -54,6 +54,15 @@ function Base.copy(b::Buffer)
   return b.data
 end
 
+Base.eltype(b::Buffer) = eltype(b.data)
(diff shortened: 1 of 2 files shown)


MikeInnes:
Thanks – we could probably use `@forward` from MacroTools to make this a bit neater.

YingboMa:
Thanks for the suggestion. I have updated the code.

ChrisRackauckas:
Fixes https://github.com/JuliaDiffEq/DiffEqSensitivity.jl/pull/71

MikeInnes:
:+1: Thanks!

PR #203, opened 2019-05-26
use min in forward_jacobian

adamnemecek:


MikeInnes:
Thanks for the patch, but this does actually affect the meaning of the code (as far as optimisations are concerned); in the current implementation, when the default chunk size no dynamic dispatch will occur, whereas with this it will always occur.

PR #205, opened 2019-05-29
Update the Project.toml version

ChrisRackauckas:
Tag please?


MikeInnes:
IRTools needs a bump first, if you bump the version I can trigger registrator from the PR I think.

PR #210, opened 2019-06-01
Fix and test adjoint for hvcat

under-Peter:
This PR fixes the the adjoint for `hvcat`. The issue arises from 
```
julia> (hvcat((2,2),1,2,3,4)...,) == (1,2,3,4)
false
```
which was an implicit assumption in the implementation so far, leading to errors in functions where the arguments appear in matrices such as
```
foo(x) = [x 2x; 3x 4x]
```


MikeInnes:
:+1: Thanks!

PR #215, opened 2019-06-04
definitions for adjoints of FFTW functions

seb1000:
Adjoint definitions and accompanying tests for the most important functions of FFTW (fft and ifft).

Since deriving the adjoint of FFTs is a bit tricky, I double-checked the results with differentiating a simple DFT-implementation which is differentiable with ReverseDiff, and the results are equal down to machine-precision. This is possible since FFT is in fact only a more efficient implementation of DFT, which itself boils down to a simple matrix-multiplication.

I add this comparison as comment here for reference:
```julia

using ReverseDiff
using FFTW
using Test
using Zygote
using FillArrays: Fill


function dft(x::AbstractArray)
    """discrete fourier transform"""
    N = size(x)[1]
    out = Array{Any}(undef,N)
    for k in 0:N-1
        out[k+1] = sum([x[n+1]*exp(-2*im*π*k*n/N) for n in 0:N-1])
    end
    return out
end

function idft(x::AbstractArray)
    """discrete inverse fourier transform"""
    N = size(x)[1]
    out = Array{Any}(undef,N)
    for n in 0:N-1
        out[n+1] = 1/N*sum([x[k+1]*exp(2*im*π*k*n/N) for k in 0:N-1])
    end
    return out
end


x = randn(100)
@test dft(x) ≈ FFTW.fft(x)
@test idft(x) ≈ FFTW.ifft(x)


# FFTW functions do not work with FillArray. To make it work with FillArrays
# as well, overload the functions
FFTW.fft(x::Fill) = FFTW.fft(collect(x))
FFTW.ifft(x::Fill) = FFTW.ifft(collect(x))

# the gradient of an FFT with respect to its input is the reverse FFT of the
# gradient of its inputs.
Zygote.@adjoint function FFTW.fft(xs)
    return FFTW.fft(xs), function(Δ)
        N = length(Δ)
        return (N * FFTW.ifft(Δ),)
    end
end

Zygote.@adjoint function FFTW.ifft(xs)
    return FFTW.ifft(xs), function(Δ)
        N = length(Δ)
        return (1/N* FFTW.fft(Δ),)
    end
end


@test real(Zygote.gradient((x)->sum(abs.(FFTW.fft(x))),x)[1]) ≈ ReverseDiff.gradient((x)->sum(abs.(dft(x))),x)
@test real(Zygote.gradient((x)->sum(abs.(FFTW.ifft(x))),x)[1]) ≈ ReverseDiff.gradient((x)->sum(abs.(idft(x))),x)
```


src/lib/fft.jl
```diff
@@ -0,0 +1,57 @@
+
+using FFTW
+using FillArrays
+# FFTW functions do not work with FillArrays, which are needed
+# for some functionality of Zygote. To make it work with FillArrays
+# as well, overload the relevant functions
+FFTW.fft(x::Fill) = FFTW.fft(collect(x))
```
MikeInnes: You can use `dims...` to make this more compact.

Project.toml
```diff
@@ -16,7 +16,7 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
-
+FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
```
MikeInnes: Please use the package manager to add FFTW, so that the formatting is right and so that the manifest gets updated too.

src/Zygote.jl
```diff
@@ -35,6 +35,7 @@ include("lib/nnlib.jl")
 include("lib/broadcast.jl")
 include("lib/forward.jl")
 include("lib/utils.jl")
+include("lib/fft.jl")
```
MikeInnes: I think this file is short enough that it could just go in `array.jl`, after linear algebra.

src/lib/fft.jl
```diff
@@ -0,0 +1,57 @@
...
+end
+
+@adjoint function FFTW.fft(xs, dims)
+    # up to now only works when dims is a single integer
+    return FFTW.fft(xs, dims), function(Δ)
+        # dims can be int, array or tuple,
+        # if it is not a single int, convert to array so that we can use it
+        # for indexing
+        if typeof(dims) != Int
+            dims = collect(dims)
```
MikeInnes: I suspect this is unnecessary, since numbers act like single-element collections anyway for things like `prod`.

src/lib/fft.jl
```diff
@@ -0,0 +1,57 @@
...
+
+@adjoint function FFTW.ifft(xs)
+    return FFTW.ifft(xs), function(Δ)
+        N = length(xs)
+        return (1/N* FFTW.fft(Δ),)
+    end
+end
+
+@adjoint function FFTW.fft(xs, dims)
+    # up to now only works when dims is a single integer
```
MikeInnes: This comment isn't clear to me; you seem to be handling multiple dimensions below. If only single integers work, best to add a type restriction here.

test/gradcheck.jl
```diff
@@ -455,3 +455,18 @@ end
...
 end
 
+@testset "FFTW" begin
+  x=[-0.353213 -0.789656 -0.270151; -0.95719 -1.27933 0.223982]
+  # gradient of ifft(rfft) must be 1
+  @test gradient((x)->real(ifft(fft(x))[1]),x)[1][1] == 1.0+0.0im
+  @test gradient((x)->real(fft(ifft(x))[1]),x)[1][1] == 1.0+0.0im
+
+  # check ffts for individual dimensions
+  @test gradient((x)->sum(abs.(FFTW.fft(x))),x)[1] ≈ gradient((x)->sum(abs.(FFTW.fft(FFTW.fft(x,1),2))),x)[1]
```
MikeInnes: It would be a bit simpler to use `gradcheck` here instead. That will check against numerical gradients, which is handy for correctness.

src/lib/fft.jl
seb1000: thanks for the tip, I changed to dims... now

Project.toml
seb1000: fixed now via package manager

src/Zygote.jl
seb1000: I moved it to array.jl now

src/lib/fft.jl
seb1000: the exception for int was necessary because size(xs)[collect(1)] does not work, but I now changed to collect(size(xs))[dims] which works and seems more elegant to me and does not need different handling for ints

src/lib/fft.jl
seb1000: this comment was from an old version and I forgot to remove it, sorry for the confusion. It indeed works also with multiple dimensions. I removed the comment now

test/gradcheck.jl
seb1000: I added 4 gradchecks as well now. I think the old test still make sense since they check whether the FFTs along individual dims work and produce the same results as when doing all dimensions in one go. However if you think that the old tests are mainly bloating gradcheck.jl I can also ommit them

src/lib/array.jl
```diff
@@ -341,3 +342,53 @@ end
...
+@adjoint function FFTW.ifft(xs)
+    return FFTW.ifft(xs), function(Δ)
+        N = length(xs)
+        return (1/N* FFTW.fft(Δ),)
+    end
+end
+
+@adjoint function FFTW.fft(xs, dims)
+    return FFTW.fft(xs, dims), function(Δ)
+    # dims can be int, array or tuple,
```
MikeInnes: The indentation is off here, just needs a quick fix.

src/lib/array.jl
```diff
@@ -341,3 +342,53 @@ end
...
+@adjoint function FFTW.ifft(xs)
+  return FFTW.ifft(xs), function(Δ)
+      N = length(xs)
+      return (1/N* FFTW.fft(Δ),)
+  end
+end
+
+@adjoint function FFTW.fft(xs, dims)
+  return FFTW.fft(xs, dims), function(Δ)
+  # dims can be int, array or tuple,
```
jekbradbury: nit: indent these few lines?

seb1000: I (hopefully) fixed the indentation now.

MikeInnes:
Alright, great, this looks good.

There's a merge conflict due to the manifest; it'd be great if you could resolve that, and even better if you can use a newer Julia version to minimise the number of changes while you're at it. Other than that I think we're good to go.

seb1000:
OK. I am very new both to Julia and collaborating on github, but I will do my best to resolve the conflicts

seb1000:
I now switched to julia 1.1.1 and resolved the conflicts. However, for some reason my manifest has more entries than the current one in master (AbstractFFTs, Conda, JSON, Reexport, VersionParsing) in addition to FFTW that I added via the package manager. Should I remove them?

MikeInnes:
No, that's ok, they are just dependencies of FFTW that got pulled in.

Unfortunately the tests are now failing; looks like you just have a syntax error in the tests you added.

seb1000:
I now fixed the tests, and they pass on my local installation. I also fixed the testset for hcat which had failed as well, and some other fixes in other testsets (forgotten @test)

seb1000:
The travis CI has finished, but this does not seem to be reported to github (when I check https://travis-ci.org/FluxML/Zygote.jl/builds/543743653?utm_source=github_status&utm_medium=notification all tests have passed, but here it is still pending.) Is there any way to re-trigger the travis-ci build?

Project.toml
```diff
@@ -4,6 +4,7 @@ version = "0.3.1"
 
 [deps]
 DiffRules = "b552c78f-8df3-52c6-915a-8e097449b14b"
+FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
```
ChrisRackauckas: Can't you just depend on AbstractFFTs.jl and write the dispatches using the high level functions?

MikeInnes:
:+1: Thanks!

ChrisRackauckas:
But can't this be done with just AbstractFFTs?

MikeInnes:
Probably, but fine to do that as a follow up. The main issue is testing but we can have FFTW as a test dependency.

Issue #213, opened 2019-06-03
Lazy adjoint/transpose gradient fails


under-Peter:
MWE:
```
$ julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.1.0 (2019-01-21)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> using Zygote, LinearAlgebra

julia> Zygote.gradient(x -> x'[1], randn(2,2))
ERROR: MethodError: no method matching adjoint(::NamedTuple{(:parent,),Tuple{Array{Float64,2}}})
Closest candidates are:
  adjoint(::Missing) at missing.jl:83
  adjoint(::Number) at number.jl:193
  adjoint(::Adjoint{#s177,#s176} where #s176<:Union{StaticArray{Tuple{N},T,1} where T where N, StaticArray{Tuple{N,M},T,2} where T where M where N} where #s177) at /home/andreas/.julia/packages/StaticArrays/VyRz3/src/linalg.jl:78
  ...
Stacktrace:
[...]

julia> Zygote.gradient(x -> transpose(x)[1], randn(2,2))
ERROR: MethodError: no method matching transpose(::NamedTuple{(:parent,),Tuple{Array{Float64,2}}})
Closest candidates are:
  transpose(::Missing) at missing.jl:83
  transpose(::ForwardDiff.Dual{T,V,N} where N where V) where T at /home/andreas/.julia/packages/ForwardDiff/N0wMF/src/dual.jl:202
  transpose(::Number) at number.jl:192
  ...
Stacktrace:
[...]
```
With `Zygote v0.3.1`. 

Workarounds:
use  `(copy∘adjoint)`

MikeInnes:
[These gradients](https://github.com/FluxML/Zygote.jl/blob/master/src/lib/array.jl#L173) just need to be overloaded to handle the named-tuple case. If the pullback receives an array, transpose it, if a named tuple just return the `.parent`.

PR #216, opened 2019-06-05
Fix and add tests for adjoint/transpose

under-Peter:
Fixes adjoint and transpose as suggested here https://github.com/FluxML/Zygote.jl/issues/213#issuecomment-498183219

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -83,7 +83,7 @@ version = "0.10.3"
 
 [[IRTools]]
 deps = ["InteractiveUtils", "MacroTools", "Test"]
-git-tree-sha1 = "4d0c6647aa89b3678d03967ca11c8a5cdc83d2b5"
+git-tree-sha1 = "09b9d429fd57ad899d579dbe6d0f68efd82e9ba4"
 repo-rev = "master"
 repo-url = "https://github.com/MikeInnes/IRTools.jl.git"
 uuid = "7869d1d1-7146-5819-86e3-90919afe41df"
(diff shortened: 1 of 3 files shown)


Manifest.toml
```diff
@@ -83,7 +83,7 @@ version = "0.10.3"
 
 [[IRTools]]
 deps = ["InteractiveUtils", "MacroTools", "Test"]
-git-tree-sha1 = "09b9d429fd57ad899d579dbe6d0f68efd82e9ba4"
+git-tree-sha1 = "4d0c6647aa89b3678d03967ca11c8a5cdc83d2b5"
```
MikeInnes: It would be best not to combine this with an IRTools version change, unless it's strictly needed for tests to pass.

src/lib/array.jl
```diff
@@ -171,8 +171,12 @@ _backmean(xs, Δ, dims) = zero(xs) .+ Δ ./ mapreduce(i -> size(xs,i),*,dims)
   end
 end
 
-@adjoint transpose(x) = transpose(x), Δ -> (transpose(Δ),)
-@adjoint Base.adjoint(x) = x', Δ -> (Δ',)
+_transpose(x) = transpose(x)
+_transpose(x::NamedTuple{(:parent,),<:Any}) = x.parent
```
MikeInnes: This looks generally good, but I'd prefer `_transpose` to be written as an inner function `back`.

Manifest.toml
under-Peter: That must've happened by accident, I undid it.

src/lib/array.jl
under-Peter: moved backward-actions to inner functions in both `adjoint` and `transpose`

MikeInnes:
This commit was included (with attribution in git) in #223. Thanks @under-Peter!

Issue #81, opened 2019-02-28
Move over adjoints from Flux


MikeInnes:
Flux has a lot of gradient definitions that we need to port over to Zygote. Copying them over would be an easy patch for anyone interested in contributing.

DoktorMike:
Good idea. Could you showcase just one such definition port? Might make it easier for someone to jump in. :)

MikeInnes:
For example, [matmul in Tracker](https://github.com/FluxML/Tracker.jl/blob/ceba6f48fe79250feb0ecdeb999f4359186eaa3e/src/lib/array.jl#L375) and [matmul in Zygote](https://github.com/FluxML/Zygote.jl/blob/086859a4162a59fc7dcaad384f715a610a7ef641/src/lib/array.jl#L115).

In general the process is: copy it over to roughly the same place in `array.jl`, change `@grad` -> `@adjoint`, and remove any calls to `data`. Let me know if there's any other questions!

c0nn3r:
I'm going to give this a try, thanks for the example!

c0nn3r:
Sorry, just getting to this now. I've ported over the following functions (the copypaste job is done ;) )

- `view`
- `cat` (might need to be changed match how `vcat` and `hcat` are now implemented)
- `mean`
- `diagm`
- `dot`
- `det`
-  `logdet`
- `logabsdet`
- `inv`

Should I also port over the ones from NNLib.jl as well?

MikeInnes:
There aren't any custom adjoints in NNlib, but there are some in Flux for NNlib – maybe that's what you mean. But yeah, it'd be good to have them.

staticfloat:
Just a note that my branch `sf/nnlib_overhaul` addresses most of the NNlib adjoints, I think.

ToucheSir:
Is this issue still relevant? The only remaining `@adjoints` I see would be [xlogy](https://github.com/FluxML/Flux.jl/blob/v0.12.4/src/losses/utils.jl#L21) (which should probably belong in NNlib if it can be rewritten as an `rrule`), and the normalization functions mentioned in https://github.com/FluxML/NNlib.jl/issues/19.

DhairyaLGandhi:
Yeah we have basically done this one.

PR #219, opened 2019-06-06
Adjoint for view

### LINKED ISSUE #52: Function with views error when taking gradient
```jl
julia> function rosenbrock(x)
           a = one(eltype(x))
           b = 100 * a
           result = zero(eltype(x))
           x2 = @view x[2:end]
           x1 = @view x[1:end-1]
           z = @. (a - x1)^2 + b * (x2 - x1^2)^2
           return sum(z)
       end
rosenbrock (generic function with 1 method)

julia> x = rand(2);

julia> Zygote.gradient(rosenbrock, x)
ERROR: MethodError: no method matching (::Zygote.Jnew{SubArray{Float64,1,Array{Float64,1},Tuple{UnitRange{Int64}},true},Nothing})(::Array{Float64,1})
Closest candidates are:
  Jnew(::Union{Nothing, NamedTuple}) where {T, G} at /Users/kristoffer/.julia/packages/Zygote/Ohw1K/src/lib/lib.jl:175
Stacktrace:
 [1] (::getfield(Zygote, Symbol("##256#back#159")){Zygote.Jnew{SubArray{Float64,1,Array{Float64,1},Tuple{UnitRange{Int64}},true},Nothing}})(::Array{Float64,1}) at /Users/kristoffer/.julia/packages/Zygote/Ohw1K/src/lib/grad.jl:37
 [2] Type at ./subarray.jl:16 [inlined]
 [3] Type at ./subarray.jl:32 [inlined]
 [4] Type at ./subarray.jl:22 [inlined]
```

mcabbott:
Closes #52 (and removes one item from the list in #81).

I'm not sure that view takes keywords, but I just copied what Tracker has.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -31,6 +31,14 @@ end
 @adjoint! setindex!(xs::AbstractArray, x...) = setindex!(xs, x...),
   _ -> error("Mutating arrays is not supported")
 
+@adjoint function view(x::AbstractArray, inds...; kw...)
+  view(x, inds...; kw...), dy -> begin
+    dx = _zero(x)
+    copyto!(view(dx, inds...; kw...), dy)
+    (dx, map(_->nothing, inds)...)
+  end
+end
+
 # General
 
 @adjoint collect(x::Array) = collect(x), Δ -> (Δ,)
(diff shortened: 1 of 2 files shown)


MikeInnes:
:+1: Thanks!

PR #220, opened 2019-06-06
Adjoint for PermutedDimsArray

mcabbott:
This treats the lazy `PermutedDimsArray` exactly like `permutedims`. I also added a line for `permutedims(x)` without `dims`, although perhaps this is not necessary, as this already works:
```
julia> rr = rand(2,2);

julia> Zygote.gradient(x -> sum(sin, transpose(x)), rr)
([0.949851 0.970093; 0.842615 0.96526],)

julia> Zygote.gradient(x -> sum(sin, permutedims(x)), rr)
([0.949851 0.970093; 0.842615 0.96526],)

julia> Zygote.gradient(x -> sum(sin, PermutedDimsArray(x,(2,1))), rr)
ERROR: MethodError: no method matching (::Zygote.Jnew{PermutedDimsArray{Float64,2,(2, 1),(2, 1),Array{Float64,2}},Nothing,false})(::Array{Float64,2})
```

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -35,8 +35,13 @@ end
 
 @adjoint collect(x::Array) = collect(x), Δ -> (Δ,)
 
+@adjoint permutedims(xs) = permutedims(xs), Δ -> (permutedims(Δ),)
+
 @adjoint permutedims(xs, dims) = permutedims(xs, dims),
   Δ -> (permutedims(Δ, invperm(dims)), nothing)
+
+@adjoint PermutedDimsArray(xs, dims) = PermutedDimsArray(xs, dims),
+  Δ -> (PermutedDimsArray(Δ, invperm(dims)), nothing)
 
 @adjoint reshape(xs, dims...) = reshape(xs, dims...),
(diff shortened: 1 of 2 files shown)


src/lib/array.jl
```diff
@@ -10,10 +10,10 @@ using Base.Broadcast: broadcasted, broadcast_shape
 
 @adjoint Base.vect(xs...) = Base.vect(xs...), Δ -> (Δ...,)
 
-@adjoint copy(x::AbstractArray) = copy(x), ȳ -> (ȳ,)
+@adjoint copy(x::AbstractArray) = copy(x), ȳ -> (ȳ,)
```
MikeInnes: Not sure what's happening here, might be an editor settings thing. Either way it'd be great to remove these from the diff.

mcabbott: Sorry about that, I tried to resolve a one-line conflict with github's editor and didn't notice. I'm hopeless at git but will see if I can un-do this.

mcabbott: Well that was fun. I think now it's all cleaned up.

MikeInnes:
:+1: Thanks!

Issue #222, opened 2019-06-09
Wrong gradient when using backslash


baggepinnen:
The following example produces different results when using Zygote and ForwardDiff
```julia
using ForwardDiff, Zygote
r(a) = norm(a[1:end-1,:]\a[2:end,:])
Zygote.refresh()
a = randn(33,3)
norm(Zygote.gradient(r,a)[1] - ForwardDiff.gradient(r,a)) # ~= 0.2
```

willtebbutt:
Yup, this is definitely a thing. Working on a fix at the minute. In short, the current implementation assumes that matrix on the lhs is square, and that therefore the regular inverse is used, rather than the pseudo-inverse.

MikeInnes:
Is this fixed by #223?

DoktorMike:
This is fixed for me on master.

```julia
using ForwardDiff, Zygote, LinearAlgebra
r(a) = norm(a[1:end-1,:]\a[2:end,:])
Zygote.refresh()
a = randn(33,3)
norm(Zygote.gradient(r,a)[1] - ForwardDiff.gradient(r,a)) # ~= 2.0503073816403344e-16
```

PR #223, opened 2019-06-09
Fix ldiv / rdiv + implement pinv

willtebbutt:
Relates to #222

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -231,6 +231,22 @@ end
   return Y, Δ->(-Y' * Δ * Y' + (I - A * Y) * Δ' * Y * Y' + Y' * Y * Δ' * (I - Y * A),)
 end
 
+@adjoint function \(A::Union{Diagonal, AbstractTriangular}, B::AbstractVecOrMat)
+  Y = A \ B
+  return Y, function(Ȳ)
+    B̄ = A' \ Ȳ
+    return (-B̄ * Y', B̄)
+  end 
+end
+
+@adjoint function /(A::AbstractMatrix, B::Union{Diagonal, AbstractTriangular})
+  Y = A / B
+  return Y, function(Ȳ)
+    Ā = Ȳ / B'
+    return (Ā, -Y' * Ā)
+  end
+end
+
 @adjoint function \(A::AbstractMatrix, B::AbstractVecOrMat)
   Z = A \ B
   return Z, function(Z̄)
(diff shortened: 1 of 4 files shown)


src/lib/array.jl
```diff
@@ -204,18 +214,55 @@ end
...
   Y = A \ B
   return Y, function(Ȳ)
-      B̄ = A' \ Ȳ
-      return (-B̄ * Y', B̄)
+    B̄ = A' \ Ȳ
+    return (-B̄ * Y', B̄)
+  end 
+end
+
+@adjoint function /(A::AbstractMatrix, B::Union{Diagonal, AbstractTriangular})
```
ararslan: What about `/` for abstract matrices? Is that still covered and I'm missing it?

src/lib/array.jl
```diff
@@ -204,18 +214,55 @@ end
...
+    Ā = Ȳ / B'
+    return (Ā, -Y' * Ā)
+  end
+end
+
+@adjoint function \(A::AbstractMatrix, B::AbstractVecOrMat)
+  Z = A \ B
+  return Z, function(Z̄)
+    B̄ = A' \ Z̄
+    if size(A, 1) == size(A, 2)
```
ararslan: Don't we also need that `A` is nonsingular?

src/lib/array.jl
willtebbutt: Yup. `/` is implemented in terms of `\` for abstract matrices :)

src/lib/array.jl
willtebbutt: I just elected to follow the structure in base found [here](https://github.com/JuliaLang/julia/blob/5d02c59185ef31703891adc748e0ed414a9a5b3f/stdlib/LinearAlgebra/src/generic.jl#L956) precisely. I think base assumes that the user will ensure that `A` is well-conditioned.

MikeInnes:
LGTM. If you need the adjoint/transpose fixes can you rebase or pick the commits from #216, so we can include the credit?

willtebbutt:
> If you need the adjoint/transpose fixes can you rebase or pick the commits from #216, so we can include the credit?

Well that's uncanny. I hadn't realised #216 is a thing, and somehow we've managed to implement pretty much the same thing character-for-character... Have cherry-picked both commits (not done this before, so may have made a mess of it :grimacing:)

MikeInnes:
Funny, yeah they really do seem identical. Once tests pass I think this is good to go.

PR #226, opened 2019-06-10
FillArray broadcasting functionality

willtebbutt:
Functionality to ensure that broadcasting works properly with `AbstractFill`s.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -397,69 +397,8 @@ end
 # =======================
 
 @adjoint function broadcasted(op, r::AbstractFill{<:Real})
-    y, _back = Zygote.forward(op, getindex_value(r))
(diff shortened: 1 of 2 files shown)


MikeInnes:
Not that nice to have to list out all these case (though it may be necessary of course). Do things break without these, or is it just slow, or something else?

willtebbutt:
That's a fair point. Last time I tried to do things without them they just broke. Will see what happens if they're not there and get back to you.

willtebbutt:
Turns out that we don't need most of the definitions. One was required to ensure that the adjoint is a `Fill` when appropriate, but other than that it appears to be fine. So this PR now just looks like a large number of tests, which isn't a terrible thing I guess.

MikeInnes:
Cool. Yep, that looks good to me then.

PR #234, opened 2019-06-18
@nograd rand

baggepinnen:


baggepinnen:
Actually, `rand` has a lot of methods, some where you potentially would like to take gradients. How can the `@nograd` be limited to only `rand(), rand(::Int...)` etc.?

willtebbutt:
> Actually, rand has a lot of methods, some where you potentially would like to take gradients. How can the `@nograd` be limited to only rand(), rand(::Int...) etc.?

If you could please implement this, it would be very helpful. It's been on the wish-list for a while.

I'm not quite sure how `@nograd` is implemented, but you can already do
```julia
@adjoint rand() = rand(), _->(,)
```
(or something similar to that), so perhaps the macro could be extended using this mechanism?

baggepinnen:
How would the interface to such functionality look? Examples:
```julia
nograd(f::Function, types::Tuple)
```

```julia
@nograd fun (types...,)
```

```julia
@nograd fun(::T1, ::T2, ...)
```

MikeInnes:
I think `@nograd fun(::T1, ::T2, ...)` works nicely. We definitely need this for `rand` (we had it disabled entirely at some point, but that's buggy when you use things like Distributions).

ChrisRackauckas:
`randn` and `randexp` as well?

willtebbutt:
I guess the question is whether or not we expect there to be any methods of `randn` or `randexp` that need to propagate gradient info. If not, a blanket `@nograd` for all types is probably fine.

ChrisRackauckas:
FWIW, with Tracker and ForwardDiff we made it treat the randoms as constants and that's how we got differencing through the stochastic differential equation solvers working (then it's like differencing w.r.t. a chosen Brownian path). So for my applications it makes sense, it's whether other applications see it the same way.

baggepinnen:
Added `randexp, randperm, randcycle` as well. `randcycle` and `randperm` are unlikely to ever be differentiable.

MikeInnes:
We definitely can't `@nograd rand` since they can take differentiable arguments. But having an adjoint for a specific method, with the pullback `_ -> nothing`, is fine. Not sure about the others, but it's worth being careful.

CarloLucibello:
@baggepinnen 
On current master we have nothing grad for `rand`
https://github.com/FluxML/Zygote.jl/blob/0e1b9bbf2136ab2cf003b768ba1d2f7c4ce950d1/src/lib/array.jl#L13
and `@nograd` for `randn`.
So  I would  limit this PR to `@nograd randexp, randcycle, randperm`

CarloLucibello:
needs a rebase on master

Issue #228, opened 2019-06-12
forward gives domain error when back is called on Float 16.


thebhatman:
This simple code breaks as it gives domain error when back is called on Float16.
```
using Zygote

using Flux

y = rand(Float16, 2)

y1 = rand(Float16, 2)

fwd, back = Zygote.forward(Flux.mse, y1, y)

back(one(Float16))
```
This is the error message :  
```
ERROR: DomainError with -0.23730469:
log will only return a complex result if called with a complex argument. Try log(Complex(x)).
```
But the same piece of code works when I use Float32 or Float64 or Int64. The problem is only with Float16.

MikeInnes:
Can you try to narrow this down a bit? At a guess, it's probably due to the `^` function.

thebhatman:
I just looked up the definitions of the log function in math.jl. I think the error is arising because log is only specifically defined for Float64 and Float32. The reason is that the StackTrace of the the error I get when calling back on Float16 is this : 
```
Stacktrace:
 [1] throw_complex_domainerror(::Symbol, ::Float32) at .\math.jl:31
 [2] log(::Float32) at .\special\log.jl:321
 [3] log at .\math.jl:1018 [inlined]
 [4] (::getfield(Zygote, Symbol("##584#585")){Float16,Int64})(::Float16) at C:\Users\manju\.julia\packages\Zygote\VeaFW\src\lib\number.jl:27
 [5] #1580#back at C:\Users\manju\.julia\packages\Zygote\VeaFW\src\lib\grad.jl:46 [inlined]
```

thebhatman:
Yeah I observed that the error is mainly when `^` has negative input.  The same error can be reproduced in this manner : 
```
using Zygote

f(x) = x .^ 2

fwd, back = Zygote.forward(f, Float16(-2))

back(one(Float16))
```
Even this throws the domain error when called on Float16, but works perfectly for Float32 and Float64.

MikeInnes:
Our rule for `^` comes from diffrules here: https://github.com/JuliaDiff/DiffRules.jl/blob/a9b00c81ac0d103888b124b2b9f58a7c815cebf9/src/rules.jl#L87

Would be great if you can show the Float16/Float32 difference outside of Zygote.

thebhatman:
[...]

PR #245, opened 2019-06-24
Adjoint for literal_pow to support Float16

thebhatman:
Gradient on `literal_pow` was throwing an error with negative base, when input was Float16 as discussed in this issue https://github.com/FluxML/Zygote.jl/issues/228.  So the adjoint has been overwritten. Now the gradient works on all Floats, without breaking.

### DIFF
--- a/src/lib/number.jl
+++ b/src/lib/number.jl
@@ -28,7 +28,7 @@ for (M, f, arity) in DiffRules.diffrules()
   end
 end
 
-@adjoint Base.literal_pow(::typeof(^), x::Real, ::Val{p}) where {p} = x^p, Δ -> (nothing, p * x ^ (p - 1), nothing)
+@adjoint Base.literal_pow(::typeof(^), x::Float16, ::Val{p}) where {p} = x^p, Δ -> (nothing, p * x ^ (p - 1), nothing)
 @adjoint Base.convert(T::Type{<:Real}, x::Real) = convert(T, x), ȳ -> (nothing, ȳ)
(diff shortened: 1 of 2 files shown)


test/gradcheck.jl
```diff
@@ -152,6 +152,15 @@ end
   @test gradtest(x -> minimum(x, dims=[1, 2]), rand(2, 3, 4))
 end
 
+@testset "literal_pow" begin
+  f(x) = x .^ 2
+  h(x) = 2 .^ x
+  @test gradient(f, Float16(-2))[1] == -4
+  @test gradient(f, Float32(-2))[1] == -4
+  @test gradient(h, Float16(2))[1] == 2.772588722239781
+  @test gradient(h, Float32(2))[1] == 2.772588722239781
```
MikeInnes: Can we write this number more compactly? I assume it's log of something, or similar.

src/lib/number.jl
```diff
@@ -28,6 +28,7 @@ for (M, f, arity) in DiffRules.diffrules()
   end
 end
 
+@adjoint Base.literal_pow(::typeof(^), x::Float16, ::Val{p}) where {p} = x^p, Δ -> (nothing, p * x ^ (p - 1), nothing)
```
MikeInnes: If you're going to add an explicit adjoint for `literal_pow`, which wasn't there before, add it for all `Real`.

(I was imagining that we'd just add the forward definition `literal_pow(::Float16) = x*x` and let Zygote differentiate it, adding the adjoint works well too).

CarloLucibello:
I think this has been superseded by #512, feel free to reopen if that's not the case

PR #248, opened 2019-06-25
Fix ldiv for adjoint of triangular

willtebbutt:
Makes sure that if you take the `adjoint` or `transpose` of a triangular matrix, the appropriate adjoint rule is used.


MikeInnes:
It might be worth a note in the code on why we have a separate definition here. Presumably the `AbstractMatrix` fallback should be correct even if the specific method is more optimised (the word "fix" in the title is making me a bit suspicious there, but maybe you didn't mean correctness).

willtebbutt:
@MikeInnes happy for me to merge this?

MikeInnes:
Sure thing :+1:

PR #251, opened 2019-06-26
Fix adjoint for *(::Adjoint, ::Vector)

Keno:
By design, this returns a scalar, tripping the type limitation on
the method for AbstractMatrix.


MikeInnes:
Thanks! I assume this is fine but would like a thumbs up from @willtebbutt.

willtebbutt:
LGTM

Issue #163, opened 2019-04-20
Support sparse arrays/matrices as output?


kbarros:
Consider the code:
```
using SparseArrays
n = 5
r = rand(n, n)
h = sprandn(rng, n, n, 0.1)
gradient(h -> tr(r * h), h)[1]
```

Currently this returns a dense `Array`. Is it possible to have it return an output with same same sparsity structure as the input `h`?

It seems that the gradient with respect to a sparse matrix should generally be a sparse matrix.

Note that users could always revert to the current behavior  by explicitly promoting `h` to a dense matrix,

```gradient(h -> tr(r * h), Array(h))[1]```

**Comment by Roger-luo** [comment]
Yes, I think we just need an adjoint for `*` for sparse matrix, that returns a sparse matrix.

**Comment by Roger-luo** [comment]
I think this is because the output of `r * h` is dense thus as the adjoint `*` defines

```julia
@adjoint function(a::AbstractVecOrMat * b::AbstractVecOrMat)
  return a * b, function(Δ)
    return (reshape(Δ * b', size(a)), reshape(a' * Δ, size(b)))
  end
end
```

This returns a dense matrix since `Δ` (dense) multiplies sparse is dense, which make sense

kbarros:
Yes, I'm looking for something like:
```
@adjoint function(a::SparseMatrixCSC * b::AbstractVecOrMat)
  return a * b, function(Δ)
    return (
      # dense-dense matrix multiplication with sparse output!
      mat_mul_with_sparse_output_in_shape_of(Δ, b', a),
      # same as before
      reshape(a' * Δ, size(b)))
  end
end
```
The new thing here is `mat_mul_with_sparse_output_in_shape_of()`. It will calculate matrix elements of `Δ * b'` only if those are present in the sparsity structure of `a`. This is the dual operation to `a * b'`.

The motivation would be efficiency. I'm trying to explore whether Zygote would be able to simplify  calculations of the type in this method https://arxiv.org/abs/1711.10570 (see Appendix B for the "reverse-mode differentiation" transformation that I performed manually).

kbarros:
> I think this is because the output of `r * h` is dense thus as the adjoint `*` defines
> 
> ```julia
> @adjoint function(a::AbstractVecOrMat * b::AbstractVecOrMat)
>   return a * b, function(Δ)
>     return (reshape(Δ * b', size(a)), reshape(a' * Δ, size(b)))
>   end
> end
> ```
> 
[...]

PR #89, opened 2019-03-06
Enable tighter inference bounds for `_zero()`

staticfloat:
@MikeInnes Is this about what you envisioned?  Unfortunately I'm not quite sure how to test this, as `@test_inferred` fails on the test case given in your last PR.

PR #256, opened 2019-07-10
Non-scalar getindex

mcabbott:
This should fix the bug described in this [discourse question](https://discourse.julialang.org/t/flux-unable-to-differentiate-an-embedding-layer/26067), a more compact version of which is 
```
using Zygote
ii = rand(1:10, 2)
jj = rand(1:10, 2,2)

Zygote.gradient(x -> sum(x[ii,:]), rand(10,3))
Zygote.gradient(x -> sum(x[jj,:]), rand(10,3)) # DimensionMismatch("tried to assign 2×2×3 array to 4×3 destination")

Zygote.gradient(x -> sum(@view x[jj,:]), rand(10,3)) # fine!
```
I've just copied the gradient from `view`, where `copyto!` does the necessary reshaping.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -22,9 +22,10 @@ using Base.Broadcast: broadcasted, broadcast_shape
 @adjoint (::Type{T})(sz) where {T<:Zeros} = Zeros(sz), Δ->(nothing,)
 @adjoint (::Type{T})(sz) where {T<:Ones} = Ones(sz), Δ->(nothing,)
 
-_zero(xs::AbstractArray{<:Number}, T=float(eltype(xs))) = fill!(similar(xs, T), false)
+_zero(xs::AbstractArray{<:Number}, T=float(eltype(xs))) =
+    fill!(similar(xs, T, size(xs)), false)
 _zero(xs::AbstractArray, T=Any) = Union{Nothing, T}[nothing for x in xs]
-_wrap(dx, x) = dx
+_makelike(dx, x) = dx
 
 @adjoint getindex(x::AbstractArray, inds...) = x[inds...], ∇getindex(x, inds)
 
@@ -38,7 +39,7 @@ _wrap(dx, x) = dx
     dx = _zero(x, eltype(dy))
     @views dx[inds...] .+= dy
   end
-  (_wrap(dx, x), map(_->nothing, inds)...)
+  (_makelike(dx, x), map(_->nothing, inds)...)
 end
 
 @adjoint! setindex!(xs::AbstractArray, x...) = setindex!(xs, x...),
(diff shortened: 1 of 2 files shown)


MikeInnes:
Looks like a test is failing.

mcabbott:
Indeed, sorry, this may need far more care. This is the error, [link](https://github.com/FluxML/Zygote.jl/blob/master/test/features.jl#L249), it's not immediately obvious to me how I've broken it: 
```
[ Info: Testing Zygote in normal mode.
Features: Error During Test at /home/travis/build/FluxML/Zygote.jl/test/features.jl:249
  Test threw exception
  Expression: gradient((x->begin
                (([x])[1]).a
            end), Foo(1, 1)) == ((a = 1, b = nothing),)
  ArgumentError: destination has fewer elements than required
```

BTW this is the error I get locally, after saying `] free Zygote`, which I presume is not my fault:
```
[ Info: Testing Zygote in normal mode.
map: Error During Test at /Users/me/.julia/packages/Zygote/fuj2C/test/gradcheck.jl:125
  Test threw exception
  Expression: gradient((v->begin
                sum([x for x = v])
            end), [1.1, 2.2, 3.3]) == ([1, 1, 1],)
  Mutating arrays is not supported
```

mcabbott:
Maybe this fixes it. I stole #117's trick of checking for integer indices. And I added a test directly from https://github.com/invenia/Nabla.jl/issues/139 too.

mcabbott:
This should now address #376, by explicitly promoting in `_zero`. And borrows the idea of #89 too.

I don’t think the test failure on 1.0 is my doing. And I’m not sure how I messed up the merge to cause a huge diff, sorry.

mcabbott:
Latest commit is an attempt to address #402 but is not right yet!

It also has the undesirable side effect of replacing the first with the second of these:
```julia
julia> fill!(similar(@SVector([1,2]), Float64,), 0)
2-element MArray{Tuple{2},Float64,1,2} with indices SOneTo(2):
 0.0
 0.0

julia> fill!(similar(@SVector([1,2]), Float64, (2,)), 0)
2-element Array{Float64,1}:
 0.0
 0.0
```

mcabbott:
This could use a review, @MikeInnes, if you have a minute. It closes lots of things. 

(It probably wants a commit adding co-authors, as it borrows from many other PRs.)

MikeInnes:
Thanks a lot for this, it looks generally solid to me.

I guess the idea of `_project` is to convert e.g. gradients of Diagonals back to Diagonals, etc. If so I think we should avoid that for now, since the discussion on #163 isn't totally settled, and if we do do this we'd should do it consistently everywhere.

I'm also curious why the `_zeros` overloads are needed, doesn't `similar` do the right thing in those cases? In general it's a little concerning to need a lot of special cases for things like this, since it suggests custom arrays might not work with Zygote by default.

mcabbott:
The idea for indexing `Diagonal` & `Symmetric`. is to make a matrix, write into it, and then throw away any parts which ought to be zero. Indeed this treats the off-diagonals as structurally zero. The overload of _zero is because `similar(Diagonal(rand(3)))` returns something which can’t be written into at `[1,2]`. 

Older versions only had a gradient for `getindex(::Array,...`, in which case indexing passed through directly to the parent. This also treats Diagonal as structurally zero off-diag. (Some examples here  https://github.com/FluxML/Zygote.jl/issues/402#issuecomment-557604195). I am not sure whether there are other problems with reviving that approach, haven’t tried.

For now, should I drop the last 2 commits, leave Diagonal etc. for later, but solve the other problems?

MikeInnes:
Yeah, I think it might be best to solve the stuff we know needs fixing right now. We should definitely have some more discussion of the other things but they might be trickier to work out carefully.

mcabbott:
Done! Closes #272 and #376. Replaces #89 and #117. Does not fix #402.

MikeInnes:
Thanks!

bors r+

PR #257, opened 2019-07-12
forwarddiff(f, xs...)

mcabbott:
This should make `Zygote.forwarddiff(f,x,y)` work, by simply evaluating `f` twice. 

Test failure is Unexpected Pass `Expression: (((sin')')')(1.0) == -(cos(1.0))` at test/features.jl:155. I don't think that's my doing?

### DIFF
--- a/src/lib/forward.jl
+++ b/src/lib/forward.jl
@@ -47,10 +47,9 @@ reshape_scalar(x, y) = reshape(y, size(x))
 reshape_scalar(x::Real, y) = y[]
 
 """
-    forwarddiff(f, x) -> f(x)
     forwarddiff(f, xs...) -> f(xs...)
 
-Runs `f(x)` as usual, but instructs Zygote to differentiate `f` using pullback
+Runs `f(xs...)` as usual, but instructs Zygote to differentiate `f` using forward
 mode, rather than the usual reverse mode.
 
 Forward mode takes time linear in `length(x)` but only has constant memory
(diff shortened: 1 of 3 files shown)


MikeInnes:
Does this have any performance disadvantage in the single-arg case?

mcabbott:
I did not test for that, it’s possible that `f(xs[1:i-1]..., x, xs[i+1:end]…)` ought to be some `ntuple` thing to infer better? Will take a look when I get a moment.

mcabbott:
I had a look, and there are small differences... not sure exactly what to test, but the example below becomes `10.691 μs (59 allocations: 4.00 KiB)` when using the multi-arg case:
```
julia> using Zygote, ForwardDiff

julia> @btime Zygote.gradient(x -> Zygote.forwarddiff(y -> sum(y .+ 1)^2, x), ones(10))
  9.277 μs (48 allocations: 3.70 KiB)
([40.0, 40.0, 40.0, 40.0, 40.0, 40.0, 40.0, 40.0, 40.0, 40.0],)

julia> @btime ForwardDiff.gradient(y -> sum(y .+ 1)^2, ones(10))
  571.478 ns (5 allocations: 3.08 KiB)
10-element Array{Float64,1}:
 40.0
 40.0
```

I messed around a bit with things like `ifun(::Val{i}) where {i} = x -> f(ntuple(d -> xs[d], i-1)..., x, ntuple(d -> xs[d+i], length(xs)-i)...)` but could only make it slower.

PR #259, opened 2019-07-15
fixed incorrect value in docs and demonstrated the use of Zygote.literal_getproperty

kraftpunk97-zz:
This PR makes changes in the docs by fixing an incorrect output in one of the code examples and brings to light the `Zygote.literal_getproperty` function which I believe, from experience, will come in handy when writing code normally and that the reader should be informed about it.

### DIFF
--- a/docs/src/adjoints.md
+++ b/docs/src/adjoints.md
@@ -99,7 +99,7 @@ dist(p::Point) = sqrt(width(p)^2 + height(p)^2)
 
 ```julia
 julia> gradient(a -> dist(a), Point(1, 2))[1]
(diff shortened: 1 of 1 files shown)


CarloLucibello:
worth checking if this is still needed

ToucheSir:
The incorrect value was fixed at some point. We probably shouldn't expose `literal_getproperty` since it's an internal function and could theoretically go away at any moment if e.g. constprop gets good enough. Formatting changes like replacing `ȳ` with `ȳ` are a good idea but would be better served by a separate PR because this one doesn't even cover all of them in one file.

PR #262, opened 2019-07-16
@nograd for integer division

thebhatman:


MikeInnes:
It would be better to have this in `number.jl`.

MikeInnes:
Thanks!

PR #264, opened 2019-07-18
Added adjoint for dropdims

thebhatman:
I have added the adjoint for dropdims, which puts back the dropped singleton dimensions.


MikeInnes:
Can you add a gradient check? Otherwise looks good.

thebhatman:
Done. I have added the tests.

PR #268, opened 2019-07-21
Wct/fix cholesky again

willtebbutt:
So my last cholesky-related PR #230 was horribly misguided and changed a perfectly good implementation to do totally the wrong thing 🍾 .

The reason for this screw up stems from the fact that checking the adjoint for cholesky is actually slightly tricky -- since the argument to the cholesky must be a (strictly) positive definite matrix, you can't just naively apply finite differencing. One strategy is therefore to check the composition of `cholesky` and some transformation that renders an arbitrary matrix positive definite, hence the tests for the function
```julia
A->cholesky(A * A' + I).U
```
that you'll find in the test suite. Unfortunately, this can actually mask some bugs, because we're not testing the `cholesky` factorisation directly. As it turns out the vjp of this function is entirely unaffected by this PR, despite the changes to the `adjoint`, but the vjp of
```julia
A->cholesky(A).U
```
is somewhat different. To ensure that this error doesn't creep back in I've added tests for the function
```julia
B->cholesky(Symmetric(B)).U
```
which pass provided that `B` is positive definition enough (i.e. it's eigenvalues are sufficiently positive) that when one perturbs `B` when finite differencing the symmetrised-perturbed matrix remains positive definite. This new function _is_ sensitive to the changes in this PR, so should prevent accidental reversion.

Again, apologies for screwing this up before, and thanks to @mohamed82008 for picking up on this in Turing's tests.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -354,8 +354,12 @@ end
     Σ̄ = copytri!(Σ̄, 'U')
     Σ̄ = ldiv!(U, Σ̄)
     BLAS.trsm!('R', 'U', 'T', 'N', one(eltype(Σ)), U.data, Σ̄)
-    Σ̄ ./= 2
-    return (Σ̄,)
+    @inbounds for n in diagind(Σ̄)
+      Σ̄[n] /= 2
+    end
+    return (UpperTriangular(Σ̄),)
+    # Σ̄ ./= 2
+    # return (Σ̄,)
   end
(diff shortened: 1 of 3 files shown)


willtebbutt:
Note that this should fix [this Turing issue](https://github.com/TuringLang/Turing.jl/issues/827), which was what brought this to my attention in the first place.

PR #270, opened 2019-07-24
AbstractFFTs, rfft, plans, and a couple of tests

dsweber2:
I'm looking to use fft to make Fourier domain convolution. I added adjoints for `rfft`, `irfft`, and while I was at it added one for `bfft` and `brfft`.  `rfft` goes from reals to complex numbers, so  require some care on the user's part. Specifically, the array being multiplied by the adjoint should be conjugate symmetric, otherwise the imaginary part of the `irfft` will be dropped, leading to an incorrect derivative. I was tempted to add something that would enforce the Fourier domain being conjugate symmetric, but figured it would be more hassle than it's worth, since anyone who's bothering to use `rfft` in the first place should probably know they'd better be doing real operations (e.g. convolution). As it is, it's something that will need to be mentioned in documentation, whenever/however that should happen.

While I was at it I switched it over to `AbstractFFTs`; I was able to get gradient running on both normal arrays and cu arrays, so I'm fairly sure this worked, though it could definitely use more testing.

Given that we'd expect the same fft gradients to get used again and again, I've put together something that works for forward plans (`plan_fft` and `plan_rfft`) with both `*` and `\`. I'm not sure how to implement the backwards plans without writing one for every plan in `FFTW` and `CuArrays`, though realistically those would mostly be for completeness sake anyways.

I added some tests that are a bit more explicit about what the expected gradients are, and for a couple of sizes, and compared the plan versions to the explicit versions. Could probably use higher and lower dimensional examples, as I mostly left that section.


CarloLucibello:
Is FFTW.jl still needed?

dsweber2:
Looks like no, assuming the user has it installed in whatever environment they're running the tests from. I've removed it. I'm still a bit sketchy on how exactly Manifest/Project are supposed to work.

Project.toml
```diff
@@ -3,8 +3,8 @@ uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
 version = "0.3.2"
 
 [deps]
+AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
 DiffRules = "b552c78f-8df3-52c6-915a-8e097449b14b"
-FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
```
mcabbott: I think if you include FFTW under [extras] instead, then it should be available only when testing. If that doesn't work then try adding also a section like this at the end:  
```
[targets]
test = ["Test", "Distances", "StatsFuns", "FFTW"]
```

CarloLucibello:
@dsweber2 as discussed in #553 it would be better to move over to AbstractFFT and have the FFTW dependency only for testing. Would you be willing to rebase and complete this PR?

dsweber2:
Glad we're pushing forward on this. Not sure why the tests aren't passing; when I got tired of living on a dev branch, I made a standalone package https://github.com/dsweber2/ZygoteFFTs.jl that contains just the fft adjoints, which does pass its tests.

src/lib/array.jl
```diff
@@ -774,54 +774,163 @@ end
...
+end
+
+@adjoint function ifftshift(x)
+    return ifftshift(x), function(Δ)
+        return (fftshift(Δ),)
+    end
+end
+
+
+# to actually use rfft, one needs to insure that everything that happens in the Fourier domain could've been done in the space domain with real numbers. This means enforcing conjugate symmetry along all transformed dimensions besides the first. Otherwise this is going to result in *very* weird behavior.
```
CarloLucibello: this comment should be broken in a few lines

src/lib/array.jl
```diff
@@ -774,54 +774,163 @@ end
...
+
+
+
+
+
+
+
+
+
+
```
CarloLucibello: no need for these whitelines

dsweber2:
alright, formatting changes made.

CarloLucibello:
nice, thanks!

bors r+

CarloLucibello:
@dsweber2 could your rebase? (hopefully for the last time)

PR #285, opened 2019-08-01
getindex(tuple, range)

### LINKED ISSUE #263: Gradient error when trying to use getindex on size of array
Here's a failing example. It throws an error saying mutating arrays is not supported : 
```
using Zygote
f(x) = prod(size(x)[1:end - 2])
x = rand(10, 10, 10, 10)
g = Zygote.gradient(()->f(x))
```
But if I try getindex on an array directly like this, it works : 
```
using Zygote
f(x) = prod(x[1:end - 2])
g = Zygote.gradient(()->f(x))
``` This doesn't throw an error.

mcabbott:
Fixes #263, closes #280. 

Changes what error you get from [this example](https://github.com/FluxML/Flux.jl/issues/820#issuecomment-515703945) in Flux#820. 

If I understand right `Zygote.literal_getindex` is only called when the index is a literal integer, so this is only for normal `getindex`.


MikeInnes:
:+1: Thanks!

PR #301, opened 2019-08-15
fix #300

### LINKED ISSUE #300: gradient gives nothing when parameters are inside a tuple of struct
MWE:

```jl
using Zygote

struct Linear{T}
    ps::Tuple{Matrix{T}, Vector{T}}
end

(l::Linear)(x) = l.ps[1] * x .+ l.ps[2]

l = Linear((rand(10, 10), rand(10)))

x = rand(10)
gs = gradient(()->sum(l(x)), Params([l.ps...]))

gs[l.ps[1]] # nothing
```

Roger-luo:


Roger-luo:
I'm not sure where to add a test for this tho.

Roger-luo:
@MikeInnes any chance to review it? I think the CI fails because some other reason.

MikeInnes:
Thanks a lot for fixing this! It'd be great if you could tweak it to apply `accum_param` when using `getproperty` on a tuple, though, analogous to how we do this for structs. (This patch makes it looks like the tuple itself is the parameter being accumulated, which gets the job done here but could fail in other cases).

Roger-luo:
Like this? @MikeInnes

cossio:
Merge?

MikeInnes:
Thanks!

PR #307, opened 2019-08-19
Pin IRTools version

mohamed82008:
The latest IRTools causes errors when loading `Zygote`. This PR tries to fix this.


MikeInnes:
Thanks for the effort -- but it would be best to add this to the `compat` section for the current version in the registry. In the mean time I'll just make the quick patch so this isn't an issue.

MikeInnes:
Thanks for sending this – Zygote is now up to date with IRTools master.

Issue #303, opened 2019-08-18
Support for tuple and function argument destructuring


oschulz:
Named tuples are really useful to pass around structured parameter sets, etc. Currently, this works:

```julia
using Zygote

function foo(x::NamedTuple{(:a, :b)})
    a, b = x.a, x.b
    a + b
end

foo((a = 1, b = 2))
gradient(foo, (a = 1, b = 2))
```

but using tuple de-structuring

```julia
function bar(x::NamedTuple{(:a, :b)})
    a, b = x
    a + b
end

bar((a = 1, b = 2))
gradient(bar, (a = 1, b = 2))
```

or function argument de-structuring (*so* convenient when passing structured parameters, esp. with @tpapp's `EponymTuples.@eponymargs`)

```julia
function baz((a, b)::NamedTuple{(:a, :b)})
    a + b
end

baz((a = 1, b = 2))
gradient(baz, (a = 1, b = 2))
```

results in 

```
ERROR: Non-differentiable function getfield
```

Would it be possible to support tuple and function argument destructuring?

unrealwill:
Hello, 

You can try defining the adjoint in your code.

The following works for the simple cases, (not tested with nested tuples, or performance). 

```
using Zygote:@adjoint

@adjoint function getfield( t::NamedTuple{N,T}, d ) where {N,T}
return getfield(t,d), Δ -> ( NamedTuple{N,T}(Tuple( i==d ? Δ : zero( typeof(t[i])) for i in 1:length(N) ) ) ,nothing)
end
```

```
function baz( (a, b) )
    a + b*b
end
baz((a = 1.0, b = 3.0))
g = gradient(baz, (a = 1.0, b = 3.0))
println( "g : ")
println( g )
```

> g : 
> ((a = 1.0, b = 6.0),)

oschulz:
Thanks, @unrealwill ! This would kinda lead to type piracy though, if I do it in a  package, right? Could Zygote include a generic version of this at some point?

unrealwill:
@oschulz I share the point of view that the place for this kind of code is inside the Zygote library.
 
Additionally, there are probably edge cases to take into consideration. 
For example the case where the gradient is independent of the parameter. In my above code, I returned zero. But it's probably not consistent with the rest of Zygote, which returns a nothing instead of a zero. I'm not quite sure about the convention.

willtebbutt:
> Could Zygote include a generic version of this at some point?

[...]

PR #308, opened 2019-08-21
NamedTuple destructuring case #303

unrealwill:
Allow the gradient to be computed properly when destructuring NamedTuple

Tested, and type-stable checked with @code_warntype

See case #303 for more info


MikeInnes:
Good to have this fix, but can you mention why this didn't work previously, and why we need a custom `getfield` to fix it? Also, does this still work for tuple deconstruction?

PR #310, opened 2019-08-21
add adjoint for var and std

chengchingwen:


src/lib/array.jl
```diff
@@ -201,6 +201,21 @@ end
 _backmean(xs, Δ, ::Colon) = zero(xs) .+ Δ ./ length(xs)
 _backmean(xs, Δ, dims) = zero(xs) .+ Δ ./ mapreduce(i -> size(xs,i),*,dims)
 
+@adjoint function Statistics.var(xs::AbstractArray; corrected::Bool=true, mean=nothing, dims=:)
+    return Statistics.var(xs; corrected=corrected, mean=mean, dims=dims), Δ -> _backvar(xs, Δ, corrected, mean, dims)
```
MikeInnes: I'm curious why you haven't done `mean = mean(xs)` here. It seems like that'd save some duplicate work.

chengchingwen: You are right! I forgot I can do that

chengchingwen: updated

chengchingwen:
Bump

MikeInnes:
Thanks!

PR #315, opened 2019-08-28
correct forward of push! pop! etc

oxinabox:


vchuravy:
Would probably be good to add a test.

ToucheSir:
This was addressed by another PR at some point, see https://github.com/FluxML/Zygote.jl/blob/60f53e709d8b5bc052a20fb4fcf0228004aa4723/src/lib/array.jl#L77 :)

PR #328, opened 2019-09-13
Explain relationship of Zygote's complex gradients with the Wirtinger calculus

simeonschaub:
I was confused quite a bit at first about how Zygote handles complex gradients, and this wasn't quite as obvious from the documentation. This also came up in the development of `ChainRules.jl`.


Krastanov:
The history of the current phrasing is explained here (as a reference when considering this PR) https://github.com/FluxML/Zygote.jl/issues/29

simeonschaub:
@MikeInnes Would you mind reviewing this?

docs/src/adjoints.md
```diff
@@ -56,7 +56,7 @@ julia> mygradient(sin, 0.5)
...
+If ``x`` and ``y`` are vectors, ``\frac{\partial y}{\partial x}`` becomes a Jacobian. Importantly, because we are implementing reverse mode we actually left-multiply the Jacobian, i.e. `v'J`, rather than the more usual `J*v`. Transposing `v` to a row vector and back `(v'J)'` is equivalent to `J'v` so our gradient rules actually implement the *adjoint* of the Jacobian. This is relevant even for scalar code: the adjoint for `y = sin(x)` is `x̄ = ȳ*cos(x)'`; the conjugation is usually moot but gives the correct behaviour for complex code, if `y(x)` is holomorphic. "Pullbacks" are therefore sometimes called "vector-Jacobian products" (VJPs), and we refer to the reverse mode rules themselves as "adjoints".
```
MikeInnes: Is there an intentional change here? If not, best to remove it from the diff.

docs/src/complex.md
```diff
@@ -2,7 +2,7 @@
 
 Complex numbers add some difficulty to the idea of a "gradient". To talk about `gradient(f, x)` here we need to talk a bit more about `f`.
 
-If `f` returns a real number, things are fairly straightforward. For ``c = x + yi`` and  ``z = f(c)``, we can define the adjoint ``\bar c = \frac{\partial z}{\partial x} + \frac{\partial z}{\partial y}i = \bar x + \bar y i`` (note that ``\bar c`` means gradient, and ``c'`` means conjugate). It's exactly as if the complex number were just a pair of reals `(re, im)`. This works out of the box.
+If `f` returns a real number, things are fairly straightforward. For ``c = x + yi`` and  ``z = f(c)``, we can define the adjoint ``\bar c = \frac{\partial z}{\partial x} + \frac{\partial z}{\partial y}i = \bar x + \bar y i`` (note that ``\bar c`` means gradient, and ``c^*`` means conjugate). It's exactly as if the complex number were just a pair of reals `(re, im)`. This works out of the box.
```
MikeInnes: We should stick with `c'` consistently here, because that's what Julia uses (and it's used in a bunch of other places on this page). If we're using `c*` elsewhere we can change that.

docs/src/adjoints.md
simeonschaub: Yes, I think GitHub just can't handle such long lines. See [here](https://www.diffchecker.com/Gs0z9iqs).

docs/src/complex.md
simeonschaub: It just felt a little bit weird to me, to use `\overline` for derivatives and `'` for the complex conjugate in mathematical notation. `^*` very commonly refers to just the complex conjugate in physics, so it felt more natural here and less confusing. Of course, I left it in snippets that are actual Julia code.

docs/src/adjoints.md
MikeInnes: The original wording was a bit unclear, but this gives the correct result for complex code _in general_, not just the holomorphic case (it is redundant for real code). The adjoint is causes us to get back complex sensitivities (otherwise the output would be the conjugate of the sensitivity).

simeonschaub: It is still unclear to me, what you would be conjugating in the non-holomorphic case, since the complex derivative doesn't exist in this case.

docs/src/complex.md
```diff
@@ -56,3 +56,40 @@ julia> wirtinger(x -> 3x^2 + 2x + 1, 1+2im)
 julia> wirtinger(x -> abs2(x), 1+2im)
 (1.0 - 2.0im, 1.0 + 2.0im)
 ```
+
+The gradient definition Zygote uses can also be expressed in terms of the [Wirtinger calculus](https://en.wikipedia.org/wiki/Wirtinger_derivatives) using the operators ``\frac{\partial}{\partial z}`` and ``\frac{\partial}{\partial z^*}``:
+
+```math
+f: \mathbb{C} \rightarrow \mathbb{R}, \qquad
```
MikeInnes: Don't you mean C->C here? The below references `Re(f)` and `f*`.

docs/src/complex.md
```diff
@@ -56,3 +56,40 @@ julia> wirtinger(x -> 3x^2 + 2x + 1, 1+2im)
...
+    &= \frac{\partial (f \circ w)}{\partial z^*} + \left(\frac{\partial (f \circ w)}{\partial z}\right)^{\!*}
+    = \frac{\partial f}{\partial w} \frac{\partial w}{\partial z^*}
+        + \frac{\partial f}{\partial w^*} \frac{\partial w^*}{\partial z^*}
+        + \left( \frac{\partial f}{\partial w} \frac{\partial w}{\partial z}
+        + \frac{\partial f}{\partial w^*} \frac{\partial w^*}{\partial z} \right)^{\!*} \\
+    &= \left[ \frac{\partial f}{\partial w^*} + \left( \frac{\partial f}{\partial w} \right)^{\!*} \right]
+        \left( \frac{\partial w}{\partial z} \right)^{\!*}
+    = \overline{f} \cdot \left( \frac{\partial w}{\partial z} \right)^{\!*}
+    \qquad \text{if $w(z)$ holomorphic} \Leftrightarrow \frac{\partial w}{\partial z^*} = 0
+\end{align*}
```
MikeInnes: I basically see what the above section is saying (though it might be nice to have an implementation in code for the sake of precision/clarity). This section strikes me as confusing though; a big equation dump there and while I'm sure it's all correct, I'm not sure what it's trying to get across.

docs/src/complex.md
simeonschaub: It is C->R, but since `f` is real, I'm using `f = Re(f) = f*` here, to express this in terms of the Wirtinger derivatives

docs/src/complex.md
MikeInnes: Fair, but then we should mention the `'` notation as well. Might be better to have a note on it in context, rather than trying to get it into this paragraph.

docs/src/adjoints.md
MikeInnes: Individual pullbacks don't (and can't) know whether the function as a whole is holomorphic, which is a global property, and don't ever see complex derivatives. They _only_ work with sensitivities, as we define them, and taking the adjoint is correct for sensitivities.

docs/src/complex.md
simeonschaub: I can try explaining this a bit better in words. I'm trying to give the mathematical reason here, why we need to take the complex conjugate of the derivative in pullbacks.

docs/src/adjoints.md
simeonschaub: I probably should have clarified: I am only talking about the partial function being holomorphic. So if we're defining the pullback for `exp`, for example, we can express the partial derivative of `exp` as a complex derivative. We can't do that for `abs2`, for example. Is there anywhere, where the term `sensitivity` is mathematically defined? In my understanding, it is a vector, you pass to a differential form, so it basically specifies the linear combination of partial derivatives. If we're only limiting ourselves to scalars, this would just be the partial derivative of the output with respect to the current function in the chain.

docs/src/complex.md
simeonschaub: Where exactly are you talking about?

docs/src/adjoints.md
MikeInnes: Ok, that's fair, but still an edge case (and even then, we still have the adjoint of the linear map, it's just expressed differently). These specific docs are a reference for adjoints rather than complex AD and I'd rather not overload people with the vagaries of that before they've gotten started with the real part :) So ideally this should just mention briefly that adjoints are generally relevant for complex AD, and have a link to the other docs for more detail.

docs/src/complex.md
MikeInnes: as in, we can clarify the notation when it's first used, rather than trying to define everything in a parenthetical.

docs/src/adjoints.md
simeonschaub: Ok, sounds like a good idea

docs/src/complex.md
simeonschaub: Ok, will do

simeonschaub:
Sorry, it took me so long to implement these changes. I hope that's better now.

CarloLucibello:
closing as outdated

PR #338, opened 2019-09-24
Implement fusible forward/reverse-modes AD for broadcasting

### LINKED ISSUE #336: Use forward-mode in broadcasting?
It seems that the overhead of the way broadcastings are handled currently is quite large:

* https://github.com/FluxML/Zygote.jl/issues/323#issuecomment-533032855
* https://discourse.julialang.org/t/zygote-performance/29028/16

Would it make sense to use the ForwardDiff-based solution (`broadcast_forward`) provided that it can be handled by ForwardDiff?  That is to say:

* The function (`.f`) wrapped in all `Broadcasted`s are singleton (`Base.issingletontype`) and
* All arguments (`.args`) wrapped in all `Broadcasted`s are `Real`, `AbstractArray{<:Real}`, or non-differentiable (e.g., `Type`, `Symbol` and `Val`).

Note that, with ForwardDiff-based approach, we can fuse all broadcastings and also call style-specific materialization method.

Is this a reasonable approach?  Or would the performance improvements be implemented soon?

tkf:
I implemented the forward-mode -based fused broadcasting AD as I outlined in #336. I also constructed the code in such a way that some of the facility can be used for reverse-mode AD of fused broadcasting.

At the moment this patch is for "early review" (although the test works locally).  The code became rather non-trivial so I think I need to add some comments before asking for a serious review.  But the main idea is similar to `Broadcast.flatten`.

Also, there seems to be no speed improvement (yet).  It's a bit puzzling because the basic strategy is similar to what I used in <https://github.com/tkf/ChainCutters.jl>.  I'll look into it after the direction of the implementation is "approved."

close https://github.com/FluxML/Zygote.jl/issues/336

### DIFF
--- a/test/gradcheck.jl
+++ b/test/gradcheck.jl
@@ -552,7 +552,7 @@ end
 
 @testset "broadcast" begin
   if !Zygote.usetyped
-    @test gradient(x -> sum(sin.(x)), Diagonal(randn(3)))[1][2] == 1
+    @test gradient(x -> sum(sin.(x)), Diagonal(randn(3)))[1][2] == 0
   end
 end
 
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -34,6 +34,7 @@ using NNlib
 # =========
 
 accum_sum(xs; dims = :) = reduce(accum, xs, dims = dims)
+accum_sum(xs::Tuple; dims = :) = reduce(accum, xs)
 
 # Work around reducedim_init issue
 accum_sum(xs::Nothing; dims = :) = nothing
@@ -50,7 +51,9 @@ unbroadcast(x::AbstractArray, x̄) =
 
 unbroadcast(x::Union{Number,Ref}, x̄) = accum_sum(x̄)
 
-unbroadcast(x::AbstractArray, x̄::Nothing) = nothing
+unbroadcast(x::Union{AbstractArray, Tuple}, x̄::Nothing) = nothing
+
+unbroadcast(x::NTuple{N}, x̄::NTuple{N}) where N = x̄
 
 # Split Reverse Mode
 # ==================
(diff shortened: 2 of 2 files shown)


test/gradcheck.jl
```diff
@@ -552,7 +552,7 @@ end
 
 @testset "broadcast" begin
   if !Zygote.usetyped
-    @test gradient(x -> sum(sin.(x)), Diagonal(randn(3)))[1][2] == 1
+    @test gradient(x -> sum(sin.(x)), Diagonal(randn(3)))[1][2] == 0
```
tkf: Is it important? I should probably follow the discussion in https://github.com/FluxML/Zygote.jl/issues/163 but unlike sparse matrices, you can't change the off-diagonal elements in `Digonal`.  So it makes sense to me that off-diagonal gradients are zero.

MikeInnes: It needs to continue being 1, otherwise it's not consistent with the semantics of other broadcast implementations. See the discussion in #163 for more discussion.

MikeInnes:
This seems like a lot of additions considering we already have an implementation of forward-mode for broadcast. I was expecting that it'd more or less just involve a compatibility check (e.g. all inputs are real or real arrays) and then dispatch to that.

Fusion is a bit of an open question since it can be a net loss with a large number of inputs. I think we should start with something very simple to make the common cases faster, and then start adding more on top of that.

cossio:
Something holding this back? This would be very nice to have.

cossio:
Looking at this: https://github.com/FluxML/Zygote.jl/blob/a2f7c9731279b8ad80322c35b3a0325f03392c0e/src/lib/broadcast.jl#L159, it looks like there is already some code for forward-mode diff for broadcast in Zygote, but only for CuArrays. Can't this be used for normal arrays too and call it a day?

AzamatB:
Bump. Would be nice to have it

cossio:
Can this benefit from https://github.com/FluxML/Zygote.jl/pull/503? If so, any chance this PR can be updated? 
It would be very nice to have decent performance in broadcast.

cossio:
Bump?

DhairyaLGandhi:
Are there some benchmarks that we can come up with? There is a case of diminishing returns, but if we can annotate some functions to use forward mode, that would be more helpful

cossio:
@oschulz There was https://github.com/FluxML/Zygote.jl/pull/1001 which got merged recently. I'm not sure if that supersedes this.

DhairyaLGandhi:
No, not really. There are some caveats with the implementation we have currently. Maybe it would actually be better to dispatch to the forward mode version generically for maps.

cossio:
I don't understand what's the difference between #1001 and this.
One question: The forward mode used here is also based on ForwardDiff?

CarloLucibello:
closing as outdated

PR #339, opened 2019-09-25
Skip precompile() when Zygote is imported for compiling downstreams

tkf:
IIUC, this is not required when precompiling downstream packages?

Before (ea4d1e894af775a6783f683d851bcfb5c10e25b0):

```julia
julia> using Zygote  # make sure Zygote is precompiled

julia> @time using Example
[ Info: Precompiling Example [7876af07-990d-54b4-ab0e-23690620f79a]
 12.440068 seconds (1.05 M allocations: 50.296 MiB, 0.81% gc time)
```

After:

```julia
julia> @time using Example
[ Info: Precompiling Example [7876af07-990d-54b4-ab0e-23690620f79a]
  2.690000 seconds (487.92 k allocations: 22.979 MiB, 0.33% gc time)
```

where I edited Example.jl so that it imports `Zygote`.

### DIFF
--- a/src/Zygote.jl
+++ b/src/Zygote.jl
@@ -52,7 +52,7 @@ end
 precompile() = usetyped || include(joinpath(@__DIR__, "precompile.jl"))
 
 # precompile()
-@init precompile()
+@init Requires.isprecompiling() || precompile()
 
 # helps to work around 265-y issues
 function refresh()


MikeInnes:
How long does the first call to `gradient` take with and without this? When I was tweaking this, I found that avoiding the annoying precompile time only shifted that to first-use, which isn't better IMO.

tkf:
I don't think this patch changes the run-time behavior? (e.g., importing Zygote in REPL still runs `precompile`)

https://github.com/FluxML/Zygote.jl/blob/e7ea1f2a2feec6637b850aa6e515fdf3fbd722c6/src/Zygote.jl#L55

tkf:
Just as a sanity check, here is the timings.

After this PR (e7ea1f2a2feec6637b850aa6e515fdf3fbd722c6):

```julia
julia> using Zygote
[ Info: Precompiling Zygote [e88e6eb3-aa80-5325-afca-941959d7151f]

julia> @time Zygote.precompile()
  0.096316 seconds (96.64 k allocations: 4.567 MiB)
(Float32[1.0],)
```

Before this PR (ea4d1e894af775a6783f683d851bcfb5c10e25b0):

```julia
julia> using Zygote
[ Info: Precompiling Zygote [e88e6eb3-aa80-5325-afca-941959d7151f]

julia> @time Zygote.precompile()
  0.091458 seconds (96.65 k allocations: 4.567 MiB)
(Float32[1.0],)
```

There is no change as expected.

MikeInnes:
It's worth checking what happens when you load Zygote via another package (like Example), because the interactions in package loading can frequently violate your expectations. Anyway, I did this myself and it looks ok. Thanks for the patch.

tkf:
Thanks a lot for carefully checking the effect of this patch!

PR #344, opened 2019-09-26
Add adjoints for Matrix and Vector

sethaxen:


willtebbutt:
LGTM

MikeInnes:
Could we generalise this and just define `Array{T,N}(x)`? Or perhaps even `AbstractArray{T,N}(x)`?

sdaxen:
I tried this but it didn't work:

```julia
@adjoint Array{T,N}(xs::AbstractArray) where {T,N} = Array{T,N}(xs), ȳ -> (ȳ,)
@adjoint Array{T,N}(xs::Array) where {T,N} = Array{T,N}(xs), ȳ -> (ȳ,)
```

e.g. failure is with `Matrix(::Symmetric)`, since that function mutates:
```julia
julia> using LinearAlgebra, Random
julia> x = Symmetric(randn(3,3));
julia> y, back = Zygote.pullback(Matrix, x)
([-2.5235116023603847 1.5236150792716827 -0.45236176055356603; 1.5236150792716827 0.1624223088079216 0.8004997059735403; -0.45236176055356603 0.8004997059735403 -0.14280628662031883], getfield(Zygote, Symbol("##32#33")){typeof(∂(Array{T,2} where T))}(∂(Array{T,2} where T)))
julia> ȳ = randn(3,3);
julia> back(ȳ)
ERROR: Mutating arrays is not supported
Stacktrace:
 [1] (::getfield(Zygote, Symbol("##834#835")))(::Nothing) at /Users/saxen/projects/Zygote.jl/src/lib/array.jl:38
 [2] (::getfield(Zygote, Symbol("##2163#back#836")){getfield(Zygote, Symbol("##834#835"))})(::Nothing) at /Users/saxen/.julia/packages/ZygoteRules/Mmoki/src/adjoint.jl:49
 [3] Type at /Users/sabae/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.2/LinearAlgebra/src/symmetric.jl:257 [inlined]
 [4] (::typeof(∂(Array{T,2} where T)))(::Array{Float64,2}) at /Users/saxen/projects/Zygote.jl/src/compiler/interface2.jl:0
 [5] (::getfield(Zygote, Symbol("##32#33")){typeof(∂(Array{T,2} where T))})(::Array{Float64,2}) at /Users/saxen/projects/Zygote.jl/src/compiler/interface.jl:38
```

sdaxen:
Any advice for how to proceed with this PR? The above error causes failures in #355.

sdaxen:
What's tricky about a generic adjoint is that it overrules sensible defaults. e.g. `Diagonal(::AbstractMatrix)` doesn't need a custom adjoint because `Diagonal(::AbstractVector)` and `diag(::AbstractMatrix)` do have custom adjoints and give it the right behavior. Defining the adjoint

```julia
@adjoint (::Type{T})(x::AbstractArray) where {T<:AbstractArray} = T(x), Δ -> (Δ,)
```

then requires `Diagonal(::AbstractMatrix)` to have a custom adjoint to cope.

It looks like nearly all of the custom adjoints for calls of `(::Type{AbstractArray})(xs)` currently in Zygote follow the pattern:

```julia
@adjoint (::Type{T})(x::AbstractArray) where {T<:AbstractArray} = T(x), Δ -> (T(Δ),)
```

I wonder if this is a sensible default. It's a bit odd because the resulting adjoint is of the same type as the output, not the input, but it ensures that adjoint information that doesn't fit the sparsity pattern of the output isn't pulled back to the input. I think this would also cover cases like `Array{Float32}(x)`, since it would ensure that the adjoints pulled back are not of higher precision than the output.

The multi-argument constructors and the `Hermitian/Symmetric` cases are more complicated and still need custom adjoints.

sdaxen:
@MikeInnes do you see an issue with the proposed approach? If not, I'll implement.

sdaxen:
@MikeInnes how does this look?

MikeInnes:
I'm not sure about the `T(Δ)` part. If types were only used for sparsity it might be ok, but for e.g. numeric types it's not consistent with how scalars promote in the backwards pass.

I'm a bit unclear on how not having this would affect `Diagonal` etc. Those adjoints should take precedence over this very generic `AbstractArray` one, right?

CarloLucibello:
@sethaxen Any further thoughts on this?

sethaxen:
Closing because stale

PR #350, opened 2019-09-29
Add adjoints for symmetric/hermitian eigen/eigvals

sdaxen:
This PR adds adjoints for `eigen` and `eigvals` for `Hermitian` and real `Symmetric` matrices where all eigenvalues are unique. I don't know if/how we can support repeating eigenvalues. jax, TF, and (I think) Theano have the same limitation, but it would be nice if we didn't. 

Indirectly, this PR also adds support for `cos`, `sin`, `tan`, `cosh`, `sinh`, `tanh`, `log`, `acos`, `asin`, `atan`, `asinh`, and `sqrt` on real `Symmetric`/`Hermitian` matrices. A future PR will handle the complex Hermitian case for these functions, `acosh`/`atanh`, and repeated eigenvalues.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -447,6 +447,26 @@ end
   Ā = real.(VTF\(VT*F̄/VTF.*X)*VT)
   (Ā, )
 end
+@adjoint function LinearAlgebra.eigen(A::LinearAlgebra.RealHermSymComplexHerm)
+  dU = eigen(A)
+  return dU, function (Δ)
+    d, U = dU
+    d̄, Ū = Δ
+    if Ū === nothing
+      P = Diagonal(d̄)
+    else
+      F = inv.(d' .- d)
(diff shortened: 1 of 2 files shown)


sdaxen:
Triggering new Travis build

sdaxen:
This looks ready to merge to me unless anyone has feedback.

sdaxen:
I certainly can. I expect they'll fail though. A low rank matrix has multiple 0 eigenvalues, so it fails the unique eigenvalues constraint.

willtebbutt:
Thanks for adding the broken tests. I'm happy to merge once the failure on 1.3 is resolved. Is it obvious what's going on there?

sdaxen:
I think so. `reim` calls `real` on the matrix in non-broadcast mode. Type hacks version on some versions wants it to be broadcast.

PR #352, opened 2019-10-01
[WIP] Add adjoints for functions on Hermitian matrices

### LINKED ISSUE #340: exp broken for defective matrices
The current implementation for the adjoint of `exp` relies on the eigenvalues of its argument being distinct. This isn't always the case, and I've found examples in my work where I need the exponential of a matrix which is defective by construction, so this assumption doesn't hold.

I've hacked together a solution that works for my research (a slow but accurate differentiable Pade approximation -- this operation is not a computational bottleneck in my work) so it's not blocking for me. Would be nice to resolve though.

sdaxen:
This PR will add adjoints for the following functions of Hermitian and real symmetric matrices:
- [x] `exp`
- [x] `cos`
- [x] `sin`
- [x] `tan`
- [x] `cosh`
- [x] `sinh`
- [x] `tanh`
- [ ] `log`
- [ ] `acos`
- [ ] `asin`
- [x] `atan`
- [ ] `acosh`
- [x] `asinh`
- [ ] `atanh`
- [ ] `sqrt`

Their adjoints take an essentially identical form to one another. Along the way, I fixed #340 for generic matrices, since the current adjoint for `exp` using the eigendecomposition shares a similar form to the Hermitian cases.


sdaxen:
Ugh, this history got all screwed up somehow. Will start over with a clean branch.

PR #359, opened 2019-10-04
Add GPU CI

dhairyagandhi96:
Adds ability to work with the JuilaGPU CI setup on GitLab via Bors. Good to check compatibility with CUDA and the GPU stack. 

cc @MikeInnes

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -31,13 +31,15 @@ end
   include("compiler.jl")
 end
 
-@info "Starting GPU intgeration tests..."
-if haskey(ENV, "CI_GITLAB_CUDA")
+
+using CUDAapi
+if has_cuda()
+  @info "Starting GPU intgeration tests..."
   @testset "CUDA tests" begin
     include("cuda.jl")
   end
 else
-  @info "ENV variable CI_GITLAB_CUDA not set - Skipping CUDA Tests"
+  @info "CUDA not found - Skipping CUDA Tests"
 end
 
 end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -39,7 +39,7 @@ if has_cuda()
     include("cuda.jl")
   end
 else
-  @info "CUDA not found - Skipping CUDA Tests"
+  @warn "CUDA not found - Skipping CUDA Tests"
 end
 
 end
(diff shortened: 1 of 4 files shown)


dhairyagandhi96:
bors try

dhairyagandhi96:
Fwiw https://gitlab.com/JuliaGPU/Zygote.jl/pipelines/86468772

dhairyagandhi96:
bors try

dhairyagandhi96:
Trying out https://forum.bors.tech/t/bors-stuck-pending-when-ci-completes/386

dhairyagandhi96:
bors try

dhairyagandhi96:
bors try-

dhairyagandhi96:
bors try

dhairyagandhi96:
bors try

dhairyagandhi96:
bors try-

dhairyagandhi96:
bors try

test/runtests.jl
```diff
@@ -31,4 +31,8 @@ end
   include("compiler.jl")
 end
 
+@testset "CUDA tests" begin
```
MikeInnes: You'll need to check that CUDA is available before running these, as in the Flux tests. It would be nice to also have a similar `@info` message to show whether we are doing GPU testing or not.

test/runtests.jl
```diff
@@ -31,8 +31,13 @@ end
   include("compiler.jl")
 end
 
-@testset "CUDA tests" begin
-  include("cuda.jl")
+@info "Starting GPU intgeration tests..."
+if haskey(ENV, "CI_GITLAB_CUDA")
```
MikeInnes: You can use CUDAapi to check that CUDA is set up – then GPU tests will run locally if cuda is available, too.

test/runtests.jl
dhairyagandhi96: Right, I've added that messaging. Wondering if I should use CUDAapi like we do with Flux

test/runtests.jl
dhairyagandhi96: Got my answer 👍

dhairyagandhi96:
bors try

MikeInnes:
Thanks @dhairyagandhi96!

PR #366, opened 2019-10-07
Add ChainRules

oxinabox:
This replaces https://github.com/FluxML/Zygote.jl/pull/291

The bits from that OP that still matter

Step 1) Change Zygote to check for chainrules before doing its normal stuff,
and adapt the stuff it gets back from chainrules to play nice with Zygote's expectations

Step 2) adapt Zygote more deeply, so it can take full advantage of thunks etc.

This PR is Step 1.

<s> TODO: workout why this seems to segfault for me. </s>


oxinabox:
@MikeInnes  any idea why this segfaults?

MikeInnes:
It could be a Zygote compiler issue, but given that you haven't touched that, the second most likely option is that it's just a stackoverflow.

MikeInnes:
IME, Julia tends to segfault due to stack overflows. Doesn't happen every time, but often.

src/compiler/interface.jl
```diff
@@ -28,7 +28,37 @@ end
 
 # Wrappers
 
-_pullback(f, args...) = _pullback(Context(), f, args...)
+function _pullback(f, args...)
```
oxinabox: This is the wrong function to overload as it will not be hit for functions that occur in functions Zygote decomposes

oxinabox:
It works!

1 of the blackslists can be removed immediately with https://github.com/JuliaDiff/ChainRules.jl/pull/124

src/compiler/interface2.jl
```diff
@@ -3,7 +3,87 @@ using IRTools.Inner: argnames!, update!
 
 ignore(T) = all(T -> T <: Type, T.parameters)
 
-@generated function _pullback(ctx::AContext, f, args...)
+
+function _pullback(ctx::AContext, f, args...)
+  @assert !(f isa AContext)
```
oxinabox:
```suggestion
```

test/chainrules.jl
```diff
@@ -0,0 +1,47 @@
+using Zygote, Test, ChainRules
+const cr_inner_demo_rrule_hitcount = Ref(0)
```
oxinabox:
```suggestion

const cr_inner_demo_rrule_hitcount = Ref(0)
```

test/chainrules.jl
```diff
@@ -0,0 +1,48 @@
...
+        cr_inner_demo_rrule_hitcount[] = 0
+        cr_inner_demo_pullback_hitcount[] = 0
+        y, pb = pullback(cr_inner_demo, 11)
+        @test y == 55
+        @test cr_inner_demo_rrule_hitcount[] == 1
+        @test cr_inner_demo_pullback_hitcount[] == 0
+        @test pb(1)==(5.0,);
+        @test pb(2)==(10.0,);
+        @test pb(3)==(15.0,);
+        @test cr_inner_demo_pullback_hitcount[] == 3
```
oxinabox:
```suggestion
        @test cr_inner_demo_pullback_hitcount[] == 3
         @test cr_inner_demo_rrule_hitcount[] == 1
```

src/compiler/interface2.jl
```diff
@@ -3,7 +3,86 @@ using IRTools.Inner: argnames!, update!
...
+chainrules_blacklist(::typeof(mapreduce), args...) = true
+chainrules_blacklist(::typeof(mapfoldl), args...) = true
+chainrules_blacklist(::typeof(mapfoldr), args...) = true
+chainrules_blacklist(::typeof(sum), f, x::AbstractArray{<:Real}) = true
+# Except for sum(abs2, xs), that is fine
+chainrules_blacklist(::typeof(sum), ::typeof(abs2), x::AbstractArray{<:Real}) = false
+
+# ChainRules current Wirtinger deriviative is not compatible
+# reconsider after https://github.com/JuliaDiff/ChainRulesCore.jl/pull/29
+chainrules_blacklist(::typeof(abs), ::Complex) = true
```
mohamed82008: Use metaprogramming?

oxinabox: Could do.
But I feel like wouldn't be much shorter and would be less clear

oxinabox: Maybe should do.
This also could be done with `Union`.

`uniontypeof(fs...) = Union{typeof.(fs)...}`

src/compiler/interface2.jl
```diff
@@ -33,25 +33,18 @@ chainrules_blacklist(f, args...) = false
...
 # reconsider after https://github.com/JuliaDiff/ChainRulesCore.jl/pull/29
-chainrules_blacklist(::typeof(abs), ::Complex) = true
-chainrules_blacklist(::typeof(abs2), ::Complex) = true
-chainrules_blacklist(::typeof(conj), ::Complex) = true
-chainrules_blacklist(::typeof(adjoint), ::Complex) = true
-chainrules_blacklist(::typeof(hypot), ::Complex) = true
-chainrules_blacklist(::typeof(angle), ::Complex) = true
-chainrules_blacklist(::typeof(imag), ::Complex) = true
-chainrules_blacklist(::typeof(real), ::Complex) = true
+for f in (abs, abs2, coj, adjoint, hypot, angle, imag, real)
```
antoine-levitt: Typo here (`conj`)

oxinabox:
```suggestion
for f in (abs, abs2, conj, adjoint, hypot, angle, imag, real)
```

MikeInnes:
I'm concerned about the performance implications of this, in its current form. In particular we have to be pretty careful to manage the call stack; recursing many times into `_pullback` will likely trigger recursion limiting heuristics that cause type inference to give up. (This is particularly pernicious because it doesn't show up in `@code_typed`.)

Also, presumably many of Zygote's current adjoints are now redundant, given this PR. It'd be nice to remove those to show the overall impact on Zygote's codebase (hopefully a significant net improvement).

oxinabox:
> I'm concerned about the performance implications of this, in its current form. In particular we have to be pretty careful to manage the call stack; recursing many times into _pullback will likely trigger recursion limiting heuristics that cause type inference to give up. (This is particularly pernicious because it doesn't show up in @code_typed.)

Which bits do you think are adding levels?
We don't recurse into `_pullback` anywhere in this PR.

How every you count it, adding deeper integation and usiing ChainRules types (rather than `zextern`ing everythiing) would cut that to 1 I think,
and also give benifit of not unthunking everything.

> Also, presumably many of Zygote's current adjoints are now redundant, given this PR. It'd be nice to remove those to show the overall impact on Zygote's codebase (hopefully a significant net improvement).

Easiest way to do would be to add convience functions to ChainRulesCore
for importing Zygote Rules's `@adjoint` defintions,
and then transferiing basically everything over.
Checking as one goes.
Then look  at the method overwrite warnings.
https://github.com/JuliaDiff/ChainRulesCore.jl/issues/44

I'ld rather not be doing that alone, and I'ld rather not have that blocking this PR.

MikeInnes:
> Which bits do you think are adding levels? We don't recurse into `_pullback` anywhere in this PR.

If the user-defined call tree is `foo -> bar -> baz`, the transformed tree under this PR is `_pullback(foo) -> _pullback_sct(foo) -> _pullback(bar) -> _pullback_sct(bar) -> _pullback(baz) -> _pullback_sct(baz)` (since `_pullback_sct` recursively calls `_pullback` for each call inside the function it transforms).

Previously, it would have been `_pullback_sct(foo) -> _pullback_sct(bar) -> _pullback_sct(baz)`. Because we have a type inference hack that lets us trick the compiler into seeing `foo -> bar -> baz` again here, this doesn't kill the compiler. Whereas now the callstack has three recursions into the same function `_pullback`, which can cause issues.

oxinabox:
> Because we have a type inference hack that lets us trick the compiler into seeing foo -> bar -> baz again here, this doesn't kill the compiler.

Where can I learn about this type inference hack?
And is the solution to just inline `_pullback_sct` back into `_pullback` by hand?

MikeInnes:
I don't think it's documented, but you just set a flag on the codeinfo, as [here](https://github.com/MikeInnes/IRTools.jl/blob/969676bc7407dcfceae96ad311c843a70ab91850/src/reflection/reflection.jl#L117).

I think manual inlining / having one uber-function for `_pullback` is probably the way to go (it can still be factored of course, as long as any function calls are compile-time).

oxinabox:
Re inference
I am starting to feel like the easiest way
might be to more down to the Mike’s Little IR level
and also to just do a bunch of edge hooking myself

rather than using the intended if `nothing==res=ChainRules.rrule(…_`
thing I can just check the method it would hit,
and then check if it is the fallback,
and if not stick a call into it and and attach an edge
Or if it is hitting the fallback then attack an edge to the method table instead

2 problems:
1. this removes ability to bail out of a chainrule via returning nothing
2. this doesn’t work in 1.0

oxinabox:
Maybe could go the otherway and make `pullback_source2source(ctx, f, args....)` set its `method_for_inference_limit_heuristics` to be `pullback(ctx, f, args....)`
rather than being set to `f(args...)` like it currently seems to be?

src/compiler/interface2.jl
```diff
@@ -32,3 +43,59 @@ end
...
+  function zback(Δs...)
+    ∂s = pb(Δs...)
+    ntuple(length(∂s)) do ii
+      ∂ = ∂s[ii]
+      zextern(∂)
+    end
+  end
+end
+
+zextern(x) = ChainRules.extern(x)
```
oxinabox: Shouldn't this be:

```suggestion
zextern(x) = ChainRules.extern(conj(x))
```

should also add test for that

oxinabox:
@MikeInnes @dhairyagandhi96 how are we going to resolve this?
Will my plan from above work?
https://github.com/FluxML/Zygote.jl/pull/366#issuecomment-546476742
Do we need to get compiler people in to a Google meet call to work though things?

oxinabox:
This is much nicer now.
Because some of the things we were blacklisting are gone,
others are fixed,
and I realized we don't need to actually get rid of most of ChainRule's types since they define `+` in the way `accum` needs.

MikeInnes:
So to recap, the core problem is that the if the compiler sees a call stack like `f -> f -> f -> ...` it assumes it's over-inferring some recursive algorithm and gives up.

If user code does `f -> g -> h -> *` (where `*` is a function we have a defined adjoint for) Zygote transforms this to something like `pullback -> _pullback(f) -> _pullback(g) -> _pulback(h) -> adjoint(*)`. Because the calls to `_pullback` functions are all the same method, we risk hitting this heuristic. However, there's a workaround: we have a hack in the compiler that lets us pretend generated code came from a different method, making this stack look like `pullback -> f -> g -> h -> adjoint(*)`, which is OK.

However, ChainRules requires a _dynamic_ check to figure out if a CR rule applies and forward to either AD or `rrule`, making the stack look like `pullback -> check(f) -> _pullback(f) -> check(g) -> _pullback(g) -> check(h) -> _pullback(h) -> check(*) -> adjoint(*)`. Since `check` is the same method in each case, I'm concerned it'll hit this heuristic, and even if it doesn't it's just good to keep the callstack compact (since it becomes exponentially worse when we nest derivatives).

I can think of a few different ways to avoid that extra `check` function:

* Put the check in the generated IR. The generated code would contain the check, the call to `rrule`, _and_ the auto-generated pullback. The check should be inlined and eliminated but it could slow down if inference isn't perfect. Still a bit more work when nesting, and relies heavily on inference being reliable.
* Do the check at compile time. e.g. we just check whether a method exists for `rrule` that isn't the default fallback. We'd need to work out world age correctness etc., and I'm concerned this would lead to a lot of recompilation if done correctly, but it should generate code as good as now.
* Do it like a DiffRules/Requires hybrid. The CR database would be more symbolic and Zygote would import its rules in an `@eval` loop to define adjoint rules like it does now. But you also provide a callback so that new rules can be defined in other packages, and Zygote will add these adds methods to `adjoint` too when the package is loaded.

I'd lean towards the compile-time check if we can make it correct without loads of recompilation. If not, I think the Requires-like approach would be fine: it doesn't inherit Requires' main problems around version bounding since none of that changes compared to the current situation. Both of these options technically add a limitation to ChainRules (you can only intercept based on types, not values), but that seems pretty acceptable to me.

Any thoughts on that @oxinabox?

oxinabox:
> If user code does `f -> g -> h -> *` (where `*` is a function we have a defined adjoint for) Zygote transforms this to something like `pullback -> _pullback(f) -> _pullback(g) -> _pulback(h) -> adjoint(*)`. Because the calls to `_pullback` functions are all the same method, we risk hitting this heuristic.

Those are different methods.
Because each function type is unique.
a few ML call's ago @Keno  confirmed that that heuristic depends on method, not function.
So I am not sure I understand.

> Do the check at compile time. e.g. we just check whether a method exists for rrule that isn't the default fallback. We'd need to work out world age correctness etc., and I'm concerned this would lead to a lot of recompilation if done correctly, but it should generate code as good as now.

The current check *should* resolve at compile time already, for cases where it can be resolved only on type.
And I am pretty sure it does.
But it would still touch the stack potentially.

See also
https://github.com/JuliaDiff/ChainRulesCore.jl/issues/127
which proposes making a DuiffRules list of rules based on using `methods` and `!(Nothing <: return_type(meth))`
 
I am not hard tied to support for rules needing to be decidable as to if they exist based on types alone all our current ones are, but if they are not then it is an ugly was of time.

oxinabox:
How do we tell if this heurustic is being hit?

MikeInnes:
> that heuristic depends on method, not function.

It's a different (specialised) `MethodInstance` but not a different method. The method corresponds to a written body of code, so both `check` and `_pullback` are the same method each time they are called as it stands.

> The current check should resolve at compile time already, for cases where it can be resolved only on type.

The current checks semantically resolve at runtime, even if they can be compiled away. More specifically I'm imagining doing this at Zygote's version of compile time (i.e. doing the check _inside_ the generated function, before deciding what code to return), which would be a change to semantics. More importantly it would mean that as far as type inference and the Julia compiler is concerned, there would be no difference to the code it has to infer.

oxinabox:
@MikeInnes  and I talked in chat.

Plan is to go with Option 1, which @MikeInnes  will implement 
then once JuliaDiff/ChainRulesCore.jl#127 is done we can benchmark option 3 against it.

willtebbutt:
Is option 2 not an option if we innovate in `ChainRules` a little? In particular, @MikeInnes IIUC type info is available at `Zygote` compile time, so if `ChainRules` has a collection of functions that specify whether a rule exists for a particular type we should be grand.

Could we not, therefore, just implement `@frule` and `@rrule` macros that in addition to implementing an `frule` and `rrule` implement a function from the argument types that always returns `true` (i.e. rule exists), with a `false` fallback? This would presumably avoid world-age issues.

@oxinabox have we had this discussion before and decided it's a bad idea?

MikeInnes:
Specifically we have a signature like `Tuple{typeof(*),Int,Int}`, so we'd just need something like `hasrule(Tuple{...})`.

If the rule changes, that's easy: the code generated by Zygote will be recompiled without running Zygote again. If a new rule for `f` gets added it's trickier, because Zygote needs to recompile the pullback for `f` even though `f` itself hasn't changed. But I think if we just add another backedge for the `hasrule` method that will sort this out, without requiring _everything_ to be recompiled when `hasrule` changes.

Happy to give pointers on any of this.

oxinabox:
> Could we not, therefore, just implement `@frule` and `@rrule` macros that in addition to implementing an `frule` and `rrule` implement a function from the argument types that always returns `true` (i.e. rule exists), with a `false` fallback? This would presumably avoid world-age issues.

from experience doing something similar in TensorFlow.jl,  recording things with macros, this gets awkward fast, and the code to handle methods gets complicated.
And if one wants to handle method deletion (which in a Revise.jl world one does),
then I think you outright can't do it.

willtebbutt:
> from experience doing something similar in TensorFlow.jl, recording things about functions as they are created with macros, this gets awkward fast, and the code to handle methods gets complicated.
And if one wants to handle method deletion (which in a Revise.jl world one does),
then I think you outright can't do it.

Oh, okay, so is Revise unable to handle
```julia
@rrule (::typeof(foo), x::Tx, y::Ty) where {Tx, Ty} = ...
```
spitting out something like
```julia
rrule(::typeof(foo), x::Tx, y::Ty) where {Tx, Ty} = ...
has_rrule(::Type{typeof(foo), Tx, Ty}) where {Tx, Ty} = true
```
?

oxinabox:
Oh, actually Revise probably can handle that, i misunderstood.
I thought you were talking about keeping a actual list object (and was going say about that having other issues).

I still dislike 2.
for several other reasons.
 - any time we provide a macro, we want it to be to give a more convient syntax for something that is not awful to write fully by hand. That starts to become aweful to write by hand once you start needing to write about types of types and making sure the `<:` are in te right place.
 - Option 1 can be done right now with no change to ChainRules
 - Option 3 is something we planned on setting up anyway, https://github.com/JuliaDiff/ChainRulesCore.jl/issues/127

Option 2 just seems like clearly worse than the other two.

willtebbutt:
I definitely agree that option 1 is clearly the best option

I'm actually leaning towards option 2 over option 3 though, since option 3 is specifically something we were going to call "Legacy Mode" with big signposts telling people not to write AD tools using this kind of feature anymore, precisely because it's very hard to keep track of what methods are valid etc.

For example, how would method deletion play with option 3?

willtebbutt:
Yeah, thinking about it both options 2 and 3 are _very_ annoying. I completely agree with @oxinabox 's point that we shouldn't be baking stuff in that makes `frule`s and `rrule`s trickier to implement manually, in particular having to implement 2 functions is very annoying and bug-prone.

Again, option 1 is clearly far superior to the other options.

MikeInnes:
I don't think it's _clearly_ superior. It's something of a compromise where we make Zygote a bit worse without changing CR. I'm not a big fan of introducing all the extra branches; even if they can be optimised away in principle, these things have a lot of knock-on effects for things like inlining heuristics (and as usual it gets exponentially worse when nesting). The main benefit is something we arguably don't need (value-dependent choice of AD vs. a rule).

I'm also not totally clear on what it means to "implement a rule manually" or why we'd need that. I'd argue the other direction: if the `@frule` macro can't express everything you can do by overloading functions, the macro is broken and we should make a better one. If `@frule` and `@rrule` is _the_ interface to defining rules in ChainRules (and I think it has to be, once you get to more complex things like kwarg handling) the other issues don't come up. Am I missing something there?

oxinabox:
> I'm also not totally clear on what it means to "implement a rule manually" or why we'd need that. I'd argue the other direction: if the `@frule` macro can't express everything you can do by overloading functions, the macro is broken and we should make a better one. If `@frule` and `@rrule` is _the_ interface to defining rules in ChainRules (and I think it has to be, once you get to more complex things like kwarg handling) the other issues don't come up. Am I missing something there

You may be right there. We haven't hit that point yet, but we might soon.
That is a discussion for ChainRulesCore.

MikeInnes:
It may be worth a look at the output of Zygote's `@adjoint` macro. It's really not something you'd want to write by hand; but there's also not a whole lot there that isn't necessary. In general I'd definitely agree that it's nice to be able to do things without a lot of syntactic abstraction, but I think this case may be a necessary exception to that.

willtebbutt:
@MikeInnes any progress on this?

AzamatB:
Bump. `Zero` being handled as `nothing` giving me issues.
What is blocking this from getting merged?

MikeInnes:
I think the latest is that @oxinabox is going to implement a way to list out CR rules so that Zygote can overload its own `adjoint` function directly (at least for now).

For the record, me and Lyndon also discussed CR depending on ZygoteRules so that the `@frules` macro can just overload `frule` and `adjoint` directly itself. Lyndon feels that special-casing Zygote in any way is too high of a cost to pay even as a short-term hack, so it's better to explore the other options.

oxinabox:
> I think the latest is that @oxinabox is going to implement a way to list out CR rules so that Zygote can overload its own adjoint function directly (at least for now).

That is option 3 from  https://github.com/FluxML/Zygote.jl/pull/366#issuecomment-584195530.
I am going to do that (because it makes updating non-source2source things easier),
but that was not what i thought the main plan was.
I thought the main plan was the @MikeInnes  was going to implement option 1, which is to do the check in generated code.

Then later once i finish implementing option 3, can benchmark the version of Zgyote that uses option 1 against it.

oxinabox:
Option 5:
Rewrite the `_pullback` on 
https://github.com/FluxML/Zygote.jl/pull/366/files#diff-c791282c7c4a5f9a19d9225f8c222236R6-R15
as a `CodeInfo`  generated function that changed the method for type-inference heuristic.

oxinabox:
Ok so plan from that meeting, correct me if I am wrong.
Mike and I are both crazy busy within the next few weeks

Mike will look in to Option 2, with the aim to add it to this PR.
Mike will follow this comment with one of his own adding more details on option 2.
Useful if someone else with more time can take a look at it.
And so we understand what it is -- I had option 1 and option 2 confused.

Meanwhile I will look into option 3. i.e. https://github.com/JuliaDiff/ChainRulesCore.jl/issues/127
with intent not to use it in Zygote (but to use it in e.g. ForwardDiff.jl)
So it will be an option still for Zygote, if it comes it it.

Rough timeline on both is 1-2 months from now.
So lets check back in every 2 weeks on the ML call on stutus.
2 months from now is: May 21.
So hoping to have this PR merged by May 21

MikeInnes:
Ok, so I think the process roughly goes like this:

1. `_pullback`, a generated function, receives a method signature `T` for which it should either do SCT or forward to ChainRules.
2. Before getting metadata for `T` as usual we get metadata for `Tuple{typeof(rrule),T...}`. We can then check if the method this points to is the CR fallback.
    1. If it is, just generate an SCT transformed version of the method for `T` as usual.
    1. Add backedges to the generated CodeInfo for (a) the MethodInstance in the meta for `T` (we do this already) and (b) the MethodInstance in the meta for `Tuple{rrule, T...}` (so that defining a new rules causes Zygote to re-run).
3. If not, produce code that does nothing more than call `rrule` with the appropriate arguments (or perhaps a wrapper for `rrule` that shims the Zygote `adjoint` interface).
4. Add the usual `T` backedge (Julia will figure out the CR one since it's called in the generated code).

willtebbutt:
Regarding point 2, have I understood correct that that really has to be done within Zygote? It would be nice to have a generic implementation in ChainRulesCore, but I'm assuming that the need to add back edges in the generated function that Zygote produces means it's not really feasible?

oxinabox:
current status:
I am fixing things in chainrules that the tests are showing are broken.

am having issues with
`@testset "Hermitian/Symmetric power series functions"`

oxinabox:
current issue: `A^p` for `A` being a symmetric matrix and `p=-0.5`
see this reproducer
https://gist.github.com/oxinabox/fad7bae6dc11a7f0de31eff8666656cd

sethaxen:
I'm not certain what's going on with the matrix. If there's a bug, it'll probably be here: https://github.com/FluxML/Zygote.jl/blob/713715faf0362f91b3a56e1967e3e29ba7740d52/src/lib/array.jl#L619-L627

The adjoint for the power is handled by https://github.com/FluxML/Zygote.jl/blob/713715faf0362f91b3a56e1967e3e29ba7740d52/src/lib/array.jl#L646 and may be off by a `conj`.

I won't have a chance to check the math until the end of the week.

sethaxen:
By comparison with the second adjoint in
https://github.com/FluxML/Zygote.jl/blob/4f7a2eeeb98eabf43dd70c8698aadb0ec6050c48/src/lib/number.jl#L33-L34

This line *does* look to be missing a `conj`:
https://github.com/FluxML/Zygote.jl/blob/713715faf0362f91b3a56e1967e3e29ba7740d52/src/lib/array.jl#L626

sethaxen:
If this can't wait until the end of the week, I dropped my notes on the derivation here in case someone else wants to check the math or that the code is compatible with it: https://gist.github.com/sethaxen/000d164e515014fdda70601be1ecfb56.

antoine-levitt:
Sorry, can't resist the nitpicking since I've run into this exact problem recently : using a Taylor series to compute accurately terms like (f(a) - f(b)) / (a-b) doesn't get you to O(eps) accuracy, it gets you sqrt(eps) when using a first order expansion. You can increase the order to n but then you get something like eps^(1-1/n) in the worst case. I don't know of any method that gives full accuracy here for a general function f.

sethaxen:
> Sorry, can't resist the nitpicking since I've run into this exact problem recently : using a Taylor series to compute accurately terms like (f(a) - f(b)) / (a-b) doesn't get you to O(eps) accuracy, it gets you sqrt(eps) when using a first order expansion. You can increase the order to n but then you get something like eps^(1-1/n) in the worst case. I don't know of any method that gives full accuracy here for a general function f.

I appreciate the nitpick! Yeah, I wasn't 100% certain about that, but I don't know how to do better here and would happily take recommendations. That doesn't seem to be the issue here though, since we're hitting the error even for a matrix where the eigenvalues are not close.

antoine-levitt:
I don't have anything better: I ended up with https://github.com/JuliaMolSim/DFTK.jl/blob/master/src/Smearing.jl#L44, with O(sqrt(eps)) accuracy using the first derivative
(sorry for derailing this topic)

src/compiler/chainrules.jl
```diff
@@ -0,0 +1,31 @@
...
+  # `nothing->nothing` can be deleted after https://github.com/FluxML/Zygote.jl/issues/603
+  # though it might be worth keeping as a performance optimization (benchmarking pending)
+  zpullback(::Nothing) = nothing
+
+  y, zpullback
+end
+
+# Don't need to conjugate these rrules, because of reasons I don't understand
+chain_rrule(f::typeof(abs), x::Number) = rrule(f, x)
+chain_rrule(f::typeof(abs2), x::Number) = rrule(f, x)
```
MikeInnes, line 31: This is concerning. I know it mirrors how we used DiffRules, but that ruleset wasn't user-extensible. In effect we're assuming all rules added to CR are analytic and will silently get incorrect results if that's not the case, which seems certain to cause problems down the line.

oxinabox: Right now it is a requirement that all rules in ChainRules are analytic.

MikeInnes: Doesn't this imply you have to delete the rules for `abs` and `abs2` and define them in Zygote instead?

oxinabox: probably  should ideally.
Right now happy mirroring DIffRules.
Definately thinking more on complex numbers has to happen still.
I am hoping @willtebbutt  is thinking about that right now

oxinabox: I think they do need to be put in Zygote as hittind the fallback in Zygote to compute `abs(::Complex)` goes to bad places

oxinabox:
All gradient tests are now passing.
Even the Power Series ones.
(But also I still think they shouldn't be passing in this branch nor in master, if they were good tests. See #608)

oxinabox:
current problem:
am trying to make `sin'''` work, but I am getting an error.

```julia
Expression: (((sin')')')(1.0) == -(cos(1.0))
 Need an adjoint for constructor NamedTuple{(:back,),Tuple{NamedTuple{(:back,),Tuple{NamedTuple{(:x,),Tuple{Float64}}}}}}. Gradient is of type Tuple{Tuple{Tuple{Float64}}}
```

`sin'` and `sin''` work fine.

oxinabox:
So if I delete the `rrule` for `identity` it works.
https://github.com/JuliaDiff/ChainRules.jl/pull/175/files#diff-61fe5ffefb8f5d33c006125d43a7fcfaR129-R134
as well as the other corrections in the linked branch to rules

oxinabox:
#### Here are the remaining rules in ChainRules and Zygote that overlap.

```
[1] _pullback(__context__::ZygoteRules.AContext, 288::typeof(abs), x::Real) in Zygote at /Users/oxinabox/JuliaEnvs/ChainRulesWorld/ZygoteRules.jl/src/adjoint.jl:47
[2] _pullback(__context__::ZygoteRules.AContext, 290::typeof(abs2), x::Number) in Zygote at /Users/oxinabox/JuliaEnvs/ChainRulesWorld/ZygoteRules.jl/src/adjoint.jl:47
[3] _pullback(__context__::ZygoteRules.AContext, 385::Type{Symmetric}, A::AbstractArray{T,2} where T) in Zygote at /Users/oxinabox/JuliaEnvs/ChainRulesWorld/Zygote.jl/src/lib/array.jl:430
[4] _pullback(__context__::ZygoteRules.AContext, 401::typeof(tr), x::AbstractArray{T,2} where T) in Zygote at /Users/oxinabox/JuliaEnvs/ChainRulesWorld/Zygote.jl/src/lib/array.jl:702
```

#### Reasons
 - `abs`/`abs2` because ChainRules doesn't actually handle non-analytic complex functions.
 - `Symmetric` because ChainRules only has the 1-arg version (see https://github.com/JuliaDiff/ChainRules.jl/issues/178), so it doesn't fully overlap.
 - `tr` because the Zygote version makes use of FillArrays which has been in ChainRules todo for ages (see https://github.com/JuliaDiff/ChainRules.jl/issues/46)



### Script to find them
```julia
using Pkg
Pkg.activate(@__DIR__)
using Zygote
using ChainRules


using ExprTools

function primal_sig(mm; skip=0)
    start_ind = 2 + skip  # always start from at least 2, so as to skip the pullback/rrule itself
    sig_vec = ExprTools.parameters(mm.sig)[start_ind:end] # aly=
    return Tuple{sig_vec...}
end

zy_meths = collect(methods(Zygote._pullback))
@time zy_sigs = primal_sig.(zy_meths, skip=1)

cr_meths = collect(methods(ChainRules.rrule))
cr_sigs = primal_sig.(cr_meths)


###
#=
Now we want to find things in Zygote that are special cases of things in ChainRulest
these can mostly be deleted

We don't need to be smart as can just do a loop over each it not a huge number
=#

zy_specials = NamedTuple[]
for (zy_meth, zy_sig) in zip(zy_meths, zy_sigs)
    for (cr_meth, cr_sig) in zip(cr_meths, cr_sigs)
        cr_sig == Tuple{Any,Vararg{Any,N} where N}  && continue  # skip the fallbakc
        if zy_sig <: cr_sig
            push!(zy_specials, (zy=zy_meth, cr=cr_meth))
        end
    end
end


first.(zy_specials)
last.(zy_specials)

```

willtebbutt:
> abs/abs2 because ChainRules doesn't actually handle non-analytic complex functions.

I'm slightly confused as to why we can't handle these. In particular I would have thought that our `abs2` definition would be fine, while the `abs` definition just needs some modification to be complex-aware, since the `sign` of a `Complex` number isn't something that's sensibly defined.

I guess I'm asking: when you say that CR doesn't handle non-analytic complex functions, is that it currently has definitions that don't play nicely with complex numbers, or that there's some Zygote interfacing thing that doesn't work properly?

oxinabox:
> I'm slightly confused as to why we can't handle these. In particular I would have thought that our abs2 definition would be fine, while the abs definition just needs some modification to be complex-aware, since the sign of a Complex number isn't something that's sensibly defined.

I amn not actually fully sure,
but I think it does come down to neither `abs` nor `abs2` are holomorphic.
so there gradients are just not defined, unless you make certain calls as to what to do under those circumstances.
We have not made those calls since we don't have Wirtinger anymore.
We do however still have those last 2 rules, because DIffRules has them.
But the definations we have (and that diffrules has) don't really play nice with our ability to interact with Zgyote for holomorphic complex functions (which is to conjugate on the way in, and on the way out).
Not 100% sure why, but didn't want to investigate too much since I feel like the root of the answer is that those gradients are not defined without deciding how to handle nonanalistic functions

willtebbutt:
Ah, okay, the conjugating on the way in and out will do it -- specifically the conjugating on the way out bit (reverse-pass way out of course).

The sensitivities _definitely_ make sense in both cases, you just have to think about it as partials w.r.t. the real and complex components. If you follow that logic, our `abs2` implementation would be completely fine if we weren't conjugating on the way out.

oxinabox:
Ok this is ready for (hopefully) final review

docs/src/adjoints.md
```diff
@@ -1,5 +1,13 @@
 # Custom Adjoints
 
+!!! note "Prefer to use ChainRules to define custom adjoints"
+    Zygote supports the use of [ChainRulesCore](http://www.juliadiff.org/ChainRulesCore.jl/stable/) to define custom sensitivities.
+    It is prefered to define the custom sensitivities using ChainRulesCore as they will work for many AD systems, not just Zygote.
+    This documentation exists to descibe how Zygote works, and how adjoints can be directly defined for Zygote.
+    Defining adjoints this way does not make them accessable to other AD systems, but does let you do things that directly depend on how Zygote works.
+    It allows for specific definations adjoints that are only defined for Zgyote (which might work differently to more generic definations defined for all AD)
```
AzamatB, line 8: definations -> definitions of

docs/src/adjoints.md
```diff
@@ -1,5 +1,13 @@
 # Custom Adjoints
 
+!!! note "Prefer to use ChainRules to define custom adjoints"
+    Zygote supports the use of [ChainRulesCore](http://www.juliadiff.org/ChainRulesCore.jl/stable/) to define custom sensitivities.
+    It is prefered to define the custom sensitivities using ChainRulesCore as they will work for many AD systems, not just Zygote.
```
nickrobinson251, line 5: Can we clarify a little more how and where to do this. e.g.

```suggestion
    It is prefered to define the custom sensitivities using `ChainRulesCore.rrule` as they will work for many AD systems, not just Zygote.
    These sensitivities can be added in your own package, or for Base functions they can be added to ChainRules.jl.
     
```

test/chainrules.jl
```diff
@@ -0,0 +1,156 @@
...
+        @test (1,) == h(1)
+
+        a3, pb3 = Zygote.pullback(h, 1)
+        @test ((1,),) == pb3(1)
+    end
+end
+
+@test_broken gradient(2.0) do x
+  @fastmath x^2.0
+end == (4.0,)
```
nickrobinson251, line 156: related to the commented out code above? could do with a comment maybe?

src/compiler/chainrules.jl
```diff
@@ -0,0 +1,57 @@
...
+end
+
+wrap_chainrules(f, args...) = wrap_chainrules_output(f(wrap_chainrules_input(args)...))
+
+
+
+function chain_rrule(f, args...)
+  #@info "Using ChainRule" f, typeof.(args)
+#  Core.println("Using ChainRule ", f," ", typeof.(args))
+
```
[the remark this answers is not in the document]
oxinabox: probably, but they were really useful

src/lib/number.jl
```diff
@@ -71,28 +23,35 @@ end
...
+#     end
+#   elseif DiffRules.hasdiffrule(:Base, f, 2)
+#     dx, dy = DiffRules.diffrule(:Base, f, :x, :y)
+#     @eval begin
+#       @adjoint Base.FastMath.$fastf(x::Number, y::Number) =
+#         Base.FastMath.$fastf(x, y),
+#         Δ -> (Δ * make_fastmath(conj($dx)), Δ * make_fastmath(conj($dy)))
+#     end
+#   end
+# end
```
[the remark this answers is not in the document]
oxinabox: Good question,
There is an issue in ChainRules to reimplement these over there

src/lib/number.jl
```diff
@@ -71,28 +23,35 @@ end
...
-    dx, dy = DiffRules.diffrule(:Base, f, :x, :y)
-    @eval begin
-      @adjoint Base.FastMath.$fastf(x::Number, y::Number) =
-        Base.FastMath.$fastf(x, y),
-        Δ -> (Δ * make_fastmath(conj($dx)), Δ * make_fastmath(conj($dy)))
-    end
-  end
-end
+# we intentionally define these here rather than falling back on ChainRules.jl
+# because ChainRules doesn't really handle nonanalytic complex functions
```
willtebbutt, line 27: This should probably be phrased differently. We don't currently handle non-analytic complex functions, but it's not a massive change to make it work properly with them. You could refer to [this issue that I just opened about this](https://github.com/JuliaDiff/ChainRulesCore.jl/issues/159)

sethaxen:
I tried switching some adjoints using ZygoteRules over to ChainRulesCore using this PR, and I ran into an issue where it appears Zygote is missing the ChainRules `rrule`s when kwargs are passed. Here's an example adapting the inner/outer test included in this PR:

```julia
julia> using ChainRules, Zygote, Test

julia> cr_inner_demo_rrule_hitcount = Ref(0);

julia> cr_inner_demo_pullback_hitcount = Ref(0);

julia> cr_inner_demo(x; kwarg = 1) = 5x;

julia> function ChainRules.rrule(::typeof(cr_inner_demo), x; kwarg = 1)
           cr_inner_demo_rrule_hitcount[] += 1
           function cr_inner_demo_pullback(Δx)
               cr_inner_demo_pullback_hitcount[] += 1
               return ChainRules.NO_FIELDS, 5.0*Δx
           end
           return cr_inner_demo(x), cr_inner_demo_pullback
       end

julia> function cr_outer_demo(x; kwarg = 1)
           2 + 10cr_inner_demo(x; kwarg = kwarg)
       end
cr_outer_demo (generic function with 1 method)

julia> @testset "gradient inner" begin
           cr_inner_demo_rrule_hitcount[] = 0
           cr_inner_demo_pullback_hitcount[] = 0
           @test (5.0,) == gradient(x -> cr_inner_demo(x; kwarg = 1), 11)
           @test cr_inner_demo_rrule_hitcount[] == 1
           @test cr_inner_demo_pullback_hitcount[] == 1
       end
gradient inner: Test Failed at REPL[7]:5
  Expression: cr_inner_demo_rrule_hitcount[] == 1
   Evaluated: 0 == 1
Stacktrace:
 [1] top-level scope at REPL[7]:5
 [2] top-level scope at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Test/src/Test.jl:1113
 [3] top-level scope at REPL[7]:2
gradient inner: Test Failed at REPL[7]:6
  Expression: cr_inner_demo_pullback_hitcount[] == 1
   Evaluated: 0 == 1
Stacktrace:
 [1] top-level scope at REPL[7]:6
 [2] top-level scope at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Test/src/Test.jl:1113
 [3] top-level scope at REPL[7]:2
Test Summary:  | Pass  Fail  Total
gradient inner |    1     2      3
```

If I omit the `kwarg` in `@test (5.0,) == gradient(x -> cr_inner_demo(x; kwarg = 1), 11)
`, then the test passes fine.

oxinabox:
> I tried switching some adjoints using ZygoteRules over to ChainRulesCore using this PR, and I ran into an issue where it appears Zygote is missing the ChainRules rrules when kwargs are passed. Here's an example adapting the inner/outer test included in this PR:

ah yes, I forgot I needed to workout what was going on with kwargs.
Since the `Zygote.chain_rrule` doesn't accept them.
I noticed that before and was meaning to ask @MikeInnes  about it abut it slipped my mind.
Thanks for reminder.

@MikeInnes where are kwargs gone at this point of the code? I can't find them

docs/src/adjoints.md
```diff
@@ -1,5 +1,15 @@
 # Custom Adjoints
 
+!!! note "Prefer to use ChainRules to define custom adjoints"
+    Zygote supports the use of [ChainRulesCore](http://www.juliadiff.org/ChainRulesCore.jl/stable/) to define custom sensitivities.
+    It is prefered to define the custom sensitivities using `ChainRulesCore.rrule` as they will work for many AD systems, not just Zygote.
+    These sensitivities can be added in your own package, or for Base functions they can be added to ChainRules.jl.
+
+    This documentation exists to descibe how Zygote works, and how adjoints can be directly defined for Zygote.
+    Defining adjoints this way does not make them accessable to other AD systems, but does let you do things that directly depend on how Zygote works.
```
piever, line 9:
```suggestion
    Defining adjoints this way does not make them accessible to other AD systems, but does let you do things that directly depend on how Zygote works.
```

docs/src/adjoints.md
```diff
@@ -1,5 +1,15 @@
...
+    It allows for specific definations of adjoints that are only defined for Zgyote (which might work differently to more generic definations defined for all AD)
```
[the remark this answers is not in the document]
oxinabox:
```suggestion
    It allows for specific definitions of adjoints that are only defined for Zgyote (which might work differently to more generic definitions defined for all AD).
```

src/lib/array.jl
```diff
@@ -729,6 +715,8 @@ end
   end
 end
 
+# ChainRules has this also but does not use FillArrays, so we have out own defination
```
AzamatB, line 718: defination -> definition

nickrobinson251: you can use the "suggestion" feature
image

AzamatB: Sorry, I did not realize I can do that – GitHub's UI keeps getting more complicated. Will do so going forward, thanks.

oxinabox:
OK, I think I have what i need to do worked out for kwargs.
Need to:

- [ ] reassemble pullback correctly
- [ ] add tests

Since chainrules does not support derivative wrt kwargs the reassembled pullback should just have some `nothing` in the right places

cossio:
> Since chainrules does not support derivative wrt kwargs

So if `f(x) = g(; kwarg=x)`, then `f(x)` won't get differentiated?

oxinabox:
> > Since chainrules does not support derivative wrt kwargs
> 
> So if `f(x) = g(; kwarg=x)`, then `f(x)` won't get differentiated?

No, this doesn't have any affect on what gets differentiated.
Its just about what can be expressed in chainrules.
You currently an not write a chainrule specifing the pullback of `g` with respect to `kwarg`.
You also can't AFAIK write a `Zgyote.@adjoint` for that either.
but AD passes though it fine.

sethaxen:
I can confirm that with the latest commit, Zygote now handles correctly my `rrule`s with `kwargs`.

src/compiler/chainrules.jl
```diff
@@ -0,0 +1,119 @@
...
+"""
+function has_chain_rrule(T)
+  m = meta(Tuple{typeof(rrule),T.parameters...})
+  if m.method !== chainrules_fallback
+    # found a rrule, no need to add any edges
+    return true, nothing
+  end
+
+  # Could be a kwarg function, handle that case
+  ft_parts = T.parameters
```
oxinabox, line 20:
```suggestion
```

src/compiler/chainrules.jl
```diff
@@ -0,0 +1,118 @@
...
+  # did not find anything, will have to attach edges so it recompiles if one is added
+  @assert m.code.edges !== nothing
+  return false, m.code.edges
+end
+
+"""
+    is_kwfunc(sigt...)
+
+Determines if `sigt` is the type signature of a kwfunction.
+each element of `sigt` should be a type.
```
oxinabox, line 35:
```suggestion
Each element of `sigt` should be a type.
```

oxinabox:
Would we like to move `zback` into its own type?
So as to avoid potential closure issues?

```julia
struct ZBack{F}
  back::F
end
(s::ZBack)(dy) = wrap_chainrules_pullback(s.back, dy)
# Dispatch here handles chainrules considing pullbacks to have multiple input if Tuple.
# TODO: this could be removed if: https://github.com/JuliaDiff/ChainRulesCore.jl/issues/152
(s::ZBack)(dy::Tuple) = wrap_chainrules_pullback(s.back, dy...)
# `nothing->nothing` can be deleted after https://github.com/FluxML/Zygote.jl/issues/603
# though it might be worth keeping as a performance optimization (benchmarking pending)
(s::ZBack)(::Nothing) = nothing
```

oxinabox:
I once again think this is ready to go.

I note that about half this code will get to be deleted once we do #603 and https://github.com/JuliaDiff/ChainRulesCore.jl/issues/152
but I think that is definately for a follow up PR. 
This one is big enough

src/compiler/chainrules.jl
```diff
@@ -0,0 +1,127 @@
+const chainrules_fallback = which(rrule, Tuple{Any})
+
+"""
+  has_chain_rrule(T)
+
+For a type-tuple `T` e.g. `Tuple{typeof(f), Int, Float64}`, checks if there is a `rrule` defined for it.
+Excluding the generic fallback.
```
CarloLucibello, line 7: maybe this is meant to be expanded

src/compiler/chainrules.jl
```diff
@@ -0,0 +1,127 @@
...
+The first return value is a Bool is whether or not the `rrule` exists.
```
CarloLucibello, line 8:
```suggestion
The first return value is `true` if the `rrule` exists, `false` otherwise.
```

src/compiler/chainrules.jl
```diff
@@ -0,0 +1,127 @@
...
+Determines if `sigt` is the type signature of a kwfunction.
+Each element of `sigt` should be a type.
+Either the first 3 types are a kwfunc type, a NamedTuple and the matching base function type,
+or the first argument is the base function type and it is not a kwfunction.
+the remaining types in `sigt` are the types of the argument.
+
+"""
+is_kwfunc(::Vararg{Any}) = false
+# Needs `@pure` because else will not run during type inference.
+# This is pure enough, the only generic function it calls is in `Core` overloading `Core.kwftype` will no doubt break many other things also
```
CarloLucibello, line 43: break this comment

CarloLucibello:
does it make sense to do some performance tests on this? e.g. time to first derivative, time to run all tests,....

src/compiler/chainrules.jl
oxinabox: how so?
Note this is an internal docstring, so it is for people who are areading the code along-side.

src/compiler/chainrules.jl
oxinabox:
```suggestion
# This is pure enough, the only generic function it calls is in `Core` 
# overloading `Core.kwftype` will no doubt break many other things also
```

oxinabox:
> does it make sense to do some performance tests on this? e.g. time to first derivative, time to run all tests,....

I do not think it is worth doing a half-baked job of benchmarking Zygote to do this PR.
Without a well considered benchmarking method I can't say that we would get information out of it that we can act on.
There are very few performance related things in this PR that we *can* change in this PR.
It is already almost fully happenning  at compile-time.
The the runtime work is mostly very fast (and sometimes compiles away) in the wrapping.
and that will be gone once we do #603 and JuliaDiff/ChainRulesCore.jl#152 which we want to do anyway but beyond the scope of this PR.

Further,If we want to benchmark Zygote then we should have meaningful benchmarks that capture things like compile-time and run-time, and should have a well chosen set of trial functions.
There is some intent to make some for https://github.com/JuliaDiff/DiffTests.jl/issues/7
if for example we just timing the whole thing will tell us some things are slower, or faster but not what.
Of course we can do better, but that is beyond the scope of this PR.
This PR has on the go for 7 months.
It woule be pretty frustrating to see it blocked again while we argue about how to benchmark or how to interpret benchmarks.

src/compiler/chainrules.jl
```diff
@@ -0,0 +1,128 @@
...
+  # Could be a kwarg function, handle that case
+  if is_kwfunc(T.parameters...)
+    # Need to check for rrule for function not the kwfunction.
+    base_T = Tuple{T.parameters[3:end]...}
+    return has_chain_rrule(base_T)
+  end
+
+  # did not find anything, will have to attach edges so it recompiles if one is added
+  @assert m.code.edges !== nothing
+  return false, m.code.edges
```
oxinabox, line 28: This is not going to work preJulia-1.3
I guess we can just return `[]` in that case and not support the loading of `rrules` after the fact (without `Zygote.refresh()`)

Will have to add that condition here and in `_pullback`

nickrobinson251:
Yes, benchmarks are clearly out of scope of this PR (in the absence of evidence of significant performance regressions caused by these changes)

CarloLucibello:
> Yes, benchmarks are clearly out of scope of this PR (in the absence of evidence of significant performance regressions caused by these changes)

I entirely agree, and this part of @oxinabox's comment is quite satisfying already
>There are very few performance related things in this PR that we can change in this PR.
>It is already almost fully happenning at compile-time.
>The the runtime work is mostly very fast (and sometimes compiles away) in the wrapping.
>and that will be gone once we do #603 and JuliaDiff/ChainRulesCore.jl#152 which we want to do ? >anyway (but beyond the scope of this PR.)

MikeInnes:
Why are kwargs a special case? We could just call `rrule(kwfunc(f), kw, f, args...)` as usual.

MikeInnes:
One useful (obviously not comprehensive) 'benchmark' is to check that

```julia
julia> f(x) = 5x + 10

julia> @code_llvm f'(1)
;  @ /Users/mike/projects/flux/Zygote/src/compiler/interface.jl:58 within `#38'
define i64 @"julia_#38_18169"(i64) {
top:
  ret i64 5
}
```

is optimised down to a constant. I believe that worked on an earlier version of this branch but it seems like `chain_rrule` is now too heavy.

oxinabox:
> Why are kwargs a special case? We could just call `rrule(kwfunc(f), kw, f, args...)` as usual

Because we define kwargs in `rrule` in the way one expected kwargs to be defined.
Tht is to say we define  `rrule(::typeof(f), args...; kwargs)` 
this is much more natural as a user interface for defining rules.

oxinabox:
> is optimised down to a constant. I believe that worked on an earlier version of this branch but it seems like `chain_rrule` is now too heavy.

Wow, I did not know that ever worked.
Thats pretty cool.
I will have to do some digging to findout where where broke it.

I learned to my surrprise the following does constant fold, so it is not within ChainRules itself, so it can work in theory.
```julia
@code_llvm (s->unthunk(ChainRules.rrule(*, 5,5)[2](s)[2]))(π)
```

oxinabox:
Fixed it:

```julia
;  @ /Users/oxinabox/JuliaEnvs/ChainRulesWorld/Zygote.jl/src/compiler/interface.jl:47 within `#43'
define i64 @"julia_#43_21487"(i64) {
top:
  ret i64 5
}
```

We definately need to do https://github.com/JuliaDiff/ChainRulesCore.jl/issues/152
I think that is now highest priority ChainRules change to follow up after this PR.

MikeInnes:
> its definately a thing that could be changed in the future, but seems like it would be a lot of work for not a clear gain to do it now.
> Unless there is a huge issue with this approach?

Although not as bad as the previous runtime rule check, it's still quite a lot of extra code to add to Zygote's output which could easily be done at load time instead. It's unfortunate that CR's interface is coupled to its database representation so that this limitation exists. Yes, on small examples these branches will tend to compile away, but it has all the usual costs: the code has to be specialised and optimised separately at almost every call site by our 50 loc/sec compiler, and where inference fails (frequently) it will have a runtime cost. It won't be possible to optimise it away in the interpreter. As usual all this gets worse when nesting AD.

Although I agree that requiring a benchmark suite is an unfairly high bar for this PR, I do think it's important not to add big regressions to an already-poor cold start experience, so we should have a couple of quick checks at least. On my machine I find that this branch currently adds about 1.7x to compile time and 2.5x to runtime on `sin'''(0.5)`, which is not ideal.

oxinabox:
> Although not as bad as the previous runtime rule check, it's still quite a lot of extra code to add to Zygote's output which could easily be done at load time instead.

Actually i can change this to be at load_time fairly easily I think.
I will do that.

oxinabox:
> Although I agree that requiring a benchmark suite is an unfairly high bar for this PR, I do think it's important not to add big regressions to an already-poor cold start experience, so we should have a couple of quick checks at least.

You had literally 7 months to bring this up.
But ok a couple of quick checks:
Restarting Zygote between each.


# Normal usage
Basically negligible change.
Not much more to say.

## sin
### This PR
```
julia> @time sin'(0.5);
  0.000000 seconds

julia> @btime sin'(0.5);
  1.313 ns (0 allocations: 0 bytes)
```

### Current release
```
julia> @time sin'(0.5);
  0.000000 seconds

julia> @btime sin'(0.5);
  1.503 ns (0 allocations: 0 bytes)
```

## Simple polynomial
```
f(x) = 4x^3 + 3x^2 + 2y + 1
```


### This PR
```
julia> @time f'(π);
  0.000000 seconds (1 allocation: 16 bytes)

julia> @btime f'(π);
  7.761 ns (1 allocation: 16 bytes)
```


### Current release
```
julia> @time f'(π);
  0.000000 seconds (1 allocation: 16 bytes)

julia> @btime f'(π);
  7.719 ns (1 allocation: 16 bytes)
```



## Simplish Function
```
function foo_inner(x)
    a = max(x, x^2)*x
    if a > 0
        return prod([a, 2a, 3a])
    else
        return sum([a, 2a, 3a])
    end
end

foo(x) = foo_inner(x) + foo_inner(-x)
```

### This PR
```
julia> @time foo'(0.5);
  0.167321 seconds (470.05 k allocations: 24.904 MiB, 12.23% gc time)

julia> @btime foo'(0.5);
  15.030 μs (161 allocations: 5.28 KiB)
```

### Current master
```
julia> @time foo'(0.5);
  0.126286 seconds (411.58 k allocations: 21.645 MiB)

julia> @btime foo'(0.5);
  14.737 μs (161 allocations: 5.34 KiB)
```



# Nested AD tests
This is where it hurts.
Because we add a bunch of machinery for converting conventions.
And that has to be AD'ed through.
and it breaks a lot of constant folding and inference.

Basically all this machinery will go away once we do #603 and JuliaDiff/ChainRulesCore.jl#152.
Which will eliminate much of the overhead.

Also nested reverse mode is always a mistake.
It should almost always be forward-over-reverse or forward-over-forward.
In the rare case reverse-over-reverse is wanted people already are missing Zygote with ReverseDiff.jl.
(Though I think real solution would be Taylor mode reverse, do the double transform at each step)

With this PR is is a larger mistake.
At least til we get those future changes in.

### sin''
### This PR
```
julia> @time sin''(0.5);
  0.991807 seconds (1.78 M allocations: 95.639 MiB, 15.34% gc time)

julia> @time sin''(0.5);
  0.000108 seconds (167 allocations: 3.812 KiB)
```

### Current Master
```
julia> @time sin''(0.5);
  0.000002 seconds

julia> @btime sin''(0.5);
  10.937 ns (0 allocations: 0 bytes)
```

## sin'''
### This PR
```
julia> @time sin'''(0.5);
 31.932736 seconds (75.77 M allocations: 3.821 GiB, 3.71% gc time)

julia> @time sin'''(0.5);
  0.005229 seconds (6.44 k allocations: 160.266 KiB)
```

### Current release
```
julia> @time sin'''(0.5)
 23.562748 seconds (57.35 M allocations: 2.896 GiB, 4.55% gc time)
-0.8775825618903728

julia>

julia> @time sin'''(0.5)
  0.002546 seconds (4.66 k allocations: 114.312 KiB)
-0.8775825618903728
```

src/compiler/chainrules.jl
```diff
@@ -0,0 +1,135 @@
...
+Either the first 3 types are a kwfunc type, a NamedTuple and the matching base function type,
+or the first argument is the base function type and it is not a kwfunction.
+the remaining types in `sigt` are the types of the argument.
+
+"""
+is_kwfunc(::Vararg{Any}) = false
+# Needs `@pure` because else will not run during type inference.
+# This is pure enough, the only generic function it calls is in `Core`
+# overloading `Core.kwftype` will no doubt break many other things also
+Base.@pure is_kwfunc(k, ::Type{<:NamedTuple}, f, args...) = k===Core.kwftype(f)
```
oxinabox, line 43: Now that it only is used at parse time probably does not need `@pure` anymore

oxinabox: checked, it is indeed not needed so i removed it

oxinabox:
This is passing  1.5.0-DEV.639 (2020-04-16) for me.
Not sue if something changed in nightly since then


I get the 1.0 failure locally also though. Not sure what is happening  there.

AzamatB:
Can this get merged?

oxinabox:
I have now removed the thing where primals returning tuples had multiple input pullbacks.
It cleans up the code nicely

CarloLucibello:
some errors on julia 1.0, can they be fixed/worked around?

oxinabox:
> some errors on julia 1.0, can they be fixed/worked around?

Those errors are https://github.com/FluxML/Zygote.jl/issues/638
Same thing happens on master, but the tests on master don't quiet hit that code path.
Even if you run our new tests on master they still don't hit that code path (but if lkike in the post for #638 you hit that path directly you get the same error.

oxinabox:
> some errors on julia 1.0, can they be fixed/worked around?

Those errors are https://github.com/FluxML/Zygote.jl/issues/638
Same thing happens on master, but the tests on master don't quiet hit that code path.
Even if you run our new tests on master they still don't hit that code path (but if lkike in the post for #638 you hit that path directly you get the same error.

MikeInnes:
Ok, I guess we should go ahead with this as is. I find the kwarg handling pretty ugly but I trust that can be improved in future; in any case Mjolnir should be able to get rid of it without issues.

Can we remove any more adjoints? Should we also remove tests that only test Zygote adjoints, since that presumably overlaps with ChainRules?

oxinabox:
> Can we remove any more adjoints?

No i already did a check programatically of the method tables looking for overlaps.
They are nicely distinct, surprisingly little overlap. I guess because ChainRules got most of its rules for Nabla, and most of Nabla's rules are linked to needing to support a particular direction of Gaussian Process research.

> Should we also remove tests that only test Zygote adjoints, since that presumably overlaps with ChainRules?

I kind of like them for now as a kind of double check of the ChainRules handling.
They are mostly on very fundermental things, since that is where the ChainRules and (now removed) Zygote rules overlap. So it is nice to be confidant that they are working right.

In the future, once we start moving more large things over we should remove the tests also.
Since they will be less fundermental operations

oxinabox:
I will give this one last rebase, and will add a `if` to make this pass with a `@test_broken` on 1.0 for the problem that relates to #638

PR #112, opened 2019-03-12
Simplest prod(x; dims) gradient

mcabbott:
The current gradient for `prod(x; dims)` gives incorrect results, this PR fixes it (parallel to  https://github.com/FluxML/Tracker.jl/pull/1 ):
```
julia> using Zygote, ForwardDiff

julia> r = rand(2,3,2);

julia> ForwardDiff.gradient(w->sum(prod(w, dims=(2,3))), r)
2×3×2 Array{Float64,3}:
[:, :, 1] =
 0.00131643  0.000954347  0.0051387 
 0.0177437   0.0354628    0.00934587

[:, :, 2] =
 0.00434307  0.0140455   0.00152818
 0.0151417   0.00464615  0.00451601

julia> Zygote.gradient(w->sum(prod(w, dims=(2,3))), r)[1] # wrong answer!
2×3×2 Array{Float64,3}:
[:, :, 1] =
 5.93867e-6  4.30525e-6  2.31817e-5
 1.60301e-5  3.2038e-5   8.44331e-6

[:, :, 2] =
 1.95925e-5  6.33622e-5  6.89391e-6
 1.36795e-5  4.19746e-6  4.07989e-6

julia> Zygote.@adjoint function prod(xs; dims = :) # as in this PR
         p = prod(xs; dims = dims)
         p, Δ -> (p ./ xs .* Δ,)
       end

julia> Zygote.refresh()

julia> Zygote.gradient(w->sum(prod(w, dims=(2,3))), r)[1] # now matches ForwardDiff
2×3×2 Array{Float64,3}:
[:, :, 1] =
 0.00131643  0.000954347  0.0051387 
 0.0177437   0.0354628    0.00934587

[:, :, 2] =
 0.00434307  0.0140455   0.00152818
 0.0151417   0.00464615  0.00451601
```
This does not handle zeros in the array correctly -- see https://github.com/FluxML/Flux.jl/pull/524 for attempts to do that. The `circshift(...` operation deleted here was a correct (but slow) gradient for `prod(x)`, but is clearly independent of `dims`. 

The example above is almost the same as the one in the tests, which strangely passes, without this PR. Perhaps something is wrong with `gradtest`?
```
julia> @test gradtest(x -> prod(x, dims = (2, 3)), (3,4,5))
Test Passed

julia> @test gradtest(x -> prod(x), (3,4,5))
Test Passed
```

PR #367, opened 2019-10-09
add @adjoin for relu to increase speed

findmyway:
```julia

using Flux, Zygote, BenchmarkTools

x = rand(128)

# before
@benchmark gradient((x) -> sum(relu.(x)), $x)

#=
BenchmarkTools.Trial: 
  memory estimate:  15.41 KiB
  allocs estimate:  425
  --------------
  minimum time:     9.980 μs (0.00% GC)
  median time:      10.649 μs (0.00% GC)
  mean time:        12.495 μs (4.12% GC)
  maximum time:     2.701 ms (98.24% GC)
  --------------
  samples:          10000
  evals/sample:     1
=#

# after

@benchmark gradient((x) -> sum(relu.(x)), $x)
#=
BenchmarkTools.Trial: 
  memory estimate:  2.30 KiB
  allocs estimate:  3
  --------------
  minimum time:     258.013 ns (0.00% GC)
  median time:      303.099 ns (0.00% GC)
  mean time:        368.242 ns (16.51% GC)
  maximum time:     4.796 μs (85.99% GC)
  --------------
  samples:          10000
  evals/sample:     394
=#
```


MikeInnes:
This is useful, but you probably don't need the loop vs. just using another broadcast in the backwards pass. See also the gradients in `broadcast.jl` for examples.

src/lib/nnlib.jl
```diff
@@ -1,17 +1,9 @@
...
-    return y, Δ -> begin
-        res = zero(Δ)
-        @inbounds for i in 1:length(res)
-            if y[i] > 0
-                res[i] = Δ[i]
-            end
-        end
-        (nothing, res)
-    end
+@adjoint function Base.Broadcast.broadcasted(::typeof(relu), xs::Array{T}) where T<:Real
```
willtebbutt: Could we relax this to make it work with `CuArray`s? Or is that covered elsewhere?

MikeInnes: Yes, this should definitely cover all array types. @findmyway there's a `Numeric` type in broadcast.jl that would be useful here, since it covers abstract arrays and scalars.

findmyway: I relaxed the constraint to `AbstractArray`. However, the improvement with `CuArray` is not that obvious compared to `Array`.

```julia

# before

#=
BenchmarkTools.Trial: 
  memory estimate:  6.52 KiB
  allocs estimate:  163
  --------------
  minimum time:     32.256 μs (0.00% GC)
  median time:      34.455 μs (0.00% GC)
  mean time:        39.182 μs (2.60% GC)
  maximum time:     38.206 ms (26.64% GC)
  --------------
  samples:          10000
  evals/sample:     1
=#

# after

#=
BenchmarkTools.Trial: 
  memory estimate:  4.50 KiB
  allocs estimate:  107
  --------------
  minimum time:     24.777 μs (0.00% GC)
  median time:      26.802 μs (0.00% GC)
  mean time:        30.936 μs (3.09% GC)
  maximum time:     36.959 ms (25.84% GC)
  --------------
  samples:          10000
  evals/sample:     1
=#
```

findmyway: OK, I'll switch to `Numeric`

findmyway: @MikeInnes Should I move the change to `broadcast.jl`, or better to keep it in the `nnlib.jl`?

MikeInnes: nnlib.jl is fine, given that it is nnlib-specific; just needs to follow the same template as the definitions in `broadcast.jl`.

src/lib/nnlib.jl
```diff
@@ -1,5 +1,10 @@
 using NNlib
-import NNlib: softmax, ∇softmax, logsoftmax, ∇logsoftmax, conv, depthwiseconv, ∇conv_data, ∇depthwiseconv_data, maxpool, meanpool, σ
+import NNlib: softmax, ∇softmax, logsoftmax, ∇logsoftmax, conv, depthwiseconv, ∇conv_data, ∇depthwiseconv_data, maxpool, meanpool, σ, relu
+
+@adjoint function Base.Broadcast.broadcasted(::typeof(relu), xs::AbstractArray{T}) where T<:Number
+    ys = relu.(xs)
+    ys, Δ -> (nothing, broadcast((y, d) -> y > 0 ? d : zero(T), ys, Δ))
```
MikeInnes: I'd prefer this to be written as `ifelse.(...)`, rather than an explicit `broadcast` function call. Also, it may as well be `zero(y)` rather than `zero(T)`.

findmyway: Do you mean

```julia
@adjoint function Base.Broadcast.broadcasted(::typeof(relu), x::Numeric)
    y = relu.(x)
    y, Δ -> (nothing, ifelse.(y .> 0, Δ, zero(y)))
end
```
?
I know that the above code will cause extra memory allocation (`y .> 0` and `zero(y)`, although we can use `FillArray` for the latter). But sorry that I'm not very familiar with broadcast, I can't come up with a more efficient code using `ifelse.(...)`.

MikeInnes: If you use `zero.(y)`, both of those intermediates will actually get compiled away, so it's just as good.

findmyway: Ah, I see. Thanks!

src/Zygote.jl
```diff
@@ -24,8 +24,8 @@ include("lib/number.jl")
 include("lib/base.jl")
 include("lib/array.jl")
 include("lib/buffer.jl")
-include("lib/nnlib.jl")
 include("lib/broadcast.jl")
+include("lib/nnlib.jl")
```
findmyway: I move the `include("lib/nnlib.jl")` after `include("lib/broadcast.jl")` to make sure that `Numeric` visiable in `nnlib.jl`. But now the result of CI indicates that the `gradcheck` of `prod` failed.

Any idea?

mcabbott: I have no idea why these things should be coupled, but... I don't know why the `prod` test ever passed, as its gradient is independent of `dims`. One attempt to fix it was here: https://github.com/FluxML/Zygote.jl/pull/112

findmyway: Interesting. If I comment out the newly added `gradtest` of `identity`, then all tests pass as they used to be. So my best guess is the randomness.

I'll set that test case as broken and leave if for #112 to fix it.

src/lib/nnlib.jl
```diff
@@ -1,5 +1,12 @@
 using NNlib
-import NNlib: softmax, ∇softmax, logsoftmax, ∇logsoftmax, conv, depthwiseconv, ∇conv_data, ∇depthwiseconv_data, maxpool, meanpool, σ
+import NNlib: softmax, ∇softmax, logsoftmax, ∇logsoftmax, conv, depthwiseconv, ∇conv_data, ∇depthwiseconv_data, maxpool, meanpool, σ, relu
+
+@adjoint Base.Broadcast.broadcasted(::typeof(identity), x::Numeric) = x, Δ -> (nothing, Δ)
```
MikeInnes: This should definitely go in `broadcast.jl`, since it's not nnlib specific.

findmyway: Done

pevnak:
I have handwritten gradients for sigmoid, tanh and selu, as I needed those for one of my projects. Would it make sense to add those?
```
explicitgrad(::typeof(tanh), x) = 1f0 - tanh(x)^2
explicitgrad(::typeof(NNlib.σ), x) = σ(x)*(1f0 - σ(x))
function explicitgrad(::typeof(NNlib.selu), x) 
  λ = oftype(x/1, 1.0507009873554804934193349852946)
  α = oftype(x/1, 1.6732632423543772848170429916717)
  λ * ifelse(x > 0, 1, α * exp(x))
end
```

MikeInnes:
Thanks @findmyway!

@pevnak yeah, those would be great to have as a PR.

bors r+

PR #372, opened 2019-10-14
Add haskey adjoint

dhairyagandhi96:
Ref FluxML/Flux.jl#878

This and the fix to splatting named tuples on master should work for most cases like it.

Test cases incoming
cc @MikeInnes

### DIFF
--- a/src/lib/base.jl
+++ b/src/lib/base.jl
@@ -48,7 +48,6 @@ end
 end
 
 @nograd haskey
-@nograd Base.RefValue{Val{f}} where f
 
 # Channels
 
--- a/src/lib/base.jl
+++ b/src/lib/base.jl
@@ -48,7 +48,7 @@ end
 end
 
 @nograd haskey
-@nograd Base.RefValue
+@nograd Base.RefValue{Val{f}} where f
 
 # Channels
 
--- a/src/lib/base.jl
+++ b/src/lib/base.jl
@@ -48,6 +48,7 @@ end
 end
 
 @nograd haskey
+@nograd Base.RefValue
 
 # Channels
 
--- a/src/lib/base.jl
+++ b/src/lib/base.jl
@@ -47,8 +47,7 @@ end
   end
 end
 
-@adjoint Base.haskey(collection, x) =
-  Base.haskey(collection, x), _ -> nothing
+@nograd haskey
 
 # Channels
 
--- a/src/lib/base.jl
+++ b/src/lib/base.jl
@@ -47,6 +47,9 @@ end
   end
 end
 
+@adjoint Base.haskey(collection, x) =
+  Base.haskey(collection, x), _ -> nothing
+
 # Channels
 
 @nograd Channel, schedule
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -172,10 +172,4 @@ end
     y, back = broadcast_forward(f, args...)
     y, ȳ -> (nothing, nothing, back(ȳ)...)
   end
-
-  @adjoint CuArrays.cu(xs) = CuArrays.cu(xs), Δ -> (CuArrays.cu(Δ),)
-
-  @adjoint CuArrays.CuArray{T,N}(xs) where {T,N} =
-    CuArrays.CuArray{T, N}(xs), Δ -> (CuArrays.CuArray{T, N}(Δ),)
-
 end
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -172,4 +172,10 @@ end
     y, back = broadcast_forward(f, args...)
     y, ȳ -> (nothing, nothing, back(ȳ)...)
   end
+
+  @adjoint CuArrays.cu(xs) = CuArrays.cu(xs), Δ -> (CuArrays.cu(Δ),)
+
+  @adjoint CuArrays.CuArray{T,N}(xs) where {T,N} =
+    CuArrays.CuArray{T, N}(xs), Δ -> (CuArrays.CuArray{T, N}(Δ),)
+
 end


MikeInnes:
How about doing `@nograd haskey` somewhere. That seems pretty safe, since it's unlikely anyone would want a differentiable `haskey`.

dhairyagandhi96:
```Julia
julia> garr = gpu(rand(3,3))
3×3 CuArrays.CuArray{Float32,2}:
 0.823063  0.140359  0.272329
 0.279044  0.2645    0.089738
 0.221039  0.531816  0.897089

julia> gradient(x -> cpu(x) |> sum, garr)
(Float32[1.0 1.0 1.0; 1.0 1.0 1.0; 1.0 1.0 1.0],)
```
This seems alright

```Julia
julia> gradient(x -> gpu(x) |> sum, rand(3,3))
ERROR: TypeError: in ccall: first argument not a pointer or valid constant expression, expected Ptr, got Tuple{Symbol,String}
Stacktrace:
 [1] macro expansion at /home/dhairyagandhi96/.julia/packages/CUDAdrv/ADRHQ/src/base.jl:140 [inlined]
 [2] #copy!#11 at /home/dhairyagandhi96/.julia/packages/CUDAdrv/ADRHQ/src/memory.jl:344 [inlined]
 [3] _pullback(::Zygote.Context, ::CUDAdrv.Mem.var"##copy!#11", ::Nothing, ::Bool, ::typeof(CUDAdrv.Mem.copy!), ::CUDAdrv.Mem.Devi
ceBuffer, ::Ptr{Float32}, ::Int64) at /home/dhairyagandhi96/.julia/dev/Zygote/src/compiler/interface2.jl:0
 [4] copy! at /home/dhairyagandhi96/.julia/packages/CUDAdrv/ADRHQ/src/memory.jl:335 [inlined]
 [5] _pullback(::Zygote.Context, ::typeof(CUDAdrv.Mem.copy!), ::CUDAdrv.Mem.DeviceBuffer, ::Ptr{Float32}, ::Int64) at /home/dhairy
agandhi96/.julia/dev/Zygote/src/compiler/interface2.jl:0
 [6] copyto! at /home/dhairyagandhi96/.julia/packages/CuArrays/0KFHW/src/array.jl:188 [inlined]
 [7] _pullback(::Zygote.Context, ::typeof(copyto!), ::CuArrays.CuArray{Float32,2}, ::Int64, ::Array{Float32,2}, ::Int64, ::Int64)
at /home/dhairyagandhi96/.julia/dev/Zygote/src/compiler/interface2.jl:0
 [8] copyto! at /home/dhairyagandhi96/.julia/packages/GPUArrays/fLiQ1/src/abstractarray.jl:104 [inlined]
 [9] _pullback(::Zygote.Context, ::typeof(copyto!), ::CuArrays.CuArray{Float32,2}, ::Array{Float32,2}) at /home/dhairyagandhi96/.j
ulia/dev/Zygote/src/compiler/interface2.jl:0
 [10] convert at /home/dhairyagandhi96/.julia/packages/GPUArrays/fLiQ1/src/construction.jl:80 [inlined]
 [11] _pullback(::Zygote.Context, ::typeof(convert), ::Type{CuArrays.CuArray{Float32,N} where N}, ::Array{Float64,2}) at /home/dha
iryagandhi96/.julia/dev/Zygote/src/compiler/interface2.jl:0
```

This is with current masters of Flux, Zygote, IRTools, ZygoteRules, CuArrays

dhairyagandhi96:
Sure, that sounds good

src/lib/base.jl
```diff
@@ -47,6 +47,9 @@ end
   end
 end
 
+@nograd haskey
+@nograd Base.RefValue
```
MikeInnes: Why wouldn't this need a gradient?

dhairyagandhi96: Perhaps `@nograd Base.RefValue{Val{f}} where f` be more suited?

I really just want to say that if (after assignment) we are pointing to memory expecting some value, it must keep pointing to that value.

dhairyagandhi96: ` @nograd Base.RefValue{Val{f}} where f` doesn't look like it will do much for that

MikeInnes: > if (after assignment) we are pointing to memory expecting some value, it must keep pointing to that value.

I'm not sure what you mean by this, can you give more context?

The reason it's an issue is that'd it drop gradients for *all* refs, when things like `RefValue(1)[]` should just propagate gradients as usual. The `Val` version is better since `Val`s can't have gradients anyway, but it'd still be useful to understand why we need it.

dhairyagandhi96: Fair enough. The motivation comes from trying to run the [vision/mlp/conv.jl](https://github.com/FluxML/model-zoo/blob/master/vision/mnist/conv.jl) example using Zygote and on the GPU and replacing the loss to use MSE instead of crossentropy. The square-ing `2` (I assume) complained about needing an adjoint for constructor `RefValue{Val{2}}(::Float32)`.

Really I'm only interested in making sure we don't take gradients for `Val`s here

MikeInnes: Ah, I see. That's a bit tricky. I suspect this should be handled by `unbroadcast`; it can dispatch on `Ref{Val}` there and either reduce or throw away gradients as needed. In future we could extend that to allow broadcasting things like `Ref{Int}`, which currently wouldn't work as intended for the same reasons.

dhairyagandhi96: Ah, I was wondering what could have been triggering that. Makes sense.

MikeInnes: Should be fixed with https://github.com/FluxML/Zygote.jl/pull/384

PR #373, opened 2019-10-14
Fixed bcasting issue on GPU

kshyatt:
Should fix Flux.jl#889 also.

```
julia> using Zygote, CuArrays

julia> v(x, n) = x .^ n;

julia> Zygote.gradient(x -> v(x, 7) |> sum, cu(rand(3,3)))
(Float32[4.804972e-5 0.15564603 7.750634e-6; 0.0015771488 0.6221976 0.32504037; 0.3631478 0.00038040814 0.00063724036],)
```
Also checked this works for `log`.

### DIFF
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -169,7 +169,7 @@ end
 
 @init @require CuArrays="3a865a2d-5b23-5a0f-bc46-62713ec82fae" begin
   @adjoint function broadcasted(::Broadcast.ArrayStyle{CuArrays.CuArray}, f, args...)
-    y, back = broadcast_forward(f, args...)
+    y, back = broadcast_forward(CuArrays.cufunc(f), args...)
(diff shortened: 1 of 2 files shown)


MikeInnes:
Awesome, thanks.

Would be great to have a quick test case, even if it's commented out. We're hoping to add GPU CI soon so we could enable it then.

kshyatt:
TFW you're notorious for testing and forget to include a test 😰 . My brand!

Where would be the best place for a test for this?

MikeInnes:
Totally understandable in this case, because it's not easy to add a working test case if we don't have CUDA set up on CI.

I suggest just adding a `cuda.jl` file that can be run manually (doesn't need to be included by `runtests.jl` yet). We can automate it as part of #359.

kshyatt:
Tests added, and I checked that they do pass on the GPU

MikeInnes:
:+1: Thanks!

PR #375, opened 2019-10-18
Added adjoint for sum for CuArrays

dhairyagandhi96:
The current adjoint for `sum` returns a `FillArray` which ends up playing weirdly with `CuArrays`. The `CuArray{N,T}` adjoint may not be necessary in its current form, as also the allocation for the array can probably be done in a smarter way.

MWE:

```Julia
julia> gradient(x -> sum(x .* 7), cu(rand(3,3))) |> typeof
Tuple{Array{Float32,2}} # should be a CuArray
```

### DIFF
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -172,4 +172,15 @@ end
     y, back = broadcast_forward(f, args...)
     y, ȳ -> (nothing, nothing, back(ȳ)...)
   end
+
+  @adjoint CuArrays.cu(xs) =
+    CuArrays.cu(xs), Δ -> (CuArrays.cu(Δ), )
+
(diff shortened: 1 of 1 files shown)


MikeInnes:
I think we should avoid the `cu` constructor and just use the lowest level ones.

It seems weird that we'd convert the gradient back to the type of the output. Why is that? I'd expect that we only need custom adjoints for cpu/gpu conversions, and then we'd likely just want to reverse the conversion.

dhairyagandhi96:
The conversion back to `Array` is the correct one. I will make update it to the right one

MikeInnes:
I think you want an `Array` type restriction on the input in that case (since if you call this function on a GPU array you don't want the gradient to be moved to the CPU).

MikeInnes:
Thanks!

bors r+

PR #383, opened 2019-10-23
update the version of IRTools in compat

findmyway:


MikeInnes:
Thanks – on your way, I think we may as well use the release of IRTools, rather than master.

findmyway:
Right, now updated.

MikeInnes:
Looks like the Travis build is failing, any idea if the manifest changes could have caused that?

findmyway:
investigating

findmyway:
The Travis build is passed now. Only the dependencies of `ITRools`  are updated.

However, it still seems very weird to me. If I just run `] up`, the `CUDAapi` seems downgraded to an old version, which caused the `has_cuda` not found [error](https://travis-ci.org/FluxML/Zygote.jl/jobs/601818138?utm_medium=notification&utm_source=github_status).

MikeInnes:
Thanks for looking into that. We can figure out the CUDA issue if it comes up again.

Issue #90, opened 2019-03-06
fastmath support


chriselrod:
```julia
julia> foo(x) = 2.0x
foo (generic function with 1 method)

julia> bar(x) = @fastmath 2.0x
bar (generic function with 1 method)

julia> Zygote.derivative(foo, 1e6)
2.0

julia> Zygote.derivative(bar, 1e6)
ERROR: Non-differentiable function IntrinsicFunction
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] (::Zygote.Pullback{Tuple{Core.IntrinsicFunction,Float64,Float64},Tuple{Core.IntrinsicFunction}})(::Int8) at /home/chriselrod/.julia/packages/Zygote/wYgcz/src/compiler/interface2.jl:0
 [3] mul_fast at ./fastmath.jl:163 [inlined]
 [4] (::Zygote.Pullback{Tuple{typeof(Base.FastMath.mul_fast),Float64,Float64},Tuple{Zygote.Pullback{Tuple{Core.IntrinsicFunction,Float64,Float64},Tuple{Core.IntrinsicFunction}}}})(::Int8) at /home/chriselrod/.julia/packages/Zygote/wYgcz/src/compiler/interface2.jl:0
 [5] bar at ./REPL[55]:1 [inlined]
 [6] (::Zygote.Pullback{Tuple{typeof(bar),Float64},Tuple{getfield(Zygote, Symbol("##265#back#163")){getfield(Zygote, Symbol("#back#162")){:mul_fast,Zygote.Context,Module,typeof(Base.FastMath.mul_fast)}},Zygote.Pullback{Tuple{typeof(Base.FastMath.mul_fast),Float64,Float64},Tuple{Zygote.Pullback{Tuple{Core.IntrinsicFunction,Float64,Float64},Tuple{Core.IntrinsicFunction}}}}}})(::Int8) at /home/chriselrod/.julia/packages/Zygote/wYgcz/src/compiler/interface2.jl:0
[...]

PR #388, opened 2019-11-01
Add Fastmath operators

Satya758:
Fixes issue #90 

1. Basic idea is to apply @adjoint macro to all defined fastmath operators.
2. This is done by looping over fastmath [operators](https://github.com/JuliaLang/julia/blob/master/base/fastmath.jl#L31) and then retrieving defined differentiation expression for each from [DiffRules](https://github.com/JuliaDiff/DiffRules.jl/blob/master/src/rules.jl).
3. Using differentiation expression create adjoint function simillar to https://github.com/FluxML/Zygote.jl/blob/master/src/lib/number.jl#L7

I have added bunch of test cases covering unary and binary operators treating normal operators as a expected value.

### DIFF
--- a/src/lib/number.jl
+++ b/src/lib/number.jl
@@ -15,7 +15,8 @@ for (M, f, arity) in DiffRules.diffrules()
     dx = :(conj($dx))
   end
   @eval begin
-    @adjoint $M.$f(x::Number) = $M.$f(x), Δ -> ($Δ * $dx,)
+    @adjoint $M.$f(x::Number) = $M.$f(x),
+      Δ -> ($Δ * $dx,)
   end
 end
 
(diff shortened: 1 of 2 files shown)


src/lib/number.jl
```diff
@@ -45,19 +46,57 @@ end
...
+
+  for f in keys(fast_op)
+    unarydx, binarydx = DiffRules.hasdiffrule(:Base, f, 1) ?
+                        DiffRules.diffrule(:Base, f, :x) : nothing,
+      DiffRules.hasdiffrule(:Base, f, 2) ?
+      DiffRules.diffrule(:Base, f, :x, :y) : nothing
+
+    fastf = fast_op[f]
+
+    if unarydx != nothing
```
MikeInnes: This seems like a strange way to write it. Why not just test `hasdiffrule` directly here?

src/lib/number.jl
```diff
@@ -45,19 +46,57 @@ end
...
+        @adjoint Base.FastMath.$fastf(x::Number, y::Number) =
+          Base.FastMath.$fastf(x, y),
+          Δ -> (Δ * make_fastmath(conj($dx)), Δ * make_fastmath(conj($dy)))
+      end
+    end
+  end
+end
 
-DiffRules._abs_deriv(x::Complex) = x/abs(x)
+fastmath()
```
MikeInnes: Why split this out rather than using a top-level loop as above?

MikeInnes:
This is very useful, thanks!

There are quite a few whitespace changes unrelated to the main issue being fixed. While many are useful it would be better to have them as a separate PR, to make review easier.

Satya758:
I have updated the code with given suggestions

src/lib/number.jl
```diff
@@ -61,3 +62,31 @@ end
 @adjoint imag(x::Number) = imag(x), ī -> (real(ī)*im,)
 
 DiffRules._abs_deriv(x::Complex) = x/abs(x)
+
+ # Function add adjoint for Fastmath operations
+ # DiffRules is used to find derivatives for each operation
+for f in keys(fast_op)
```
MikeInnes:
```suggestion
for (f, fastf) in fast_op
```

MikeInnes:
:+1: Thanks!

bors r+

PR #391, opened 2019-11-04
pkg compat + updates

MikeInnes:


andreasnoack:
Is the 0.8 bound on SepcialFunctions required. It's causing some trouble for Turing.jl so it would be great if you could allow 0.7.

willtebbutt:
Seconded. Causing problems for Stheno by way of Distributions, which (currently) also required 0.7

PR #394, opened 2019-11-06
looser bounds

MikeInnes:
bors r+


MikeInnes:
The problem here appears to be that FillArrays gets updated to 0.8, which is incompatible with GPUArrays, which forces CuArrays and CUDAapi to use a really old version instead. Pinning FillArrays should do the trick for now.

PR #398, opened 2019-11-08
Relax bound for DiffRules to 0.1 for compat with SpecialFunctions v0.8

torfjelde:
Zygote.jl already allows `SpecialFunctions@0.8` but also suffers from the this issue: https://github.com/JuliaMath/SpecialFunctions.jl/issues/186 (though different error message):
```julia
julia> using Zygote, SpecialFunctions

julia> Zygote.gradient(a -> SpecialFunctions.loggamma(a[1]), [1.0])
ERROR: TypeError: in _pullback, in ccall: first argument not a pointer or valid constant expression, expected Ptr, got Tuple{Symbol,String}
Stacktrace:
 [1] logabsgamma at /var/home/tef30/.julia/packages/SpecialFunctions/ne2iw/src/gamma.jl:606 [inlined]
 [2] loggamma at /var/home/tef30/.julia/packages/SpecialFunctions/ne2iw/src/gamma.jl:650 [inlined]
 [3] #9 at ./REPL[9]:1 [inlined]
 [4] _pullback(::Zygote.Context, ::getfield(Main, Symbol("##9#10")), ::Array{Float64,1}) at /var/home/tef30/.julia/packages/Zygote/8dVxG/src/compiler/interface2.jl:0
 [5] _pullback(::Function, ::Array{Float64,1}) at /var/home/tef30/.julia/packages/Zygote/8dVxG/src/compiler/interface.jl:31
 [6] pullback(::Function, ::Array{Float64,1}) at /var/home/tef30/.julia/packages/Zygote/8dVxG/src/compiler/interface.jl:37
 [7] gradient(::Function, ::Array{Float64,1}) at /var/home/tef30/.julia/packages/Zygote/8dVxG/src/compiler/interface.jl:46
 [8] top-level scope at none:0
```

DiffRules recently got their version bumped to support SpecialFunctions v0.8 (https://github.com/JuliaDiff/DiffRules.jl/commit/49129c0c9993a7206965ead8d6564d3afdc011cd) which fixes this issue.

EDIT: Together with `DiffRules@0.1.0`:
```julia
julia> using Zygote, SpecialFunctions

julia> Zygote.gradient(a -> SpecialFunctions.loggamma(a[1]), [1.0])
([-0.577216],)
```

PR #400, opened 2019-11-13
Bump Patch

willtebbutt:
Bump patch to propagate new DiffRules compat

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.4.1"
+version = "0.4.2"
 
 [deps]
 DiffRules = "b552c78f-8df3-52c6-915a-8e097449b14b"


MikeInnes:
ref https://github.com/FluxML/Zygote.jl/pull/398#issuecomment-553433577. I don't think we need a tag for this (but lmk if I'm missing something).

torfjelde:
@MikeInnes Could there be an issue if someone already has `Zygote@0.4.1` available locally? I.e. when I do `] instantiate` and the "old" verision of `Zygote@0.4.1` exists locally, will `Pkg` actually download the updated version?

If not, then I suggest bumping the version to `0.4.2` anyways so that packages depending on Zygote can put the bounds to `0.4.2` and know that it will work with `DiffRules@0.1`. Otherwise, seems like you could also do what you suggest:) Honestly not familiar enough with semvar to know whether or not a version-dependency-bump is suitable here or not.

MikeInnes:
Ok, I went ahead and added it to the changes [here](https://github.com/JuliaRegistries/General/pull/5135).

Issue #402, opened 2019-11-19
Adjoints for functions of specialized matrices


sdewaele:
I encountered an error with the adjoint of `getindex` of a [special
matrix](https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#Special-matrices-1). The issue is that the current `getindex` adjoints tries to assign to an element in the special matrix, which is not always allowed. It points to a more general problem/opportunity for adjoints of function of special matrices. Using `Diagonal` as an example, this is the error:
```julia
using LinearAlgebra
using Zygote
using Random
rng = MersenneTwister(54754)
n = 3
A = rand(rng,n,n)

y,B_getindex = Zygote.pullback(x->getindex(x,2,1),Diagonal(A))
bA = B_getindex(1)[1] |> display
#ArgumentError: cannot set off-diagonal entry (2, 1) to a nonzero value (1)
```
Two potential possible are:

**Solution A** : adjoint is a tuple with an unconstrained array
```julia
# Copy of getindex(D::Diagonal,...)
using LinearAlgebra:diagzero
@inline function getindex_duplicate(D::Diagonal, i::Int, j::Int)
  @boundscheck checkbounds(D, i, j)
  if i == j
      @inbounds r = D.diag[i]
  else
      r = diagzero(D, i, j)
  end
  r
end

yA,∂getindexA = Zygote.pullback(x->getindex_duplicate(x,3,3),Diagonal(A))
D̄A = ∂getindexA(1)[1]
display(D̄A)
# (diag = [0.0, 0.0, 1.0],)
```
Pro: Simple.
Con: The adjoint does not return a `Diagonal`. This reduces the efficiency of the adjoint code. This can perhaps be a fallback, as we implement solution B.

**Solution B** : Insert orthogonal projection and return a special matrix
```julia
getindexB(D::Diagonal,i::Int,j::Int) = getindex(A,i,j)
Zygote.@adjoint getindexB(D::Diagonal,i::Int,j::Int) = begin
  y,∂U = Zygote.pullback(x->getindex_duplicate(x,i,j),D)
  y, function(Ȳ)
    D̄U = ∂U(Ȳ)[1]
    D̄ = Diagonal(D̄U.diag)
    return D̄, nothing, nothing
  end
end

yB,∂getindexB = Zygote.pullback(x->getindexB(x,3,3),Diagonal(A))
D̄B = ∂getindexB(1)[1]
display(D̄B)
[...]

PR #407, opened 2019-12-01
lyap adjoint more robust to different inputs

sdewaele:
A difference between the adjoint input type and the types to the original lyap can cause issues in computing the adjoint. They are address by converting the adjoint argument to the original type.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -447,10 +447,10 @@ end
   end
 end
 
-@adjoint function lyap(A::AbstractMatrix, C::AbstractMatrix)
+Zygote.@adjoint function lyap(A::T, C::T) where {T<:AbstractMatrix}
   X = lyap(A, C)
   return X, function (X̄)
-    C̄ = lyap(collect(A'), X̄)
+    C̄ = lyap(collect(A'), convert(T,X̄))
     Ā = C̄*X' + C̄'*X
     return (Ā, C̄)
   end
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -447,10 +447,10 @@ end
   end
 end
 
-Zygote.@adjoint function lyap(A::T, C::T) where {T<:AbstractMatrix}
+Zygote.@adjoint function lyap(A::StridedArray, C::StridedArray)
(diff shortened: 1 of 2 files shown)


MikeInnes:
Might be good to have a test to demonstrate this specific issue, which the current one doesn't seem to (?). Also, I wonder if it'd be better to do the conversion the other way around, i.e. convert the input to the type of the gradient, since this tends to come up because of numeric promotion, and the gradient will have the wider type. Open to discussion on that though.

sdewaele:
I had an issue with this that occurred when using indexing. However, this issue does not occur on master anymore... This did make me realize that this adjoint could fail as a result of changing types.

I'll think about how to change the types in response to your comment.

sdewaele:
An issue occurs after fixing #402 using commit https://github.com/FluxML/Zygote.jl/commit/73fdfa9db68941509937d76dcddc98aecdf421dd.
In the example below, I copy the code for `_zero` from this commit, specialized to `Diagonal`.

```julia
using Zygote
using LinearAlgebra
using Random
rng = MersenneTwister(7850147)
n = 3 
A = rand(rng,n,n)
C = rand(rng,n,n)
X = lyap(A,C)

# `_zero` based on that in commit https://github.com/FluxML/Zygote.jl/commit/73fdfa9db68941509937d76dcddc98aecdf421dd
Zygote._zero(xs::Diagonal, T=float(eltype(xs))) = fill!(similar(xs, T, size(xs)), false)

function f(A)
  X = lyap(A,C)
  Xd = Diagonal(X)
  return Xd[1,1]
end

g = gradient(f,A)[1]
```
yields the error:
```julia
ERROR: LoadError: MethodError: no method matching lyap(::Array{Float64,2}, ::Diagonal{Float64,SparseArrays.SparseVector{Float64,Int64}})
```
This is caused by the fact that a `similar(xs, T, size(xs))` returns a sparse array.

CarloLucibello:
The example above still gives an error. It would be good to rebase this PR or even better to open a corresponding one in ChainRules.jl

CarloLucibello:
closing as outdated

Issue #376, opened 2019-10-18
getindex for real input fails with complex sensitivity


sethaxen:
`getindex`'s adjoint assumes the sensitivity is of the same type as the input. This causes `InexactError`s when the input to `getindex` is real but the sensitivity is complex.

```julia
julia> using Zygote
[ Info: Precompiling Zygote [e88e6eb3-aa80-5325-afca-941959d7151f]

julia> x = randn(2)
2-element Array{Float64,1}:
 -1.5316563904446463
 -0.7974154569041462

julia> y, back = Zygote._pullback(x->x[1]*im, x)
(-0.0 - 1.5316563904446463im, ∂(#7))

julia> back(1.0)
ERROR: InexactError: Float64(0.0 - 1.0im)
Stacktrace:
 [1] Real at ./complex.jl:37 [inlined]
 [2] convert at ./number.jl:7 [inlined]
 [3] setindex! at ./array.jl:780 [inlined]
 [4] #838 at /Users/saxen/.julia/packages/Zygote/lRotY/src/lib/array.jl:32 [inlined]
 [5] #2174#back at /Users/saxen/.julia/packages/ZygoteRules/Mmoki/src/adjoint.jl:49 [inlined]
 [6] literal_getindex at /Users/saxen/.julia/packages/Zygote/lRotY/src/lib/lib.jl:77 [inlined]
 [7] #7 at ./REPL[7]:1 [inlined]
 [8] (::typeof(∂(#7)))(::Float64) at /Users/saxen/.julia/packages/Zygote/lRotY/src/compiler/interface2.jl:0
 [9] top-level scope at REPL[8]:1
```

The culprit appears to be these lines:
https://github.com/FluxML/Zygote.jl/blob/def6b45724daa2d05d4794212a9f6baa86a0fc97/src/lib/array.jl#L31-L32

MikeInnes:
There's no great reason to use `setindex!` here except that there's no convenient out-of-place `setindex`. This would be a good reason to implement that function.

sdaxen:
You mean to implement for arrays in Julia base or to have our own `_setindex` for that purpose?

MikeInnes:
Ideally, it would be in base, but that will take too long; it'd be fine to have something in Zygote and switch out to Base if it ever gets added.

sdaxen:
Okay, I guess a simple Zygote-friendly `setindex` default would be something like this:

```
function Base.setindex(A::AbstractArray, X, inds...)
[...]

Issue #342, opened 2019-09-26
Complex gradient on real function with complex intermediates


sethaxen:
I came across something odd while working with complex numbers:

```julia
julia> using Zygote
julia> Zygote.gradient(x->real(x + 2.0*im), 3.0)
(1.0,)
julia> Zygote.gradient(x->imag(x + 2.0*im), 3.0)
(0.0 + 1.0im,)
```

While the inputs and outputs of both functions are real, the first produces a real gradient, while the second produces a complex gradient. Is this Zygote's intended behavior?

Version info:

```julia
julia> versioninfo()
Julia Version 1.2.0
Commit c6da87ff4b (2019-08-20 00:03 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin18.6.0)
  CPU: Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, haswell)

(v1.2) pkg> st --manifest Zygote
    Status `~/.julia/environments/v1.2/Manifest.toml`
  [b552c78f] DiffRules v0.0.10
  [7a1cc6ca] FFTW v0.3.0
  [1a297f60] FillArrays v0.6.4
  [f6369f11] ForwardDiff v0.10.3
  [7869d1d1] IRTools v0.2.3
  [1914dd2f] MacroTools v0.5.1
  [872c559c] NNlib v0.6.0
  [77ba4419] NaNMath v0.3.2
  [ae029012] Requires v0.5.2
  [276daf66] SpecialFunctions v0.7.2
  [e88e6eb3] Zygote v0.3.4
  [700de1a5] ZygoteRules v0.1.0
  [b77e0a4c] InteractiveUtils 
  [37e2e46d] LinearAlgebra 
  [9a3f8284] Random 
  [10745b16] Statistics 
```

sdaxen:
FWIW, Tracker and ForwardDiff do what I'd expect:

```julia
julia> using Tracker, ForwardDiff
julia> Tracker.gradient(x->imag(x + 2.0*im), 3.0)[1]
0.0 (tracked)
julia> ForwardDiff.derivative(x->imag(x + 2.0*im), 3.0)
0.0
```

sdaxen:
For anyone coming across this, a workaround is to directly call `complex`:

```julia
[...]

PR #409, opened 2019-12-03
matrix exp(A) returns real-valued adjoint for real-valued A

sdewaele:
The current adjoint for the matrix `exp(A)` where `A` is real-valued can return a complex-valued matrix. The complex components are only small numerical errors, close to machine precision. However, the fact that it is complex-valued leads to issues in the adjoint computation, see below for an example. This PR converts the return value to a real-valued array for real-valued arguments.

BTW - This issue has some similarities to #402, in that
1) Having indexing in the test code revealed the issue. For that reason, it may be an idea to add a `gradcheck_getindex`, to complement the current `gradcheck`? Or a test of compatibility of the return value of the adjoint with the input type?

2) An alternative fix would be to allow the adjoint of the indexed array to be complex-valued. This is similar to allowing unconstrainted matrices as adjoints in #402. This would also require changes to the `getindex` adjoint.

```julia
# Note: A randomly generated 8 × 8 also fails (almost?) always. Using this confirmed example to be sure.
A = [ 0.0    1.0    0.0
      0.0    0.0    1.0
     -4.34 -18.31  -0.43]
x = [1.0]

f(A,x) = exp(A*x[1])
Y,Bf = Zygote.pullback(f,A,x)
Ȳ = ones(3,3)
Ā,x̄ = Bf(Ȳ) # fails
```
Error:
```julia
ERROR: InexactError: Float64(14.999150171071268 + 9.678155693021492e-16im)
Stacktrace:
 [1] Type at .\complex.jl:37 [inlined]
 [2] convert at .\number.jl:7 [inlined]
 [3] setindex!(::Array{Float64,1}, ::Complex{Float64}, ::Int64) at .\array.jl:767
 [4] #980 at C:\Users\user\.julia\packages\Zygote\ycnjm\src\lib\array.jl:32 [inlined]
 [5] #2619#back at C:\Users\user\.julia\packages\ZygoteRules\6nssF\src\adjoint.jl:49 [inlined]
 [6] literal_getindex at C:\Users\user\.julia\packages\Zygote\ycnjm\src\lib\lib.jl:77 [inlined]
 [7] (::typeof(∂(Zygote.literal_getindex)))(::Complex{Float64}) at C:\Users\user\.julia\packages\Zygote\ycnjm\src\compiler\interface2.jl:0
 [8] f at .\REPL[7]:1 [inlined]
 [9] (::typeof(∂(f)))(::Array{Float64,2}) at C:\Users\user\.julia\packages\Zygote\ycnjm\src\compiler\interface2.jl:0
 [10] (::getfield(Zygote, Symbol("##28#29")){typeof(∂(f))})(::Array{Float64,2}) at C:\Users\user\.julia\packages\Zygote\ycnjm\src\compiler\interface.jl:38
 [11] top-level scope at none:0
```


mcabbott:
One issue here is whether real parameters should by default be allowed to return complex gradients, as is the current behaviour, see #342 for a discussion. 

The other is that, given this behaviour, the gradient of `getindex` should promote as necessary. This is precisely #376, which is fixed by #256.

src/lib/array.jl
```diff
@@ -485,7 +485,8 @@ end
   X = _pairdiffquotmat(exp, n, w, ew, ew, ew)
   V = E.vectors
   VF = factorize(V)
-  Ā = (V * ((VF \ F̄' * V) .* X) / VF)'
+  Āc = (V * ((VF \ F̄' * V) .* X) / VF)'
+  Ā = eltype(A) <: Real ? real(Āc) : Āc
```
mcabbott:
```suggestion
  Ā = eltype(A) <: Real && eltype(F̄) <: Real ? real(Āc) : Āc
```

sdewaele:
Thanks for your comments! I will update the PR accordingly, and fix the test issue.

src/lib/array.jl
```diff
@@ -485,7 +485,8 @@ end
   X = _pairdiffquotmat(exp, n, w, ew, ew, ew)
   V = E.vectors
   VF = factorize(V)
-  Ā = (V * ((VF \ F̄' * V) .* X) / VF)'
+  Āc = (V * ((VF \ F̄' * V) .* X) / VF)'
+  Ā = eltype(A) <: Real && eltype(F̄) <: Real ? real(Āc) : Āc
```
MikeInnes: Would `isreal(A) && isreal(F)` work?

sdewaele: Yes. I added it. Thanks!

sdewaele:
The check [failed](https://travis-ci.org/FluxML/Zygote.jl/jobs/625158623#L254) for Julia 1.3 with error 
```
No output has been received in the last 10m0s,
this potentially indicates a stalled build or something wrong with the build itself.
```
I am not sure why this is happening; when I test on my machine with Julia 1.3 running the tests takes 5 minutes. Versioninfo:
```
julia>  versioninfo()
Julia Version 1.3.0
Commit 46ce4d7933 (2019-11-26 06:09 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, skylake)
```

MikeInnes:
Travis probably just has a slow CPU. Bors should be fine:

bors r+

Issue #98, opened 2019-03-08
Jacobian and Hessian


MikeInnes:
Should be easy to port from [Tracker's versions](https://github.com/FluxML/Tracker.jl/blob/bd98dab59026eb753c296a5f19367a408c2cb8b6/src/back.jl#L174).

baggepinnen:
The current hessian function gives stack overflow on the following
```julia
using Flux
inp     = randn(30)      
f = Chain(Dense(30, 30, tanh),Dense(30, 30, tanh),Dense(30, 30, tanh),Dense(30, 1, tanh))
Zygote.hessian(f, inp)
```

gdkrmr:
I need a jacobian, this works:
```julia
function jacobian(f, x)
    y = f(x)
    n = length(y)
    m = length(x)
    T = eltype(y)
    j = Array{T, 2}(undef, n, m)
    for i in 1:n
        j[i, :] .= gradient(x -> f(x)[i], x)[1]
    end
    return j
end
```
I am sure this can still be optimized but I don't know enough about the internals.

ferrolho:
I am trying to get the Jacobian of a `R^n -> R^m` function, where `n > 1` and `m > 1` (vector input and vector output); for example, a function like `func_1` below.

```julia
function func_1(x::Array{T}) where T  # out-of-place
    dx = zeros(T, 3)
    dx[1] = 2x[1] + √x[3]
    dx[2] = 3x[1] + sin(x[4])
    dx[3] = 4x[1] + cos(x[5])
    dx
end
```

When I try to get the Jacobian with the function posted above by @gdkrmr with `jacobian(func_1, rand(5))` I get the following error:

```zsh
Mutating arrays is not supported

Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] (::getfield(Zygote, Symbol("##794#795")))(::Nothing) at /home/henrique/.julia/packages/Zygote/bdE6T/src/lib/array.jl:37
 [3] (::getfield(Zygote, Symbol("##2078#back#796")){getfield(Zygote, Symbol("##794#795"))})(::Nothing) at /home/henrique/.julia/packages/ZygoteRules/Jn4LO/src/adjoint.jl:48
 [4] func_1 at ./In[16]:5 [inlined]
 [5] (::typeof(∂(func_1)))(::Array{Float64,1}) at /home/henrique/.julia/packages/Zygote/bdE6T/src/compiler/interface2.jl:0
[...]

Issue #51, opened 2019-01-01
How to calculate Jacobians (or just derivatives of vector functions) with Zygote?


Krastanov:
Why does
```julia
Zygote.forward(a -> [2.,1.]*a, 1)[2](1)
```
return `3` instead of `[2,1]`?

Am I missing a big philosophical or computational reason for the omission of non-scalar functions? Zygote is doing a form of forward AD, so the derivative of a function that maps a scalar to a vector should be cheap, right?

MikeInnes:
Zygote is reverse mode, so the output of `forward` calculates a vector-Jacobian product. For forward you should probably check out ForwardDiff.jl.

It would be good to add utilities for full Jacobians though.

MikeInnes:
Incidentally it's easy to do forward-over-reverse with Zygote's API.

```julia
using Flux

y, back = Tracker.forward(x -> [3x^2, 5x+2], 5)

Tracker.gradient(x -> back(x)[1], [1, 1]) # [30, 5]
```

This probably won't actually work with Zygote yet, since nested AD doesn't work in all cases, but it will eventually.

PR #414, opened 2019-12-05
A Jacobian which understands Params

### LINKED ISSUE #413: Jacobian in Zygote with Params
I simply love the Flux+Zygote flexibility for extracting parameters from object and get the appropriate gradients directly in the same structure (`Grads`).
However it would be nice to have the same feature for `jacobian`s with `Params`. I saw the already open issues #98 and #51. But the proposed workaround is only valid for arrays.

mcabbott:
Closes #413, and #51, and half of #98. 

Questions:

* Should `jacobian(f, ps::Params)` return something other than an `IdDict()`?

* Should `matrix=true` be the default? Needs to be added to Params method.

Needs tests obviously.

### DIFF
--- a/src/lib/utils.jl
+++ b/src/lib/utils.jl
@@ -79,3 +79,63 @@ a real.
      1  0
 """
 hessian(f, x::AbstractArray) = forward_jacobian(x -> gradient(f, x)[1], x)[2]
+
+"""
+    jacobian(f, args...)
+
+If `y = f(args...)` is a vector, then for each vector `a ∈ args`
+this returns a matrix with `J_a[o,i] = ∂yₒ/∂aᵢ`.
+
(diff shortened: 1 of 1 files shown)


Roger-luo:
XRef: I implemented one in #188 as well but I didn't do anything on Params

mcabbott:
Oh I completely missed that, sorry. Perhaps something to handle Params should just be bolted onto that.

theogf:
@mcabbott I am trying to use your solution until #235 is merged and has a `Params` option, but I am confused by what you refer to when using `Zygote.forward`, is it `Zygote.forwarddiff`?

mcabbott:
Ah no, sorry, it's now called `Zygote.pullback`. The name changed in v0.4 I think.

theogf:
Thanks! It's working perfectly now!

PR #415, opened 2019-12-06
Make the definition of the pullback more precise

ChrisRackauckas:
B_y(ybar) had extra embedded information that isn't in the definition of the function. Higher order AD concerns can ask things like "what is the derivative of B_y(ybar) with respect to x?" which seemed to confuse some 18.337 students since x is left off in that notation. The notation B_f^x(ybar) on y=f(x) makes it explicit that there is an x-dependence in the pullback function, and this corresponds to how information about the forward pass is embedded in the closures that Zygote is generating. This notation tweak is quite small but seemed to make it a lot more clear, at least to me and for the way I was teaching it, so I thought I'd upstream it back to Zygote.

### DIFF
--- a/docs/src/adjoints.md
+++ b/docs/src/adjoints.md
@@ -13,17 +13,21 @@ julia> y
 0.479425538604203
 ```
 
-`pullback` gives two outputs: the result of the original function, `sin(0.5)`, and a *pullback*, here called `back`. `back` implements the gradient computation for `sin`, accepting a derivative and producing a new one. In mathematical terms, it implements a vector-Jacobian product. Where ``y = f(x)`` and the gradient ``\frac{\partial l}{\partial x}`` is written ``\bar{x}``, the pullback ``\mathcal{B}_y`` computes:
+`pullback` gives two outputs: the result of the original function, `sin(0.5)`, and a *pullback*, here called `back`. `back` implements the gradient computation for `sin`, accepting a derivative and producing a new one. In mathematical terms, it implements a vector-Jacobian product. Where ``y = f(x)`` and the gradient ``\frac{\partial l}{\partial x}`` is written ``\bar{x}``, the pullback ``\mathcal{B}_f^x`` computes:
 
 ```math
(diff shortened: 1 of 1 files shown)


MikeInnes:
The current notation is already precise, but a bit hidden in the text and code (`B_f = back = J(f, x)`, which clearly shows dependence on `x`). I think it'd be useful to add the latex version of that line to make it clearer, like in the paper, and then it's not necessary to make the pullback notation noisier.

ChrisRackauckas:
>gradient takes a function l = f(x) and assumes l̄ = \frac{\partial l}{\partial l} = 1 and feeds this in to the pullback.

I also really don't like this explanation because it relies on knowing the context in standard backpropagation speak. I think it's much more clear to understand vjps as directional derivatives which pull out rows of Jacobians, and an R^n -> R function has a Jacobian of 1 row so the vjp with value 1 is the gradient. For this reason, I wouldn't put the vjp discussion below a

>The rest of this section contains more technical detail. It can be skipped if you only need an intuition for pullbacks; you generally won't need to worry about it as a user.

and would feature vjps more precisely. It makes sparse reverse-mode and Hessian-vector products much easier to define!

ChrisRackauckas:
>The current notation is already precise, but a bit hidden in the text and code (B_f = back = J(f, x), which clearly shows dependence on x).

Yes, this version from the newer diff prog papers is precise, so if you prefer the `J` form more than the docs should at least be updated to that. `J(f,x)(ybar)` is then fairly close to code.

MikeInnes:
The docs above that line are in fact clear that the pullback is a vector-Jacobian product, but only in passing because it's not aimed at mathematicians. In order to be helpful for users in general it has to target someone who is familiar with the idea of a derivative, but not necessarily with things like Jacobians, so starting with the more advanced thing and working backwards isn't the right way to do it. Users really don't need to understand sparse-reverse mode or HVPs to write their own adjoints.

CarloLucibello:
closing as outdated

PR #423, opened 2019-12-12
Add CuArray convert adjoint

dhairyagandhi96:
Fixes FluxML/Flux.jl#947

### DIFF
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -187,4 +187,8 @@ end
     sum(xs, dims = dims), Δ -> (placeholder .= Δ,)
   end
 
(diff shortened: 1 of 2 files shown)


src/lib/broadcast.jl
```diff
@@ -187,4 +187,8 @@ end
     sum(xs, dims = dims), Δ -> (placeholder .= Δ,)
   end
 
+  @adjoint function Base.convert(::Type{T}, xs::Array{K,N})  where {T<:CuArrays.CuArray, K, N}
```
MikeInnes: The `K, N` could be removed here, right? Otherwise LGTM.

dhairyagandhi96: Should be fixed now, apologies for the delay

PR #430, opened 2019-12-18
Accumulate over Params

DhairyaLGandhi:
Implicit parameters can be tricky to reason with, since some cases may need to collect the gradients from non bitstype cases (https://github.com/JuliaDiffEq/DiffEqFlux.jl/pull/96), such as how we have in the `Params` object. We should really start deprecating the implicit params soon.

Good to get some eyes on this.

### DIFF
--- a/src/lib/lib.jl
+++ b/src/lib/lib.jl
@@ -49,6 +49,13 @@ end
   end
 end
 
+function accum_param(cx::Context, x::Vector{<:AbstractArray}, Δ)
+  for (p,g) in zip(x, Δ)
+    haskey(cache(cx), p) && (cache(cx)[p] = accum(cache(cx)[p],g))
+  end
+  nothing
+end
+
 function accum_global(cx::Context, ref, x̄)
   gs = globals(cx)
   gs[ref] = accum(get(gs, ref, nothing), x̄)


MikeInnes:
It's not totally clear to me why this is needed, and it'd be good to have a test case. The semantics of `accum_param` should be that it's called on anything that is itself a param, not things that are collections of params.

Separately, it's not that likely that we'll deprecate implicit params any time soon. It needs to happen in tandem with an overhaul of Flux's API and particularly optimisers that we don't have a ton of bandwidth for right now.

dhairyagandhi96:
Right, but in general, collections of params that don't get accumulated cause incorrect gradients to be reported. Similar in vein to having the params in a Tuple or some such. I'll come up with a simpler reproducer.

MikeInnes:
I see. The way this works for structs is that when you do a `getfield`, we call `accum_param` on the field at that point. So to solve this similarly for arrays and tuples we'd have to add the same `accum_param` call in the adjoints for (scalar) `getindex`.

The current solution could work in principle, but it's weird that it makes accessing arrays from global scope an `O(n)` operation.

CarloLucibello:
close as stale / missing clear description / missing tests

PR #444, opened 2020-01-05
Use dispatch for accum nothing cases

oxinabox:
Very minor change.
It will compile down to the same code.
But I find this easier to read.

* technically there is an edge case, that is someone had defined a non-`Nothing` value to be `==nothing` then this would change things.
But I don't think we want to deal with that case like this.

### DIFF
--- a/src/lib/lib.jl
+++ b/src/lib/lib.jl
@@ -5,10 +5,10 @@ using Base: RefValue
 accum() = nothing
 accum(x) = x
 
-accum(x, y) = x + y
-accum(::Nothing, y) = y
-accum(x, ::Nothing) = y
-accum(::Nothing, ::Nothing) = nothing
+accum(x, y) = 
+  x === nothing ? y :
+  y === nothing ? x :
+  x + y
 
 accum(x, y, zs...) = accum(accum(x, y), zs...)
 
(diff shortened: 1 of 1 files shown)


MikeInnes:
This is the same if the types are all known but not if they aren't. If they aren't I suspect a branch can be significantly quicker than a method dispatch (though I haven't tested this carefully).

(Changing `==` to `===` would be fine though).

oxinabox:
> This is the same if the types are all known but not if they  aren't. If they aren't I suspect a branch can be significantly quicker  than a method dispatch (though I haven't tested this carefully).

All types are known by the type you reach your branch.
Because there is the function barrier.

Inside a function, the type of all your arguments are known.
If you do some (e. nontype inferable) computation on them the result of that might not be known.

oxinabox:
I guess the dispatch might optimize for this case?
And then avoid this being a dynamic dispatch to ` accum`.

But there are other methods for `accum` defined.
So it needs to work out that it isn't hitting one of them.

MikeInnes:
When Zygote infers it makes heavy use of `Union{T,Nothing}`; in principle `accum` can be inlined and specialised on that type. I don't know if it actually does this (it probably doesn't), but a dispatch version would preclude without much cleverer inlining of dispatch than we have. (And of course that doesn't work for tuples etc. for reasons you have pointed out.)

oxinabox:
Ok, lets play it safe.
I am dubious as to if that optimization exists, and even if it did exist, as to if it would be hit in the case of `accum` (so many methods to branch against potentially).

But sure

PR #447, opened 2020-01-06
Push for buffers

MikeInnes:


dhairyagandhi96:
Would we need a similar construct for IdSet?

MikeInnes:
Not sure; `IdSet`s might be differentiable already since they just use `IdDict`s under the hood. But if we have a use case for it we can certainly take a look.

PR #452, opened 2020-01-09
Add adjoint for `pairs(::NamedTuple)` and `Pair`

YingboMa:
Fix https://github.com/JuliaDiffEq/DiffEqFlux.jl/issues/105


YingboMa:
```julia
julia> @nograd pairs

julia> Zygote.gradient(()->sum(neural_ode(dudt,x,tspan,Tsit5(),save_everystep=false,save_start=false,sensealg=BacksolveAdjoint())),p)
Grads(...)
```

src/lib/base.jl
```diff
@@ -104,3 +104,14 @@ end
 # Make @sync work
 # Safe as long as other kinds of mutation are disallowed
 @adjoint push!(refs::Vector{Any}, t::Task) = push!(refs, t), _ -> nothing
+
+# named tuple
+@adjoint function pairs(t::NamedTuple{N}) where N
+    fun = function pairs_namedtuple(Δ)
```
MikeInnes: You can just refer to the function as `pairs_namedtuple`, so you don't need the assignment to `fun`.

MikeInnes:
Thanks! FWIW, a minimal reproducer is

```julia
julia> foo(;kw...) = 1
foo (generic function with 1 method)

julia> gradient(() -> foo(a=1,b=2.0))
ERROR: Can't differentiate gc_preserve_end expression
```

This works if all kwargs are of the same type; it would be great to understand why that is, to make sure we've got the whole issue fixed.

YingboMa:
CI failure looks unrelated. @MikeInnes could you review again?

MikeInnes:
The patch looks good to me. I'd still like to understand what when wrong in detail before merging, but I can look into that myself if it's not already clear. 

bors try

CarloLucibello:
@MikeInnes what should we do here?

cossio:
bump

CarloLucibello:
related to #669 , but of the tests in this PR
```julia
    @test (x->10*pairs((a=x, b=2))[1])'(100) === 10
    @test (x->10*pairs((a=x, b=2))[2])'(100) === 0
    foo(;kw...) = 1
    @test gradient(() -> foo(a=1,b=2.0)) === ()
    @test (x->10*(x => 2)[1])'(100) === 10
    @test (x->10*(x => 2)[2])'(100) === 0
```
only the 3rd passes on current master, so we should merge this as well

CarloLucibello:
@YingboMa could your rebase on master and remove the pairs adjoint from #669 ?

src/lib/base.jl
```diff
@@ -115,3 +115,11 @@ end
     end
     return pairs(t), pairs_namedtuple
 end
+
+@adjoint function Base.getfield(p::Pair, i::Int)
```
Yansf677, line 121: How to write the adjoint method for LinearAlgebra.diagm of Pair{Int64,Array{Float64,1}}

PR #459, opened 2020-01-14
Limit maximum and minimum adjoints to arrays

mohamed82008:
This PR limits the adjoint definition of `minimum` and `maximum` to arrays. `minimum` and `maximum` are also defined for distributions in `Distributions.jl`, so the general adjoint was causing troubles.


MikeInnes:
Seems sensible, thanks. 

bors r+

PR #461, opened 2020-01-14
Include rank in conv and pooling gradcheck testset names

oxinabox:
When these tests fail, it is useful to know what rank was being tested when it failed

### DIFF
--- a/test/gradcheck.jl
+++ b/test/gradcheck.jl
@@ -112,23 +112,21 @@ end
 end
 
 @testset "conv: spatial_rank=$spatial_rank" for spatial_rank in (1, 2, 3)
-    x = rand(repeat([10], spatial_rank)..., 3, 2)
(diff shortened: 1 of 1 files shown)


MikeInnes:
I thought when these fail it already prints something like `spatial_rank=2` in the test report?

oxinabox:
nope.
Example failure:

```
Test Failed at /Users/oxinabox/JuliaEnvs/ChainRulesWorld/Zygote.jl/proto.jl:37
  Expression: gradtest(((x, w)->begin
            conv(x, w, cdims)
        end), x, w)
```

MikeInnes:
That seems equally problematic since the `conv` string isn't shown there either. When I ran it that string appeared at the beginning though. Could've sworn I saw the iteration being printed somewhere but I can't seem to reproduce that now.

bors r+

PR #465, opened 2020-01-16
Adds a nograd function to encapsulate an operation without gradient

cossio:
As suggested by @MikeInnes, the idea is to be able to write something like this:

```julia
A = randn(5)
x = randn(5)

function f(x)
      nograd() do
          A .= x
      end
      sum(A + x)
end

f'(x) # returns [1, 1, 1, 1, 1]
```

Even though `f(x)` is computing `sum(2x)`, the gradient is a vector of ones because `A` is treated as a constant. This way we are able to mutate arrays for which we don't need a gradient.

### DIFF
--- a/src/lib/grad.jl
+++ b/src/lib/grad.jl
@@ -1,5 +1,11 @@
 using MacroTools: @q
 
+"""
+    @nograd(f...)
+
+The output of the argument functions will be considered as constant
+and will not contribute gradients.
+"""
 macro nograd(ex)
   isexpr(ex, :tuple) || (ex = Expr(:tuple, ex))
   blk = @q begin end
@@ -9,6 +15,23 @@ macro nograd(ex)
   end
   return blk
 end
+
+"""
+    nograd(x)
+
+Treats `x` as a constant. For example:
+
+```julia-repl
+julia> f(x) = sum(nograd(x) + x); gradient(f, randn(5))
+([1.0, 1.0, 1.0, 1.0, 1.0],)
+```
+
+In the example above, the gradient is an array of ones even though
+`f(x)` computes `sum(2x)`, because `nograd(x)` uses the value of `x`
+as if it were a constant.
+"""
+nograd(x) = x
+@nograd nograd
 
 macro which(ex)
   @capture(ex, f_(args__)) || error("Zygote.@which f(args...)")
--- a/test/gradcheck.jl
+++ b/test/gradcheck.jl
@@ -1312,6 +1312,11 @@ end
   @test gradient(x -> findall(ismissing, x)[1], [1, missing]) == (nothing,)
 end
 
+@testset "nograd" begin
+  f(x) = sum(Zygote.nograd(x) + x)
+  @test gradient(f, randn(5)) == (ones(5),)
+end
+
 @testset "fastmath" begin
   @test gradient(x -> begin @fastmath sin(x) end, 1) == gradient(x -> sin(x), 1)
   @test gradient(x -> begin @fastmath tanh(x) end, 1) == gradient(x -> tanh(x), 1)
--- a/docs/src/utils.md
+++ b/docs/src/utils.md
@@ -11,4 +11,5 @@ Zygote.dropgrad
 Zygote.hessian
 Zygote.Buffer
 Zygote.forwarddiff
+Zygote.nograd
 ```


src/lib/grad.jl
```diff
@@ -1,5 +1,11 @@
 using MacroTools: @q
 
+"""
+  @nograd(f...)
```
MikeInnes: I think this needs to be four spaces to be recognised as code.

src/lib/grad.jl
```diff
@@ -10,6 +16,27 @@ macro nograd(ex)
...
+Calls `f()` without accumulating gradients.
+
+    julia> A = randn(5);
+    julia> function f(x)
+             nograd() do
+               A .= x
+             end
+             return sum(A + x)
+           end
+    julia> julia> x = randn(5); println(f'(x))
```
MikeInnes: Extra julia prompt here

cossio:
The name `ignore` is too generic? I think `nograd` is more explicit. Or maybe `ignoregrad`?

I would also like to be able to do something like:

```
@nograd A .= x
```

But I wouldn't not know how to write that macro. In that case we would definitely need a different name to not collide with the existing `@nograd`.

src/lib/grad.jl
cossio: Fixed.

src/lib/grad.jl
cossio: Fixed.

cossio:
Merge?

MikeInnes:
My thinking was that if you write `Zygote.ignore() do ...` it's not particularly ambiguous; you're telling Zygote to ignore the gradient. (Of course, if you want a shorter version you can import it explicitly, and it's still clear where it comes from). Not adamant about that if you think `nograd` is better though.

`@nograd ...` could be implemented by lowering to `nograd() do ...` pretty easily, I think.

cossio:
> `@nograd ...` could be implemented by lowering to `nograd() do ...` pretty easily, I think.

Overwriting the current definition of `@nograd`?

MikeInnes:
I just meant that that would be a path to implementing the macro; I'm not taking any stance on the naming or on overwriting `@nograd`.

src/lib/grad.jl
```diff
@@ -10,6 +16,27 @@ macro nograd(ex)
   return blk
 end
 
+"""
+    nograd(f)
+
+Calls `f()` without accumulating gradients.
+
+    julia> A = randn(5);
```
CarloLucibello: can you give an example not mutating globals (it doesn't feel very julian). Maybe just place A within f(x)?

test/gradcheck.jl
```diff
@@ -1219,6 +1219,18 @@ end
...
 
+@testset "nograd" begin
+  let n = 5, A = randn(n), x = randn(n)
+    function f(x)
+      Zygote.nograd() do
+        A .= x
+      end
+      sum(A + x)
+    end
+    @test f'(x) == ones(n)
```
CarloLucibello: let's use `gradient`  instead of `'` as we may deprecate it in the future

CarloLucibello:
This could also be used as the standard way to "detach" part of the computational graph:
```julia
julia> function f2(x)
            y = nograd(() -> x) 
            return sum(y + x)
       end
julia> x = randn(5); println(f2'(x))
[1.0, 1.0, 1.0, 1.0, 1.0]
```
It may be convenient to specialize `nograd` for `AbstractArrays` as follows
```
nograd(f) = f()
nograd(x::AbstractArrays) = x
@nograd nograd

julia> function f3(x)
            y = nograd(x) 
            return sum(y + x)
       end
julia> x = randn(5); println(f3'(x))
[1.0, 1.0, 1.0, 1.0, 1.0]
```

cossio:
> It may be convenient to specialize `nograd` for `AbstractArrays` as follows

Actually, it might be better to just define `nograd` like this:

```
nograd(x) = x
@nograd nograd
```

Then your example works and we don't have to do the weird wrapping in the do-block. We just call `nograd` on *values* that we want to use as constants. This seems more readable to me. What do you think? I can rewrite `nograd` like this?

src/lib/grad.jl
cossio: Fixed.

test/gradcheck.jl
cossio: Fixed

src/lib/grad.jl
```diff
@@ -10,6 +16,22 @@ macro nograd(ex)
...
 end
 
+"""
+    nograd(x)
+
+Treats `x` as a constant. For example:
+
+    julia> f(x) = sum(nograd(x) + x)
+    julia> x = randn(5); gradient(f, x)
+    ([1.0, 1.0, 1.0, 1.0, 1.0],)
```
CarloLucibello: use a non-indented ```julia-repl  block here

CarloLucibello:
I think this last proposal is fine. 

> The downside is that it no longer helps with array mutation, but maybe that is not so important.

You can still wrap your code into a mutating function `f!` and then `@nograd f!`. Also, in the future we could add a `@nograd` block: 

```julia
@nograd begin
..... 
mutating and non-mutating operations
....
end
```

src/lib/grad.jl
cossio: Done

CarloLucibello:
thanks!

bors r+

CarloLucibello:
wait, I just found out that we already have `Zygote.dropgrad`, documented in the Utils section and unexported, doing exactly the same thing

cossio:
That's right! I missed `dropgrad`

MikeInnes:
Yes, the original proposal in this issue provides something different from `dropgrad` (which doesn't help with the array mutation issues discussed in the OP). With the definition that's been merged, it's redundant.

PR #466, opened 2020-01-17
Exception handlers (happy path only)

MikeInnes:
In theory, all we have to do to support exceptions now is to stop complaining about them. Unfortunately the following test case is crashing, so further investigation is needed:

```julia
function foo(f)
  y = 1
  try
    y = f()
  catch
    y
  end
  y
end

gradient(x -> foo(() -> x), 1) # 1
pullback(foo, () -> 0//0) # segfault
```

We could just be violating some `enter`/`leave` invariant in lowered code.

For now, we won't support differentiating through `catch` blocks at all. Full support for exceptions would require us to build some kind of continuation-like object as we unwind the stack; that's heavy machinery that will likely slow everything else down. An obvious compromise is to drop the gradient for the exception itself so that we can ignore all operations further down the stack; unfortunately, while this works for pure programs, in Julia the stack might have carried out mutations that are visible elsewhere. Dropping those gradients might be an acceptable compromise, or it might lead to extremely hard-to-debug incorrect gradients. For now, refusing to go there at all is a safe option.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -23,9 +23,9 @@ version = "0.2.0"
 
 [[DataStructures]]
 deps = ["InteractiveUtils", "OrderedCollections"]
-git-tree-sha1 = "f784254f428fb8fd7ac15982e5862a38a44523d3"
+git-tree-sha1 = "b7720de347734f4716d1815b00ce5664ed6bbfd4"
 uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
-version = "0.17.7"
+version = "0.17.9"
 
 [[Dates]]
 deps = ["Printf"]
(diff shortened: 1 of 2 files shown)


oxinabox:
Don't we need to still throw an error if we enter a catch block?
Otherwise we can get some very hard to debug incorrect gradients

```
function foo(x)
    y=2x
    try
        y*=asin(x)  # DomainError may be thrown here if   !*(-1 < x < 1)
        y*=10
    catch
        y*=1000
    end
    return y
end

foo'(1.5)
```

MikeInnes:
IIRC that's exactly what this does, and what I meant about not supporting catch blocks at all. Obviously it's not clear from the patch that this is the behaviour and it needs tests etc, but at any rate I do agree.

oxinabox:
> IIRC that's exactly what this does, and what I meant about not supporting catch blocks at all. Obviously it's not clear from the patch that this is the behaviour and it needs tests etc, but at any rate I do agree.

Great.
Maybe can use the code I posted as a test.

oxinabox:
@DhairyaLGandhi what is the status on this?

CarloLucibello:
closing as outdated

PR #471, opened 2020-01-19
make Params construction differentiable

piever:
This is the companion PR to https://github.com/FluxML/Flux.jl/pull/997 (but they can be merged separately). With both PRs the following becomes possible:

```julia
gradient(params(model)) do
  sum(norm, params(model))
end
```

The constructors of `Params` should be exactly the same, with the same output, I only changed which one is the inner constructor.

### DIFF
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -55,8 +55,20 @@ Base.adjoint(f::Function) = x -> gradient(f, x)[1]
 struct Params
   order::Vector{Any}
   params::IdSet{Any}
-  Params() = new([], IdSet())
+  function Params(xs)
+    order = Buffer([])
+    params = IdSet()
+    for x in xs
+      if !(x in params)
+        push!(order, x)
+        push!(params, x)
+      end
+    end
+    return new(copy(order), params)
+  end
 end
+
+Params() = Params(())
 
 @forward Params.order Base.iterate, Base.length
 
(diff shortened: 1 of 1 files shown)


MikeInnes:
What if `order` was always stored as a buffer? There's no particular reason `Buffer`s have to be short-lived, and it seems like that way we should be able to leave the other operations as-is.

piever:
I thought about that too, but that seemed like a more "impacting change". In that case, `push!(::Params, x)` would be differentiable here I guess, which would make https://github.com/FluxML/Flux.jl/pull/997 unnecessary.

I'm happy to try a PR for that. Is the idea that `Params` should still implement iteration but not `getindex` and that iteration would in turn be implemented via indexing on `Buffer`?

MikeInnes:
Not sure what you mean about `getindex`; I think we should just be able to replace `Vector` for `Buffer` in the `order` field and nothing else needs to change. `Buffer` should behave equivalently to a vector for our purposes here.

piever:
I was a bit unclear. At the moment `Params` forwards iteration to its `order` field. If `order` becomes a `Buffer` it can no longer iterate (as `Buffer` does not implement `Base.iterate`), but I think it's important that `Params` stays iterable.

I'm not sure if there are fundamental reasons for `Buffer` not to be iterable (or if it is just that the iteration method has not been added yet), so I was wondering whether one should:
- make `Buffer` iterable, or
- implement iteration in `Params` using the fact that `Buffer` has `getindex` (feels hacky though).

MikeInnes:
Ah, I see, that makes complete sense. Yes, it's totally fine to just make `Buffer` iterable. I think you'd want to do that explicitly (i.e. by indexing into the array) rather than by just forwarding to the underlying array for gradients to work, but it should still just be a couple lines.

piever:
superseded by #476

PR #474, opened 2020-01-20
use IdDict in grad_mut fallback for IdDict

ChrisRackauckas:
Fixes hashing issues with CuArrays, i.e. https://github.com/FluxML/Flux.jl/issues/989

### DIFF
--- a/src/lib/base.jl
+++ b/src/lib/base.jl
@@ -24,6 +24,7 @@ end
 # Dictionaries
 
 grad_mut(d::AbstractDict) = Dict()
+grad_mut(d::IdDict) = IdDict()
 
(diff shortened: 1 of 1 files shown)


CarloLucibello:
good to go?

ChrisRackauckas:
It should be. I don't see an issue. All it's doing is making things that should've been an IdDict and IdDict.

dhairyagandhi96:
bors try

dhairyagandhi96:
Aside: Ah, found it, bors is syncing from an old state on Zygote. Flux is fine, but I need to figure out how to turn syncing toggle

PR #476, opened 2020-01-21
make Buffer iterable and use it for Params

piever:
This is a cleaner approach to #471, which simply uses `Buffer` to store `Params`. To do so it defines iteration on `Buffer` by mirroring `AbstractArray`.

There is a minor annoyance: to specify the type of the `order` field in `Params`, one would need to use the `Buffer` type which isn't defined there yet, as the `src/lib` folder is included after the `src/compiler` folder. Should I move the `Buffer` definition above the `Params` definition, while leaving everything else in the `lib/buffer.jl` file?


src/compiler/interface.jl
```diff
@@ -53,9 +53,9 @@ Base.adjoint(f::Function) = x -> gradient(f, x)[1]
 
 # TODO store ids only
 struct Params
-  order::Vector{Any}
+  order #::Buffer{Any, Vector{Any}}
```
piever: Uncommenting this errors during precompilation as `Buffer` is not defined yet.

MikeInnes:
I suggest putting Buffer into `tools/buffer.jl` and keeping the adjoints in `lib/buffer.jl`. That better mirrors the setup for other code anyway, i.e. with the core definitions separate from their adjoints.

Issue #446, opened 2020-01-06
gradient of keyword argument?


chengchingwen:
Is there any ways to compute the backward gradient of a keyword argument? For example, the `var` function can pass a pre-computed mean value as a keyword argument, but right now it seems like we can't get the gradient of it. This will have an effect on something like `Flux.normalise`.

MikeInnes:
No, there isn't currently, at least not directly. The best way to do this is instead to create a normal-argument wrapper around the function and differentiate that:

```julia
_foo(a, b) = a*b

@adjoint _foo(a, b) = ...

foo(; a = 1, b = 2) = _foo(a, b)
```

The setup might have to change a bit for a function built in to base, but should be doable; if not we can work out another option.

chengchingwen:
Then we probably need to change the implementation of `Flux.normalise`
https://github.com/FluxML/Flux.jl/blob/e92da0cf850a982c425b83c92d6274174e52b02c/src/layers/stateless.jl#L82-L86
or maybe add some other function for computing mean and variance.
 
Is it possible to support keyword argument in the future?

MikeInnes:
Is the issue that the gradient for `normalise` ends up incorrect because the gradient for `std` is incorrect (with respect to the kwarg `mean`)?

If so that's a Zygote bug we should certainly fix; those things really need to work well out of the box, rather than relying on workarounds by users.

chengchingwen:
I think so. Because Zygote can't get the gradient of the keyword argument right now, that gradient is missing. We might need to use/define a different set of statistics api that don't require kwarg (or make kwarg work in Zygote).

chengchingwen:
probably need a custom definition for support `mean` kwarg in Zygote (but won't work for other kwarg functions)
```julia
function _pullback(__context__::AContext, ::Core.kwftype(typeof(var)), kw, typeof(var), xs) end
```

MikeInnes:
Zygote can actually get gradients of kwargs, so it's not an issue there as such; the problem is in [this definition](https://github.com/FluxML/Zygote.jl/blob/0e25138a3c3c88600de2a4ae2999f5a5d4676f02/src/lib/array.jl#L224) for `var` (along with `std` below) which assumes the default definition of `mean` will be used and calculates a gradient only for `xs` consistent with that. Actually telling Zygote about the gradient of `mean` is one issue, and modifying `_backvar` to calculate that gradient is another.

[...]

PR #478, opened 2020-01-23
add var/std gradient wrt kw mean

chengchingwen:
fix problem mentioned in #446

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -238,6 +238,27 @@ _backvar(xs, Δ, N::Int, mean) = (convert(eltype(xs), 2/N) .* Δ .* (xs .- mean)
     return s, Δ -> _backvar(xs, Δ ./ (2 .* s), corrected, mean, dims)
 end
 
+function _pullback(ctx::AContext, ::Core.kwftype(Core.Typeof(Statistics.var)), kw, ::typeof(Statistics.var), xs::AbstractArray)
+  y, _back = adjoint(ctx, Statistics.var, xs; kw...)
+  y, Δ -> begin
+    backvar = _back(Δ)[1]
+    if haskey(kw, :mean)
+      dims = get(kw, :dims, :)
+      backmean = -sum(backvar; dims=dims)
+      kwb = NamedTuple{keys(kw)}(nothing for i = 1:length(kw))
+      kwb = merge(kwb, (; :mean=>backmean))
+      (nothing, kwb, nothing, backvar)
+    else
+      (nothing, nothing, nothing, backvar)
+    end
+  end
+end
+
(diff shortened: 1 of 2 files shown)


CarloLucibello:
looks fine and tests are exhaustive, merge?

CarloLucibello:
@chengchingwen is this giving warnings as you seem to mention in https://github.com/FluxML/Zygote.jl/issues/446#issuecomment-577769735?

chengchingwen:
@CarloLucibello  yes, the warnings are still there.

chengchingwen:
if we want to avoid those warnings, we need to avoid using `@adjoint` for `var` and `std` by manually define the `_pullback` functions.

DhairyaLGandhi:
Defining internal methods like so seems sketchy. The adjoint definition from ChainRules as well as within zygote can handle it. The adjoint can accept kwargs

chengchingwen:
I don't really get what you mean. The current [`@adjoint`](https://github.com/FluxML/ZygoteRules.jl/blob/master/src/adjoint.jl#L42) return `nothing` for all the keyword arguments.

DhairyaLGandhi:
Let me rephrase that, we don't want to differentiate kwargs here rather let those be passed to the functions appropriately.

chengchingwen:
But the default `std`/`var` behavior only accept `mean` as keyword argument. If we want to make `mean` be differentiable as position argument, then we will result in a `std`/`var` api that only work when Zygote is imported. Otherwise, we might need to make a PR to Statistics for the `std`/`var` that accept `mean` as position argument.

mcabbott:
What's an example of when this would matter, and would give correct results? On simple things, and the example from here: https://github.com/FluxML/Zygote.jl/issues/446#issuecomment-571489945 it gives zero. With `@show` inserted:
```julia
julia> gradient([1,2,3]) do x
         m = mean(x)
         std(x; mean=m)
       end
backmean = -0.0
([-0.5, 0.0, 0.5],)

julia> function normalise(x::AbstractArray; dims=1)
         μ′ = mean(x, dims = dims)
         σ′ = std(x, dims = dims, mean = μ′, corrected=false)
         return (x .- μ′) ./ σ′
       end
normalise (generic function with 1 method)

julia> gradient(x -> sum(sin, normalise(x)), [1,2,3.0])
backmean = [-0.0]
([-0.2697761903106471, 0.5395523806212941, -0.2697761903106471],)
```
Stumbled on this while thinking about https://github.com/JuliaDiff/ChainRules.jl/issues/567

chengchingwen:
@mcabbott Yes, `normalise` was not an good example for this as the gradient of `std(x, mean=μ)` wrt `μ` is 0 if `μ == mean(x)`. It would only matter when `std(x, mean=μ)` where `μ != mean(x)` and it's probably arguable whether you would ever do that.

CarloLucibello:
closing as outdated

Issue #260, opened 2019-07-15
`Base.adjoint` overload for `::Function` is type piracy


tkoolen:
https://github.com/FluxML/Zygote.jl/blob/4d12379872cb4744b38de54f0b5ef6938b20892e/src/compiler/interface.jl#L50

is type piracy.

PR #481, opened 2020-01-24
Remove the `Base.adjoint(f::Function)` type piracy

DilumAluthge:
Zygote does not own the `Function` type, and it does not own the `Base.adjoint` function, so it has no business defining the  `Base.adjoint(f::Function)`. As a general rule, we should discourage this kind of type piracy.

If a user wants the convenience of writing something like `f'(x)`, then they can simply add this line to their script:
```julia
Base.adjoint(f::Function) = x -> Zygote.gradient(f, x)[1]
```

But Zygote shouldn't be defining this method.

This is a breaking change, so I have incremented the version number from `0.4.7` to `0.5.0`, which indicates a breaking change.

cc: @ChrisRackauckas, who lit the fire under me


ChrisRackauckas:
This is just a nasty piracy to enforce (it has already caused issues in the past with Calculus.jl for example), yet at the same time Zygote doesn't even seem to want you to use it, as in the documentation on calculating gradients doesn't mention or use it at all. So it's cute but I don't see why it's so crucial to enforce on every installation. If a user wants it, they can add that to the top of their script.

dhairyagandhi96:
I think I'd be okay with this change but would rather see it with a depwarn.

Otherwise, a lot of it is syntactic sugar but this does sound like we should be able to get the required meaning within the `gradient` context, rather than have it be global as it is now? Cc @MikeInnes

oxinabox:
Yes, lets do  deprecation.
and tag a nonbreaking change.
Then later remove it

test/features.jl
```diff
@@ -164,7 +164,7 @@ D(f, x) = grad(f, x)[1]
 @test D(x -> x*D(y -> x*y, 1), 4) == 8
 
 if VERSION >= v"1.1"
-  @test sin'''(1.0) ==  -cos(1.0)
+  @test Zygote.gradient(y -> Zygote.gradient(x -> Zygote.gradient(sin, x)[1], y)[1], 1.0)[1] ==  -cos(1.0)
```
MikeInnes: Would be better to use the `D` operator here.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.4.6"
+version = "0.5.0"
```
MikeInnes: I think we can keep to a patch version; this is really more of a repl feature that packages don't depend on.

MikeInnes:
I actually think we can just go ahead and remove this. I don't think downstream packages tend to use it.

Project.toml
oxinabox: Don't do it, follow SemVer.
Don't chose to do the bad thing.
We have semver ant it is a nice thing.

Either deprecate or mark breaking

DilumAluthge:
It is times like this that it would be really to have `NewPkgEval` but for testing against package reverse dependencies.

Project.toml
MikeInnes: The thing about semver is that there's no well-defined boundary for what's breaking in Julia. Essentially any change could be breaking (including adding new exports and renaming or refactoring private functions), so in general we have to make a judgement about what will break in practice.

So I think a positive argument is needed for why this should be treated differently from an internal or interactive-usage-only function, if it should be.

ChrisRackauckas:
Fixes https://github.com/FluxML/Zygote.jl/issues/260

test/features.jl
DilumAluthge: @MikeInnes What is the syntax of the `D` operator?

MikeInnes: It's defined and used right above; it's essentially just a one-argument version of `gradient`.

DilumAluthge:
@oxinabox Want to pick this PR up where it left off? I think some tests need to be fixed to no longer use the `'` syntax.

PR #482, opened 2020-01-26
improve support for nested array

stev47:
These definitions are handy when working with `Vector{NTuple{d,T}}` types which can be interpreted as `Vector{T}`.


CarloLucibello:
closing as outdated

PR #404, opened 2019-11-25
Widen type constraints in accum method for tuples

AzamatB:
Fixes [#399](https://github.com/FluxML/Zygote.jl/issues/399)

PR #489, opened 2020-01-30
Change adjoint for _apply

### LINKED ISSUE #399: Error during gradient calculation
I'm getting an error during gradient calculation and don't understand what is causing it.
My actual code, where this happens is quite large, but I was able to reduce it to the following MRE:
```Julia
using Flux
using Zygote
using LinearAlgebra

mutable struct State
   context
   decoding
   prediction
end

m = (state = State(zeros(Float32, 8, 7), zeros(Float32, 8, 7), zeros(Float32, 4, 7)),
     listen = RNN(5, 8),
     attention_ϕ = Dense(8, 8),
     attention_ψ = Dense(8, 8),
     spell = RNN(20, 8),
     infer = Chain(Dense(16, 4), logsoftmax))

function f(xs::AbstractVector{<:AbstractMatrix})
   hs = m.listen.(xs)
   Hs = cat(hs...; dims=3)
   ψHs = m.attention_ψ.(hs)

   ŷs = map(1:length(xs)) do _
      ϕSᵢᵀ = permutedims(m.attention_ϕ(m.state.decoding))
      Eᵢs = diag.((ϕSᵢᵀ,) .* ψHs)
      αᵢs = softmax(hcat(Eᵢs...)')
      m.state.context = dropdims(sum(reshape(αᵢs, 1, 7, :) .* Hs; dims=3); dims=3)
      m.state.prediction = m.infer([m.state.decoding; m.state.context])
      m.state.decoding = m.spell([m.state.decoding; m.state.prediction; m.state.context])
      return sum(m.state.prediction)
   end
   return sum(ŷs)
end

xs = [rand(Float32, 5,7) for _ ∈ 1:3]

θ = params(m)

Zygote.gradient(θ) do
   f(xs)
end
```
the error message I get is:
```Julia
ERROR: MethodError: no method matching +(::Array{Array{Float32,2},1}, ::Tuple{Array{Float32,2},Array{Float32,2},Array{Float32,2}})
Closest candidates are:
  +(::Any, ::Any, ::Any, ::Any...) at operators.jl:529
  +(::Array, ::Array...) at arraymath.jl:44
  +(::Array, ::SparseArrays.SparseMatrixCSC) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/SparseArrays/src/sparsematrix.jl:1647
  ...
Stacktrace:
 [1] accum(::Array{Array{Float32,2},1}, ::Tuple{Array{Float32,2},Array{Float32,2},Array{Float32,2}}) at /home/azamat/.julia/packages/Zygote/ycnjm/src/lib/lib.jl:8
 [2] (::typeof(∂(f)))(::Float32) at ./untitled-6447fe34214b7f2f51d14ffb8e06011c:20
 [3] (::Zygote.var"#28#29"{typeof(∂(f))})(::Float32) at /home/azamat/.julia/packages/Zygote/ycnjm/src/compiler/interface.jl:38
 [4] gradient(::Function, ::Array{Array{Float32,2},1}) at /home/azamat/.julia/packages/Zygote/ycnjm/src/compiler/interface.jl:47
 [5] top-level scope at untitled-6447fe34214b7f2f51d14ffb8e06011c:37
```
while `f(xs)` works.

I tried to reduce the example further, but the error starts to evade.

AzamatB:
Fixes #399. Proposed by @mohamed82008 in slack.

### DIFF
--- a/src/lib/lib.jl
+++ b/src/lib/lib.jl
@@ -153,9 +153,10 @@ unapply(t, xs) = _unapply(t, xs)[1]
   y, back = Core._apply(_pullback, (__context__, f), args...)
   st = map(_empty, args)
   y, function (Δ)
+    maybetoarray(arg, Δ) = arg isa Array ? reshape(collect(Δ), size(arg)) : Δ
     Δ = back(Δ)
     Δ === nothing ? nothing :
-      (first(Δ), unapply(st, Base.tail(Δ))...)
+      (first(Δ), maybetoarray.(args, unapply(st, Base.tail(Δ)))...)
   end
 end
 
--- a/test/compiler.jl
+++ b/test/compiler.jl
@@ -16,7 +16,8 @@ trace_contains(st, func, file, line) = any(st) do fr
     fr.line == line
 end
 
-@test gradient(x -> vcat(x...)'x, [1,1]) == ([2,2],)
+@test gradient(x -> vcat(x...)'x, [1, 1]) == ([2, 2],)
+@test gradient(x -> sum(hcat(x...) * sum(x)), [[1, 2], [3, 4]]) == ([[7, 11], [9, 13]],)
 
 bad(x) = x
 @adjoint bad(x) = x, Δ -> error("bad")
--- a/test/compiler.jl
+++ b/test/compiler.jl
@@ -16,6 +16,8 @@ trace_contains(st, func, file, line) = any(st) do fr
     fr.line == line
 end
 
+@test gradient(x -> vcat(x...)'x, [1,1]) == ([2,2],)
+
 bad(x) = x
 @adjoint bad(x) = x, Δ -> error("bad")
 


MikeInnes:
This could do with a minimal test case so it's clear what it's fixing.

AzamatB:
Done

mohamed82008:
I think some of the tests assumed the previous behavior which is why they are failing. Disclaimer: I said on slack I am not sure this "fix" is even correct, but it does fix the issue of splatting arrays.

MikeInnes:
Ok, I understand what's going on here now. This is a tricky case and we need to think carefully about what the right fix is.

The problem generally is that we can splat non-tuple objects (arrays, sets, custom types etc) and this currently means they'll get a tuple gradient. This is potentially surprising and doesn't give a good error message if it goes wrong.

```julia
julia> gradient(x -> +(x...), [1, 2, 3])[1]
(1, 1, 1)
```

There are a few different options here:

1. Define a way for custom types to participate in `_empty` and `_unapply` (similar to what you have here, but user-extensible and without closing over inputs). This has the downside that any user-defined iterable type will break if splatted, unless people hook into a somewhat obscure Zygote API.
2. Define tuples as valid gradients for iterables, in which case the fix is closer to #404. Still potentially surprising, especially in the REPL, but possibly more natural to support.
3. Rewrite `Core._apply(f, xs)` to `Core._apply(f, collect_tuple(xs))`. The idea would be to differentiate through the `collect_tuple` call, which uses iteration etc. as normal, meaning it will always produce a gradient of the correct type. This has the least fragile behaviour but potentially adds overhead (though possibly only in cases that would be slow anyway; it's pretty easy to make the tuple case fast).

PoC for the last option, which I'm leaning towards:

```julia
julia> function collect_tuple(xs, state...)
         next = iterate(xs, state...)
         next === nothing && return ()
         return (next[1], collect_tuple(xs, next[2])...)
       end

julia> gradient(x -> +(collect_tuple(x)...), [1, 2, 3])[1]
3-element Array{Int64,1}:
 1
 1
 1
```

CarloLucibello:
should this be revived in order to fix https://github.com/FluxML/Zygote.jl/issues/599?

CarloLucibello:
closing as outdated

PR #504, opened 2020-02-07
Remove unnecessary splatting from the code fo adjoint of [hv]cat

AzamatB:


oxinabox:
If we are going to microoptimize, I suggest we do it a bit more heavily:

### Benchmarking:
```
julia> const xs = (ones(4,4), ones(5,5), ones(8,7));

julia> f(x) = cumsum(collect(size.(x,1)))

julia> g(x) = cumsum([size.(x,1)...])

julia> function h(x)
       r = collect(size.(x,1))
       cumsum!(r,r)
       end
h (generic function with 1 method)

julia> function i(x)
       k = Ref(0)
       ntuple(length(x)) do i
       @inbounds k[]+=size(x[i],1)
       end
       end
i (generic function with 1 method)
```

```
julia> @btime f($xs)
  60.529 ns (2 allocations: 224 bytes)

julia> @btime g($xs)
  59.682 ns (2 allocations: 224 bytes)

julia> @btime h($xs)
  36.109 ns (1 allocation: 112 bytes);

julia> @btime i($xs);
  1.251 ns (0 allocations: 0 bytes)
```

mcabbott:
Sounds like a use case for https://github.com/JuliaLang/julia/pull/34654.

CarloLucibello:
closing as stale

PR #505, opened 2020-02-08
delete! for Params

CarloLucibello:
This is needed to expose some parameter freezing  functionality in Flux, as suggested in https://github.com/FluxML/Flux.jl/issues/1022

### DIFF
--- a/src/tools/buffer.jl
+++ b/src/tools/buffer.jl
@@ -66,7 +66,15 @@ function Base.copy(b::Buffer)
   return b.data
 end
 
-@forward Buffer.data Base.eltype, Base.length, Base.ndims, Base.size, Base.axes, Base.eachindex, Base.stride, Base.strides
+function Base.deleteat!(b::Buffer, i)
+  b.freeze && error("Buffer is frozen")
+  deleteat!(b.data, i)
+  return b
+end
+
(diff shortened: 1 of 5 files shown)


src/tools/buffer.jl
```diff
@@ -66,7 +66,15 @@ function Base.copy(b::Buffer)
   return b.data
 end
 
-@forward Buffer.data Base.eltype, Base.length, Base.ndims, Base.size, Base.axes, Base.eachindex, Base.stride, Base.strides
+function Base.deleteat!(b::Buffer, i)
```
oxinabox: Can we have tests on this please?

CarloLucibello:
@oxinabox thanks for the review,  can you merge this?

MikeInnes:
FWIW, the intended way to freeze parameters is to make sure they never turn up in `Params` in the first place; ideally with something like a `Freeze(m)` layer that behaves like `m` but doesn't have params. It might be a good idea to implement that as part of Flux.

Anyway, this can't do any harm in itself, so it LGTM.

CarloLucibello:
Yes, I guess we can have both. Also, my use case is a bit more fine-grained, I need to freeze only a specific parameter (the bias term) within a layer.

PR #508, opened 2020-02-09
Add adjoint for repeat(x::AbstractVecOrMat, m, n)

AzamatB:


src/lib/array.jl
```diff
@@ -124,6 +124,13 @@ end
   end
 end
 
+@adjoint function repeat(x::AbstractVecOrMat, m::Integer, n::Integer=1)
+   size₁, size₂ = size(x, 1), size(x, 2)
+   begin₁, begin₂ = firstindex(x, 1), firstindex(x, 2)
+   end₁,   end₂   =  lastindex(x, 1),  lastindex(x, 2)
+   return repeat(x, m, n), ȳ -> (sum(@view ȳ[(begin₁ + i*size₁):(end₁ + i*size₁), (begin₂ + j*size₂):(end₂ + j*size₂)] for i ∈ 0:(m-1), j ∈ 0:(n-1)), nothing, nothing)
```
CarloLucibello:
```suggestion
   return repeat(x, m, n), ȳ -> begin
      ȳ = reshape(ȳ, size(x,1), m, size(x,2), n)
      dropdims(sum(ȳ, dims=(2,4)), dims=(2,4))
   end
```

CarloLucibello:
Can we also have a gradcheck test for this?

src/lib/array.jl
AzamatB: This (like the original version) will return matrix for vector input, so `AbstractVector` needs a separate method

CarloLucibello: `reshape(sum(ȳ, dims=(2,4)), size(x))` will produce vector output for vector input, so this version should be fine for VecOrMat

AzamatB: Indeed. Thanks, done

AzamatB: This will still produce a matrix for a vector input in the forward pass. So looks like we do need a separate method

CarloLucibello:
needs some tests and then it should be ready

AzamatB:
Done.

src/lib/array.jl
```diff
@@ -124,6 +124,16 @@ end
   end
 end
 
+@adjoint repeat(x::AbstractVector, m::Integer) =
+   repeat(x, m), ȳ -> (dropdims(sum(reshape(ȳ, length(x), :); dims=2); dims=2), nothing)
```
CarloLucibello: do we need this method? Doesn't the VecOrMat version cover both?

mcabbott: Here's a test which fails without it:
```
gradient(x -> sum(1 .+ repeat(x, 2)::Vector), ones(2)) == ([2,2],)
```

CarloLucibello: right, although I fail to see why. Without the type cast it works, and the return from `repeat(x, 2)` for x vector is already a vector, so the cast is useless :confused:

mcabbott: No, the return from `repeat(x, 2)` is not a vector here, because `@adjoint function repeat(x::AbstractVecOrMat, ...` replaces it with `repeat(x, 2, 1)`.

CarloLucibello:
Thanks!

bors r+

Issue #494, opened 2020-02-01
Differentiating ForwardDiff Expressions Fails


ChrisRackauckas:
```julia
using Zygote, ForwardDiff
p = [1f0]; ps = Zygote.Params(p)
f(x) = sin(p[1]*x)
g(x) = ForwardDiff.derivative(f,x)
Zygote.gradient(()->sum(g(1f0)),ps)
```

PR #510, opened 2020-02-10
Differentiation of Dual numbers

MikeInnes:
This PR shows how to make `Dual` constructors and `partials` to be differentiable in Zygote. Because `Dual`s are `Number`s, we have to differentiate them naturally rather than structurally (this works already):

```julia
julia> gradient(x -> x*x, Dual(5, 1))
(Dual{Nothing}(10,2),)
```

This is easy to interpret if we're doing forward-over-reverse; we get the second derivative here. It's harder to interpret that if we do something like

```julia
gradient(x -> partials(x*x, 1), Dual(5, 1))
```

In a case like this it makes more sense to take a structural derivative (calculating the derivatives of the partials, not the partials of the derivative). That's what Zygote currently does for `partials`, but this breaks as in #494 when we try to mix structural and natural gradients.

The way to resolve this is to work out how derivatives-of-partials and partials-of-derivatives are related and map between them in the adjoints for `partials` and `Dual`. This PR currently just assumes they are the same, which actually works in some simple cases, but we should be much more careful that this is valid before merging.

```julia
julia> gradient(x -> partials(sin(Dual(5, x)), 1), 1)
(0.28366218546322625,)
```

### DIFF
--- a/src/lib/forward.jl
+++ b/src/lib/forward.jl
@@ -8,6 +8,9 @@ using ForwardDiff: Dual
 @adjoint literal_getproperty(d::Dual{T}, ::Val{:partials}) where T =
   d.partials, ṗ -> (Dual{T}(0, ṗ...),)
 
+@adjoint literal_getproperty(d::Dual{T}, ::Val{:value}) where T =
+  d.value, ẋ -> (Dual{T}(ẋ, zeros(d.partials.values)...),)
+
 # Mixed mode
 
 seed(x::Real, ::Val) = Dual(x, true)
--- a/src/lib/forward.jl
+++ b/src/lib/forward.jl
@@ -1,6 +1,15 @@
 using ForwardDiff
 using ForwardDiff: Dual
 
+# ForwardDiff integration
+
+@adjoint Dual{T}(x, ẋ::Tuple) where T = Dual{T}(x, ẋ), ḋ -> (ForwardDiff.value(ḋ), (ForwardDiff.partials(ḋ)...,))
+
+@adjoint literal_getproperty(d::Dual{T}, ::Val{:partials}) where T =
+  d.partials, ṗ -> (Dual{T}(0, ṗ...),)
+
+# Mixed mode
+
 seed(x::Real, ::Val) = Dual(x, true)
 
 function seed(x, ::Val{N}, offset = 0) where N
--- a/src/lib/forward.jl
+++ b/src/lib/forward.jl
@@ -3,13 +3,16 @@ using ForwardDiff: Dual
 
 # ForwardDiff integration
 
-@adjoint Dual{T}(x, ẋ::Tuple) where T = Dual{T}(x, ẋ), ḋ -> (ForwardDiff.value(ḋ), (ForwardDiff.partials(ḋ)...,))
+@adjoint function Dual{T}(x, ẋ::Tuple) where T
+  @assert length(ẋ) == 1
+  Dual{T}(x, ẋ), ḋ -> (ḋ.partials[1], (ḋ.value,))
+end
 
 @adjoint literal_getproperty(d::Dual{T}, ::Val{:partials}) where T =
-  d.partials, ṗ -> (Dual{T}(0, ṗ...),)
+  d.partials, ṗ -> (Dual{T}(ṗ[1], 0),)
 
 @adjoint literal_getproperty(d::Dual{T}, ::Val{:value}) where T =
-  d.value, ẋ -> (Dual{T}(ẋ, zeros(d.partials.values)...),)
+  d.value, ẋ -> (Dual{T}(0, ẋ),)
 
 # Mixed mode
 


ChrisRackauckas:
Seems like the MWE ends up with a `nothing` gradient:

```julia
using Zygote, ForwardDiff
p = [1f0]; ps = Zygote.Params(p)
f(x) = sin(p[1]*x)
g(x) = ForwardDiff.derivative(f,x)
grads = Zygote.gradient(()->sum(g(1f0)),ps)
grads.grads

IdDict{Any,Any} with 1 entry:
  1.0 => nothing
```

ForwardDiff's more complicated seeding code may be causing something.

MikeInnes:
That's because you aren't using `Params` correctly; best to make it work without ForwardDiff first, and then introduce the `derivative` call once it does.

MikeInnes:
So this is pretty interesting. This works, at least in a couple of cases, if we define the structural sensitivity of the dual `d = x + pϵ` to be `d̄ = p̄ + x̄ϵ`, and use the normal natural adjoints for Dual numbers. I have no real justification for this but can prove it works in a couple of specific cases:

<details>

```julia
a::Dual * b::Dual = Dual(a.x*b.x, a.p * b.x + b.p * a.x)

ā.x = c̄.x * b.x + c̄.p * b.p
    = (Dual(c̄.p, c̄.x) * b).p

ā.p = c̄.p * b.x
    = (Dual(c̄.p, c̄.x) * b).x

sin(a::Dual) = Dual(sin(a.x), cos(a.x)*a.p)

ā.x = cos(a.x) * b̄.x - sin(a.x) * b.p * a.p
    = (-cos(a)*Dual(b̄.p, b̄.x)).p

ā.p = cos(a.x) * b̄.p
    = (-cos(a)*Dual(b̄.p, b̄.x)).x
```
</details>

Is it problematic to use `Dual` as an adjoint to represent two quite different things? I think it may not be, since in correct programs the `Dual` constructor and `partials` should provide a boundary for the structural part. If you write `partials(x, 1) + x` you'll get nonsense, but your forward pass is nonsense anyway.

cc @oxinabox who likes these kinds of puzzles.

ChrisRackauckas:
Okay, still issue:

```julia
using Zygote, ForwardDiff
p = [1f0]; ps = Zygote.Params([p])
f(x) = sin(p[1]*x)
g(x) = ForwardDiff.derivative(f,x)
grads = Zygote.gradient(()->sum(g(1f0)),ps)
grads.grads
```

```
IdDict{Any,Any} with 1 entry:
  Float32[1.0] => ForwardDiff.Dual{ForwardDiff.Tag{typeof(f),Float32},Float32,1…
```

Needs to value that.

MikeInnes:
If you want to write code like that you're going to have to drop the `Dual` component of the gradient yourself. There's no other way for Zygote to know that you don't want the gradient there.

ChrisRackauckas:
But duals only show up in the intermediate: they are not the input or output to the function, so why are they the result of the gradient? The result of the gradient here is just a real number.

MikeInnes:
For the same reason there's a complex component to the gradient

```julia
julia> gradient(x -> abs2(x*1+2im), 1)
(2 + 4im,)
```

even though it's R -> R. Similarly in your code there's a dual component to the gradient. You can avoid this easily enough by not capturing things in the `f` you pass to forwarddiff. Or you can capture them but explicitly unwrap the dual gradient easily enough.

ChrisRackauckas:
>Or you can capture them but explicitly unwrap the dual gradient easily enough.

But that would mean that `Flux.train!` won't work in any case where you're nesting with ForwardDiff? The real case here is https://github.com/JuliaDiffEq/NeuralNetDiffEq.jl/pull/55 . This would mean putting a Chain into a ForwardDiff call (for nesting derivatives of a neural network) will make all gradients dual numbers and then cause an error in the standard training loop?

ChrisRackauckas:
Looks like it's working, but would need to document how to properly convert the outputs of the gradient:

```julia
using Zygote, ForwardDiff, FiniteDiff
p = [1f0,2f0]
f(x,p) = sin(p[1]*p[2]*x)
g(x,p) = ForwardDiff.derivative(x->f(x,p),x)
res = getindex.(ForwardDiff.partials.(Zygote.gradient((p)->sum(g(1f0,p)),p)[1]),1)

g2(x,p) = FiniteDiff.finite_difference_derivative(x->f(x,p),x)
res = Zygote.gradient((p)->sum(g2(1f0,p)),p)
```

are the same.

ChrisRackauckas:
This has been reviewed in the context of a physics-informed neural network (https://github.com/JuliaDiffEq/DiffEqFlux.jl/pull/175 + https://github.com/JuliaDiffEq/NeuralNetDiffEq.jl/pull/55) and it seems like it's correct. Can't guarantee all cases with that test of course, but that's a pretty non-trivial case already.

dhairyagandhi96:
Quick bump for the failing test on nightly :)

oxinabox:
> The way to resolve this is to work out how derivatives-of-partials and partials-of-derivatives are related and map between them in the adjoints for partials and Dual.

Or map between then in `accum`.

ChrisRackauckas:
bump?

MikeInnes:
This was mainly just a PoC; rather than doing something potentially incorrect/confusing here we're just going to implement Zygote-compatible forward mode, #503.

PR #515, opened 2020-02-14
Fix Flux.flip by providing an adjoint for Base.reverse

tanhevg:
The main motivation behind this PR is to address various issues concerning `Flux.flip()` (used mainly for bRNNs), e.g.  FluxML/Flux.jl#962, FluxML/Flux.jl#990 and FluxML/model-zoo#179

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -65,8 +65,7 @@ end
 
 @adjoint function reverse(x::AbstractArray, args...; kwargs...)
   _reverse(t) = reverse(t, args...; kwargs...)
-  _nothings(t) = map(_->nothing, keys(t))
-  _reverse(x), Δ->(_reverse(Δ), _nothings(args)..., _nothings(kwargs)...)
+  _reverse(x), Δ->(_reverse(Δ), map(_->nothing, args)...)
 end
 
 @adjoint permutedims(xs) = permutedims(xs), Δ -> (permutedims(Δ),)
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -63,6 +63,12 @@ end
   circshift(A, shifts), Δ -> (circshift(Δ, map(-, shifts)), nothing)
 end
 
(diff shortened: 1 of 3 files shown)


CarloLucibello:
Does it fix `flip`? Can we have a test for it?

tanhevg:
@CarloLucibello 
> Does it fix `flip`? Can we have a test for it?

Yes it does. Tests would have to go in `Flux`, so this would be a separate PR. One of us could put one in once this is merged and tagged in a Zygote release. Adding PRs that rely on unreleased versions of dependencies is messy.

dhairyagandhi96:
Having the PR up would be helpful, I think it's one of the tail end of fixes remaining to update the zoo as well

CarloLucibello:
> @CarloLucibello
> 
> > Does it fix `flip`? Can we have a test for it?
> 
> Yes it does. Tests would have to go in `Flux`, so this would be a separate PR. One of us could put one in once this is merged and tagged in a Zygote release. Adding PRs that rely on unreleased versions of dependencies is messy.

oh right, thought flip was a Base function

src/lib/array.jl
```diff
@@ -63,6 +63,12 @@ end
   circshift(A, shifts), Δ -> (circshift(Δ, map(-, shifts)), nothing)
 end
 
+@adjoint function reverse(x::AbstractArray, args...; kwargs...)
+  _reverse(t) = reverse(t, args...; kwargs...)
+  _nothings(t) = map(_->nothing, keys(t))
+  _reverse(x), Δ->(_reverse(Δ), _nothings(args)..., _nothings(kwargs)...)
```
MikeInnes: You don't need to return adjoints for the kwargs here.

tanhevg: Thanks, duly noted

CarloLucibello:
this seems good to go

tanhevg:
Fix FluxML/Flux.jl#962

Issue #513, opened 2020-02-13
Literal arrays are not differentiable


ChrisRackauckas:
MWE:

```julia
using Zygote
function f(u)
    M = [1 0
         0 u[1]][4]
end
Zygote.gradient(f,[1.0])
```

```julia
Mutating arrays is not supported
error(::String) at error.jl:33
(::Zygote.var"#1009#1010")(::Nothing) at array.jl:49
(::Zygote.var"#2685#back#1011"{Zygote.var"#1009#1010"})(::Nothing) at adjoint.jl:49
hvcat_fill at abstractarray.jl:1708 [inlined]
(::typeof(∂(hvcat_fill)))(::Array{Float64,2}) at interface2.jl:0
typed_hvcat at abstractarray.jl:1730 [inlined]
(::typeof(∂(typed_hvcat)))(::Array{Float64,2}) at interface2.jl:0
(::Zygote.var"#165#166"{typeof(∂(typed_hvcat)),Tuple{Tuple{Nothing,Nothing},NTuple{4,Nothing}}})(::Array{Float64,2}) at lib.jl:156
#321#back at adjoint.jl:49 [inlined]
hvcat at abstractarray.jl:1715 [inlined]
(::typeof(∂(hvcat)))(::Array{Float64,2}) at interface2.jl:0
f at test.jl:3 [inlined]
(::typeof(∂(f)))(::Float64) at interface2.jl:0
(::Zygote.var"#36#37"{typeof(∂(f))})(::Float64) at interface.jl:38
gradient(::Function, ::Array{Float64,1}) at interface.jl:47
top-level scope at test.jl:6
```

The literal syntax lowers to `typed_hvcat` which generates then fills the matrix, but that mutation is not allowed by Zygote. The adjoint may need to be manually defined.

mcabbott:
There is an adjoint defined:
https://github.com/FluxML/Zygote.jl/blob/f70a8ad1eab511c3ec9e0ded92eee629043cccf5/src/lib/array.jl#L79 
But it doesn't get called in your case, when the numbers have different types:
```julia
Zygote.gradient(f,[1]) # ([1],)

@which [1 0
        0 u[1]] # hvcat(rows::Tuple{Vararg{Int64,N} where N}, xs::Number...)

g(xs::Number...) = "numbers"
g(1, 1.0) #  "numbers"
h(xs::Vararg{T}) where {T<:Number} = "uniform"
h(1, 1.0) # MethodError
```

[...]

PR #518, opened 2020-02-16
Special adjoint for broadcasted literal pow

haampie:
Currently taking the gradient of anything that contains a broadcasted
literal pow adds RefValue{typeof(^)}(^) and a similar entry for the
literal power itself to the IdDict. This is probably because of the
special signature in the broadcasting machinery:

```
Base.broadcasted(Base.literal_pow, Main.:^, vec, %2)
```

where %2 is a Val{N} instance.

By adding a special adjoint for broadcasting literal_pow, not only
do we reduce the noise in the param's IdDict, but it also speeds
up taking the gradient of basic loss functions like sum(err.^2).

Ref #513 and also solves https://github.com/FluxML/Flux.jl/issues/1018 (most of it)

### DIFF
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -76,6 +76,11 @@ Numeric{T<:Number} = Union{T,AbstractArray{<:T}}
   res, Δ -> (nothing, unbroadcast(x, Δ ./ y), unbroadcast(y, -Δ .* res ./ y))
 end
 
+@adjoint function broadcasted(::typeof(Base.literal_pow), ::typeof(^), x::Numeric, exp::Val{p}) where p
+  y = Base.literal_pow.(^, x, exp)
+  y, ȳ -> (nothing, nothing, ȳ .* p .* conj.(x .^ (p - 1)), nothing)
+end
+
 @adjoint broadcasted(::typeof(identity), x::Numeric) = x, Δ -> (nothing, Δ)
 
 @adjoint function broadcasted(::typeof(σ), x::Numeric)


mcabbott:
Minimal timing test, which improves by 100x:
```julia
@btime gradient(x -> sum(x .^ 2), x0)  setup=(x0=collect(1.0:10^6))
```
However to get the right answers, I think it needs to be more like this:
```julia
import Base: broadcasted
using Zygote: @adjoint
Numeric{T<:Number} = Union{T,AbstractArray{<:T}}

@adjoint function broadcasted(::typeof(Base.literal_pow), ::typeof(^), x::Numeric, ::Val{p}) where {p}
  y = Base.literal_pow.(^, x, Val(p))
  y, ȳ -> (nothing, nothing, ȳ .* p .* conj.(x .^ (p-1)), nothing)
end
```

haampie:
Oops, yeah. Should be captured by @mcabbott tests from #513 :)

Should be fine now!

CarloLucibello:
@oxinabox could you merge this?

src/lib/broadcast.jl
```diff
@@ -76,6 +76,11 @@ Numeric{T<:Number} = Union{T,AbstractArray{<:T}}
   res, Δ -> (nothing, unbroadcast(x, Δ ./ y), unbroadcast(y, -Δ .* res ./ y))
 end
 
+@adjoint function broadcasted(::typeof(Base.literal_pow), ::typeof(Main.:^), x::Numeric, y::Val{p}) where p
```
CarloLucibello:
```suggestion
@adjoint function broadcasted(::typeof(Base.literal_pow), ::typeof(Main.:^), x::Numeric, ::Val{p}) where p
```

mcabbott: Wait `y` is the result in the function body? 

Although I also don't see why `Main.:^` not just `^`. (In fact see my comment.)

haampie: Good catch, that `y` type instability 😅.

haampie:
@mcabbott any reason to write `Base.literal_pow.(^, x, Val(p))` instead of `x .^ p`?

mcabbott:
Mostly a general feeling that one shouldn't change the forward pass without good reason. But I think `literal_pow` is also often quite a bit quicker, it may even be worthwhile using this for the backward pass too:
```
julia> A = randn(10,2,5,7);

julia> my_power(A, p=2) = A .^ p;

julia> @btime my_power($A);
  14.790 μs (2 allocations: 5.67 KiB)

julia> @btime Base.literal_pow.(^, $A, Val(2)); 
  698.965 ns (2 allocations: 5.67 KiB)
```

haampie:
Oh... of course. Somehow I thought the compiler would fill in the `p` as a constant and it would result in `literal_pow` anyways. But it's not. I'll just make sure it uses literal_pow explicitly then.

mcabbott:
It may actually catch it here, compare maximal:
```
julia> @adjoint function broadcasted(::typeof(Base.literal_pow), ::typeof(^), x::Numeric, ::Val{p}) where {p}
         y = Base.literal_pow.(^, x, Val(p))
         y, ȳ -> (nothing, nothing, ȳ .* p .* conj.(Base.literal_pow.(^, x, Val(p-1))), nothing)
       end

julia> @btime gradient(x -> sum(x .^ 2), x0)  setup=(x0=collect(1.0:10^6))
  2.247 ms (20 allocations: 15.26 MiB)
([2.0, 4.0, 6.0, 8.0, 10.0, 12.0, 14.0, 16.0, 18.0, 20.0  …  1.999982e6, 1.999984e6, 1.999986e6, 1.999988e6, 1.99999e6, 1.999992e6, 1.999994e6, 1.999996e6, 1.999998e6, 2.0e6],)
```
and minimal:
```
julia> @adjoint function broadcasted(::typeof(Base.literal_pow), ::typeof(^), x::Numeric, ::Val{p}) where {p}
         y = x .^ p
         y, ȳ -> (nothing, nothing, ȳ .* p .* conj.(x .^ (p-1)), nothing)
       end

julia> @btime gradient(x -> sum(x .^ 2), x0)  setup=(x0=collect(1.0:10^6))
  2.307 ms (20 allocations: 15.26 MiB)
([2.0, 4.0, 6.0, 8.0, 10.0, 12.0, 14.0, 16.0, 18.0, 20.0  …  1.999982e6, 1.999984e6, 1.999986e6, 1.999988e6, 1.99999e6, 1.999992e6, 1.999994e6, 1.999996e6, 1.999998e6, 2.0e6],)
```
So maybe I'm just superstitious! Both a bit quicker than `113.114 ms` without this.

haampie:
Let's just go for the exact same expression in the forward pass, then there's no surprises.

haampie:
Actually, there is a difference for `-1`, in the case of literal_pow it is computed via Base.div_float. So, let's definitely stick to `literal_pow` :)

CarloLucibello:
Nice, thanks!

bors r+

PR #519, opened 2020-02-19
fix a few adjoints for complex arrays

marius311:
These were giving wrong answers before. 

I didn't write tests for the `det` fixes mostly because I couldn't figure out how to use `gradtest` for complex matrices, although I welcome any help, but perhaps its ok since these are correct in [ChainsRules](https://github.com/JuliaDiff/ChainRules.jl/blob/d3464c2c86b59ee766b10c9e496fb3cde9053b93/src/rulesets/LinearAlgebra/dense.jl#L53) and, as I understand, this will switch to using that soon.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -301,11 +301,11 @@ end
 
 @adjoint diag(A::AbstractMatrix) = diag(A), Δ->(Diagonal(Δ),)
 
-@adjoint det(xs) = det(xs), Δ -> (Δ * det(xs) * transpose(inv(xs)),)
+@adjoint det(xs) = det(xs), Δ -> (Δ * det(xs) * inv(xs)',)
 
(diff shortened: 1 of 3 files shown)


CarloLucibello:
maybe add some tests on small and simple matrices, comparing the analytical gradient with zygote's one?

marius311:
Could you provide some guidance on how `gradtest` is supposed to work, which presumably is what I'm supposed to use to write these tests?

CarloLucibello:
just bypass gradtest  and for some small matrix `x` of which you know the true gradient, do something like
```julia
g = gradient(logabsdet, x)[1]
@test norm(g  .- gtrue) < 1e-10
```julia

marius311:
Ok, just pushed, ready from my end.

CarloLucibello:
great!

bors r+

PR #532, opened 2020-03-02
fix conv on cpu with fillarays

CarloLucibello:
fix https://github.com/FluxML/Flux.jl/issues/1055 thanks to @dlfivefifty's suggestion in https://github.com/JuliaArrays/FillArrays.jl/issues/88

Some tests for corner cases are broken, but a fix for that will have to wait for some future PR


dhairyagandhi96:
So the idea is to force allocating the whole array?

CarloLucibello:
> So the idea is to force allocating the whole array?

yep, there is no way around it unless we want to write a specialized convolutional kernel, but this is not worth the effort since we are talking about a corner case that is unlikely to show up in real scenarios

PR #545, opened 2020-03-17
Restrict some linear algebra adjoints to matrices

mohamed82008:
This PR restricts the adjoints that make sense for matrices to matrices. This PR fixes https://github.com/TuringLang/Bijectors.jl/issues/90.


CarloLucibello:
`inv`, `det` and `logdet` are also defined for `Number`s in julia. Can you make sure we don't loose those adjoints and add a test?

mohamed82008:
Done.

PR #565, opened 2020-03-28
Fix broadcast for CuArray with CuArrays 2.0

### LINKED ISSUE #567: Gradients of broadcasted `exp` broken with CuArrays v2.0.1
The following works on CuArrays v1.7 but doesn't work on v2.0.1.

```julia
Flux.gradient(x -> sum(exp.(x)), CuArrays.ones(1,1))
```

```julia
┌ Warning: calls to Base intrinsics might be GPU incompatible
│   exception =
│    You called exp(x::T) where T<:Union{Float32, Float64} in Base.Math at special/exp.jl:75, maybe you intended to call exp(x::Float64) in CUDAnative at /vol/bitbucket/at6617/dot_julia/packages/CUDAnative/cnQli/src/device/cuda/math.jl:100 instead?
│    Stacktrace:
│     [1] exp at special/exp.jl:75
│     [2] #28 at /vol/bitbucket/at6617/dot_julia/packages/GPUArrays/QDGmr/src/host/broadcast.jl:64
└ @ CUDAnative /vol/bitbucket/at6617/dot_julia/packages/CUDAnative/cnQli/src/compiler/irgen.jl:266
```

ararslan:
`CuArray`s now have their own `BroadcastStyle`.

cc @maleadt

EDIT: Fixes #567

### DIFF
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -179,7 +179,13 @@ end
 end
 
 @init @require CuArrays="3a865a2d-5b23-5a0f-bc46-62713ec82fae" begin
-  @adjoint function broadcasted(::Broadcast.ArrayStyle{CuArrays.CuArray}, f, args...)
+  if isdefined(CuArrays, :CuArrayStyle)  # Introduced in CuArrays v2.0
+    using CuArrays: CuArrayStyle
+  else
+    CuArrayStyle = Broadcast.ArrayStyle{CuArrays.CuArray}
+  end
+
+  @adjoint function broadcasted(::CuArrayStyle, f, args...)
     y, back = broadcast_forward(CuArrays.cufunc(f), args...)
     y, ȳ -> (nothing, nothing, back(ȳ)...)
   end
--- a/Project.toml
+++ b/Project.toml
@@ -23,7 +23,7 @@ ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 [compat]
 AbstractFFTs = "0.5"
 ArrayLayouts = "0.1, 0.2"
-CuArrays = "1.7"
+CuArrays = "1.7, 2"
 DiffRules = "0.0, 0.1, 1"
 FillArrays = "0.8"
 ForwardDiff = "0"


dhairyagandhi96:
Thanks ! bors r+

ararslan:
It bors sick? 🤒

dhairyagandhi96:
bors ping

CarloLucibello:
@maleadt bors' error seems a regression in the cuda stack

maleadt:
> @maleadt bors' error seems a regression in the cuda stack

These broadcasts are getting materialized on the CPU because the element types are not supported by the math functions in CUDAnative. I don't think this was supported in earlier versions either?

```
julia> log.(CuArray([1]))
ERROR: MethodError: no method matching log(::Int64)
You may have intended to import Base.log
Closest candidates are:
  log(::Complex{Float32}) at /home/tim/Julia/pkg/CUDAnative/src/device/cuda/math.jl:77
  log(::Complex{Float64}) at /home/tim/Julia/pkg/CUDAnative/src/device/cuda/math.jl:76
  log(::Float32) at /home/tim/Julia/pkg/CUDAnative/src/device/cuda/math.jl:73
  ...
Stacktrace:
 [1] _broadcast_getindex_evalf at ./broadcast.jl:631 [inlined]
 [2] _broadcast_getindex at ./broadcast.jl:604 [inlined]
 [3] getindex at ./broadcast.jl:564 [inlined]
 [4] copy at ./broadcast.jl:854 [inlined]
 [5] materialize(::Base.Broadcast.Broadcasted{CuArrays.CuArrayStyle{1},Nothing,typeof(CUDAnative.log),Tuple{CuArray{Int64,1,Nothing}}}) at ./broadcast.jl:820
 [6] top-level scope at REPL[16]:1
```

ararslan:
Do the GPU integration tests work with CuArrays 1.7?

dhairyagandhi96:
Same Manifest as usual tests

ararslan:
So I guess that means this PR must have broken the tests?

src/lib/broadcast.jl
```diff
@@ -179,7 +179,13 @@ end
 end
 
 @init @require CuArrays="3a865a2d-5b23-5a0f-bc46-62713ec82fae" begin
-  @adjoint function broadcasted(::Broadcast.ArrayStyle{CuArrays.CuArray}, f, args...)
+  if isdefined(CuArrays, :CuArrayStyle)  # Introduced in CuArrays v2.0
```
dhairyagandhi96: Since we claim compliance with CuArrays 2.x do we need to check for this anymore? Old versions of flux and Zygote would just never get new CuArrays, and the new ones wouldn't get old CuArrays, right?

ararslan: > Since we claim compliance with CuArrays 2.x

Where do we do that? I didn't think Zygote had a direct dependency on CuArrays.

CarloLucibello:
gpu integration tests are failing in all PRs. CuArrays 2.0 is being installed, which means the current manifest is not used in gitlab's CI

ararslan:
I believe the CI will be fixed by #574.

CarloLucibello:
Should we support only julia 1.3+ upon merging this?

CarloLucibello:
> Should we support only julia 1.3+ upon merging this?

we don't, I see now this also supports CuArrays < 2. 

@ararslan could you add v2 to CuArrays' compat in Project.toml?

ararslan:
Done!

CarloLucibello:
Since https://github.com/JuliaGPU/CuArrays.jl/issues/666 is not a CuArrays 2 regression, we can merge this. I will perform some compatibility tests with Flux before tagging a new release

PR #569, opened 2020-03-31
fix gpu integration tests

CarloLucibello:
bu adding a compat entry for CuArrays

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -6,6 +6,12 @@ git-tree-sha1 = "051c95d6836228d120f5f4b984dd5aba1624f716"
 uuid = "621f4979-c628-5d54-868e-fcf4e3e8185c"
 version = "0.5.0"
 
+[[Adapt]]
+deps = ["LinearAlgebra"]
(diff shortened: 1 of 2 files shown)


dhairyagandhi96:
We already have an entry in the extras for a test dependency, and we need to be compatible with CuArrays 2 right?

CarloLucibello:
> We already have an entry in the extras for a test dependency, and we need to be compatible with CuArrays 2 right?

sorry adding to deps was a mistake, check now. I'm not updating to cuarrays 2 here, just making sure that test pass as they used to do

dhairyagandhi96:
Even so, this would limit the CuArrays version Flux will be able to use right? An `up` on the environment should have been fine otherwise.

CarloLucibello:
I tried `add Flux#master`  and the compat here doesn't block the installation of cuarrays 2.0 required by flux

CarloLucibello:
@dhairyagandhi96 why you say up the environment? CuArrays is not part of the environment...
You mean we should create a  Manifest.toml in test/?

dhairyagandhi96:
> I tried add Flux#master and the compat here doesn't block the installation of cuarrays 2.0 required by flux

Does that install release Zygote or master?

PR #574, opened 2020-03-31
Fix GPU broadcasting tests

ararslan:
Tests for the correctness of gradients when doing basic broadcasting over arrays on the GPU were added in #373. However, that was before GPU-enabled CI was set up on the repository, so the tests as written were not put through CI.

The issue with the tests as written are just that they use integers. This PR changes them to `Float32`, which should fix the `MethodError`s that occur when running these tests.

Can somebody with the appropriate permissions bors try this?

### DIFF
--- a/test/cuda.jl
+++ b/test/cuda.jl
@@ -7,7 +7,7 @@ using CuArrays
 end
 
 @testset "basic bcasting" begin
-  a = cu(collect(1:9))
+  a = cu(Float32.(1:9))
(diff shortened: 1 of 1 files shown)


maleadt:
Oh right, that was one breaking change in the CuArrays release: Integers used to be eagerly converted by `cu` to Float32. That's more than questionable, so we got rid of it. However, many NVIDIA intrinsics are not available for integers, so that broke some tests.

PR #581, opened 2020-04-02
Make Cholesky adjoint GPU-friendly

aterenin:
This PR modifies the Cholesky adjoint to replace loop with broadcast, which makes it work on GPU once appropriate low-level linear algebra methods are available. I'm making a number of PRs for these (for example: https://github.com/JuliaGPU/GPUArrays.jl/pull/262).

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -495,14 +495,12 @@ end
   return C, function(Δ::NamedTuple)
     issuccess(C) || throw(PosDefException(C.info))
     U, Ū = C.U, Δ.factors
-    Σ̄  = similar(U.data)
-    mul!(Σ̄ , Ū, U')
+    Σ̄ = similar(U.data)
+    Σ̄ = mul!(Σ̄, Ū, U')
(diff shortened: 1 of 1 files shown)


CarloLucibello:
this can be merged right now irrespective of the GPUArrays changes, right?
Then once your other PR is merged and gpuarrays tagged it would be nice if you could file another PR testing this on gpu

aterenin:
Yes, this can be merged now. For GPU, it needs other PRs for `mul!` and `trsm!` which I will make soon once other PRs get merged in.

CarloLucibello:
what about using non-mutating operations? is `BLAS.trsm` supported right now?

aterenin:
Not yet. The CUBLAS versions are all available from `CuArrays.CUBLAS`, but have a very slightly different API from BLAS (for instance, CUBLAS `trsm!` allows non-in-place, whereas BLAS `trsm!` is in-place only). This means that wrappers are needed to translate between these - will be making a PR with these soon.

CarloLucibello:
ok, let's merge this then

bors r+

PR #582, opened 2020-04-05
tag new version

CarloLucibello:


dhairyagandhi96:
Does Zygote still give the warning about the CuArrays dependency?

CarloLucibello:
it does, only when testing. Probably having a compact entry for an extras dependency is not kosher. Should I remove it?

dhairyagandhi96:
Correct, yes please

dhairyagandhi96:
Registering it now

PR #586, opened 2020-04-06
Add keyword by to adjoint for sort

baggepinnen:


dhairyagandhi96:
Can we add a test here that uses a kwargs?

Thanks a lot!

baggepinnen:
added some proper tests for sorting with and without the `by` kwargs

dhairyagandhi96:
Should we just go with the `kwargs...` Notation instead of just `by`, so we can cover the other kwargs too?

baggepinnen:
The PR adding sort argued against that

dhairyagandhi96:
Okay, hopefully it's alright for now.

bors r+

Thanks!

PR #587, opened 2020-04-07
 Fix 'MethodError: no method matching ∇maxpool' error for NNlib.maxpool

nirmal-suthar:
Fix [FluxML/NNlib.jl/issues/188](https://github.com/FluxML/NNlib.jl/issues/188)

- Added type-assertion to `pdims` which caused error while computing gradient for NNlib.maxpool and NNlib.meanpool
- Added test for maxpool and meanpool which inputs `k::NTuple{N, Integer}` instead of `pdims::PoolDims`

### DIFF
--- a/src/lib/nnlib.jl
+++ b/src/lib/nnlib.jl
@@ -67,12 +67,12 @@ colmajor(::AbstractColumnMajor, x) = x
        )
    end
 
-@adjoint function maxpool(x, pdims; kw...)
+@adjoint function maxpool(x, pdims::NNlib.PoolDims; kw...)
   y = maxpool(x, pdims; kw...)
   y, Δ -> (NNlib.∇maxpool(Δ, y, x, pdims; kw...), nothing)
 end
 
-@adjoint function meanpool(x, pdims; kw...)
+@adjoint function meanpool(x, pdims::NNlib.PoolDims; kw...)
   y = meanpool(x, pdims; kw...)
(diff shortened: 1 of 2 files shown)


CarloLucibello:
nice, this looks good. If you don't mind, before merging this, could you investigate if we have similar problems with the convolution functions? If we do, you could fix those in this PR

nirmal-suthar:
Sure, will do it.

nirmal-suthar:
For conv:
```
julia> using Flux
julia> weight = rand(2,2,3,4) # equivalent to weights of Conv((2,2),3=>4)
2×2×3×4 Array{Float64,4}:
julia> gradient(() -> sum(conv(rand(2,2,3,1), weight)), params(weight))
Grads(...)
```
For depthwiseconv:
```
julia> using Flux
julia> weight  = rand(2,2,2,3) # equivalent to weights of DepthwiseConv((2,2),3=>6)
2×2×2×3 Array{Float64,4}:
julia> gradient(()-> sum(depthwiseconv(rand(2,2,3,1), weight)), params(weight))
Grads(...)
```
I am not able to find any problem with calculating gradients of conv and depthwiseconv (also Grads(...) are non-empty). 
Is there any need to add type-assertion to `cdims` in conv and depthwiseconv? (Don't know they behave normally and pool doesn't)

CarloLucibello:
thank you!

bors r+

PR #591, opened 2020-04-08
update project and tag new version

CarloLucibello:


dhairyagandhi96:
Why are we removing the docs manifest?

CarloLucibello:
> Why are we removing the docs manifest?

Documenter's docs just suggest to have Project.toml with a compact entry. Why do we have a manifest?

dhairyagandhi96:
So we can push out docs updates with a reproducible environment and so we can push doc updates without a new release everytime

PR #593, opened 2020-04-08
Fixes bug in Zeros / Ones with non-Float64 element type

willtebbutt:


dhairyagandhi96:
Happy with the patch!
Do we need the release immediately yet?

willtebbutt:
> Do we need the release immediately yet?

Would be very helpful :)

PR #605, opened 2020-04-20
remove eval from Hermitian/Symmetric power series functions gradchecks

oxinabox:
no need for `eval` 
(no need even for `getfield(Base, func)`)

I have tested this on my branch which fails those tests right now, and so I can see it displays them the same


dhairyagandhi96:
Not sure why the Travis build errored like that. Hopefully just a network issue

Lgtm! Will have the bot merge it once gitlab is done with it's thing. Extra queues often just slow things down.

PR #606, opened 2020-04-20
Add SnoopCompile bot

ianshmean:
Adding the [SnoopCompile bot](https://timholy.github.io/SnoopCompile.jl/stable/bot/) action, to explore improving load times

PR #607, opened 2020-04-21
stop running functions in init

KristofferC:
Zygote currently does the same thing as CSV used to do (https://github.com/JuliaData/CSV.jl/issues/324) which is to run some representative functions in `__init__` to make the first call look faster. In reality, this just shifts the latency in the first call to package load time. The problem is that Zygote can be loaded in a Julia session without necessarily getting called. In those scenarios, users have to pay the compilation cost anyway which makes it more costly to have Zygote as a dependency.

### DIFF
--- a/src/Zygote.jl
+++ b/src/Zygote.jl
@@ -45,9 +45,6 @@ end
 
 precompile() = include(joinpath(@__DIR__, "precompile.jl"))
 
-# precompile()
-@init Requires.isprecompiling() || precompile()
-
 # helps to work around 265-y issues
 function refresh()
   include(joinpath(@__DIR__, "compiler/interface2.jl"))


dhairyagandhi96:
If we are removing the code to be called, then does `precompile.jl` need to be removed as well?

ianshmean:
This makes sense, and would address the issue that motivated me to look into #606 
Also, on reflection, the contents of `precompile.jl` is probably what the SnoopCompile bot should be running in #606 rather than the unit tests

ianshmean:
As a case study, loading is currently quite painful on a modern aarch64 system
```
julia> @time using Zygote
[ Info: Precompiling Zygote [e88e6eb3-aa80-5325-afca-941959d7151f]
155.025100 seconds (76.22 M allocations: 3.948 GiB, 2.12% gc time)

julia> exit()
ian@nx1:~/Documents/julia-1.4.1/bin$ ./julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.4.1 (2020-04-14)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> @time using Zygote
 90.218599 seconds (56.06 M allocations: 2.980 GiB, 2.39% gc time)
```

This PR improves things a lot (but aarch64 does seem disproportionately slow at loading for some reason..)
```
julia> @time using Zygote
[ Info: Precompiling Zygote [e88e6eb3-aa80-5325-afca-941959d7151f]
 58.374538 seconds (23.72 M allocations: 1.380 GiB, 0.99% gc time)

julia> exit()
ian@nx1:~/Documents/julia-1.4.1/bin$ ./julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.4.1 (2020-04-14)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> @time using Zygote
 19.383780 seconds (22.75 M allocations: 1.336 GiB, 2.19% gc time)
```
And adding the SnoopCompile bot to run the `precompile.jl` script and provide baked in precompile statements should restore the benefit while keeping load time lower. New PR incoming, @aminya is working in upcoming changes to the SnoopCompile bot api

CarloLucibello:
so this definitely seems worth doing, and we can keep `precompile.jl` for SnoopCompile

bors r+

MikeInnes:
Yes, we may as well remove `precompile.jl` here too.

aminya:
@MikeInnes this is done in https://github.com/FluxML/Zygote.jl/pull/615

PR #609, opened 2020-04-22
Tweaks euclidean distance

willtebbutt:
Makes the `SqEuclidean` computation recompute if there are potentially roundoff errors in the forwards pass. This makes it possible to use Zygote to propagate gradient information when inputs are close together. Prior to this change you wind up with negative arguments to `sqrt`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.4.15"
+version = "0.4.16"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
--- a/src/lib/distances.jl
+++ b/src/lib/distances.jl
@@ -65,7 +65,7 @@ end
 end
 
 @adjoint function pairwise(::Euclidean, X::AbstractMatrix, Y::AbstractMatrix; dims=2)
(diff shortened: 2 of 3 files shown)


dhairyagandhi96:
Sgtm

willtebbutt:
Am I good to bors @dhairyagandhi96 ?

dhairyagandhi96:
I think so

dhairyagandhi96:
Might need to think about types here, maybe eps(eltype(x))?

willtebbutt:
> Might need to think about types here, maybe eps(eltype(x))?

Good point.

willtebbutt:
@dhairyagandhi96 looks good to you now?

dhairyagandhi96:
Probably just stylistic, also could just use the eltype like `X::AbstractMatrix{T} where T` instead if it seems too long for one line?

willtebbutt:
That also doesn't fit on one line 😂 . Maybe

dhairyagandhi96:
Hah

I am alright with the actual patch, so probably let's get this in and hopefully that's alright

willtebbutt:
Cool. I'll bump the patch version and bors then :)

devmotion:
I think this is fixed upstream in https://github.com/JuliaStats/Distances.jl/pull/161. Hence the special implementation in this PR should not be needed anymore if the upstream PR is merged.

PR #610, opened 2020-04-22
@nograd randn!

aterenin:
It's very commonplace when using Flux with CuArrays to write `x = randn!(similar(y, (16,32)))` rather than `randn(16,32)` because this works equally well whether or not the input `y` is an `Array` or `CuArray`.

This PR adds `@nograd` for `randn!`, whereas right now it is only reserved for `randn`.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -1,13 +1,14 @@
 using FillArrays, AbstractFFTs
 using FillArrays: AbstractFill, getindex_value
 using Base.Broadcast: broadcasted, broadcast_shape
+import Random.randn!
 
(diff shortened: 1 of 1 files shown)


dhairyagandhi96:
Seems fine to me

I'm assuming it doesn't mess with the gradients, we should be accumulating the gradients till `similar` like usual.

bors r+

aterenin:
Yep. And on that note, any objections if I add `@nograd similar` to this too, or should I make a separate PR for that? I think I needed this in my project code as well.

dhairyagandhi96:
I think differentiating through similar sound be fine. Why do we need that?

PR #611, opened 2020-04-23
copyto! for Params and Grads

CarloLucibello:
Implement `copyto!`, copying params and gradients to flat vectors (and viceversa). This can be convenient for interoperability with other libraries in the julia ecosystem, e.g. [see this for Optim.jl](https://github.com/baggepinnen/FluxOptTools.jl/blob/master/src/FluxOptTools.jl)

I had to add an extra field to Grads containing the original params object, since Grads sometimes contain some extra elements. I don't think this is a problem. cc @baggepinnen

### DIFF
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -85,8 +85,40 @@ function Base.show(io::IO, ps::Params)
   print(io, "])")
 end
 
+
+"""
+    copyto!(ps::Params, x::AbstractVector)
+    copyto!(x::AbstractVector, ps::Params)
+    
+Copies the content of array `x` into the parameters `ps` or viceversa.
+The length of `x` has to be equal to the sum of the lenghts
+of all parameters.
+"""
+function Base.copyto!(ps::Params, x::AbstractVector)
+  @assert length(x) == sum(length(p) for p in ps)
+  i = 0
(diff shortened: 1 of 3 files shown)


CarloLucibello:
test error on julia nightly is unrelated and due to `@timed` returning a namedtuple now instead of a tuple

CarloLucibello:
@dhairyagandhi96 any objection to this?

src/compiler/interface.jl
```diff
@@ -85,8 +85,40 @@ function Base.show(io::IO, ps::Params)
   print(io, "])")
 end
 
+
+"""
+    copyto!(ps::Params, x::AbstractVector)
+    copyto!(x::AbstractVector, ps::Params)
+    
+Copies the content of array `x` into the parameters `ps` or viceversa.
+The length of `x` has to be equal to the sum of the lenghts
```
MikeInnes, line 94:
```suggestion
The length of `x` has to be equal to the sum of the lengths
```

MikeInnes:
Seems reasonable to me, thanks.

bors r+

dhairyagandhi96:
Speculatively related to https://julialang.slack.com/archives/C7LFJTXV5/p1587743941283300

PR #612, opened 2020-04-23
utils for Optim.jl

CarloLucibello:
a convenience function for interoperability with Optim.jl, essentially taken from https://github.com/baggepinnen/FluxOptTools.jl. cc @baggepinnen  
It would be nice to have this here since it is just a few lines of code carrying no extra dependencies. 
Needs #611 to be merged first

### DIFF
--- a/src/Zygote.jl
+++ b/src/Zygote.jl
@@ -31,6 +31,7 @@ include("lib/nnlib.jl")
 include("lib/forward.jl")
 include("lib/utils.jl")
 include("lib/range.jl")
+include("lib/optim.jl")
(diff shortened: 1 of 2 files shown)


MikeInnes:
I'm uneasy about adding this. It's a bit different from #611, where there was an existing API that couldn't have another meaning, since anything done outside Zygote would have to either be clumsier or commit type piracy.

For this I think having a dedicated package like FluxOptTools is quite a nice idea, especially since then it can have dedicated docs, it's easier for it to change over time and so on.

PR #613, opened 2020-04-24
reducedim_init for accum

cossio:
Fix #594.

Issue #594, opened 2020-04-09
reducedim_init MethodError with structs


cossio:
```julia
using Zygote
struct A
    x::Float64
end
f(a,v) = a.x + v
g(A,V) = sum(f.(A,V))
gradient(g, A.(randn(2)), randn(2,2)) # Error
```

Error message and stacktrace:

```
ERROR: LoadError: MethodError: no method matching reducedim_init(::typeof(identity), ::typeof(Zygote.accum), ::Array{NamedTuple{(:x,),Tuple{Float64}},2}, ::Tuple{Int64,Int64})
Closest candidates are:
  reducedim_init(::Any, ::Union{typeof(+), typeof(Base.add_sum)}, ::AbstractArray, ::Any) at reducedim.jl:109
  reducedim_init(::Any, ::Union{typeof(*), typeof(Base.mul_prod)}, ::AbstractArray, ::Any) at reducedim.jl:112
  reducedim_init(::Any, ::typeof(min), ::AbstractArray, ::Any) at reducedim.jl:132
  ...
Stacktrace:
 [1] _mapreduce_dim(::Function, ::Function, ::NamedTuple{(),Tuple{}}, ::Array{NamedTuple{(:x,),Tuple{Float64}},2}, ::Tuple{Int64,Int64}) at ./reducedim.jl:317
 [2] #mapreduce#580 at ./reducedim.jl:307 [inlined]
 [3] #reduce#582 at ./reducedim.jl:352 [inlined]
 [4] accum_sum(::Array{NamedTuple{(:x,),Tuple{Float64}},2}; dims::Tuple{Int64,Int64}) at /home/cossio/.julia/packages/Zygote/4tJp5/src/lib/broadcast.jl:36
 [5] unbroadcast at /home/cossio/.julia/packages/Zygote/4tJp5/src/lib/broadcast.jl:48 [inlined]
 [6] map(::typeof(Zygote.unbroadcast), ::Tuple{Array{Main.ZygoteStruct.A,1},Array{Float64,2}}, ::Tuple{Array{NamedTuple{(:x,),Tuple{Float64}},2},Array{Float64,2}}) at ./tuple.jl:177
[...]

PR #614, opened 2020-04-24
accum_zero

cossio:
https://github.com/FluxML/Zygote.jl/pull/613 is not type-stable. This version should be type-stable.
See https://github.com/FluxML/Zygote.jl/issues/594.

Currently tests fail, because `accum_zero` is capturing cases it shouldn't.

@CarloLucibello @MikeInnes


MikeInnes:
Just noting that this is effectively an API change (since it may not be obvious). We should think carefully about the edge cases for this since it may break some cases.

I'm not sure the `Union{T,Nothing}` is that big of a problem here. Lots of work has gone into Julia to make this case fast. It won't prevent things like BLAS (since the numeric case is handled separately). Making the compiler or runtime figure out precisely when the union is needed might have more cost than assuming it always is.

It might be useful to have a concrete example of where we believe this will be useful.

CarloLucibello:
fixed elsewhere

PR #615, opened 2020-04-24
SnoopCompile bot

aminya:
It uses the whole test suit of Zygote (except Cuda and Abstract FFT tests) for generating OS-Version specific precompilation sentences.

In running the tests shows that we get:
 - `10s` raw improvement  
 - `23 M` less memory usage
 - `1.16GiB` less memory allocation.

Tests:
- A successful run: https://github.com/aminya/Zygote.jl/actions/runs/95945010
- A successful PR created by the bot: https://github.com/aminya/Zygote.jl/pull/4/files which has precompilation sentences for Windows, Linux, macOS for Julia 1.4 and Julia 1.3.
The CI also passes for that generated PR: https://travis-ci.com/github/aminya/Zygote.jl/builds/162029060

cc: @ianshmean

related: https://github.com/FluxML/Zygote.jl/pull/607

The changes made in the tests are just adding two simple checks for existance of the `SnoopCompile_ENV`. GitHub diff isn't working properly here.

# Ready for review

### DIFF
--- a/.github/workflows/SnoopCompile.yml
+++ b/.github/workflows/SnoopCompile.yml
@@ -73,7 +73,7 @@ jobs:
           # committer: YOUR NAME <yourEmail@something.com> # Change `committer` to your name and your email.
           title: "[AUTO] Update precompiles'"
           labels: SnoopCompile
-          branch: "SnoopCompile"
+          branch: "SnoopCompile_AutoPR"
 
 
   Skip:
--- a/.github/workflows/SnoopCompile.yml
+++ b/.github/workflows/SnoopCompile.yml
@@ -1,0 +1,84 @@
+name: SnoopCompile
+
+# Only runs on pushes. Edit based on your taste.
+on:
+  push:
+    branches:
+      # - 'master'
+  pull_request:
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  SnoopCompile:
+    if: "!contains(github.event.head_commit.message, '[skip ci]')"
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        # Uncomment other versions if you want multi-version signatures (should exactly match BotConfig.version):
+        version:
+          - '1.4.1'
+          # - '1.2.0'
+          # - '1.0.5'
+        # Uncomment other options if you want multi-os signatures (currently only these are supported by Github):
+        os:
+          - ubuntu-latest
+          # - windows-latest
+          # - macos-latest
+        arch:
+          - x64
+    steps:
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@latest
+        with:
+          version: ${{ matrix.julia-version }}
+      - name: Install dependencies
+        run: |
+          julia --project -e 'using Pkg; Pkg.instantiate();'
+          julia -e 'using Pkg; Pkg.add(PackageSpec(url = "https://github.com/aminya/SnoopCompile.jl", rev = "multios")); Pkg.develop(PackageSpec(; path=pwd())); Pkg.add("JuliaFormatter"); using SnoopCompile; SnoopCompile.addtestdep();'
+      - name: Generating precompile files
+        run: julia --project -e 'include("deps/SnoopCompile/snoopi_bot.jl")'
+      - name: Running Benchmark
+        run: julia --project -e 'include("deps/SnoopCompile/snoopi_bench.jl")'
+      - name: Format precompile_includer.jl
+        run: julia --project -e 'using JuliaFormatter; format_file("src/precompile_includer.jl")'
+      - name: Upload all
+        uses: actions/upload-artifact@v2-preview
+        with:
+          path: ./
+
+  Create_PR:
+    if: "!contains(github.event.head_commit.message, '[skip ci]')"
+    needs: SnoopCompile
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download all
+        uses: actions/download-artifact@v2-preview
+      - name: Move the content of the directory to the root
+        run: |
+          mv -v artifact/* ./
+          mv -v artifact/.[^.]* ./
+      - name: Discard unrelated changes
+        run: |
+          git checkout -- Project.toml
+          git diff -w --no-color | git apply --cached --ignore-whitespace && git checkout -- . && git reset && git add -p
+      - name: Create Pull Request
+        # https://github.com/marketplace/actions/create-pull-request
+        uses: peter-evans/create-pull-request@v2
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: Update precompile_*.jl file
+          # committer: YOUR NAME <yourEmail@something.com> # Change `committer` to your name and your email.
+          title: "[AUTO] Update precompiles'"
+          labels: SnoopCompile
+          branch: "SnoopCompile"
+
+
+  Skip:
+    if: "contains(github.event.head_commit.message, '[skip ci]')"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Skip CI 🚫
+        run: echo skip ci
--- a/.github/workflows/SnoopCompile.yml
+++ b/.github/workflows/SnoopCompile.yml
@@ -1,11 +1,11 @@
 name: SnoopCompile
 
-# Only runs on pushes. Edit based on your taste.
+# Edit based on your repository.
 on:
   push:
     branches:
-      # - 'master'
-  pull_request:
+      # - 'master' # Uncomment for big repositories
+  pull_request:  # Comment for big repositories
 
 defaults:
   run:
@@ -21,12 +21,12 @@ jobs:
         version:
           - '1.4.1'
           # - '1.2.0'
-          # - '1.0.5'
+          - '1.0.5'
         # Uncomment other options if you want multi-os signatures (currently only these are supported by Github):
         os:
           - ubuntu-latest
-          # - windows-latest
-          # - macos-latest
+          - windows-latest
+          - macos-latest
         arch:
           - x64
     steps:
--- a/.github/workflows/SnoopCompile.yml
+++ b/.github/workflows/SnoopCompile.yml
@@ -22,7 +22,7 @@ jobs:
         version:
           - '1.4.1'
           - '1.3.1'
-          - '1.2.0' # min
+          # - '1.2.0' # min
         # Uncomment other options if you want multi-os signatures (currently only these are supported by Github):
         os:
           - ubuntu-latest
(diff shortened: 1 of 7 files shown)


aminya:
I found a bug in GitHub or Julia. Despite running 1.3.1, `VERSION` is given as `1.4.1`! How is this even possible?? 😨 

https://github.com/aminya/Zygote.jl/runs/616453405#step:5:6
https://github.com/aminya/SnoopCompile.jl/blob/2bcc1c4205e59963fd0a3e9f51be46ee79a43d96/src/bot/snoopi_bot.jl#L68
https://github.com/aminya/SnoopCompile.jl/blob/2bcc1c4205e59963fd0a3e9f51be46ee79a43d96/src/bot/snoopi_bot.jl#L37

aminya:
There is an issue with the compatibility bounds of packages on Julia 1.2:
```julia
ERROR: Unsatisfiable requirements detected for package CuArrays [3a865a2d]:
```
https://github.com/aminya/Zygote.jl/runs/616498612?check_suite_focus=true#step:4:132

I will remove Julia 1.2 from the bot.

CarloLucibello:
So, what are the consequences of merging this? People will get precompiled files when updating Zygote? 
Another question: I see there is a benchmark script, what are the results?

aminya:
> So, what are the consequences of merging this? People will get precompiled files when updating Zygote?

Yes, once you merge this PR, the bot will run and will make another PR like [this](https://github.com/aminya/Zygote.jl/pull/4/files). After you merge that PR, Julia will cache the inference result for the common operations of the Zygote. 

In the future, the same precompilation signatures can be used by PackageCompiler to make an almost static Zygote. Not only for `using Zygote` but also for calling its functions.

> Another question: I see there is a benchmark script, what are the results?
The current benchmark only measures the inference time:
```
 Benchmarking the inference time of `using Zygote`
┌ Info: *******************
│ Benchmark Started
└ *******************
┌ Info: Precompile Deactivated Inference Benchmark
└ ------------------------
[ Info: 170.793 ms
┌ Info: Precompile Activated Inference Benchmark
└ ------------------------
[ Info: 159.906 ms
┌ Info: *******************
│ Benchmark Finished
└ *******************
Benchmarking the inference time of `using Zygote` & basic function test
┌ Info: *******************
│ Benchmark Started
└ *******************
┌ Info: Precompile Deactivated Inference Benchmark
└ ------------------------
[ Info: 19043.853 ms
┌ Info: Precompile Activated Inference Benchmark
└ ------------------------
[ Info: 18622.542 ms
┌ Info: *******************
│ Benchmark Finished
└ *******************
```

CarloLucibello:
So the real gain is larger than what appears in the benchmark?

Is it reasonable to trigger bot for each PR merged? What about doing it only when tagging a new release requires, so to avoid much maintenance burden?

dhairyagandhi96:
Is it a ji file Julia can load directly?

Are there security concerns we should be aware of of shipping precompiled files not produced on the local machine?

dhairyagandhi96:
This is for my own education as anything else

dhairyagandhi96:
If this is successful, I would like to explore doing a similar build for Flux

aminya:
> 
> 
> So the real gain is larger than what appears in the benchmark?
> 

Yes. I plan to make a PackageCompiler bot, that will use these to make static system images.

> Is it reasonable to trigger bot for each PR merged? What about doing it only when tagging a new release, so to avoid much maintenance burden?

It is better to run it every time because as functions change, a new precompilation is needed. 

To minimize the number of PRs created, for bigger repositories I recommend to use: 
```yaml
on:
  push:
    branches:
       - 'master'
  # pull_request:  # comment for big repositories
```
which runs only when people push/merge something to master. 

> 
> 
> Is it a ji file Julia can load directly?
> 

Yes, the precompilations are calls to `Base.precompile`. Many packages already use this. For example, [Plots.jl](https://github.com/JuliaPlots/Plots.jl/blob/master/src/precompile.jl).

You can see a sample bot PR here: 
https://github.com/aminya/Zygote.jl/pull/4/files


> Are there security concerns we should be aware of shipping precompiled files not produced on the local machine?

IMO, it is better than a local machine. Since we cover different OS and Julia versions. Usually, you can run your script on different configurations. The bot reruns the tests internally in its benchmark stage and will error if anything goes wrong. You will also check the CI before merging,

aminya:
> 
> 
> If this is successful, I would like to explore doing a similar build for Flux

I will create a PR for Flux too.

aminya:
If you want to run the bot only before publishing you can remove the following altogether
https://github.com/FluxML/Zygote.jl/pull/615/files#diff-0b4f3d0a2175dd678c65e865e174454aR4
```yaml
# Edit based on your repository.
on:
  push:
    branches:
      # - 'master'
  # pull_request:  # comment for big repositories
```

and use something like 

```yaml
on:
  watch
    types: [started]
```
Based on the following post, the workflow will be executed each time you star, or unstar and star again the repo.
https://dev.to/s_abderemane/manual-trigger-with-github-actions-279e

You may do this only before registering. When registration is done, you can set `should_precompile` to `false` again to let people develop without any precompilation happening.
https://github.com/aminya/Zygote.jl/pull/4/files#diff-2270ce88ed8290c3b808f5ea76728946R1

CarloLucibello:
ok, this is nicer. I wonder if we can achieve something essentially maintenance-free. E.g. reduce the publishing process to the following comment on the last commit:
```
@SnoopCompile precompile()  # precompile, automatically merge, set should_precompile=false
```
and then
```
@JuliaRegistrator register()
```
on the SnoopCompile commit

aminya:
> 
> 
> ok, this is nicer. I wonder if we can achieve something essentially maintenance-free. E.g. reduce the publishing process to the following comment on the last commit:
> 
> ```
> @SnoopCompile precompile()  # precompile, automatically merge, set should_precompile=false
> ```
> 
> and then
> 
> ```
> @JuliaRegistrator register()
> ```
> 
> on the SnoopCompile commit

If you want a commit message trigger, you can do the following

```yaml
on:
  push:
    branches:
       - 'master'
jobs:
  SnoopCompile:
    if: "contains(github.event.head_commit.message, '[precompile]')"
```

```yaml
  Skip:
    if: "!contains(github.event.head_commit.message, '[precompile]')"
```

However, you need to merge the PR and start the registration. When registration is done, set the `should_precompile` to `false`. Automating the last part requires interfacing with `JuliaRegsitrator`, which I don't think I can implement any time soon.

aminya:
It is possible to make the PR merge itself, however, I don't think that is safe. Human review is needed in this process.

MikeInnes:
One option would be to have a release branch; there are no precompile files on master, but on each tag we build a set of precompile files, commit on that branch, and register.

I'd also like to more clearly understand how this affects (cold/hot) package load time and 'time to first gradient'. Last time I tried this kind of thing it didn't achieve much, so we should clearly understand how it'll help before adding the maintenance burden.

aminya:
> 
> 
> One option would be to have a release branch; there are no precompile files on master, but on each tag we build a set of precompile files, commit on that branch, and register.
> 

That works too. But I think setting `should_precompile = false` after registration is easier than creating another branch. It is up to you. If you are already planning for such a branch, then we can go for it.

> I'd also like to more clearly understand how this affects (cold/hot) package load time and 'time to first gradient'. Last time I tried this kind of thing it didn't achieve much, so we should clearly understand how it'll help before adding the maintenance burden.

As I mentioned previously, the gain is very high when we create a sysimage (I am planning to make a bot for that too). Because we can store compiled functions.

For the raw improvements in a dynamic way (that Julia only stores inference result), there is already a benchmark implemented.

I also created an issue to track other possibilities for benchmarking.
https://github.com/timholy/SnoopCompile.jl/issues/78

MikeInnes:
We should definitely not add this for the benefit when creating a sysimg. If you're doing that it's really easy to just call `gradient` and all this stuff will be compiled.

aminya:
> 
> 
> We should definitely not add this for the benefit when creating a sysimg. 

That is not the case. We are not doing this **only** for sysimage. Using current timing tools we still get performance improvements.

>If you're doing that it's really easy to just call `gradient` and all this stuff will be compiled.

Using a simple `gradient` case, only 90 precompile signatures were generated. But now 1190 signatures are generated.
https://github.com/aminya/Zygote.jl/pull/4/files

MikeInnes:
Right, an improvement of 2% for the time it takes to load zygote and take a gradient? If it was a 2x difference it might be worth considering.

> Using a simple `gradient` example, only 90 precompile signatures were generated. But now 1190 signatures are generated.

So what? 13× more precompile signatures are clearly not improving the performance by 13×. It matters that the Zygote compiler itself is cached, not that everything it has ever compiled is cached. And it seems hard to get that benefit, with the current Julia compiler, unless you call `gradient` at the time that native code is being generated.

aminya:
@MikeInnes Currently, only Zygote is using its precompile signatures (SnoopCompile generates many more for other packages that are called). Once every package starts to use them, this number will show itself more.

MikeInnes:
If other packages decide this is worthwhile, great, we'll take that improvement. But the marginal impact from this patch will still itself be a very small improvement and we're unlikely to consider it worth the additional hassle.

aminya:
Running `@timev` in running the tests shows that we get:
 - 10s raw improvement  
 - 23 MB less memory usage
 - 1GiB less memory allocation.

```jl
│ Precompile Deactivated Benchmark
# for using Zygote:
└ Inference time (ms): 	23654.842
@timev result (This has some noise).
391.374697 seconds (628.72 M allocations: 31.522 GiB, 4.16% gc time)
elapsed time (ns): 391374697041
gc time (ns):      16295209095
bytes allocated:   33846029551
pool allocs:       628513544
non-pool GC allocs:190530
malloc() calls:    11983
realloc() calls:   85
GC pauses:         503
```
```
│ Precompile Activated Benchmark
┌ Info: 
└ Inference time (ms): 	22565.431
@timev result (This has some noise).
382.785964 seconds (605.11 M allocations: 30.360 GiB, 3.93% gc time)
elapsed time (ns): 382785964109
gc time (ns):      15042733929
bytes allocated:   32598547445
pool allocs:       604912736
non-pool GC allocs:184276
malloc() calls:    11737
realloc() calls:   81
GC pauses:         467
full collections:  6
```

CarloLucibello:
As things stand right now, I think this has to be entirely maintenance-free to be worth doing

aminya:
> As things stand right now, I think this has to be entirely maintenance-free to be worth doing

If you want you can have a release branch, and register from that branch.

MikeInnes:
Thanks a lot for looking into this. Given the cost/benefit ratio I don't think it makes sense for us at the moment, though.

PR #618, opened 2020-04-25
adjoints for Transpose and Adjoint constructors

simeonschaub:
Add adjoints for the lower level `Transpose` and `Adjoint` constructors. I also added an adjoint for `parent(::Transpose)` while I was at it. These shouldn't get called a lot directly, but I think we might as well have them. This came up in https://discourse.julialang.org/t/zygote-gradient-error/38206.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -317,13 +317,27 @@ end
   return transpose(x), back
 end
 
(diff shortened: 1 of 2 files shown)


CarloLucibello:
thanks!

bors r+

Issue #571, opened 2020-03-31
Why does Zygote differentiate unnecessary code?


jessebett:
This discussion was started on Slack, but not resolved:

Okay, so I know I’ve asked this a few times. This is related to Zygote failing to support `@info` or string interpolation inside a function being differentiated. Though this issue has been solved with clever `@nograd` definitions, I think it indicates a general problem that I would like to understand.

Why does Zygote attempt to differentiate code that does not affect the primal value?

```
using Zygote
f(x) = 2x
losses=[]
Zygote.gradient(1.) do x
  loss = f(x)
  append!(losses,loss)
  return loss
end
```

For example, the above code triggers the `mutating arrays not supported` error. This kind of problem can't be solved like above with `nograd` because in general we might want to differentiate through mutating arrays. However, this case the primal value is unaffected by it being appended to an array, mutation should not be part of the expression being differentiated. 

My understanding is that Zygote can’t guarantee in general that functions called don’t have side effects or closures, but what I don’t understand is why that information isn’t available to be inferred by Zygote’s IR in this case (or the case where we’re logging or printing).

Specifically this should not require any ability to mutate arrays, it only requires realizing that the returned value isn’t affected by that array mutation. Same with differentiating `@info`, shouldn’t involve anything with `try/catch`. That this is being solved by `nograd` on specific functions is fine for printing and logging, but what is preventing Zygote from inferring what code is actually affecting the primal value?

@oxinabox @MikeInnes can you shed some light for me why Zygote can't infer the information required to address this more generally? (I pinged on Slack already, but here as well) @

This is quite a common pattern. e.g logging or appending the primal value of a loss function during training. Also that something like `@nograd` can’t be used inside the do block to manually tell Zygote to ignore it. @oxinabox can you link the PR for the function you were describing that could be used inside the gradient like `@ignore`?

```
Zygote.gradient(1.) do x
  loss = f(x)
  Zygote.@ignore append!(losses,loss)
  return loss
end
```

jessebett:
For example, using Zygote's IR:

```
vals = []
f(x) = 2x

function g(x)
  y = 2. * x
  append!(vals,y)
  return y
end

Zygote.@code_ir g(1.)
```

We get:

```
julia> Zygote.@code_ir g(1.)
1: (%1, %2)
  %3 = 2.0 * %2
  %4 = Main.append!(Main.vals, %3)
  return %3
```

[...]

Issue #621, opened 2020-04-26
nograd not working properly?


cossio:
```julia
using Zygote
isderiving() = false  # https://fluxml.ai/Zygote.jl/latest/adjoints/#Gradient-Reflection-1
Zygote.@adjoint isderiving() = true, _ -> nothing
f(x) = (@show isderiving(); sum(x))
g(x) = Zygote.nograd(f(x))
g'(randn())  # isderiving() = true   (!)
```

I was expecting the `f` not to be differentiated here. What is going on?

cossio:
Once https://github.com/FluxML/Zygote.jl/pull/623 is merged, we will be able to write:

```julia
using Zygote; using Zygote: @adjoint, ignore
isderiving() = false  # https://fluxml.ai/Zygote.jl/latest/adjoints/#Gradient-Reflection-1
@adjoint isderiving() = true, _ -> nothing
f(x) = (@show isderiving(); sum(x))
g(x) = ignore(() -> f(x))
g'(randn())  # isderiving() = false
```

So with `ignore` we really prune the computation of the gradient of `f`.

CarloLucibello:
Closing as deprecated functionality. `dropgrad` is now deprecated (`Base.@deprecate dropgrad(x) ChainRulesCore.ignore_derivatives(x)` in `src/deprecated.jl`); use `ChainRulesCore.ignore_derivatives` instead, which avoids differentiating the discarded branch. We won't be optimizing the deprecated `dropgrad` path.

PR #623, opened 2020-04-28
Ignore utility

MikeInnes:
Remove the `nograd` utility (which duplicated `dropgrad` and add an `ignore` for blocks of code which e.g. uses mutation. Should help with cases like #571, cc @jessebett.


cossio:
This also helps with https://github.com/FluxML/Zygote.jl/issues/621.

```
using Zygote; using Zygote: @adjoint, ignore
isderiving() = false  # https://fluxml.ai/Zygote.jl/latest/adjoints/#Gradient-Reflection-1
@adjoint isderiving() = true, _ -> nothing
f(x) = (@show isderiving(); sum(x))
g(x) = ignore(() -> f(x))
g'(randn())  # isderiving() = false
```

test/gradcheck.jl
```diff
@@ -1450,11 +1450,7 @@ end
   @test gradient(x -> findfirst(ismissing, x), [1, missing]) == (nothing,)
   @test gradient(x -> findlast(ismissing, x), [1, missing]) == (nothing,)
   @test gradient(x -> findall(ismissing, x)[1], [1, missing]) == (nothing,)
-end
-
-@testset "nograd" begin
-  f(x) = sum(Zygote.nograd(x) + x)
-  @test gradient(f, randn(5)) == (ones(5),)
+  @test gradient(x -> Zygote.ignore(x -> x*x), 1) == (nothing,)
```
cossio, line 1453: The test is failing because `ignore` calls a function without arguments. So the test should be written as:

```julia
gradient(x -> Zygote.ignore(() -> x*x), 1) == (nothing,)
```

MikeInnes: You may as well make that a github suggestion, and I'll commit it :)

cossio:
```suggestion
  @test gradient(x -> Zygote.ignore(() -> x*x), 1) == (nothing,)
```

cossio: First time I use the github suggestion feature! Hope it's ok.

src/lib/grad.jl
```diff
@@ -16,23 +10,6 @@ macro nograd(ex)
...
-([1.0, 1.0, 1.0, 1.0, 1.0],)
-```
-
-In the example above, the gradient is an array of ones even though
-`f(x)` computes `sum(2x)`, because `nograd(x)` uses the value of `x`
-as if it were a constant.
-"""
-nograd(x) = x
-@nograd nograd
-
```
oxinabox, line 35: why are these docstring gone?
Are these being deprecated ? If so why not be ising `@deprecate`/`@deprecate_binding` ?

MikeInnes: I don't think anyone was using them; `nograd` was only around temporarily and duplicated the existing `dropgrad`, which still exists.

CarloLucibello:
so we'll have `@nograd`, `dropgrad` and `ignore` doing related work, can we do a bit better in merging names in order to improve discoverability, i.e. deprecating `dropgrad` instead of `nograd`? And how bad would be to use `nograd` also for `ignore`:
``` julia
nograd(x) = x
nograd(f::Function) = f()
```
?

src/lib/grad.jl
CarloLucibello: still we can add a deprecation

dhairyagandhi96:
Ignore would supersede the other two, right?

MikeInnes:
We shouldn't use the name `nograd` because it does something quite different from `@nograd`, so that would be misleading. `@nograd` and `ignore` are closer, so we could potentially use `@ignore`. For a while I've wanted to deprecate `@nograd` in favour of something that takes a type signature, so that would be a good opportunity to do that.

> And how bad would be to use `nograd` also for `ignore`

In general the meaning of a function shouldn't change based on type. In particular, Julia functions don't have to be `::Function` (e.g. callable objects) and `::Function`s can be used as values (e.g. `dropgrad` on a closure is meaningful), so we need both names to understand what the user intended.

cossio:
We should consider also adding a macro (with whatever name is decided):

```julia
macro ignore(ex)
	return :(ignore() do
		$ex
	end)
end
```

so that

```julia
ignore() do
    ...
end
```

can also be written as:

```julia
@ignore ...
```

WIth that I think we should not rename the current `@nograd` to `@ignore`.

cossio:
What if we remove the currently defined `dropgrad` (which acts on values)? It has the issue that it still computes gradients so it is wasteful (https://github.com/FluxML/Zygote.jl/issues/621). Anything it does can be accomplished with the other functions here. 

If we do that, a way forward could be to rename the `ignore` (and `@ignore`) in this PR to `dropgrad`, `@dropgrad`.

Another option is to rename the currently defined `@nograd` to `@noadjoint`.

MikeInnes:
It's true that `dropgrad(x)` can be written `ignore(() -> x)`. We should probably discuss dropping `dropgrad` separately from this PR. (I think it may be worth keeping since for most use cases it has the more primitive semantics we want, and more compiler work will solve the problem of gradients being redundantly computed).

cossio:
Maybe just merge this? Since Zygote is still pre-1.0, we can keep all three `dropgrad`, `ignore` and `@nograd` for now and see later on which one to keep.

For now adding `ignore` can be useful.

cossio:
Bump?

PR #624, opened 2020-04-28
Fix the docs build

MikeInnes:
Unfortunately it looks like we can't document `copyto!` without breaking Documenter on travis.


MikeInnes:
@dhairyagandhi96 do you know if it's easy to test the docs build as part of the bors run?

dhairyagandhi96:
Yes, we just need to add it as part of the gitlab run

Issue #627, opened 2020-04-30
Zygote errors on loading in Julia 1.0


mohamed82008:
The latest Zygote gives an error upon loading on Julia 1.0.
```julia
ERROR: LoadError: LoadError: UndefVarError: broadcasted not defined
Stacktrace:
 [1] getproperty(::Module, ::Symbol) at .\sysimg.jl:13
 [2] top-level scope at none:0
 [3] include at .\boot.jl:317 [inlined]
 [4] include_relative(::Module, ::String) at .\loading.jl:1044
 [5] include at .\sysimg.jl:29 [inlined]
 [6] include(::String) at C:\Users\user\.julia\dev\Zygote\src\Zygote.jl:1
 [7] top-level scope at none:0
 [8] include at .\boot.jl:317 [inlined]
 [9] include_relative(::Module, ::String) at .\loading.jl:1044
 [10] include(::Module, ::String) at .\sysimg.jl:29
 [11] top-level scope at none:2
 [12] eval at .\boot.jl:319 [inlined]
 [13] eval(::Expr) at .\client.jl:393
 [14] top-level scope at .\none:3
in expression starting at C:\Users\user\.julia\dev\Zygote\src\compiler\interface.jl:82
in expression starting at C:\Users\user\.julia\dev\Zygote\src\Zygote.jl:20
ERROR: Failed to precompile Zygote [e88e6eb3-aa80-5325-afca-941959d7151f] to C:\Users\user\.julia\compiled\v1.0\Zygote\4kbLI.ji.
Stacktrace:
 [1] error(::String) at .\error.jl:33
 [2] compilecache(::Base.PkgId, ::String) at .\loading.jl:1203
 [3] _require(::Base.PkgId) at .\loading.jl:960
 [4] require(::Base.PkgId) at .\loading.jl:858
 [5] require(::Module, ::Symbol) at .\loading.jl:853
```
This is true for master as well.

PR #629, opened 2020-04-30
CI for julia 1.0

CarloLucibello:
Either we support julia 1.0 or we don't. And we should test on the lowest version we support in order to prevent breakage as in #627


mohamed82008:
I say remove it. If you ignore it for Julia 1.0, then for consistency, we should also ignore it for Julia 1.4. The Manifest.toml is awkward to maintain anyways because it may not support all the Julia versions tested and we can't have a manifest for each version.

dhairyagandhi96:
We use it to keep the environment consistent for development frequently, esp when we have to test with non release version of packages, I think we keep it in. Julia 1.0 support isn't that hard to do, even with the couple of hiccups we see.

We still have to get nnpack support in nnlib for example, which needs corresponding changes and consistent testing through the changes.

CarloLucibello:
> We use it to keep the environment consistent for development frequently, esp when we have to test with non release version of packages, I think we keep it in. Julia 1.0 support isn't that hard to do, even with the couple of hiccups we see.

how do we do it while keeping a Manifest? now we are stuck with SpecialFunctions 0.10 not supporting julia 1.0

andreasnoack:
> ...I think we keep it in. Julia 1.0 support isn't that hard to do, even with the couple of hiccups we see.

I think this doesn't recognize the scope of #627. It's not just that it breaks Zygote. The issue caused all 135 packages that depend on Zygote to break since they couldn't be loaded. So having a Manifest checked in in Zygote is effectively making the 135 downstream packages your 1.0 testing and I don't think that is reasonable. You can temporarily check-in a Manifest on a branch and use that for development but having it on the master branch is just moving a nuisance from the few to the many.

CarloLucibello:
let's see how far we go on julia 1.0 removing the Manifest

CarloLucibello:
Tests are passing, we have to reach a decision. It's a little annoyance we have to lose the Manifest, but clearly we cannot avoid CI if we claim to support julia 1.0. So I suggest we merge this @MikeInnes @dhairyagandhi96

oxinabox:
Getting rid of the manifest is a feature.

dhairyagandhi96:
Do we need to depend on Future? If it's a test only dep, let's just have it in extras

dhairyagandhi96:
Let's also come up with a decent alternative so we can CI test unreleased changes together, rather than relying on local tests. Thoughts on this @oxinabox?

CarloLucibello:
> Do we need to depend on Future? If it's a test only dep, let's just have it in extras

It's needed to support `copy!` in julia 1.0, it slipped from #631

dhairyagandhi96:
Let's add a test that would work on 1.0 because of this?

CarloLucibello:
> Let's add a test that would work on 1.0 because of this, or like mark the one in tests already

test is already there, is the one using copy!, don't think there is need to mark it

oxinabox:
> Let's also come up with a decent alternative so we can CI test unreleased changes together, rather than relying on local tests. Thoughts on this @oxinabox?

You run local tests.
Then you register merge each PR and tag a release.
Each in turn.
Retriggering CI each time on the thing you are to merge next.

Why would you need more than that?
The local tests help you during implementation.
The individual CI tests for each package shows that it works on all platforms.
The per package unit tests each check the whole API that any other package uses.
So there is no need for CI running extra integration tests of unreleased things.

MikeInnes:
> having it on the master branch is just moving a nuisance from the few to the many.

I don't agree that having a manifest checked in is bad per se. Not doing CI on 1.0 is not ideal, yes; but there's more than one way to resolve that.

> Why would you need more than that?

CI can take quite a long time and/or require resources you don't have locally (e.g. the GPU setup we use for bors); `bors try` makes it a lot easier to work asynchronously with a reproducible setup. Zygote can also depend on quite large changes to dependent packages; if you have to roundtrip through PRs and releases of multiple dependencies multiple times, that rapidly becomes a lot more friction than pushing to a couple of branches and seeing what CI says.

It's also nice to support very WIP and draft PRs that touch multiple packages; the API of the dependent might be very experimental and not releasable, but you nonetheless want to make sure it's easy for devs to try out a branch that uses it. And large, experimental, far-reaching changes are exactly where reproducible CI is needed most.

Of course, these things are more of an issue early in development when there's a lot of churn, so it may be less of an issue now for Zygote.

I wonder if bors can help here: for example, we could ignore the checked-in manifest on CI by default, but run `bors try master` or something for PRs in the draft stage.

andreasnoack:
> It's also nice to support very WIP and draft PRs that touch multiple packages; the API of the dependent might be very experimental and not releasable, but you nonetheless want to make sure it's easy for devs to try out a branch that uses it.

In this case, I think the solution is to check in a Manifest in the PR that should be removed before the PR is merged.

aminya:
Using Manifest breaks backward compatibility, and it is a bad practice to add it to the master. You should allow Pkg to generate Manifest. By checking Manifest, you will bypass all of its goodness for different users and Julia versions.

You should instead use `compat` entry for limiting the package bounds.
```
PackageX = "0.4, 1" # min supported to max supported
```

For example, If a dep start to use a Jil package, all the Julia versions older than 1.3 will break.
```
[[IntelOpenMP_jll]]
deps = ["Libdl", "Pkg"]
git-tree-sha1 = "fb8e1c7a5594ba56f9011310790e03b5384998d6"
uuid = "1d5cc7b8-4909-519e-a0f8-d0f5ad9712d0"
version = "2018.0.3+0"
```

oxinabox:
Right, i never considered the fact that CI was testing stuff like GPU that you might not test locally.

PR #635, opened 2020-05-03
Add nograd for nthreads and threadid

mohamed82008:
This PR adds `Threads.nthreads` and `Threads.threadid` to the nograd list.


mohamed82008:
Can I please get a release 🙏

PR #647, opened 2020-05-14
Adds `@nograd <<, >>`

cjdoris:
Or maybe it's better to restrict it to just `<<(::Integer, ::Integer)`?


CarloLucibello:
why `@nograd`?  bit shifting is equivalent to multiplication by `2^n`, so we should define the appropriate adjoint (using the inverse bit shifting)

cjdoris:
Because the inputs and outputs are integers, and so not differentiable. `x<<y` is specifically a bit operation, and shouldn't be treated the same as `x*2^y`.

PR #650, opened 2020-05-18
Make map 3.5-4.8 times faster

oxinabox:
### Without this PR
```julia
julia> @btime (x->last(map(sin, fill(x,100))))'(0.1)
  15.635 μs (769 allocations: 23.53 KiB)
```

### With this PR
```julia
julia> @btime (x->last(map(sin, fill(x,100))))'(0.1)
  3.182 μs (35 allocations: 8.23 KiB)
```

### Without this PR
```julia
julia> @btime (x->last(map(sin, fill(x,100))))'(0.1)
  16.815 μs (769 allocations: 22.88 KiB)
0.9950041652780258
```

### With this PR
```julia
julia> @btime (x->last(map(sin, fill(x,100))))'(0.1)
  4.751 μs (57 allocations: 8.14 KiB)
````

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -158,6 +158,9 @@ function unzip(tuples)
       map(tuple -> tuple[i], tuples)
   end
 end
+function unzip(tuples::AbstractVector{<:Tuple{Any, Any}})
+  map(first, tuples), map(last, tuples)
+end
 function ∇map(cx, f, args...)
   ys_and_backs = map((args...) -> _pullback(cx, f, args...), args...)
   if isempty(ys_and_backs)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/lib/array.jl [hunk 1]
```diff
@@ -158,6 +158,9 @@ function unzip(tuples)
       map(tuple -> tuple[i], tuples)
```
oxinabox, line 158: we could replace the outer `map` here with `ntuple`. Not sure if that would have same overall effect.
Not sure we ever actually hit this map function

oxinabox: Ok tested that on
`@btime (x->last(map(atan, fill(x,100), fill(2x, 100))))'(0.1)`
nothing I can do makes it perform better than it currently does

oxinabox:
Bump
CI being broken on nightly is unrelated to this PR afaict

src/lib/array.jl
```diff
@@ -158,6 +158,9 @@ function unzip(tuples)
...
   end
 end
+function unzip(tuples::AbstractVector{<:Tuple{Any, Any}})
```
oxinabox, line 161:
```suggestion
# Special-case for this particularly commonly called varient that has no loop 
function unzip(tuples::AbstractVector{<:Tuple{Any, Any}})
```

dhairyagandhi96:
Failure to build docs because of probably wants to be run on julia 1.3+

cossio:
Would it make sense to generalize this to:

```julia
@generated function unzip(tuples::AbstractVector{<:NTuple{N}}) where {N}
  Expr(:tuple, (:(getindex.(tuples, $i)) for i ∈ 1:N)...)
end
``

PR #654, opened 2020-05-20
Copy, broadcasted tangents

DhairyaLGandhi:
Helps with getting NNs to be used with forward mode.

cc @MikeInnes

### DIFF
--- a/src/forward/lib.jl
+++ b/src/forward/lib.jl
@@ -37,7 +37,7 @@ zerolike(x::Union{Module,Type}) = x # false -> x
 zerolike(x::Core.Box) = isdefined(x, :contents) ? Core.Box(zerolike(x.contents)) : Core.Box()
 @tangent Core.Box() = Core.Box(), () -> Core.Box()
 @tangent Core.Box(x) = Core.Box(x), ẋ -> Core.Box(x)
-@tangent Base.copy(x) = copy(x), ẋ -> copy(x)
+@tangent Base.copy(x) = copy(x), ẋ -> copy(ẋ)
 
 @tangent __new__(T, s...) =
   __new__(T, s...), (_, ṡ...) -> NamedTuple{fieldnames(T)}(ṡ)
--- a/src/forward/lib.jl
+++ b/src/forward/lib.jl
@@ -7,12 +7,13 @@ zerolike(x::Tuple) = zerolike.(x)
 end
 
 # TODO figure out why this made a test fail
-zerolike(x::Union{Module,Type}) = false
+zerolike(x::Union{Module,Type}) = x # false -> x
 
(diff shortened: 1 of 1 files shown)


src/forward/lib.jl
```diff
@@ -7,12 +7,13 @@ zerolike(x::Tuple) = zerolike.(x)
 end
 
 # TODO figure out why this made a test fail
-zerolike(x::Union{Module,Type}) = false
+zerolike(x::Union{Module,Type}) = x # false -> x
```
MikeInnes, line 10: I don't think the derivative of `x` should ever just be `x`; if you do `foo = getfield(Main, :foo)` then `foo` should have no derivative. So I think it might be better to overload operations on `Module` and `Type` so that they don't throw.

src/forward/lib.jl
```diff
@@ -62,5 +65,15 @@ _pushforward(dargs, ::typeof(getproperty), x, f) =
 @tangent literal_getproperty(t, ::Val{i}) where i =
   getproperty(t, i), (ṫ, _) -> getproperty(ṫ, i)
 
+@tangent literal_getproperty(t::DataType, ::Val{i}) where i =
```
MikeInnes, line 68: Types should never have gradients, so this can just return a zero.

src/forward/lib.jl
```diff
@@ -62,5 +65,15 @@ _pushforward(dargs, ::typeof(getproperty), x, f) =
...
+  getproperty(t, i), (ṫ, _) -> getproperty(t, i)
+
 @tangent literal_getindex(t, ::Val{i}) where i =
   getindex(t, i), (ṫ, _) -> getindex(ṫ, i)
+
+import Base.Broadcast.broadcasted
+@tangent function broadcasted(f, args...)
+  broadcasted(f, args...), (ḟ, dargs...) -> begin
+    broadcasted(f, dargs...)
```
MikeInnes, line 77: This doesn't seem correct at the moment, e.g. for `a .* b`.

src/forward/lib.jl
dhairyagandhi96: Could you give me an idea of what operations we should overload?

src/forward/lib.jl
dhairyagandhi96: Gotcha, that would be indeed right!

src/forward/lib.jl
dhairyagandhi96: Yeah, sorry I will fix this one

src/forward/lib.jl
MikeInnes: Whichever operations fail when the derivative is a zero type. Probable `getfield` at least.

CarloLucibello:
closing as stale

PR #655, opened 2020-05-21
nograd randexp

cossio:
Just like `randn`, we should have a nograd rule for `randexp`.

We should also extend the rule for `rand` to cover different number types and / or random generators.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -1,6 +1,7 @@
 using Random, FillArrays, AbstractFFTs
 using FillArrays: AbstractFill, getindex_value
 using Base.Broadcast: broadcasted, broadcast_shape
+using Random: default_rng
 
 @adjoint (::Type{T})(::UndefInitializer, args...) where T<:Array = T(undef, args...), Δ -> nothing
 
@@ -8,7 +9,7 @@ using Base.Broadcast: broadcasted, broadcast_shape
 @adjoint Array(xs::Array) = Array(xs), ȳ -> (ȳ,)
 
 @nograd size, length, eachindex, axes, Colon(), findfirst, findlast, findall, ones, zeros, one, zero, any, all
-@nograd randn, randexp, randn!, randexp!, Random.default_rng
+@nograd randn, randexp, randn!, randexp!, default_rng
 
 @adjoint Base.rand(rng::AbstractRNG, ::Type{T}, dims...) where {T<:Number} =
   rand(rng, T, dims...), _ -> nothing
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -1,17 +1,17 @@
-using FillArrays, AbstractFFTs
+using Random, FillArrays, AbstractFFTs
 using FillArrays: AbstractFill, getindex_value
 using Base.Broadcast: broadcasted, broadcast_shape
-import Random.randn!
 
 @adjoint (::Type{T})(::UndefInitializer, args...) where T<:Array = T(undef, args...), Δ -> nothing
 
 @adjoint Array(xs::AbstractArray) = Array(xs), ȳ -> (ȳ,)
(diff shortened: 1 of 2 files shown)


src/lib/array.jl
```diff
@@ -1,17 +1,17 @@
...
 
 @adjoint Array(xs::AbstractArray) = Array(xs), ȳ -> (ȳ,)
 @adjoint Array(xs::Array) = Array(xs), ȳ -> (ȳ,)
 
-@nograd size, length, eachindex, axes, Colon(), findfirst, findlast, findall, randn, randn!, ones, zeros, one, zero,
+@nograd size, length, eachindex, axes, Colon(), findfirst, findlast, findall, randn, randexp, randn!, randexp!, ones, zeros, one, zero,
   any, all
 
-@adjoint rand(dims::Integer...) = rand(dims...), _ -> nothing
+@adjoint Base.rand(rng::AbstractRNG, ::Type{T}, dims::Integer...) where {T<:Number} =
```
CarloLucibello, line 13:
```suggestion
@adjoint Base.rand(rng::AbstractRNG, ::Type{T}, dims...) where {T<:Number} =
```
so that we catch also the case of `dims` given as a tuple

CarloLucibello:
can you also add some tests for both `rand` and `randexp`? e.g.
```
gradient(x -> sum(rand(rng, Float32, 1,1)), [1])[1] === nothing
gradient(x -> sum(rand(rng, Float32, (1,1))), [1])[1] === nothing
```

CarloLucibello:
We also have this little problem:
```julia
julia> gradient(x->sum(rand(Random.default_rng())),[1])[1] === nothing
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] isassigned at ./array.jl:229 [inlined]
 [3] (::typeof(∂(isassigned)))(::Nothing) at /home/carlo/.julia/packages/Zygote/YeCEW/src/compiler/interface2.jl:0
 [4] default_rng at /home/carlo/julia/julia-1.4.1/share/julia/stdlib/v1.4/Random/src/RNGs.jl:300 [inlined]
 [5] (::typeof(∂(default_rng)))(::Nothing) at /home/carlo/.julia/packages/Zygote/YeCEW/src/compiler/interface2.jl:0
 [6] default_rng at /home/carlo/julia/julia-1.4.1/share/julia/stdlib/v1.4/Random/src/RNGs.jl:297 [inlined]
 [7] #65 at ./REPL[53]:1 [inlined]
 [8] (::Zygote.var"#36#37"{typeof(∂(#65))})(::Float64) at /home/carlo/.julia/packages/Zygote/YeCEW/src/compiler/interface.jl:46
 [9] gradient(::Function, ::Array{Int64,1}) at /home/carlo/.julia/packages/Zygote/YeCEW/src/compiler/interface.jl:55
 [10] top-level scope at REPL[53]:1
 [11] eval(::Module, ::Any) at ./boot.jl:331
 [12] eval_user_input(::Any, ::REPL.REPLBackend) at /home/carlo/julia/julia-1.4.1/share/julia/stdlib/v1.4/REPL/src/REPL.jl:86
 [13] run_backend(::REPL.REPLBackend) at /home/carlo/.julia/packages/Revise/AMRie/src/Revise.jl:1023
 [14] top-level scope at none:0
```
I think it's fine to `@nograd Random.default_rng`  as well

src/lib/array.jl
cossio: Yep, thanks

cossio:
Thanks, I added some tests.

cossio:
Hmm tests pass locally but something fails on the CI.

cossio:
Oh, `default_rng` was not available in Julia 1.0 (https://github.com/JuliaLang/julia/commit/c49ba0d487b2f3c276c7b9599e8875ed7a888b03)

cossio:
@CarloLucibello Looks good to go!

PR #656, opened 2020-05-21
Make map with more than one argument ~2x faster

cossio:
Generalize https://github.com/FluxML/Zygote.jl/pull/650 to more than one argument.

Before this PR:

```
julia> @btime (x->last(map(+, fill(2x,100), fill(x,100))))'(0.1);
  12.993 μs (580 allocations: 25.70 KiB)
```

After this PR:

```
julia> @btime (x->last(map(+, fill(2x,100), fill(x,100))))'(0.1);
  6.746 μs (73 allocations: 12.80 KiB)
```

CC: @oxinabox

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -160,15 +160,16 @@ end
 
 @adjoint getindex(i::Int, j::Int) = i[j], _ -> nothing
 
+struct StaticGetter{i} end
+(::StaticGetter{i})(v) where {i} = v[i]
+@generated function _unzip(tuples, ::Val{N}) where {N}
+  Expr(:tuple, (:(map($(StaticGetter{i}()), tuples)) for i ∈ 1:N)...)
+end
 function unzip(tuples)
-  map(1:length(first(tuples))) do i
-      map(tuple -> tuple[i], tuples)
-  end
+  N = length(first(tuples))
+  _unzip(tuples, Val(N))
 end
-# Special-case for this particularly commonly called varient that has outer loop unrolled 
-function unzip(tuples::AbstractVector{<:Tuple{Any, Any}})
-  map(first, tuples), map(last, tuples)
-end
+
 function ∇map(cx, f, args...)
   ys_and_backs = map((args...) -> _pullback(cx, f, args...), args...)
   if isempty(ys_and_backs)


oxinabox:
I tried out using `ntuple` which should have compile to the same code as you post but it didn't seem to improve things.
But you are welcome (encouraged even) to run own experiments on that.
I could ahve screwed it up, it it might be that ntuple doesn't workout for this.

cossio:
In my experience manually unrolling the tuple (using `@generated`, like here) is faster than `ntuple`. Sometimes `ntuple` is not even inferrable.

oxinabox:
Can you also test 
```julia
@btime (x->last(map(sin, fill(x,100))))'(0.1)
```
to make sure we haven't regressed somehow?

src/lib/array.jl
```diff
@@ -160,15 +160,16 @@ end
...
 
-function unzip(tuples)
-  map(1:length(first(tuples))) do i
-      map(tuple -> tuple[i], tuples)
-  end
+struct StaticGetter{i} end
+(::StaticGetter{i})(v) where {i} = v[i]
+@generated function _unzip(tuples, ::Val{N}) where {N}
+  Expr(:tuple, (:(map($(StaticGetter{i}()), tuples)) for i ∈ 1:N)...)
 end
```
oxinabox, line 167: How about we get rid of the functor and just put the getindex into the source.
Does that end up with the same performance?

```suggestion
@generated function _unzip(tuples, ::Val{N}) where {N}
  args = ntuple(N) do ii
    :(map(x->x[$ii], tuples))
  end
  Expr(:tuple, args...)
end
```

cossio:
@oxinabox To make sure we don't regress from https://github.com/FluxML/Zygote.jl/pull/650, here I get:

```julia
julia> @btime (x->last(map(sin, fill(x,100))))'(0.1);
  4.814 μs (57 allocations: 8.14 KiB)
```

while with https://github.com/FluxML/Zygote.jl/pull/650 (that is, on current master):

```
julia> @btime (x->last(map(sin, fill(x,100))))'(0.1);
  5.518 μs (57 allocations: 8.14 KiB)
```

There is no discernible difference.

(I am on Julia 1.4.1)

src/lib/array.jl
cossio: Did you try running that? You can't use closures inside a generated function (so can't do `x->x[$i]`). I did try some variations of this and always ended up making things worse. 

For example, the closest thing to your suggestion is:

```julia
@generated function _unzip(tuples, ::Val{N}) where {N}
  args = ntuple(N) do i
    :(map($(x -> x[i]), tuples))
  end
  Expr(:tuple, args...)
end

julia> @btime (x->last(map(+, fill(2x,100), fill(x,100))))'(0.1);
  31.180 μs (974 allocations: 36.31 KiB)
```

oxinabox: > Did you try running that?

No, it was quite late for me by then.

> You can't use closures inside a generated function

I always forget this (even though it is a blocking issue for Arborist.jl).
I alway think it is that you can't call them in the body of the generated function, but can return them.

oxinabox:
I can't do any better.
I tried to use `Base.Fix2(getindex, i)` but did not perform same as the function.

LGTM

cossio:
Thanks!

cossio:
Just for future reference, there should be a `Base.unzip` function at some point, https://github.com/JuliaLang/julia/pull/33515. So when that gets merged we should probably use it.

cossio:
@oxinabox Any chance we tag a release with this? Thanks!

Issue #636, opened 2020-05-05
xlogx adjoint at zero should be -Inf


cossio:
https://github.com/FluxML/Zygote.jl/blob/9c03dd06a82a06eca99663e2b4a8c13b2199c7cf/src/lib/statsfuns.jl#L7

The derivative of `xlog(x) = x * log(x)` at `x=0` should be `-Inf`, not zero.

cossio:
This might be a bit subtle.

Mathematically the derivative of xlogx(x) at x=0 is -Inf. But if you differentiate the branch as if it were a piecewise function, then you get xlogx'(0)=0. What's the correct answer here?

oxinabox:
Transfering from slack discussion:


@jiahao:
> Depends on the direction of the derivative? from x->0+ it's -Inf; from x->0- it's 0, no?

> Another data point: the real part of the complex function Real( d(z log (z))/dz ) at z=0  is -Inf. So that might be the more "natural" derivative (edited) 

@sethaxen 
> Yeah, I second that it should be -Inf.

@oxinabox
> a claim infavor of 0.
For any point nondifferentaiblility:
Take the subgradient, i.e the set of tangents from.all approaches.
The only thing people care about from the subgradient is "does this set include zero", this is it a potential minima to be checked.
Thus if the subgradient does include zero you should claim the gradient is zero.

sethaxen:
Another example is the unnormalized `sinc` function. The naive way to implement this is:

```julia
usinc(x) = x == 0 ? one(x) : sinc(x) / x
```

This is fine for first derivatives, because the first derivative of `usinc(x)` at `x==0` is 0, so you get the right first derivative. But if you want 2nd derivatives, then I would say it's broken, because it will give 0, when the derivative goes to -1/3 as `x` goes to 0. So if I wanted to support AD up to 2nd order, I would usually implement this as:

```julia
usinc(x) = x == 0 ? 1 - x^2 / 6 : sinc(x) / x
```

where I've inserted the power series expansion around `x==0`. The two codes give the exact same output but have different derivatives with AD. Which is "correct"? What does "correct" mean here?

mcabbott:
Besides the choice of what derivative to provide at `x==0`, for which zero seems fine, shouldn't the gradient of `f(x)` be of consistent type? 

E.g. `g(x) = x<0 ? zero(x) : x^2` behaves well under `@code_warntype g(1f0)`, but `g'(-1f0) === nothing`. 

[...]

PR #657, opened 2020-05-22
adjoints for StatsFuns

cossio:
Add some adjoints for functions from [StatsFuns.jl](https://github.com/JuliaStats/StatsFuns.jl). We need these rules because some of these functions use `ifelse` or `max/min` statements which make sense mathematically but confuse Zygote.

For example, this line:

https://github.com/JuliaStats/StatsFuns.jl/blob/069cbb0924640c466a58a331d4513a5f6bdb5a8b/src/basicfuns.jl#L211

results in a wrong gradient computed by Zygote:

```
julia> gradient(logaddexp, 5, 5)  # should be (0.5, 0.5)
(0.0, 1.0)
```

This and similar issues are fixed in this PR for the functions addressed here.

I also added rules for broadcasting of these functions, since Zygote is slow differentiating broadcasts (until something like https://github.com/FluxML/Zygote.jl/pull/338 gets merged). In the meantime we need these broadcasting rules for the functions we want to broadcast and differentiate fast. For example, take:

```julia
julia> using StatsFuns, Zygote, BenchmarkTools
julia> f(x) = sum(xlogx.(x))
julia> x = rand(100);
```

Before this PR:

```julia
julia> @btime gradient(f, $x);
  17.104 μs (342 allocations: 12.33 KiB)
```

After this PR:

```julia
julia> @btime gradient(f, $x)
  1.029 μs (7 allocations: 4.42 KiB)
```

is **17x faster**.

The functions addressed in this PR are: `xlogx`, `xlogy`, `logaddexp`, `logsubexp`, `log1pexp`.

### DIFF
--- a/src/lib/statsfuns.jl
+++ b/src/lib/statsfuns.jl
@@ -8,13 +8,11 @@ using Base.Broadcast: broadcasted
     back(δ) = (dx * δ,)
     return result, back
 end
-
 @adjoint function broadcasted(::typeof(xlogx), x::Numeric)
     result, dx = ∇xlogx(x)
     back(δ) = (nothing, unbroadcast(x, δ .* dx))
     return result, back
 end
-
 function ∇xlogx(x::Numeric)
     logx = log.(x)
     xlogx = x .* logx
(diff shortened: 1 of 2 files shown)


dhairyagandhi96:
I would rather get a broadcasting kernel be merged so we can get a better sense of benefits of writing one specific to a particular function like in here.

dhairyagandhi96:
Also, we should identify the issue with incorrect gradients from `ifelse` and fix that, since that would catch a lot more of the code we expect to differentiate

cossio:
> I would rather get a broadcasting kernel be merged so we can get a better sense of benefits of writing one specific to a particular function like in here.

What do you mean? There is already some generic broadcasting machinery in place, https://github.com/FluxML/Zygote.jl/blob/master/src/lib/broadcast.jl. But it is slow. So for some functions it is faster to have the explicit rule, e.g. 

https://github.com/FluxML/Zygote.jl/blob/3e0a904cae5565bc7b458ad6d575c531ce70dd96/src/lib/nnlib.jl#L6.

https://github.com/FluxML/Zygote.jl/blob/a2f7c9731279b8ad80322c35b3a0325f03392c0e/src/lib/broadcast.jl#L94

I agree a more generic solution would be better. In the meantime it is a pain having to write all of these by hand (automatic differentiation doesn't feel so "automatic" if you have to do this!).

> Also, we should identify the issue with incorrect gradients from `ifelse` and fix that, since that would benefit a lot more of the code we expect to differentiate

I think `ifelse` gradients work fine, except for https://github.com/FluxML/Zygote.jl/issues/636 (which is an issue with all conditionals, not just `ifelse`). In the case of `logaddexp`, we have the singular check `x == y`. I don't see how that can be fixed in general, though.

Anyway, for now, maybe a better approach is to have a separate package `StatsFunsAD` with these rules?

dhairyagandhi96:
> There is already some generic broadcasting machinery in place, 

Just to clarify, I mean we should quantify the difference between performance of broadcasted kernels written by hand and on top of #338. If the ad overhead can be mitigated (or minimal), we probably wouldn't need the hand coded rule. I am fine with the hand coded rule here, but would want to know the performance difference with a fix to the broadcasted ad to know the overhead.

dhairyagandhi96:
Cc @MikeInnes

cossio:
I haven't tested, but just quoting the OP from https://github.com/FluxML/Zygote.jl/pull/338, 

> Also, there seems to be no speed improvement (yet). 

It seems https://github.com/FluxML/Zygote.jl/pull/338 still requires some work. And probably it take some rewriting after https://github.com/FluxML/Zygote.jl/pull/503. I'd say we don't block having these rules, since these more general issues might still take some time? For now this is a net improvement.

dhairyagandhi96:
bors r+

thanks! We should be careful to let performance sensitive stuff be brought up.

cossio:
Great, thanks! Any chance we could tag a release?

Issue #537, opened 2020-03-07
add `@nograd` block


CarloLucibello:
As suggested in #465,  it can be of some use the possibility to add blocks of code where gradients are not computed (detached) and mutations are allowed. The proposed syntax is
```julia
@nograd begin
    ...
end
```

MikeInnes:
I still think we should just call this `Zygote.ignore` (or `Zygote.@ignore`; we can have both). It should be pretty clear then that you're just asking Zygote to ignore that block of code.

Consistency with `@nograd` isn't too important as we want to get rid of it eventually anyway.

PR #659, opened 2020-05-26
ignore macro

cossio:
A macro version of https://github.com/FluxML/Zygote.jl/pull/623. With this PR,

```julia
ignore() do
    ...
end
```

can also be written as:

```julia
@ignore ...
```

fix https://github.com/FluxML/Zygote.jl/issues/537

CC: @MikeInnes @CarloLucibello 

I bumped the version here so if we merge we can tag a release.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.4.20"
+version = "0.4.21"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.4.21"
+version = "0.4.20"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
--- a/src/lib/utils.jl
+++ b/src/lib/utils.jl
@@ -26,6 +26,17 @@ ignore(f) = f()
 @adjoint ignore(f) = ignore(f), _ -> nothing
 
 """
+    @ignore (...)
+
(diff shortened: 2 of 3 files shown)


CarloLucibello:
LGTM

cossio:
Merge? Also maybe tag a release

src/lib/utils.jl
```diff
@@ -25,6 +25,23 @@ Obviously, you run the risk of incorrect gradients if you use this incorrectly.
...
 @adjoint ignore(f) = ignore(f), _ -> nothing
 
+"""
+    @ignore (...)
+
+Tell Zygote to ignore an expression. Equivalent to `ignore() do (...) end`.
+Example:
+
+```julia-repl	
+julia> f(x) = (y = Zygote.@ignore x; x * y); f'(1)
```
MikeInnes, line 35: Would be great to have a slightly more realistic example here – and it's allowed to use multiple lines.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.4.20"
+version = "0.4.21"
```
MikeInnes, line 3: Let's keep release management separate from this PR.

MikeInnes:
Good to merge with the changes above.

dhairyagandhi96:
Also could we export it?

cossio:
@dhairyagandhi96 I'm not sure about exporting, since we haven't converged on how to name this things (`@nograd`, `ignore`, `@ignore`, `@dropgrad`)

dhairyagandhi96:
Fair enough. I was just going to comment saying on second thoughts it's a bit too aribtrary to export, so I concur.

src/lib/utils.jl
cossio: I think this gets it accros but if you suggest anything I'll add it. I did fix the newline.

MikeInnes:
Yeah, I think it's important that this is clearly associated with `Zygote` in code (either `Zygote.@ignore` or `using Zygote: @ignore`) so that it's clear what it means.

MikeInnes:
Thanks @cossio!

PR #660, opened 2020-05-28
fill can't handle nothing as a recieved gradient

dsweber2:
I was using fill as a way of indexing into another matrix, e.g. `x[fill(1,N)...]`, and this causes `(nothing,)` to be handed to the pullback of `fill`. As all the gradient does is sum at the moment, this causes errors trying to sum `nothing`. I address this by checking the type, and passing on any received `nothing`, and if the entries are mixed it ignores any `nothing`s in the sum.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -75,7 +75,11 @@ end
 
 @adjoint collect(x::Array) = collect(x), Δ -> (Δ,)
 
-@adjoint fill(x::Real, dims...) = fill(x, dims...), Δ->(sum(Δ), map(_->nothing, dims)...)
+@adjoint function fill(x::Real, dims...) 
+    back(Δ) =(sum(x->((x==nothing) ? 0 : x), Δ), map(_->nothing, dims)...)
+    back(Δ::NTuple{N, <:Nothing}) where N = (nothing, map(_->nothing, dims)...)
+    return fill(x, dims...), back
+end
 
 @adjoint function circshift(A, shifts)
   circshift(A, shifts), Δ -> (circshift(Δ, map(-, shifts)), nothing)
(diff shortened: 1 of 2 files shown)


MikeInnes:
Why would `fill` get a tuple of nothings as its gradient? That seems fishy and probably points to a different issue. 

`nothing` (not a tuple) is a valid gradient, but the `@adjoint` macro should already be handling that.

dsweber2:
here's a minimal example of what I was doing that caused the problem.
```
function example(x,N)
    ax = axes(x)
    extraAxe = ax[2+N:end]
    filledLoc = fill(1, N)
    return x[:, filledLoc..., extraAxe...]
end
y, back = pullback(example, randn(5,3,4,3), 2)
back(y)
```
which attempts to sum `(nothing, nothing)`. Not sure where the code for it is, but it does make sense that the adjoint of splatting could result in multiple `nothing`s grouped together.

MikeInnes:
Ok, thanks for the clarification, I see what's happening. Unfortunately this is two things conspiring, the `nothing` gradient from indexing and the tuple-ing of those `nothing`s from the splat (so easy to forget that iteration means you have to support tuples as a gradient). I think you already understood that but just so it's spelt out.

The best way to resolve this is probably to use `accum_sum` which handles nothings; I'll think about it bit I suspect we can't avoid passing the tuple of nothings here.

As an aside, this is quite a strange use of `Fill`; it would be more typical to use `ntuple` here since the compiled code will get more benefit from knowing the size of `filledLoc` rather than its particular value. This is not to say we shouldn't support it, of course.

dsweber2:
> it would be more typical to use `ntuple`

Good point, I should get more in the habit of using tuples. Thanks for looking into it.

CarloLucibello:
closing as stale

Issue #663, opened 2020-05-28
pullback of norm causes forward pass to be type unstable


dsweber2:
Minimal example
```
y, back = pullback(norm, randn(Float32, 10))
typeof(y)
typeof(back(y))
```
as expected, `back(y)` also has an eltype of `Float64`. I'm also not sure why the pullback has length 2, with the second entry a single `Complex{float64}`. Perhaps something strange is happening with `p`? Not sure how that would end up as a complex number though.

sethaxen:
This is not unique to `norm` but rather to any function that computes the pullback of an exponent of a potentially negative number. A minimal example:

```julia
julia> y, back = Zygote.pullback(^, -2.0, 2.0);

julia> y
4.0

julia> back(1.0)
(-4.0, 2.772588722239781 - 12.566370614359172im)
```

See also ChainRules' definition of this rule: https://github.com/JuliaDiff/ChainRules.jl/blob/master/src/rulesets/Base/fastmath_able.jl#L49-L52

The issue is that only for integer powers is the power of a negative real well-defined. Otherwise you must make it complex first:

```julia
julia> (-2.0)^2.1
ERROR: DomainError with -2.0:
Exponentiation yielding a complex result requires a complex argument.
Replace x^y with (x+0im)^y, Complex(x)^y, or similar.
Stacktrace:
 [1] throw_exp_domainerror(::Float64) at ./math.jl:37
 [2] ^(::Float64, ::Float64) at ./math.jl:872
 [3] top-level scope at REPL[30]:1
```

Probably the only type-stable thing to do here is to add a rule for integer exponents to avoid complexifying the pullback of the exponent. But if the exponent must be exactly an integer for this, then strictly speaking, its pullback should be 0 or `nothing`. This is what we do for exponents of `Symmetric`/`Hermitian` matrices.

Related proposal for `norm`: https://github.com/JuliaDiff/ChainRules.jl/issues/204

dsweber2:
the `Float32` -> `Float64` also appears to happen for square rooting via `(x).^(1/2)`

sethaxen:
> Perhaps I should file this on ChainRules.jl? Briefly looking at [the norm rrule in the LinearAlgebra rulesets](https://github.com/JuliaDiff/ChainRules.jl/blob/3a969a2319a07f21ce29334659ef5609684562ce/src/rulesets/LinearAlgebra/dense.jl#L173) it's not clear to me why it is switching a Float64 to a Float32.

[...]

PR #666, opened 2020-05-30
Make norm type-stable with Float32 arguments

sethaxen:
The current adjoint for `norm` promotes `Float32` inputs to `Float64`. This PR makes sure that doesn't happen. I also couldn't find a test for `norm` with array arguments, so I added one.

This PR fixes https://github.com/FluxML/Zygote.jl/issues/663

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -424,7 +424,7 @@ end
 end
 
 function _pullback(cx::AContext, ::typeof(norm), x::AbstractArray, p::Real = 2)
-  fallback = (x, p) -> sum(abs.(x).^p .+ eps(0f0))^(1/p) # avoid d(sqrt(x))/dx == Inf at 0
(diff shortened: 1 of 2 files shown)


CarloLucibello:
@sethaxen could you avoid using `only` in tests, since it's not present in julia 1.0?

src/lib/array.jl
```diff
@@ -424,7 +424,7 @@ end
 end
 
 function _pullback(cx::AContext, ::typeof(norm), x::AbstractArray, p::Real = 2)
-  fallback = (x, p) -> sum(abs.(x).^p .+ eps(0f0))^(1/p) # avoid d(sqrt(x))/dx == Inf at 0
+  fallback = (x, p) -> sum(abs.(x).^p .+ eps(0f0))^(1f0 / p) # avoid d(sqrt(x))/dx == Inf at 0
```
dhairyagandhi96, line 427: Maybe use `one(T)`?

Issue #664, opened 2020-05-29
keyword argument splatting can cause gc_preserve_end issues?


ChrisRackauckas:
```julia
using Zygote
f(args...;sensealg=nothing,kwargs...) = g(sensealg,args...;kwargs...)
g(args...;x=1,save_idxs=Colon(),kwargs...) = x[save_idxs]
Zygote.gradient(x->sum(f(;x=x,save_idxs=1:1)),ones(2))
```

```
Can't differentiate gc_preserve_end expression
error(::String) at error.jl:33
getindex at essentials.jl:591 [inlined]
(::typeof(∂(getindex)))(::Nothing) at interface2.jl:0
typejoin at promotion.jl:103 [inlined]
(::typeof(∂(typejoin)))(::Nothing) at interface2.jl:0
_promote_typejoin at promotion.jl:131 [inlined]
promote_typejoin at promotion.jl:130 [inlined]
(::typeof(∂(promote_typejoin)))(::Nothing) at interface2.jl:0
_compute_eltype at tuple.jl:117 [inlined]
(::typeof(∂(_compute_eltype)))(::Nothing) at interface2.jl:0
eltype at tuple.jl:110 [inlined]
eltype at namedtuple.jl:145 [inlined]
Pairs at iterators.jl:169 [inlined]
pairs at iterators.jl:226 [inlined]
(::typeof(∂(pairs)))(::NamedTuple{(:data, :itr),Tuple{NamedTuple{(:x, :save_idxs),Tuple{Array{Float64,1},Nothing}},Nothing}}) at interface2.jl:0
f at test.jl:2 [inlined]
(::typeof(∂(f##kw)))(::FillArrays.Fill{Float64,1,Tuple{Base.OneTo{Int64}}}) at interface2.jl:0
#78 at test.jl:4 [inlined]
(::typeof(∂(#78)))(::Float64) at interface2.jl:0
(::Zygote.var"#36#37"{typeof(∂(#78))})(::Float64) at interface.jl:46
gradient(::Function, ::Array{Float64,1}, ::Vararg{Array{Float64,1},N} where N) at interface.jl:55
top-level scope at test.jl:4
```

Interestingly, the `save_idxs=1:1` thing seems to be required in order for this to trigger.

dhairyagandhi96:
Interesting, is that a new occurrence or have we seen something like this before?

[...]

Issue #584, opened 2020-04-06
Cant differentiate gc_preserve_end due to keyword args


baggepinnen:
I get an error in this function
```julia
function evaluate(d::AbstractRootDistance,w1::AbstractModel,w2::AbstractModel; kwargs...)
    evaluate(d, preprocess_roots(d,w1), preprocess_roots(d,w2); kwargs...)
end
```
that goes away if I remove the `kwargs...`
The error only appears if some keyword arguments are sent to the functoin, if none are provided it's fine. 

The error is
```
Can't differentiate gc_preserve_end expression
error(::String) at error.jl:33
getindex at essentials.jl:591 [inlined]
(::typeof(∂(getindex)))(::Nothing) at interface2.jl:0
iterate at essentials.jl:603 [inlined]
(::typeof(∂(iterate)))(::Nothing) at interface2.jl:0
_compute_eltype at tuple.jl:117 [inlined]
(::typeof(∂(_compute_eltype)))(::Nothing) at interface2.jl:0
eltype at tuple.jl:110 [inlined]
eltype at namedtuple.jl:145 [inlined]
Pairs at iterators.jl:169 [inlined]
pairs at iterators.jl:226 [inlined]
(::typeof(∂(pairs)))(::Nothing) at interface2.jl:0
evaluate at losses.jl:356 [inlined]
(::typeof(∂(evaluate##kw)))(::Float64) at interface2.jl:0
#evaluate#68 at losses.jl:478 [inlined]
(::typeof(∂(#evaluate#68)))(::Float64) at interface2.jl:0
evaluate at losses.jl:476 [inlined]
(::typeof(∂(evaluate##kw)))(::Float64) at interface2.jl:0
#19 at test_diff.jl:255 [inlined]
(::typeof(∂(#19)))(::Float64) at interface2.jl:0
(::Zygote.var"#36#37"{typeof(∂(#19))})(::Float64) at interface.jl:36
gradient(::Function, ::Array{Float64,1}) at interface.jl:45
top-level scope at util.jl:175
```

gszep:
I am also encountering this issue

dhairyagandhi96:
Could we please check master, and bump? Thanks @baggepinnen !

baggepinnen:
Does not appear to be a problem any longer!

PR #669, opened 2020-05-30
Pairs adjoint

dhairyagandhi96:
Related to #664, #584

cc @MikeInnes @ChrisRackauckas

### DIFF
--- a/src/lib/base.jl
+++ b/src/lib/base.jl
@@ -49,6 +49,9 @@ end
 
 @nograd haskey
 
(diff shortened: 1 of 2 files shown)


dhairyagandhi96:
There is an error on testing with Julia 1.0 pointing at chain rules, which seems to be unrelated to this PR. @oxinabox  could you take a peek at it? I'm assuming we should try and define more `rrule`s in general, but this seems like it should be handled too. Either way, I am going to here it for now

bors r+

PR #673, opened 2020-06-03
CuArrays -> CUDA

dhairyagandhi96:


dhairyagandhi96:
bors try

src/lib/broadcast.jl
```diff
@@ -186,27 +186,30 @@ end
   return y, back
 end
 
-@init @require CuArrays="3a865a2d-5b23-5a0f-bc46-62713ec82fae" begin
-  if isdefined(CuArrays, :CuArrayStyle)  # Introduced in CuArrays v2.0
-    CuArrayStyle = CuArrays.CuArrayStyle
+# For CUDA integeration
+@nograd Core.Intrinsics.checked_sdiv_int
```
dhairyagandhi96, line 190: Seems to be wrong. Will probably need an adjoint here.

CarloLucibello:
there is still a `using CUDAapi` in `tests/`

dhairyagandhi96:
Done 👍

dhairyagandhi96:
cool, only need to get around the docs now.

CarloLucibello:
CUDA is not available for julia <1.3, so if this gets merged as it is people on <1.3 won't get cuda support even if `using CuArrays`. I see 2 solutions:
- we drop < 1.3 support
- we keep both a CuArrays and CUDA require. It could be easy to keep testing 1.0 though. Maybe
something runtests like 
```julia
if VERSION >= v"1.3"
     Pkg.add("CUDA")
     using CUDA
else
       Pkg.add("CuArrays")
      using CuArrays
end
```

dhairyagandhi96:
CUDA supports 1.3 on master, and will in the next release.

CarloLucibello:
> We might have to bump Julia dependency to 1.3 for a clean solution

yes I would do that

dhairyagandhi96:
bors try

CarloLucibello:
bump on this. Needs a julia > 1.3 compat

dhairyagandhi96:
I was wondering if we should wait merging until cuda.jl 1.0 is tagged?

CarloLucibello:
CUDA 1.0 is out. We just need some compat fixes dropping julia 1.3 in project and CI here

src/lib/broadcast.jl
```diff
@@ -186,27 +186,27 @@ end
...
-@init @require CuArrays="3a865a2d-5b23-5a0f-bc46-62713ec82fae" begin
-  if isdefined(CuArrays, :CuArrayStyle)  # Introduced in CuArrays v2.0
-    CuArrayStyle = CuArrays.CuArrayStyle
+@init @require CUDA="052768ef-5323-5732-b1bb-66c8b64840ba" begin
+  if isdefined(CUDA, :CuArrayStyle)
+    CuArrayStyle = CUDA.CuArrayStyle
   else
-    CuArrayStyle = Broadcast.ArrayStyle{CuArrays.CuArray}
+    CuArrayStyle = Broadcast.ArrayStyle{CUDA.CuArray}
   end
```
CarloLucibello, line 194: we can remove this branch

PR #674, opened 2020-06-05
Define adjoint for softplus

AStupidBear:
fix https://github.com/FluxML/Flux.jl/issues/876

### DIFF
--- a/src/lib/nnlib.jl
+++ b/src/lib/nnlib.jl
@@ -10,6 +10,11 @@ end
     return y, Δ -> (Δ * y * (1 - y),)
 end
 
+@adjoint function softplus(x::Real)
+    y = softplus(x)
+    return y, Δ -> (Δ * σ(*),)
+end
+
 @adjoint softmax(xs; dims=1) = softmax(xs, dims=dims), Δ -> (∇softmax(Δ, xs, dims=dims),)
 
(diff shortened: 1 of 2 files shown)


src/lib/nnlib.jl
```diff
@@ -10,6 +10,11 @@ end
     return y, Δ -> (Δ * y * (1 - y),)
 end
 
+@adjoint function softplus(x::Real)
+    y = softplus(x)
+    return y, Δ -> (Δ * σ(*),)
```
dhairyagandhi96, line 15: This doesn't look right. Maybe it should be 
```suggestion
    return y, Δ -> (Δ * σ(x),)
```

AStupidBear: Sorry, a stupid typo. Now fixed.

CarloLucibello:
tests are failing due to some tiny numerical error here, @AStupidBear can we fix that or make tests more robust?

devmotion:
Couldn't/shouldn't the custom adjoint be defined in NNlib using ChainRulesCore such that other AD frameworks could use it as well?

dhairyagandhi96:
We can move over the adjoints to frules but that would need to be done for a lot more than functions than just this. We can have this in now and move the adjoints as a different issue if that works

devmotion:
Sure, that could probably done for other definitions as well. I was wondering mainly because this PR adds a new definition, so one could avoid having to move it around later by adding it to NNlib right away.

CarloLucibello:
this should be filed to NNlib.jl

PR #676, opened 2020-06-05
Stateful map

MikeInnes:
See https://github.com/FluxML/Flux.jl/issues/1209.

It would be good if we could come up with a reasonable test case. It may be tricky to do that, and if so it may make sense to merge this to avoid the issues people are seeing.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -171,9 +171,14 @@ function unzip(tuples)
   _unzip(tuples, Val(N))
 end
 
-_tryreverse(backs, Δ) = backs, Δ
-_tryreverse(backs, Δ::AbstractVector) = reverse(backs), reverse(Δ)
-
+# Reverse iteration order when ∇map is applied to vector,
+# needed for stateful functions. 
+# See https://github.com/FluxML/Flux.jl/issues/1209
+# Should be generalized to abstract array, but reverse takes a dims keyword there
+_tryreverse(m, backs, Δ) = backs, Δ
+_tryreverse(m::typeof(map), backs, Δ::AbstractVector) = reverse(backs), reverse(Δ)
+_tryreverse(m, x) = x
+_tryreverse(m::typeof(map), x::AbstractVector) = reverse(x)
 
 for (mapfunc,∇mapfunc) in [(:map,:∇map),(:pmap,:∇pmap),(:vmap,:∇vmap)]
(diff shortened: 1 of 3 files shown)


CarloLucibello:
should we merge this? the issue is sneaky and comes up quite often

MikeInnes:
Yeah, we probably should just merge it and tag a patch.

dhairyagandhi96:
bors r-

We should add tests for this case too

CarloLucibello:
`reverse` has a mandatory `dims` keyword for multi-dimensional array. 
The easiest fix seems just to reverse vectors,  and leave higher-order arrays non-reversed

AzamatB:
Bump, [Flux.jl#1209](https://github.com/FluxML/Flux.jl/issues/1209) is a really nasty and silent bug, can we please merge this fix?

src/lib/broadcast.jl
```diff
@@ -154,6 +154,15 @@ end
   end
 end
 
+# # Use the `map` adjoint in this special case, which is the same but applies
+# # pullbacks in reverse order.
+# # This leaves regular `broadcast` technically incorrect when the broadcasted
+# # function is stateful.
+# # Look, I'm not proud of it, but this is extremely rare in practice.
```
DhairyaLGandhi, line 161: Double comments

src/lib/array.jl
```diff
@@ -179,8 +188,9 @@ for (mapfunc,∇mapfunc) in [(:map,:∇map),(:pmap,:∇pmap),(:vmap,:∇vmap)]
     else
       ys, backs = unzip(ys_and_backs)
       ys, function (Δ)
-        Δf_and_args_zipped = $mapfunc((f, δ) -> f(δ), backs, Δ)
-        Δf_and_args = unzip(Δf_and_args_zipped)
+        # Apply pullbacks in reverse order. Needed for correctness if `f` is stateful.
+        Δf_and_args_zipped = $mapfunc((f, δ) -> f(δ), _tryreverse($mapfunc, backs, Δ)...) 
+        Δf_and_args = unzip(_tryreverse($mapfunc, Δf_and_args_zipped))
```
DhairyaLGandhi, line 193: Any particular reason for the refactoring? Might be better to split out the adjoint for pmap etc to just do the sync and pmap in the pullback.

PR #688, opened 2020-06-15
Respect linearity when accumulating implicit params

MikeInnes:
All gradients need to be linear, i.e. consumed only once. Previously we allowed gradients to be consumed by the `accum_param` (tracking gradients of parameters given to the `Params` object) _and then_ continue backpropagating. Backpropagated gradients were always correct, but `accum_param` might be triggered more than once on the same object, leading to multiplied gradients. We now backpropagate `nothing` for a value instead.

### DIFF
--- a/src/lib/lib.jl
+++ b/src/lib/lib.jl
@@ -46,10 +46,14 @@ end
 # backpropagated as values; these should be mutually exclusive options.
 
 @generated function accum_param(cx::Context, x, Δ)
-  isbitstype(x) && return
+  isbitstype(x) && return :(Δ)
   quote
-    haskey(cache(cx), x) && (cache(cx)[x] = accum(cache(cx)[x],Δ))
-    return
+    if haskey(cache(cx), x)
+      cache(cx)[x] = accum(cache(cx)[x],Δ)
+      return
+    else
+      return Δ
+    end
   end
 end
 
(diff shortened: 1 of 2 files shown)


src/lib/lib.jl
```diff
@@ -46,10 +46,14 @@ end
 # backpropagated as values; these should be mutually exclusive options.
```
oxinabox, line 46: Can this TODO be removed now?

MikeInnes: Huh, guess I already thought of this issue at some point :D Yeah I think this is the fix though.

cossio:
Can we merge this before the next tagged release? I think this is an important (silent!) bug. 
This probably also fixes https://github.com/FluxML/Zygote.jl/issues/691 (though I haven't tried yet).

CarloLucibello:
tests are failing due to a return type inference problem. Can we fix this @MikeInnes ? Or just remove the test?

MikeInnes:
Our type inference tests are pretty minimal and we should try to preserve them, since a performance issue with something like tuple indexing can easily become a big problem. I'm not sure why Julia doesn't figure it out in this case, but luckily it's easy to rewrite it in a way that can be inferred.

bors r+

CarloLucibello:
@MikeInnes another CI error. Let's try to get this into the release

PR #701, opened 2020-06-18
move to CUDA.jl

CarloLucibello:
also dropping julia <1.3 support

supercedes #673


MikeInnes:
There's also a bunch of changes to the test setup here, can you elaborate on those a bit? I prefer to have an overall test set and this is used pretty consistently over all the packages I'm aware of.

PR #721, opened 2020-06-30
Use ChainRules v0.7.0

sethaxen:
ChainRules v0.7.0 introduced a new convention for complex numbers. It is no longer necessary for Zygote to conjugate sensitivities sent to and received from ChainRules. This PR is a non-breaking change that uses ChainRules v0.7.0.

Fixes https://github.com/JuliaDiff/ChainRules.jl/issues/210 and supersedes #720.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -25,7 +25,7 @@ ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 [compat]
 AbstractFFTs = "0.5"
 ArrayLayouts = "0.1, 0.2, 0.3"
-ChainRules = "0.6.0"
+ChainRules = "0.7.0"
 DiffRules = "1.0"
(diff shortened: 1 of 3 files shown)


oxinabox:
bors+

sethaxen:
Thanks!

MikeInnes:
Is there a particular reason to use 0.7.0 here rather than 0.7? 0.7.0 seems very restrictive, and will likely cause issues when we have to update the version in future.

MikeInnes:
Scratch that, the concern about the version number is moot. But it would still be good to hear thoughts on whether a patch release is OK here. I guess Zygote's behaviour is not changing even though CR did.

sethaxen:
Yes, Zygote's behavior should not have changed at all, so a patch release is fine.

PR #723, opened 2020-07-01
ChainRules compat

MikeInnes:
Compat bounds should generally be as wide as possible; seems right that we should drop CR 0.6 as complex support has changed significantly, but I'm not sure why we shouldn't include patch versions.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -25,7 +25,7 @@ ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 [compat]
 AbstractFFTs = "0.5"
 ArrayLayouts = "0.1, 0.2, 0.3"
-ChainRules = "0.7.0"
+ChainRules = "0.7"
 DiffRules = "1.0"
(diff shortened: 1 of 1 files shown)


MikeInnes:
cc @oxinabox @sethaxen would like to get confirmation that this looks OK to you.

I assume this shouldn't prevent a patch release?

Project.toml
```diff
@@ -25,7 +25,7 @@ ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 [compat]
 AbstractFFTs = "0.5"
 ArrayLayouts = "0.1, 0.2, 0.3"
-ChainRules = "0.7.0"
+ChainRules = "0.7"
```
KristofferC, line 28: These should mean the same thing.

MikeInnes: Ah, I'd just misunderstood; the `.0` is redundant but it's not doing any harm. Thanks @KristofferC

PR #728, opened 2020-07-08
support parallel maps (tmap, pmap, vmap, etc.)

ChrisRackauckas:
This PR is quite simple and does it. It would be a shame if we had to keep waiting to support data parallelism just because we can't figure out where to put the code, so I hope this PR can drive the conversation.


ChrisRackauckas:
Note that I would very much understand not wanting the tmap here. It's really to show how it would be done. That's pulled from https://github.com/jw3126/ThreadingTools.jl which isn't registered

DhairyaLGandhi:
I am not sure adding all the utilities here is such a good idea, it bloats up the package and is arguably the domain of a different package (ideally Base) that we can support in terms of writing adjoints etc. But I get the point of having this.

CarloLucibello:
yeah, clearly `tmap` doesn't belong here. We should definitely have `pmap` though, being in Base, and also `vmap` since it's used in NNlib

ChrisRackauckas:
The issue with the maps is that, in some sense, it should be using multiple dispatch, but it doesn't. So the adjoints for all of these different maps is literally the same piece of code which makes it easier to do all in one loop. What I am thinking might be a good quick solution is:

- Fork and tag ThreadingTools.jl (or get it registered)
- Depend on ThreadingTools and LoopVectorization
- Have a map that covers all 4 cases
- Work to get `tmap` into Base, i.e. https://github.com/JuliaLang/julia/issues/36579

chriselrod:
Note that LoopVectorization also has `vmapnt` (uses nontemporal stores) and `vmapntt` (uses nontermporal stores and threads). Documented [here](https://chriselrod.github.io/LoopVectorization.jl/latest/vectorized_convenience_functions/#vmapnt-and-vmapntt-1). I'm not too excited about the non-temporal stores (I'm assuming 100 million element vectors is longer than practical applications we'll be seeing on the CPU?; non-temporal stores can severely hurt performance at smaller sizes), but I'd be happy to add the missing `vmapt` for temporal stores + threads.

But I wonder if instead of Threading the maps and threading the matrix multiplications, we should divide batches up by threads?
I'd have to do tests, but I speculate that if layers aren't too large, we could divide the batch up into chunks that fit in the L2 cache and get better cache locality by working on these mini-batches sequentially in each thread. If the matrices are small enough that we don't need much memory churn within those mini-batches, that could decrease the amount of memory churn needed.

CarloLucibello:
Until we figure out the best way forward, let's merge this with just `pmap` and `vmap`. Some tests are needed as well

Project.toml
```diff
@@ -13,6 +13,7 @@ Future = "9fa8497b-333b-5362-9e8d-4d0656e87820"
 IRTools = "7869d1d1-7146-5819-86e3-90919afe41df"
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+LoopVectorization = "bdcacae8-1622-11e9-2a5c-532679323890"
```
oxinabox, line 16: Why is this dependency added?

ChrisRackauckas: vmap

ChrisRackauckas: FWIW, NNlib and such have to use it for other reasons as well, so it's effectively becoming a common dependency for any Flux user already.

oxinabox: It isn't referenced in the code anywhere?

ChrisRackauckas: because I'm silly.

CarloLucibello: It's used in NNlib, and the way we do things right now it to provide adjoints for whatever NNlib defines/needs here in Zygote.
Probably is better to go the other way around, remove NNlib's dependence from Zygote and have NNlib define its own adjoints, but this would be stuff for separate discussion and PR

ChrisRackauckas:
Added some tests now too.  I really just wanted to fix the fact that "can Zygote differentiate parallel code?" had a solution that could be done in about 10 minutes from a cellphone screen during a call, so there's really no reason to keep waiting. This'll work on the 90% of codes which happen to be data-parallel, and we can get a `tmap` in a different package, and the rest of parallelism can take a lot longer. 

>yeah, clearly tmap doesn't belong here. We should definitely have pmap though, being in Base, and also vmap since it's used in NNlib

The nicer solution is https://github.com/JuliaLang/julia/issues/36579#issuecomment-655655292, but that doesn't exist yet, so I think it's best to just get something working until data-parallelism in Julia itself gets more organized, and then users will naturally switch to that along with Zygote.

ChrisRackauckas:
Hmm. @chriselrod is the LoopVectorization stack blocking something in LLVM.jl that CUDA.jl needs?

chriselrod:
VectorizationBase is currently set to require 1.5, 1.6, 1.7, or 2.0.

CarloLucibello:
CUDA.jl has compat
LLVM = "1.5.2, 2"
so I don't see what is causing the problem

DhairyaLGandhi:
Quick fix would be to wrap it in `@requires`

ChrisRackauckas:
How would that fix it? Seems like that just wouldn't test it?

maleadt:
CUDA 1.1 has dropped support for Julia 1.3, but only 1.1 is compatible with LLVM 2.0 (which is required explicitly by the build here -- not sure why).

DhairyaLGandhi:
bors try

DhairyaLGandhi:
bors try

chriselrod:
It needs to depend on and `using Distributed` for `pmap`.

chriselrod:
I'll generalize the vmap method and release a new LoopVectorization version shortly.

ChrisRackauckas:
Tests are passing locally with the latest LoopVectorization 🎉

ChrisRackauckas:
bors try

ChrisRackauckas:
:(

DhairyaLGandhi:
bors try

ChrisRackauckas:
Bors ran on v1.4 (https://gitlab.com/JuliaGPU/Zygote-jl/-/jobs/632198576#L50) but I forgot to change the name of the job so it said v1.3. So that last commit is just fixing the name but it should be good to go.

CarloLucibello:
still not clear what is forcing LLVM to 2.0 on julia 1.3. I don't think we had this issue before this PR

CarloLucibello:
actually the issue is already on master, see #729 

bors r+

PR #747, opened 2020-07-28
Add jacobian function

ChrisRackauckas:
Fixes https://github.com/FluxML/Zygote.jl/issues/98

### DIFF
--- a/src/lib/utils.jl
+++ b/src/lib/utils.jl
@@ -111,7 +111,7 @@ function jacobian(f,x)
     n  = length(x)
     J  = Matrix{eltype(y)}(undef,k,n)
     e_i = fill!(similar(x), 0)
-    for i = 1:k
+    @inbounds for i = 1:k
         e_i[i] = oneunit(eltype(x))
         J[i,:] = back(e_i)[1]
         e_i[i] = zero(eltype(x))
--- a/src/lib/utils.jl
+++ b/src/lib/utils.jl
@@ -110,7 +110,7 @@ function jacobian(f,x)
     k  = length(y)
     n  = length(x)
     J  = Matrix{eltype(y)}(undef,k,n)
-    e_i = fill!(similar(x), 0)
+    e_i = fill!(similar(y), 0)
     @inbounds for i = 1:k
         e_i[i] = oneunit(eltype(x))
         J[i,:] = back(e_i)[1]
--- a/src/lib/utils.jl
+++ b/src/lib/utils.jl
@@ -116,7 +116,7 @@ function jacobian(f,x)
         J[i,:] = back(e_i)[1]
         e_i[i] = zero(eltype(x))
     end
-    J
+    (J,)
 end
 
 """
--- a/src/lib/utils.jl
+++ b/src/lib/utils.jl
@@ -110,7 +110,7 @@ function jacobian(f,x)
     k  = length(y)
     n  = length(x)
     J  = Matrix{eltype(y)}(undef,k,n)
-    e_i = zero(x)
+    e_i = fill!(similar(x), 0)
     for i = 1:k
         e_i[i] = oneunit(eltype(x))
         J[i,:] = back(e_i)[1]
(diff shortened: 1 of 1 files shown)


src/lib/utils.jl
```diff
@@ -99,6 +99,26 @@ macro showgrad(x)
     end)
 end
 
+"""
+jacobian(f,x)
+
+Construct the Jacobian of `f` where `x` is a real-valued array
+and `f(x)` is real-valued.
```
CarloLucibello, line 106:
```suggestion
and `f(x)` is also a real-valued array.
```

CarloLucibello:
needs a test and a mention in the docs

cossio:
Bump?

MasonProtter:
If we split out `OneHotVector` from Flux.jl, we could have this simpler implementation that's sometimes more performant and has less weird edge cases:
```julia
using Flux: OneHotVector

"""
jacobian(f,x)
Construct the Jacobian of `f` where `x` is a real-valued array
and `f(x)` is also a real-valued array.
"""
function jacobian(f, x)
    y, ∂f = Zygote.pullback(f,x)
    n, m  = length(x), length(y)
    mapreduce(i -> ∂f(OneHotVector(i, m)), hcat, 1:m)
end
```

src/lib/utils.jl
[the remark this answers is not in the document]
ChrisRackauckas: I'm not sure. I think people either know that or "want a Jacobian", in which case it's just an odd detail.

src/lib/utils.jl
```diff
@@ -99,6 +99,26 @@ macro showgrad(x)
...
+
+Construct the Jacobian of `f` where `x` is a real-valued array
+and `f(x)` is also a real-valued array.
+"""
+function jacobian(f,x)
+    y,back  = Zygote.pullback(f,x)
+    k  = length(y)
+    n  = length(x)
+    J  = Matrix{eltype(y)}(undef,k,n)
+    e_i = fill!(similar(x), 0)
```
DhairyaLGandhi, line 113: `zero(x)` would be clearer

DhairyaLGandhi (commented):
Looks good, but needs tests

DhairyaLGandhi:
Also probably something that should hook into the forward mode

ChrisRackauckas:
Forward mode Jacobian should be kept as a separate call: which one is better depends on the application so both are required. What this should really connect to is SparseDiffTools for sparse Jacobians, but what will happen there is that SparseDiffTools will have a... more intense version that handles all of that extra machinery.

DhairyaLGandhi:
This is due to it detecting sparsity patterns I assume?

src/lib/utils.jl
mcabbott: But not mutable -- see earlier suggestion.

MasonProtter:
```suggestion
    e_i = fill!(similar(y), 0)
```
For reverse diff, your `e_i` is the wrong size unless your jacobian happens to be square.

MasonProtter (changes requested):

ChrisRackauckas:
>This is due to it detecting sparsity patterns I assume?

Well this could be made to accept a color vector even without SparseDiffTools.

src/lib/utils.jl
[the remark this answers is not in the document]
mcabbott: The point would be to remind them which way around the indices are, and that dimensions of x are flattened not kept. These are programming conventions not mathematical ones.

src/lib/utils.jl
```diff
@@ -99,6 +99,26 @@ macro showgrad(x)
...
+    k  = length(y)
+    n  = length(x)
+    J  = Matrix{eltype(y)}(undef,k,n)
+    e_i = fill!(similar(y), 0)
+    @inbounds for i = 1:k
+        e_i[i] = oneunit(eltype(x))
+        J[i,:] = back(e_i)[1]
+        e_i[i] = zero(eltype(x))
+    end
+    J
```
mcabbott, line 119:
```suggestion
    (J,)
```
Returning a 1-tuple allow for a consistent `jacobian(f, x, y)` without breaking anything, and would match what `gradient` does.

PR #748, opened 2020-07-29
Speed up of selu

racinmat:
Implementing speedup of selu based on this comment: https://github.com/FluxML/Zygote.jl/pull/367#issuecomment-541844988
Utilizing vectorized ifelse based on https://github.com/FluxML/NNlib.jl/blob/master/src/activation.jl#L159.

### DIFF
--- a/src/lib/nnlib.jl
+++ b/src/lib/nnlib.jl
@@ -1,8 +1,7 @@
 using NNlib
 import NNlib: softmax, ∇softmax, logsoftmax, ∇logsoftmax, conv, depthwiseconv, ∇conv_data, ∇depthwiseconv_data, maxpool, meanpool, σ, relu, batched_mul, batched_adjoint
-import LoopVectorization: vifelse
 
-drelu(x, Δ) = vifelse(x > 0, Δ, zero(x))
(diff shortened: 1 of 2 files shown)


src/lib/nnlib.jl
```diff
@@ -1,12 +1,23 @@
 using NNlib
 import NNlib: softmax, ∇softmax, logsoftmax, ∇logsoftmax, conv, depthwiseconv, ∇conv_data, ∇depthwiseconv_data, maxpool, meanpool, σ, relu, batched_mul, batched_adjoint
+import LoopVectorization: vifelse
```
DhairyaLGandhi, line 3: Based on current discussion, vectorisation would need some changes to how we plan to integrate lv within lv. Could we remove the dep from this PR?

racinmat: Oh, my bad, sure, I'll return it.

DhairyaLGandhi:
Thank you so much!

Would you mind adding the tanh adjoint in there as well?

racinmat:
Done.

DhairyaLGandhi:
Thanks!

racinmat:
May I expect a new patch version released, so we could start using it?

PR #750, opened 2020-07-30
Bump patch version

DhairyaLGandhi:
cc @racinmat


racinmat:
Thanks. But why are you deleting the tanh derivation whixh you specifically asked me to add?

Issue #712, opened 2020-06-23
remove DiffRules dependence


CarloLucibello:
#503 reintroduces DiffRules dependence. Can we use ChainRules instead? @oxinabox @MikeInnes

oxinabox:
Yes, we can, and we should.

oxinabox:
I believe @willtebbutt  talked to @MikeInnes  about it when #503 was openned.

willtebbutt:
But yeah, it would definitely make sense to make the forwards-mode stuff in Zygote also depend on ChainRules... you'll get maintenance for free.

sethaxen:
And we already have a decent number of `frule`s that would instantly become available.

PR #752, opened 2020-08-01
WIP: use ChainRules in Forward

willtebbutt:
Addresses https://github.com/FluxML/Zygote.jl/issues/712.

TODO:

- [ ] Get nested AD working
- [ ] Ensure that kwargs work correctly
- [ ] "Unwrap" ChainRules types into Zygote type for compatibility
- [ ] Remove all excess rules
- [ ] Tidy up

A separate PR will be needed to move rules that are currently in `Zygote.Forward` to `ChainRules`. Unlike the reverse-mode rules there aren't a particularly large number of them, so it shouldn't be a massive job.


src/forward/compiler.jl
```diff
@@ -70,7 +70,90 @@ function dual(ir)
...
+  hascr, cr_edge = has_chain_frule(base_T)
+
+  chain_frule_f = iskw ? :chain_frule_kw : :chain_frule
+  if hascr
+    return :($chain_frule_f(dargs, f, args...))
+  else
+    return :(__pushforward(dargs, f, args...))
+  end
+
+  # g = try _lookup_grad(T) catch e e end
```
willtebbutt, line 142: I'm really not sure what I'm doing with the edges stuff here -- this is currently copied from the reverse-mode definition. @oxinabox or @DhairyaLGandhi  could you comment on what sort of thing I'm going to need to do here? In particular, I'm not sure what `_lookup_grad` is doing, so I'm not sure how to reason about adapting the reverse-mode implementation.

test/forward/forward.jl
```diff
@@ -20,22 +29,22 @@ end
...
+# @test D(1) do x
+#   D(2) do y
+#     D(3) do z
+#       x = z * y
+#     end
+#   end
+#   x
+# end == 0
 
 @test D(x -> abs(x+2im), 1) == gradient(x -> abs(x+2im), 1)[1]
```
willtebbutt, line 49: This doesn't currently pass. I'm pretty sure that the `Forward` implementation is doing the correct thing, and `gradient` is wrong. `D` (correctly) returns a `Real` as the derivative w.r.t `x`, while `gradient` returns a `Complex`. @sethaxen do you agree that Zygote is doing the wrong thing here?

test/forward/forward.jl
```diff
@@ -1,13 +1,22 @@
...
+# This intentionally has the wrong definition so that we can detect if Zygote is using it.
+ChainRulesCore.frule((_, dx), ::typeof(foo), x) = x, 2 * dx
+
+D(f, x...) = pushforward(f, x...)(1)
+
+@test D(foo, 1.0) == 2
 
 @test D(x -> sin(cos(x)), 0.5) == -cos(cos(0.5))*sin(0.5)
 
 @test D(x -> D(cos, x), 0.5) == -cos(0.5)
```
willtebbutt, line 17: My implementation at the minute doesn't like nested AD. There's a bunch of debugging still to do, so it's entirely possible that it's innocuous. In any case, @oxinabox or @DhairyaLGandhi if you get a minute, could you take a look a the error that this line generates?

willtebbutt:
Most recent commit fixed one bug, but re-introduced some tests that expose us to having to differentiate through `broadcast`, which isn't something that `Zygote.Forward` currently handles. I'll need to give that some thought...

test/forward/forward.jl
sethaxen: ChainRulesCore has an issue open about this: https://github.com/JuliaDiff/ChainRulesCore.jl/issues/176. It's also related to https://github.com/JuliaDiff/ChainRules.jl/issues/232. The issue is that our rules for addition between real and complex (`+(x, 2im)`) will pull back a complex adjoint to `x`. 
So we'd expect these to do the same thing, but
```julia
julia> gradient(x -> abs(x+2im), 1)[1]
0.4472135954999579 + 0.8944271909999159im

julia> gradient(x -> abs(complex(x, 2)), 1)[1]
0.4472135954999579
```
The first issue above links to some Zygote discussions. Essentially, Zygote treats all reals as actually complex, so from that perspective it is doing the right thing. I just don't think that's the right way to go.

willtebbutt: Ahh interesting. It's interesting that `Zygote.Forward` doesn't naturally do the same thing as `Zygote`.

I guess that the question is what the right way forward for us is with `Forward`. Mike previously had some extra code for `abs` to make it do the Zygote-y thing.

I wonder whether it's acceptable to utilise a different convention in `Forward` than what `Zygote` currently does...

@DhairyaLGandhi what are your thoughts on the way that Zygote currently handles complex numbers / have you had a chance to review the various ChainRules discussions about complex numbers?

src/forward/compiler.jl
```diff
@@ -70,7 +70,91 @@ function dual(ir)
...
+
+Determines if `sigt` is the type signature of a kwfunction.
+Each element of `sigt` should be a type.
+Either the first 3 types are a kwfunc type, a NamedTuple and the matching base function type,
+or the first argument is the base function type and it is not a kwfunction.
+the remaining types in `sigt` are the types of the argument.
+
+"""
+is_kwfunc(::Vararg{Any}) = false
+is_kwfunc(k, ::Type{<:NamedTuple}, f, args...) = k===Core.kwftype(f)
```
oxinabox, line 125: this code exists already for reverse mode we should just import it from there?

willtebbutt: Yes. I'll need to refactor to avoid all of this duplication. I just copied + pasted to start with because I was tired haha

src/forward/compiler.jl
```diff
@@ -70,7 +70,91 @@ function dual(ir)
...
+  if m.method !== chainrules_frule_fallback
+    # found a frule, no need to add any edges
+    return true, nothing
+  end
+
+  return false, m.instance
+end
+
+"""
+    chain_frule(f, args...)
```
oxinabox, line 96: this doc is wrong but the code is right.
`dargs` includes `dself, darg1, darg2 ...` and `args` includes `foo, darg1, darg2`

src/forward/compiler.jl
```diff
@@ -70,7 +70,91 @@ function dual(ir)
...
+  ignore_sig(T) && return :(f(args...), Zero())
+
+  iskw = is_kwfunc(dargs, f, args...)
+  # if it is_kw then `args[1]` are the keyword args, `args[2]` is actual function
+  base_T = iskw ? Tuple{args[2:end]...} : T
+  hascr, cr_edge = has_chain_frule(base_T)
+
+  chain_frule_f = iskw ? :chain_frule_kw : :chain_frule
+  if hascr
+    return :($chain_frule_f(dargs, f, args...))
```
oxinabox, line 138: where are my kwargs gone?

src/forward/compiler.jl
```diff
@@ -70,7 +70,91 @@ function dual(ir)
...
+@noinline chain_frule(dargs, args...) = frule(dargs, args...)
+
+"""
+  chain_frule_kw(kwf, kwargs, f, args...)
+
+As per [`chain_frule`](@ref) but with support for kwargs.
+`kwf` should be the kwfunc matching to `f`, and `kwargs` are a `NamedTuple` of keyword
+arguments.
+"""
+@inline chain_frule_kw(kwf, kwargs, f, args...) = frule(f, args...; kwargs...)
```
oxinabox, line 110: This doesn't seem right.
`frule` takes `dargs` first before `f`.
Maybe:

```suggestion
@inline chain_frule_kw(kwf, kwargs, dargs, f, args...) = frule(dargs, f, args...; kwargs...)
```
which acts the same under most circumstances

src/forward/compiler.jl
```diff
@@ -70,7 +70,91 @@ function dual(ir)
...
+For a type-tuple `T` e.g. `Tuple{typeof(f), Int, Float64}`, checks if there is a `rrule` defined for it.
+Excluding the generic fallback.
+The first return value is `true` if the `rrule` exists, `false` otherwise.
+If it does not, then the second argument is a list of edges to attach to the CodeInfo for a generated function,
+such that if a suitable rule is defined later, the generated function will recompile.
+"""
+function has_chain_frule(T)
+  m = meta(Tuple{typeof(frule), T.parameters...})
+
+  if m.method !== chainrules_frule_fallback
```
oxinabox, line 87: I wonder if we should just abstract this to be `has_chainrule(kind, T)`

```suggestion
function has_chain_rule(kind, T)
  m = meta(Tuple{typeof(kind), T.parameters...})

  if m.method !== chainrules_rule_fallback(kind)
```

test/forward/forward.jl
```diff
@@ -1,7 +1,16 @@
...
+using ChainRulesCore
+using Test
+using Zygote
 using NNlib: relu
 
-D(f, x) = pushforward(f, x)(1)
+foo(x) = x
+
+# This intentionally has the wrong definition so that we can detect if Zygote is using it.
+ChainRulesCore.frule((_, dx), ::typeof(foo), x) = x, 2 * dx
```
oxinabox, line 9: the tests in the reverse mode chainrules tests set a variable that can late be read to see if it was hit.
This should be tested the same way

src/lib/lib.jl
```diff
@@ -76,6 +76,10 @@ function global_set(ref, val)
         ref.mod, ref.name, val)
 end
 
+function ChainRules.frule(dargs, ::typeof(Zygote.global_set), ref, val::Nothing)
+  return Zygote.global_set(ref, val), nothing
```
oxinabox, line 80:
```suggestion
  return Zygote.global_set(ref, val), DoesNotExist()
```

src/forward/compiler.jl
```diff
@@ -70,7 +70,91 @@ function dual(ir)
...
+  return false, m.instance
+end
+
+"""
+    chain_frule(f, args...)
+
+Returns a the (primal) value of `f(args...)` and tangent, by invoking
+`ChainRules.frule(f, args...)`.
+"""
+@noinline chain_frule(dargs, args...) = frule(dargs, args...)
```
oxinabox, line 101: we need a translation layer replacing Composites with NamedTuples and AbstractZeros with `nothing`
and visa versa

Like we have for reverse mode.
Can probably just abstract and call that

src/forward/compiler.jl
willtebbutt: Just not implemented yet, see check-boxes at the top :)

src/forward/compiler.jl
willtebbutt: Yup, I just haven't done anything with kwargs yet, thanks for pointing this out though.

test/forward/forward.jl
willtebbutt: Oh nice, I'll do that.

src/forward/compiler.jl
willtebbutt: Yup, on the TODO list :)

willtebbutt:
Thanks for the review @oxinabox . The main thing I'm still unsure about is what `_lookup_grad` was doing in the reverse mode implementation. (See my first comment)

oxinabox:
AFAICT `_lookup_grad` in reverse, dispite the name, does not look up the gradient,
it performs the core of the source code tranform to do AD. it is defined in [emit.jl](https://github.com/FluxML/Zygote.jl/blob/master/src/compiler/emit.jl)
Roughly speaking the matching function in forward is `dual` which is also terribly named

willtebbutt:
Closing because I've definitely not got time to work on this at the minute, and don't anticipate doing so for a while. If someone else wants to have a go at this, they should feel free to do so.

PR #757, opened 2020-08-05
Remove Base.parameter_upper_bound given removal from Base

ianshmean:
`Base.parameter_upper_bound` has been removed on julia master, so was breaking Flux on julia master.
It doesn't seem to be mentioned anywhere other than this in Zygote

I couldn't find where it left base, but perhaps it wasn't intended to be used anyway given https://github.com/JuliaLang/julia/pull/36769#issuecomment-662719375

I can't vouch for just deleting this being a good idea, other than I couldn't find any other mention in the Zygote source, but either way it needs fixing


DhairyaLGandhi:
Thanks!

bors r+

ianshmean:
@DhairyaLGandhi just for your radar, this is also blocking on master https://github.com/FluxML/IRTools.jl/issues/71

Issue #742, opened 2020-07-24
ERROR: LoadError: Need an adjoint for constructor SparseMatrixCSC{Float64,Int64}. Gradient is of type Array{Float64,2}.


pzhanggit:
In the following case, 
```
using SparseArrays,Zygote

function loss_rd(x::AbstractArray{T,1}) where{T}
    f = ones(T, 3)
    K = sparse(I, J, x)
    u = K \ f
    return sum(abs2, u)
end
# case that does not work
I = [1, 1, 2, 3, 3]
J = [1, 3, 2, 3, 1]
x = [3.0, 1.0, 3.0, 3.0, 1.0]

println("loss_rd=", loss_rd(x))

dldx=Zygote.gradient(loss_rd,x)
println("Zygnote, dl/dx=",dldx)
```
I got the error,
```
ERROR: Need an adjoint for constructor SparseMatrixCSC{Float64,Int64}. Gradient is of type Array{Float64,2}
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] (::Zygote.Jnew{SparseMatrixCSC{Float64,Int64},Nothing,false})(::Array{Float64,2}) at /Users/6pz/.julia/packages/Zygote/1GXzF/src/lib/lib.jl:306
 [3] (::Zygote.var"#380#back#194"{Zygote.Jnew{SparseMatrixCSC{Float64,Int64},Nothing,false}})(::Array{Float64,2}) at /Users/6pz/.julia/packages/ZygoteRules/6nssF/src/adjoint.jl:49
 [4] SparseMatrixCSC at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/SparseArrays/src/sparsematrix.jl:32 [inlined]
 [5] (::typeof(∂(SparseMatrixCSC{Float64,Int64})))(::Array{Float64,2}) at /Users/6pz/.julia/packages/Zygote/1GXzF/src/compiler/interface2.jl:0
 [6] SparseMatrixCSC at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/SparseArrays/src/sparsematrix.jl:45 [inlined]
 [7] (::typeof(∂(SparseMatrixCSC)))(::Array{Float64,2}) at /Users/6pz/.julia/packages/Zygote/1GXzF/src/compiler/interface2.jl:0
[...]

Issue #570, opened 2020-03-31
StaticArrays adjoint constructor


martinhath:
I'm trying to use Zygote.jl together with StaticArrays.jl, but am getting an error message I don't understand. Here's a very minimal example

```julia
julia> gradient(n->SMatrix{1,1}(n)[1], 1)
ERROR: Need an adjoint for constructor SArray{Tuple{1,1},Int64,2,1}. Gradient is of type MArray{Tuple{1,1},Int64,2,1}
```
I have tried to add adjoints, but I don't understand for which type the constructor should be for, take in, and whether the type of the delta matters. None of the adjoints I've tried to define helps, so there's obviously things I don't understand here. It is also very confusing that the error message mentions `MArray` when I've gone out of my way of avoiding mutable arrays, since Zygote doesn't support that.

AzamatB:
To clarify, Zygote does not support mutation operation, but it does support mutable data structures (`Array` is a prime example of that). So `MArray` should be fine as long as you don't mutate then in the code you want to differentiate.

mcabbott:
The gradient of `getindex` calls `similar(x)` to get a fresh array to write into, it needs mutation internally:
```julia
julia> similar(SA[1,2]) isa MArray
true

julia> gradient(x -> x[1], SA[1,2])[1]
2-element MArray{Tuple{2},Int64,1,2} with indices SOneTo(2):
 1
 0
```
In your example it's trying and failing to get back from that to the argument(s), the manual has a section on making this work for new structs... but StaticArrays has lots of  more complicated constructors than the examples there. However the `SA` form seems to work:
```julia
julia> gradient((x,y) -> sum(SA[x,y,x,y][1:3]), 1,2)
(2, 1)
```

martinhath:
> To clarify, Zygote does not support mutation operation, but it does support mutable data structures (Array is a prime example of that). So MArray should be fine as long as you don't mutate it in the code you want to differentiate.

Alright, I figured as much :) It seems I've found some ways to go around the need for mutation (for now). 

> The gradient of getindex calls similar(x) to get a fresh array to write into, it needs mutation internally:

I get that this is the wrong repo to ask this, but why does `getindex` without `!` need mutation? What do you mean by "a fresh array to write into" when we're getting a scalar?


> In your example it's trying and failing to get back from that to the argument(s)

Okay I see; StaticArrays is converting internally in between `S` and `MArray`

> the manual has a section on making this work for new structs.

[...]

PR #762, opened 2020-08-10
basic sparse handling

DhairyaLGandhi:
Relates to #570, SciML/DifferentialEquations.jl#649, #742


willtebbutt:
Can this be closed in favour of https://github.com/JuliaDiff/ChainRules.jl/pull/246?

test/gradcheck.jl
```diff
@@ -1571,6 +1573,12 @@ end
   end
 end
 
+<<<<<<< HEAD
```
DhairyaLGandhi, line 1576:
```suggestion
```

src/lib/array.jl
```diff
@@ -957,3 +959,15 @@ end
   back(Δ::AbstractArray) = (nothing, getindex.(_back.(Δ), 1))
   return Fill(y, size(r)), back
 end
+
+# Sparse Arrays
+
+@adjoint function SparseMatrixCSC{T,N}(arr) where {T,N}
+  SparseMatrixCSC{T,N}(arr), Δ -> (collect(Δ),)
+end
+@adjoint diagm(x) = diagm(x), Δ -> (diag(Δ), )
```
willtebbutt, line 968: Perhaps consider restricting the type here a bit? For example, this adjoint is probably innappropriate for `Fill`s since `diag(Δ)` will presumably be dense.

src/lib/array.jl
```diff
@@ -957,3 +959,15 @@ end
...
+
+@adjoint function Broadcast.broadcasted(Float32, a::SparseMatrixCSC{T,N}) where {T,N}
+  Float32.(a), Δ -> (nothing, T.(Δ), )
+end
+@adjoint issymmetric(x) = issymmetric(x), Δ -> (Δ,)
```
willtebbutt, line 973: Is this differentiable, or should it be `@nograd`ed?

willtebbutt (commented):
A couple of comments.

DhairyaLGandhi:
bors r+

Let's also add the corresponding changes in ChainRules, but I think we have some work looking to use sparse handling currently, and we can dd it to the list of things we have to move.

DhairyaLGandhi:
bors r-

Reviews 👍

src/lib/array.jl
DhairyaLGandhi: I think it needs the adjoint here, else you'll get a bunch of `nothing`s

src/lib/array.jl
DhairyaLGandhi: Shouldn't  `diagm` fail in that case?

src/lib/array.jl
willtebbutt: What's wrong with getting `nothing`s in this context?

src/lib/array.jl
willtebbutt: Well, for example,
```julia
julia> diagm(Fill(5.0, 3))
3×3 Array{Float64,2}:
 5.0  0.0  0.0
 0.0  5.0  0.0
 0.0  0.0  5.0
```
seems to be well-defined.

src/lib/array.jl
DhairyaLGandhi: It can lead to dropped gradients for `x` then I think

src/lib/array.jl
DhairyaLGandhi: Would you recommend `DenseArray`?

DhairyaLGandhi: Actually we need it for the sparse arrays, so I'm not quite sure

src/lib/array.jl
willtebbutt: Looking at this another way, `Δ` here is presumably a scalar of some kind, so isn't an appropriate gradient for `x`.

src/lib/array.jl
willtebbutt: That would work for me!

willtebbutt: Ah, yes, good point. Maybe restrict it to sparse arrays for now then, and expand if necessary later?

src/lib/array.jl
willtebbutt: Similarly, `issymmetric(randn())` will always return `true`, so the derivative should definitely be some flavour of `0`.

src/lib/array.jl
```diff
@@ -957,3 +959,21 @@ end
...
+  SparseMatrixCSC{T,N}(arr), Δ -> (collect(Δ),)
+end
+
+@adjoint function SparseVector{T,N}(v) where {T,N}
+  SparseVector{T,N}(arr), Δ -> (collect(Δ),)
+end
+
+@adjoint diagm(x::AbstractSparseArray) = diagm(x), Δ -> (diag(Δ), )
+
+@adjoint function Broadcast.broadcasted(Float32, a::SparseMatrixCSC{T,N}) where {T,N}
```
DhairyaLGandhi, line 975:
```suggestion
@adjoint function Broadcast.broadcasted(Float32, a::AbstractSparseArray{T,N}) where {T,N}
```

src/lib/array.jl
DhairyaLGandhi: Done

willtebbutt (approved):
No LGTM. Just needs a patch bump :)

test/gradcheck.jl
```diff
@@ -1571,6 +1573,12 @@ end
   end
 end
 
+@testset "Sparse" begin
+  @test gradtest(x -> sum(sparse(x)), rand(Float32, 3,3))
+  @test gradtest(x -> sum(sparse(x)), rand(Float32, 3)) # test vectors also
+  @test gradcheck(x -> sum(diagm(x)), rand(3))
```
DhairyaLGandhi, line 1579:
```suggestion
  @test gradcheck(x -> sum(diagm(x)), sparse(rand(3)))
```

src/lib/array.jl
```diff
@@ -957,3 +959,21 @@ end
...
+  SparseMatrixCSC{T,N}(arr), Δ -> (collect(Δ),)
+end
+
+@adjoint function SparseVector{T,N}(v) where {T,N}
+  SparseVector{T,N}(v), Δ -> (collect(Δ),)
+end
+
+@adjoint diagm(x::AbstractSparseArray) = diagm(x), Δ -> (diag(Δ), )
+
+@adjoint function Broadcast.broadcasted(Float32, a::AbstractSparseArray{T,N}) where {T,N}
```
ChrisRackauckas, line 975:
```suggestion
@adjoint function Broadcast.broadcasted(::Type{Float32}, a::AbstractSparseArray{T,N}) where {T,N}
```

otherwise it's just going to overload every case!

ChrisRackauckas:
That small difference would cause weird issues with `identity.(A)` and such, but also `f.(A)` would do `Float32.(A)` instead of `f.(A)` which would lead to incorrect results, so that might be what's causing gradient issues.

test/gradcheck.jl
```diff
@@ -1571,6 +1573,12 @@ end
   end
 end
 
+@testset "Sparse" begin
+  @test gradtest(x -> sum(sparse(x)), rand(Float32, 3,3))
+  @test gradtest(x -> sum(sparse(x)), rand(Float32, 3)) # test vectors also
+  @test gradcheck(x -> sum(diagm(x)), sparse(rand(3)))
+end
```
CarloLucibello, line 1580: can add a test for broadcasting

src/lib/array.jl
```diff
@@ -1,6 +1,8 @@
 using Random, FillArrays, AbstractFFTs
 using FillArrays: AbstractFill, getindex_value
 using Base.Broadcast: broadcasted, broadcast_shape
+using Distributed: pmap
```
CarloLucibello, line 4: why this import?

test/gradcheck.jl
```diff
@@ -2,6 +2,8 @@ using Zygote, Test, Random, LinearAlgebra, Statistics, FillArrays,
     AbstractFFTs, FFTW, Distances
 using Zygote: gradient
 using Base.Broadcast: broadcast_shape
+using Distributed: pmap
```
CarloLucibello, line 5: why this import?

CarloLucibello:
a test is failing

src/lib/array.jl
```diff
@@ -957,3 +959,21 @@ end
...
+end
+
+@adjoint function SparseVector{T,N}(v) where {T,N}
+  SparseVector{T,N}(v), Δ -> (collect(Δ),)
+end
+
+@adjoint diagm(x::AbstractSparseArray) = diagm(x), Δ -> (diag(Δ), )
+
+@adjoint function Broadcast.broadcasted(::Type{Float32}, a::AbstractSparseArray{T,N}) where {T,N}
+  Float32.(a), Δ -> (nothing, T.(Δ), )
```
CarloLucibello, line 976: should this try to produce a sparse `da` also in case `Δ` is non-sparse?

CarloLucibello:
Actually, I see you also filed https://github.com/JuliaDiff/ChainRules.jl/pull/246, can we close this PR in favor of that?

src/lib/array.jl
DhairyaLGandhi:
```suggestion
```

CarloLucibello:
Closing since this has been superseded by https://github.com/JuliaDiff/ChainRules.jl/pull/579

Issue #758, opened 2020-08-07
Nan gradient for elu and selu.


racinmat:
```julia
julia> using Zygote
julia> gradient(x -> sum(Zygote.selu.(x)), [1_000.0])
([1.0507009873554805],)
julia> gradient(x -> Zygote.selu(x), 1_000.0)
(NaN,)
julia> gradient(x -> sum(Zygote.elu.(x)), [1_000.0])
([NaN],)
julia> gradient(x -> Zygote.elu(x), 1_000.0)
(NaN,)
```
the broadcasted version for Selu works because of https://github.com/FluxML/Zygote.jl/blob/master/src/lib/nnlib.jl#L17
but I assume non-broadcasted version should be added and the elu derivation should also be added, right?

DhairyaLGandhi:
+1 for this

racinmat:
Created PR but it would be great if someone reviewed if it's correct.

racinmat:
Closed by #765

PR #765, opened 2020-08-13
fix for https://github.com/FluxML/Zygote.jl/issues/758

racinmat:
Added few tests, defined adjoint for scalar selu and elu.
Fixes #578 
I hope this is right, I'm still not sure if I fully got all the tricks around adjoints.

### DIFF
--- a/test/gradcheck.jl
+++ b/test/gradcheck.jl
@@ -102,6 +102,8 @@ end
 @test gradient(xs -> sum(elu.(xs, 2)), [1_000, 10_000]) == ([1., 1.],)
 @test gradient(x -> elu(x, 2), 1_000) == (1.,)
 @test gradient(x -> elu(x, 2), -1) == (2*exp(-1),)
+@test gradcheck(x->sum(selu.(x)),[1_000, 10_000])
+@test gradcheck(x->sum(elu.(x, 3.5)),[1_000, 10_000])
 
 @test gradtest((x, W, b) -> tanh.(W*x .+ b), 5, (2,5), 2)
 @test gradtest((x, W, b) -> tanh.(W*x .+ b), (5,3), (2,5), 2)
@@ -1462,7 +1464,7 @@ end
 
   x = randn(Float64,16,16)
   @test typeof(gradient(x->sum(abs2,ifft(fft(x,1),1)),x)[1]) == Array{Complex{Float64},2}
-  @test typeof(gradient(x->sum(abs2,irfft(rfft(x,1),16,1)),x)[1]) == Array{Float64,2} 
+  @test typeof(gradient(x->sum(abs2,irfft(rfft(x,1),16,1)),x)[1]) == Array{Float64,2}
 
   x = randn(Float32,16)
   P = plan_fft(x)
--- a/test/gradcheck.jl
+++ b/test/gradcheck.jl
@@ -104,13 +104,13 @@ end
 @test gradient(x -> elu(x, 2), -1) == (2*exp(-1),)
 @test gradcheck(x->sum(selu.(x)),[100., 1_000.])
 @test gradcheck(x->sum(elu.(x, 3.5)),[100., 1_000.])
+@test gradcheck(x->sum(elu.(x, 3.5)),[1_000., 10_000.]) # for elu the tests are passing but for selu not, interesting
 # numerical instability even for the linear part of such function, see:
 # julia> ngradient(x->sum(selu.(x)),[1_000., 10_000.])
 # ([1.0506591796875, 1.0506591796875],)
 # julia> gradient(x->sum(selu.(x)),[1_000., 10_000.])
 # ([1.0507009873554805, 1.0507009873554805],)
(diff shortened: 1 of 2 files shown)


racinmat:
I see the tests are planning, but according to https://travis-ci.org/github/FluxML/Zygote.jl/builds/716162039 the same thing is happening on maste too, the tests I added are passing.

racinmat:
@DhairyaLGandhi  bump

DhairyaLGandhi:
Hmm the failures are seemingly 1.5 related.

The adjoints look about right to me too.

test/gradcheck.jl
```diff
@@ -93,6 +93,15 @@ end
 @test gradtest((x, W, b) -> relu.(W*x .+ b), (5,3), (2,5), 2)
 @test gradtest((x, W, b) -> selu.(W*x .+ b), 5, (2,5), 2)
 @test gradtest((x, W, b) -> selu.(W*x .+ b), (5,3), (2,5), 2)
+@test gradtest((x, W, b) -> elu.(W*x .+ b, 2), 5, (2,5), 2)
+@test gradtest((x, W, b) -> elu.(W*x .+ b, 2), (5,3), (2,5), 2)
+
+# tests for https://github.com/FluxML/Zygote.jl/issues/758
+@test gradient(xs -> sum(selu.(xs)), [1_000, 10_000]) == ([1.0507009873554805, 1.0507009873554805],)
```
DhairyaLGandhi, line 100: Can we use gradtest to check against numerical differentiation

racinmat: I have gradtest for selu on rows 94-95, and I was kinda lost in the gradtest functions, so I didn't understand how can I pass some specific numbers to gradtest, not just matrix sizes

DhairyaLGandhi: See `gradcheck`

racinmat: Oh, I see, thanks.
Ok, now I'm a bit lost :D 
I added some tests, but they are failing because the numeric gradient is inexact:
```julia
ulia> gradcheck(x->sum(selu.(x)),[1_000, 10_000])
ERROR: InexactError: Int64(999.9999999925494)
Stacktrace:
 [1] Int64 at .\float.jl:710 [inlined]    
 [2] convert at .\number.jl:7 [inlined]   
 [3] setindex! at .\array.jl:847 [inlined]
 [4] ngradient(::Function, ::Array{Int64,1}) at C:\Projects\others\Zygote.jl\test\gradcheck.jl:14
 [5] gradcheck(::Function, ::Array{Int64,1}) at C:\Projects\others\Zygote.jl\test\gradcheck.jl:24
 [6] top-level scope at none:1
```

racinmat: And when I make them floats and not integers, the `gradcheck` is returning false
`gradcheck(x->sum(selu.(x)),[1_000., 10_000.])`

racinmat: Or should I assert that the gradcheck is returning false to prove the instability and thus different results of derivation?

racinmat: For 
```julia
gradcheck(x->sum(selu.(x)),[100., 1_000.])
gradcheck(x->sum(elu.(x, 3.5)),[100., 1_000.])
```
it's working, so should I put these smaller numbers here?

DhairyaLGandhi: It should give the right results with floats, I would think

DhairyaLGandhi: It shouldn't fail with large numbers either

DhairyaLGandhi: I would verify that it's the inexact behavior in `ngradient` causing this

racinmat: It's not working for larger numbers, see
```
julia> ngradient(x->sum(selu.(x)),[1_000., 10_000.])
([1.0506591796875, 1.0506591796875],)

julia> gradient(x->sum(selu.(x)),[1_000., 10_000.])
([1.0507009873554805, 1.0507009873554805],)

julia> gradient(x->sum(selu.(x)),[1_000., 10_000.]) .- ngradient(x->sum(selu.(x)),[1_000., 10_000.])
([4.1807667980453544e-5, 4.1807667980453544e-5],)
```

racinmat: The difference is bigger than atol, so it returns false.

DhairyaLGandhi: I see, that makes sense, that shouldn't be an issue in this case hopefully, although I would add a `@test_broken` with the reason in there so we can track this.

racinmat: ok, I'll add `@test` for the lower number and `@test_broken` for higher numbers with the comment on that

racinmat: Tests are now passing with new tests and also broken test.

racinmat:
@DhairyaLGandhi what about now?

DhairyaLGandhi:
Lgtm

Issue #717, opened 2020-06-27
Dictionary indexing failure inside closure and structs


willtebbutt:
```julia
julia> Zygote.gradient(x -> (() -> x[:y])(), Dict(:y => 0.4))
(nothing,)
```
the gradient w.r.t. the `y` element of `x` should be `1`.

This bug doesn't occur with the equivalent closure-free function
```julia
julia> Zygote.gradient(x -> x[:y], Dict(:y => 0.4))
(Dict{Any,Any}(:y => 1.0),)
```
and appears to be `Dict`-specific since
```julia
julia> Zygote.gradient(x -> (() -> x.y)(), (y = 0.4,))
((y = 1.0,),)
```

This bug was introduced in 0.4.21 -- the correct result is obtained on 0.4.20. The bug persists on 0.4.22 and 0.5.

This is breaking for [Stheno.jl](https://github.com/willtebbutt/Stheno.jl/pull/115).

@MikeInnes @CarloLucibello any thoughts on what might be causing this?

willtebbutt:
@DhairyaLGandhi do you have any thoughts on what might be causing this?

femtomc:
Here's another MWE. This one is a little more complex, because it matches a use case that I have.

```julia
module GradsMVP

using Zygote

mutable struct Foo
    store::Dict{Symbol, Float64}
    score::Float64
end

function (f::Foo)(acc::Symbol, fn::Function, args...)
    val = getindex(f.store, acc)
    ret = fn(val)
    f.score += ret
    fn(args...)
end

function get_grads(store, acc, ret_grad, call, args...)
    fn = (args, store) -> begin
        f = Foo(store, 0.0)
        ret = f(acc, call, args...)
        (f.score, ret)
    end
    _, back = Zygote.pullback(fn, args, store)
    arg_grads, store_grads = back((1.0, ret_grad))
    return arg_grads, store_grads
end

function foo(a::Float64)
    return a
end

ags, gs = get_grads(Dict(:x => 1.0), :x, 1.0, foo, 1.0)
println(ags)
println(gs)

end # module
```

femtomc:
To fix this MWE, it suffices to define the adjoint for `getindex`:

```julia
Zygote.@adjoint getindex(d::Dict, acc) = getindex(d, acc), retgrad -> (1.0, nothing)
```

I'm unsure if this will break something fundamental.

femtomc:
[...]

PR #767, opened 2020-08-15
Fix for #717 - grads for dictionaries in closures/struct fields

femtomc:
Commented out `===` check in lib `@adjoint` for `literal_getproperty`.

This is breaking grads for dicts in closures (and thus, also structs with dictionary fields).

There are a number of small MWEs for this issue:

https://github.com/FluxML/Zygote.jl/issues/717

I have no idea if this breaks other things, because master already has broken tests.

### DIFF
--- a/src/lib/lib.jl
+++ b/src/lib/lib.jl
@@ -199,7 +199,7 @@ end
 @adjoint function literal_getproperty(x, ::Val{f}) where f
   val = getproperty(x, f)
   function back(Δ)
-    accum_param(__context__, val, Δ) === nothing && return
+    accum_param(__context__, val, Δ) # === nothing && return
     if isimmutable(x)
       ((;nt_nothing(x)...,pair(Val(f), Δ)...), nothing)
     else
--- a/test/lib/base.jl
+++ b/test/lib/base.jl
@@ -1,0 +1,32 @@
+struct Maz
+    d::Dict
+end
+maz = m -> begin
+    d = getfield(m, :d)
+    z = getindex(d, :x)
+    return z
+end
+
+struct Bar
+    x::Float64
+end
+bar = m -> begin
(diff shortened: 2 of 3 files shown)


willtebbutt:
@femtomc could you add a regression test please? Would like to prevent this bug from creeping back in.

femtomc:
Absolutely.

I don’t fully understand this code - but I can transform the MWEs into a small set of tests.

DhairyaLGandhi:
Not sure this is the right thing to do here, we want to shortcut out if something doesn't have a gradient

femtomc:
Well it’s catching things which do have gradients - so the intention is not working as expected.

Is there an example of when the shortcut is applied? That would help me change my fix.

femtomc:
Transformed the MWEs into a test suite.

DhairyaLGandhi:
We do need to return the nothing there, so the "fix" would be to let it fall back to the `nt_nothing` function and accumulate there.

CarloLucibello:
> We do need to return the nothing there, so the "fix" would be to let it fall back to the nt_nothing function and accumulate there.

can you post a scratch of the approach?

DhairyaLGandhi:
I'll try to post a write-up soon

CarloLucibello:
bumpity bump

DhairyaLGandhi:
Yeah, I had written something similar for this and another project here https://github.com/Chemellia/ChemistryFeaturization.jl/pull/111, I have to convert that into a cleaned up PR to Zygote.

ToucheSir:
Ref. https://github.com/FluxML/Zygote.jl/issues/717#issuecomment-1208927447 for visibility.

CarloLucibello:
closing as outdated. Also, #717 has been closed

PR #768, opened 2020-08-16
Update ArrayLayouts and FillArrays

devmotion:
Supersedes https://github.com/FluxML/Zygote.jl/pull/763, https://github.com/FluxML/Zygote.jl/pull/759, and https://github.com/FluxML/Zygote.jl/pull/764. All these PRs didn't actually test ArrayLayouts 0.4 and FillArrays 0.9 (or even failed to install them) since FillArrays 0.9 requires ArrayLayouts 0.4 and hence both compatibilities have to be updated together.

Locally tests fail in the same way as on the master branch with Julia 1.5 and nightly.

PR #772, opened 2020-08-24
Changes needed for Julia 1.5

DhairyaLGandhi:
Passes some of the tests for Julia 1.5 locally, but gives ~incorrect gradients for some cases.~ the right result now

Ref #768

### DIFF
--- a/src/lib/base.jl
+++ b/src/lib/base.jl
@@ -58,7 +58,7 @@ grad_mut(ch::Channel) = Channel(ch.sz_max)
 @adjoint! function put!(ch::Channel, x)
   put!(ch, x), function (ȳ)
     x̄ = grad_mut(__context__, ch)
-    x̄ = x̄ isa Channel ? isready(x̄) ? take!(x̄) : nothing : x̄
+    x̄ = x̄ isa Channel ? isready(x̄) ? take!(x̄) : x̄ : nothing
     (nothing, accum(x̄, ȳ), nothing)
   end
 end
@@ -74,11 +74,10 @@ end
   t = Task(f)
   t.code = function ()
     y, back = _pullback(__context__, f)
-    cache(__context__)[t] = Task(() -> back(sensitivity(y)))
+    cache(__context__)[t] = Task(back)
     return y
   end
   t, _ -> begin
-    istaskdone(cache(__context__)[t]) || schedule(cache(__context__)[t])
     fetch(cache(__context__)[t])
   end
 end
@@ -102,10 +101,14 @@ end
 end
 
 @adjoint! function Base.sync_end(refs)
-  Base.sync_end(refs), _ -> begin
-    foreach(t -> runadjoint(__context__, t), refs)
-    return
-  end
+  Base.sync_end(refs), _ -> foreach(t -> runadjoint(__context__, t), refs)
+end
+
+@adjoint! function Base.sync_end(ch::Channel)
+  at = take!(ch)
+  c = grad_mut(__context__, ch)
+  put!(c, at)
+  Base.sync_end(c), _ -> foreach(t -> runadjoint(__context__, t), (at,))
 end
 
 # Make @sync work
--- a/src/lib/base.jl
+++ b/src/lib/base.jl
@@ -74,7 +74,7 @@ end
   t = Task(f)
   t.code = function ()
     y, back = _pullback(__context__, f)
-    cache(__context__)[t] = Task(() -> back(1.f0))
+    cache(__context__)[t] = Task(() -> back(sensitivity(y)))
     return y
   end
   t, _ -> begin
--- a/src/lib/base.jl
+++ b/src/lib/base.jl
@@ -58,7 +58,7 @@ grad_mut(ch::Channel) = Channel(ch.sz_max)
 @adjoint! function put!(ch::Channel, x)
   put!(ch, x), function (ȳ)
     x̄ = grad_mut(__context__, ch)
-    dx = !isopen(x̄) ? nothing : take!(x̄)
+    dx = isopen(x̄) ? take!(x̄) : nothing
     (nothing, accum(dx, ȳ), nothing)
   end
 end
--- a/src/lib/base.jl
+++ b/src/lib/base.jl
@@ -57,7 +57,8 @@ grad_mut(ch::Channel) = Channel(ch.sz_max)
 
 @adjoint! function put!(ch::Channel, x)
   put!(ch, x), function (ȳ)
-    x̄ = take!(grad_mut(__context__, ch))
+    x̄ = grad_mut(__context__, ch)
+    x̄ = x̄ isa Channel ? isready(x̄) ? take!(x̄) : nothing : x̄
     (nothing, accum(x̄, ȳ), nothing)
   end
 end
@@ -73,10 +74,13 @@ end
   t = Task(f)
   t.code = function ()
     y, back = _pullback(__context__, f)
-    cache(__context__)[t] = Task(back)
+    cache(__context__)[t] = Task(() -> back(1.f0))
     return y
   end
-  t, _ -> fetch(cache(__context__)[t])
+  t, _ -> begin
+    istaskdone(cache(__context__)[t]) || schedule(cache(__context__)[t])
+    fetch(cache(__context__)[t])
+  end
 end
 
 function runadjoint(cx, t, ȳ = nothing)
@@ -98,7 +102,10 @@ end
 end
 
 @adjoint! function Base.sync_end(refs)
-  Base.sync_end(refs), _ -> foreach(t -> runadjoint(__context__, t), refs)
+  Base.sync_end(refs), _ -> begin
+    foreach(t -> runadjoint(__context__, t), refs)
+    return
+  end
 end
 
 # Make @sync work
--- a/src/lib/base.jl
+++ b/src/lib/base.jl
@@ -58,7 +58,7 @@ grad_mut(ch::Channel) = Channel(ch.sz_max)
 @adjoint! function put!(ch::Channel, x)
   put!(ch, x), function (ȳ)
     x̄ = grad_mut(__context__, ch)
-    dx = isready(x̄) ? take!(x̄) : x̄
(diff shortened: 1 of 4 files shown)


DhairyaLGandhi:
The issue seems to be in differentiating `T[i for i in x]`. The definition for `getindex(::Tuple, ::AbstractVector)` changed (ref https://github.com/JuliaLang/julia/pull/31138) in Julia 1.5, to be

```julia
getindex(t::Tuple, r::AbstractArray{<:Any,1}) = (eltype(t)[t[ri] for ri in r]...,)
```

While the generator itself doesn't do mutation, this form does allocate and mutate, and hence the error.

Adding an adjoint (should be an `rrule`, but not sure if its exactly needed right now considering this was using the compiler directly to handle iteration) is a bit safer considering Base can change this on us again.

DhairyaLGandhi:
Locally this resulted in a hang, but that is orthogonal to this PR, it has to do with handling `Channel`s and `Threads` which seems to have changed with 1.5 as well

DhairyaLGandhi:
The other case is of a weird mixing differentiating the `Channel`s. In Julia 1.5, the refs used to manage the `sync_end` function is now a `Channel` as opposed to a `Vector{Any}`. This fetches prematurely some time and locks up. This is fixed, but there is currently an incorrect result. Likely due to accumulating twice over the same channel.

DhairyaLGandhi:
@willtebbutt thoughts? We can move them to `rrule`s but this can unblock the CI for now.

willtebbutt (commented):
This is a really nice fix.

While I'm fairly confident that the `getindex` fix makes sense and is appropriately tested via the existing tests, I'm not comfortable reviewing the parallelism-related code as it's an area of the Julia ecosystem that I'm fairly unfamiliar with. Is there someone else that could take a look at that?

DhairyaLGandhi:
I think the parallelism part is correct-but-not-optimal (specifically, I duplicating the `Channel` makes me uncomfortable somewhat). @oxinabox what would be a better way of expressing this?

oxinabox:
I will look tomorrow.

src/lib/base.jl
```diff
@@ -57,7 +57,8 @@ grad_mut(ch::Channel) = Channel(ch.sz_max)
 
 @adjoint! function put!(ch::Channel, x)
   put!(ch, x), function (ȳ)
-    x̄ = take!(grad_mut(__context__, ch))
+    x̄ = grad_mut(__context__, ch)
+    x̄ = isready(x̄) ? take!(x̄) : x̄
```
oxinabox, line 61: This need a comment.
I don't understand why we would do this.
`x̄ ` is a`Channel`
`take!` on a `Channel` blocks if not ready.

Are we just doing it early if it is already ready?
Don't was just absolutely have to do this before `accum` completes?
So `accum` must  do blocking `take!`?


(Also maybe not great to reassign a local variable with a different type. Mostly the optimizer can fix it, but somehting it results in being unstable. More importantly it is harder to read. I just never reassign variables :grin:)

src/lib/base.jl
```diff
@@ -101,6 +102,22 @@ end
...
+    put!(c, i)
+    put!(b, i)
+  end
+  Base.sync_end(c), _ -> begin
+    while !isempty(b)
+      t = take!(b)
+      runadjoint(__context__, t)
+    end
+  end
+end
```
oxinabox, line 119: Can we have meaningful variable names?
I have no ide what any of these variables mean, and thus this is hard to reivew.

src/lib/lib.jl
```diff
@@ -112,6 +112,14 @@ end
 @adjoint getindex(xs::NTuple{N,Any}, r::AbstractUnitRange) where N =
   (xs[r], Δ -> (ntuple(j -> j in r ? Δ[findfirst(isequal(j), r)] : nothing, Val(N)), nothing))
 
+@adjoint function getindex(xs::NTuple{N,Any}, r::AbstractVector) where N
+  val = xs[r]
+  function back(Δ)
+    (ntuple(x -> x in r ? count(isequal(x), r) * Δ[findfirst(isequal(x), r)] : nothing, Val(N)), nothing)
```
oxinabox, line 118: Faster, and IMO clearer:

```suggestion
    x̄s = ntuple(Val(length(xs))) do x
        total = zero(eltype(Δ))
        for r_i in eachindex(r)
            r[r_i] === x || continue
            total += Δ[r_i]
        end
        return total
    end
    return (x̄s, nothing)
```

Benchmarks:
```julia
function back_orig(xs, r)
    Δ = ones(length(xs[r]))  # for testing
    (ntuple(x -> x in r ? count(isequal(x), r) * Δ[findfirst(isequal(x), r)] : nothing, Val(length(xs))), nothing)
end

function back_neo(xs, r)
    Δ = ones(length(xs[r]))  # for testing
    dxs = ntuple(Val(length(xs))) do x
        total = zero(eltype(Δ))
        for r_i in eachindex(r)
            r[r_i] === x || continue
            total += Δ[r_i]
        end
        return total
    end
    return (dxs, nothing)
end


const xs1 = (10, 20, 30, 40, 50);
const r1 = [1,1,2,1,3,2,3,4,5,1];
```

```julia
julia> @btime back_neo(xs1, r1)
  391.060 ns (3 allocations: 416 bytes)
((4.0, 2.0, 2.0, 1.0, 1.0), nothing)

julia> @btime back_orig(xs1, r1)
  591.661 ns (10 allocations: 592 bytes)
((4.0, 2.0, 2.0, 1.0, 1.0), nothing)
```
```

Comment (oxinabox) [reply src/lib/lib.jl:118]: Mostly I wrote the above to convince myself that what this code was doing even made sense.
Does it have a test?

File: src/lib/base.jl (hunk 2 above)
Comment (DhairyaLGandhi) [reply src/lib/base.jl:119]: Oh sure, yes, that should have been caught sooner.

File: src/lib/base.jl (hunk 1 above)
Comment (DhairyaLGandhi) [reply src/lib/base.jl:61]: Reassigning variables is taken.

The issue is that we were taking from a consumed channel at one point.

File: src/lib/lib.jl (hunk 3 above)
Comment (DhairyaLGandhi) [reply src/lib/lib.jl:118]: Should be a function. The existing tests might be plenty in this case. I'm happy to take this suggestion too. It's a bit more verbose but I think that's alright.

Comment (oxinabox) [reply src/lib/lib.jl:118]: The suggestion is for a function.
I just didn't make it a function for the benchmark

File: src/lib/base.jl (hunk 1 above)
Comment (oxinabox) [reply src/lib/base.jl:61]: Shouldn't that error?

Comment (DhairyaLGandhi) [reply src/lib/base.jl:61]: It blocks iirc, since it may have something put back into it at a later time.

Comment (oxinabox) [reply src/lib/base.jl]: `accum(::Channel, ȳ)`  isn't defined, is it?
If we don't take from it we can't do the next line

Comment (DhairyaLGandhi) [reply src/lib/base.jl]: It falls back to `accum(<:Any, ::Nothing)` and moves on from there. The change in behaviour of using a channel for syncing was from 1.5. I wanted to use the `taskref` kwarg but for `@async` we don't control how that's constructed, so it's a bit awkward to get the references back.

If a channel is closed it shouldn't have a "partial" for it anyway.

Comment (oxinabox) [reply src/lib/base.jl]: I am still not following.
what makes it certain that `ȳ===nothing` just because it isn't ready?
If that is the case why do we check if we can take? why not just check if `ȳ===nothing` ?

Comment (DhairyaLGandhi) [reply src/lib/base.jl]: Missed this completely. Is there an `isclosed` on Channels? Further, there is a case for whether it can be taken from later, so its awkward to decide when to wait for a Channel to be populated vs when it will block endlessly.

We aren't certain that `ȳ===nothing`, we can add a check here. Gradients for Channels themselves would be a `nothing`/ `DoesNotExist` right? The elements of the Channels themselves can have the expected gradients, but the Channel itself is needed only for the references to the elements. The problem is that if we `take!` here and the channel isn't populated we will block.

Comment (oxinabox) [reply src/lib/base.jl]: > Missed this completely. Is there an isclosed on Channels?

There is `isopen`

>  Gradients for Channels themselves would be a nothing/ DoesNotExist right? 

I don't know, My instincts say that  `Channels` are basically the same as `Vector`, and so the gradient of a channel is another channel with the gradients of all the elements.
But it seems we don't do that.
So I don't really get this code, i clearly do not have enough of picture of what is going on.

**Comment by CarloLucibello** [comment]
@DhairyaLGandhi @oxinabox I think we should merge this as it is, and open an issue if any bugs in channel's handling are left. We have been with broken CI for too long

**Review by CarloLucibello — APPROVED** [review]

**Comment by CarloLucibello** [comment]
@DhairyaLGandhi bump

**Comment by oxinabox** [comment]
> @DhairyaLGandhi @oxinabox I think we should merge this as it is, and open an issue if any bugs in channel's handling are left. We have been with broken CI for too long

We need to resolve this https://github.com/FluxML/Zygote.jl/pull/772#discussion_r496649961.
Either it is wrong, or it needs comments explaining why it is right, or both.

Having been working on some of the Zygote internals a fair bit lately,
touching the places where it does stuff that doesn't error, but is conceptually nonsense have been some of the all time biggest time sinks.

File: src/lib/base.jl (hunk 1 above)
Comment (DhairyaLGandhi) [reply src/lib/base.jl]: Cool, will update it accordingly.

I agree that that makes sense.  We can have it capture all the pullbacks in a different channel which we can take from if it is populated.

File: src/lib/base.jl [hunk 4]
```diff
@@ -58,7 +58,7 @@ grad_mut(ch::Channel) = Channel(ch.sz_max)
 @adjoint! function put!(ch::Channel, x)
   put!(ch, x), function (ȳ)
     x̄ = grad_mut(__context__, ch)
-    dx = isready(x̄) ? take!(x̄) : x̄
+    dx = !isopen(x̄) ? nothing : take!(x̄)
```
oxinabox, line 61:
```suggestion
    dx = isopen(x̄) ? take!(x̄) : nothing
```

oxinabox (approved):
Makes sense to me now

PR #777, opened 2020-08-28
Patch map bug

willtebbutt:
Fixes a bug in which when the input to map is a vector, but the adjoint is a `Tuple` the reversal doesn't happen.

This is a patch rather than a proper fix -- really this situation shouldn't ever occur.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.5.5"
+version = "0.5.6"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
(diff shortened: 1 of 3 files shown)


DhairyaLGandhi:
We prolly need to keep the tests around.

DhairyaLGandhi:
Note, #777 reverses some of the `reverse` work here.

willtebbutt:
> We prolly need to keep the tests around.

😂 

> Note, dg/stateful_bcast reverses some of the reverse work here.this is needed for the lstm issue in Flux.

Could you point me towards that?

DhairyaLGandhi:
https://github.com/FluxML/Zygote.jl/blob/21f07b450ddfe2e4ba6196b3edd8ed08f656e5e3/src/lib/array.jl#L191

willtebbutt:
Okay, that looks like a nice change. The tests I've added here would be handy in that though, so it might make sense to go ahead and merge this.

This appears to be passing on 1.3. How are we going about merging stuff at the minute with master broken on 1.5?

DhairyaLGandhi:
Currently I have not updated bors to Julia 1.5, although you might want to check against https://github.com/FluxML/Zygote.jl/pull/772. I want to make this backwards compatible and fix the incorrect gradient.

DhairyaLGandhi (approved):
bors r+

PR #781, opened 2020-09-03
Test that keyword arguments are correct

ChrisRackauckas:
Because https://github.com/JuliaDiff/ChainRules.jl/issues/257 shows that there are rules that can make it not correct.


DhairyaLGandhi:
Will merge after ChainRules is tagged

DhairyaLGandhi:
bors try

Issue #221, opened 2019-06-07
Gradient doesn't work properly with `enumerate`, `zip`


avik-pal:
`ERROR: Variable %56 in block 4 is not defined.` when using enumerate, zip.

MWE: (on julia-1.2rc1, IRTools master, Zygote master)
```julia
s = rand(10)
d = rand(10)

Zygote.gradient(s) do
       sums = 0
       for (a, b) in zip(s, d)
            sums += a + b
       end
       return sums
end
```

cc @dhairyagandhi

avik-pal:
This causes trouble with a lot of Base functions like:
```julia
a = rand(10, 10)
Zygote.gradient(Params([a])) do
    return sum(reduce(hcat, a))
end
```

Error
```julia
ERROR: Compiling Tuple{typeof(Base._typed_hcat),Type{Float64},Array{Array{Float64,1},1}}: DimensionMismatch("dimensions must match")
Stacktrace:
 [1] promote_shape at ./indices.jl:154 [inlined]
 [2] _promote_shape at ./iterators.jl:317 [inlined]
 [3] axes at ./iterators.jl:316 [inlined]
 [4] map at ./tuple.jl:140 [inlined]
 [5] axes at ./iterators.jl:316 [inlined]
 [6] _array_for(::Type{Array{IRTools.Variable,1}}, ::Base.Iterators.Zip{Tuple{Array{Any,1},Base.Iterators.Zip{Tuple{Array{Any,1},Array{Any,1}}}}}, ::Base.HasShape{1}) at ./array.jl:598
 [7] collect(::Base.Generator{Base.Iterators.Zip{Tuple{Array{Any,1},Base.Iterators.Zip{Tuple{Array{Any,1},Array{Any,1}}}}},getfield(IRTools, Symbol("##126#132"))}) at ./array.jl:611
 [8] prune!(::IRTools.IR) at /home/avik-pal/.julia/packages/IRTools/L1jyI/src/passes/passes.jl:121
 [9] |> at ./operators.jl:854 [inlined]
 [10] #IR#23(::Bool, ::Bool, ::Type{IRTools.IR}, ::IRTools.Meta) at /home/avik-pal/.julia/packages/IRTools/L1jyI/src/ir/wrap.jl:195
 [11] IR at /home/avik-pal/.julia/packages/IRTools/L1jyI/src/ir/wrap.jl:190 [inlined]
[...]

PR #785, opened 2020-09-08
Iterators. Product, Filter, enumerate

mcabbott:
Could perhaps use more testing & tests. 

Iterators.Flatten is a bit harder.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -165,7 +165,7 @@ function unzip(tuples)
 end
 
 # Reverse iteration order when ∇map is applied to vector,
-# needed for stateful functions. 
+# needed for stateful functions.
 # See https://github.com/FluxML/Flux.jl/issues/1209
 # Should be generalized to abstract array, but reverse takes a dims keyword there
 _tryreverse(m, backs, Δ) = backs, Δ
@@ -184,7 +184,7 @@ for (mapfunc,∇mapfunc) in [(:map,:∇map),(:pmap,:∇pmap),(:vmap,:∇vmap)]
       ys, backs = unzip(ys_and_backs)
       ys, function (Δ)
         # Apply pullbacks in reverse order. Needed for correctness if `f` is stateful.
-        Δf_and_args_zipped = $mapfunc((f, δ) -> f(δ), _tryreverse($mapfunc, backs, Δ)...) 
+        Δf_and_args_zipped = $mapfunc((f, δ) -> f(δ), _tryreverse($mapfunc, backs, Δ)...)
         Δf_and_args = unzip(_tryreverse($mapfunc, Δf_and_args_zipped))
         Δf = reduce(accum, Δf_and_args[1])
         (Δf, Δf_and_args[2:end]...)
@@ -219,6 +219,32 @@ end
         dx[t] .= Δ
         (nothing, dx)
     end
+end
+
+# Iterators
+
+@adjoint enumerate(xs) = enumerate(xs), diys -> (map(last, diys),)
+
+@adjoint function Iterators.Filter(f, x)
+  b = map(f, x)
+  Iterators.Filter(f, x), dy -> begin
+    dx = similar(x, eltype(dy)) .= 0
+    dx[b] .= dy
+    (nothing, dx,)
+  end
+end
+
+_ndims(::Base.HasShape{d}) where {d} = d
+_ndims(x) = Base.IteratorSize(x) isa Base.HasShape ? _ndims(Base.IteratorSize(x)) : 1
+
+@adjoint function Iterators.product(xs...)
+  d = 1
+  Iterators.product(xs...), dy -> ntuple(length(xs)) do n
+    nd = _ndims(xs[n])
+    dims = ntuple(i -> i<d ? i : i+nd, ndims(dy)-nd)
+    d += nd
+    reshape(sum(y->y[n], dy; dims=dims), axes(xs[n]))
+  end
 end
 
 # Reductions
--- a/test/features.jl
+++ b/test/features.jl
@@ -361,6 +361,24 @@ end
     @test (x->10*(x => 2)[2])'(100) === 0
 end
 
+@testset "Iterators" begin
+  @test gradient(1:5) do xs
+    sum([x^i for (i,x) in enumerate(xs)])
+  end == ([1, 4, 27, 256, 3125],)
+
+  @test gradient(2:9) do xs
+    sum([x^2 for x in xs if iseven(x)])  # Iterators.Filter
+  end == ([4, 0, 8, 0, 12, 0, 16, 0],)
+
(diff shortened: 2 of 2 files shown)


tvieijra:
I tried using the adjoint of the product iterator because I needed a way to construct a 2D matrix.  I noticed that the code returns an error when one of the indices of the product iterator is not used as a value for the elements of the iterator.  For example, the following minimal example yields an error:
```
using Zygote
f(A) = sum([A[i,j] for i in 1:3, j in 1:3])
gr = f'(ones(3,3))
```
```
ERROR: MethodError: no method matching getindex(::Nothing, ::Int64)
Stacktrace:
 [1] (::Zygote.var"#555#559"{Int64})(::Nothing) at /home/tom/.julia/environments/v1.5/dev/Zygote/src/lib/array.jl:244
 [2] _mapreduce(::Zygote.var"#555#559"{Int64}, ::typeof(Base.add_sum), ::IndexLinear, ::Array{Nothing,2}) at ./reduce.jl:408
 [3] _mapreduce_dim(::Function, ::Function, ::NamedTuple{(),Tuple{}}, ::Array{Nothing,2}, ::Colon) at ./reducedim.jl:318
 [4] #mapreduce#620 at ./reducedim.jl:310 [inlined]
 [5] mapreduce at ./reducedim.jl:310 [inlined]
 [6] _sum(::Function, ::Array{Nothing,2}, ::Colon) at ./reducedim.jl:724
 [7] #sum#628 at ./reducedim.jl:720 [inlined]
 [8] sum at ./reducedim.jl:720 [inlined]
 [9] _reducedim_init(::Zygote.var"#555#559"{Int64}, ::typeof(Base.add_sum), ::typeof(zero), ::typeof(sum), ::Array{Nothing,2}, ::Tuple{Int64}) at ./reducedim.jl:121
 [10] reducedim_init(::Function, ::typeof(Base.add_sum), ::Array{Nothing,2}, ::Tuple{Int64}) at ./reducedim.jl:109
 [11] _mapreduce_dim(::Function, ::Function, ::NamedTuple{(),Tuple{}}, ::Array{Nothing,2}, ::Tuple{Int64}) at ./reducedim.jl:324
 [12] #mapreduce#620 at ./reducedim.jl:310 [inlined]
 [13] _sum at ./reducedim.jl:746 [inlined]
 [14] #sum#628 at ./reducedim.jl:720 [inlined]
 [15] (::Zygote.var"#553#557"{Array{Nothing,2},Tuple{UnitRange{Int64},UnitRange{Int64}}})(::Int64) at /home/tom/.julia/environments/v1.5/dev/Zygote/src/lib/array.jl:244
 [16] ntuple(::Zygote.var"#553#557"{Array{Nothing,2},Tuple{UnitRange{Int64},UnitRange{Int64}}}, ::Int64) at ./ntuple.jl:18
 [17] (::Zygote.var"#552#556"{Tuple{UnitRange{Int64},UnitRange{Int64}}})(::Array{Nothing,2}) at /home/tom/.julia/environments/v1.5/dev/Zygote/src/lib/array.jl:240
 [18] (::Zygote.var"#2633#back#560"{Zygote.var"#552#556"{Tuple{UnitRange{Int64},UnitRange{Int64}}}})(::Array{Nothing,2}) at /home/tom/.julia/packages/ZygoteRules/6nssF/src/adjoint.jl:49
 [19] f at ./REPL[4]:1 [inlined]
 [20] (::typeof(∂(f)))(::Float64) at /home/tom/.julia/environments/v1.5/dev/Zygote/src/compiler/interface2.jl:0
 [21] (::Zygote.var"#41#42"{typeof(∂(f))})(::Float64) at /home/tom/.julia/environments/v1.5/dev/Zygote/src/compiler/interface.jl:40
 [22] gradient(::Function, ::Array{Float64,2}) at /home/tom/.julia/environments/v1.5/dev/Zygote/src/compiler/interface.jl:49
 [23] (::Zygote.var"#43#44"{typeof(f)})(::Array{Float64,2}) at /home/tom/.julia/environments/v1.5/dev/Zygote/src/compiler/interface.jl:52
 [24] top-level scope at REPL[5]:1
```
The following example, which just uses the iterator indices in the elements in a trivial way, does work however.
```
using Zygote
f(A) = sum([(i*j*A[i,j])/(i*j) for i in 1:3, j in 1:3])
gr = f'(ones(3,3))
```
```
3×3 Array{Float64,2}:
 1.0  1.0  1.0
 1.0  1.0  1.0
 1.0  1.0  1.0
```

mcabbott:
Oh I see, thanks for the example. I haven't got around to looking at this again but it probably wouldn't be hard to fix.

DhairyaLGandhi:
Can you suggest if #801 needs to handle more cases of Zip?

mcabbott:
When last I looked, that one failed the test here, which has arrays of unequal length.

DhairyaLGandhi:
Good idea, although we should have `unzip` handle these cases gracefully rather than have it in the adjoint, so its cleaner to implement.

darsnack:
What's the status of this PR?

oschulz:
Just ran into "Need an adjoint for constructor Base.Iterators.ProductIterator" myself - would love to see this merged if/when possible.

mcabbott:
Status is that many things work, but every time I fiddle with this, I find more. 

`map` also failed to handle inputs of differing length, but should be fixed. They share some code. 

The example from https://github.com/FluxML/Zygote.jl/issues/221#issue-453450622 is in fact fixed, once you fix its bugs:
```julia
julia> d = rand(7)         # gradient w.r.t. d is not tracked
       Zygote.gradient(rand(11)) do s
              sums = 0
              for (a, b) in zip(s, d)
                   sums += 13a + 17b
              end
              return sums
       end
([13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 13.0, 0.0, 0.0, 0.0, 0.0],)

julia> gradient(rand(5)) do  # like @avik-pal's original, no argument, expect an error...
              sums = 0
              for a in d     # ... even without a zip, because do takes no arguments.
                   sums += a
              end
              return sums
       end
ERROR: MethodError: no method matching (::var"#55#56")(::Vector{Float64})
```
What doesn't work are some uses inside generators, these two are marked `@test_broken` right now. I don't like silent errors but they are the same as before this PR:
```julia
julia> gradient(10:14, 1:10) do xs, ys
           sum(x/y for (x,y) in zip(xs, ys))
         end
(nothing, nothing)

julia> gradient(1:10, 3:7) do xs, ys
           sum(x^2+y for x in xs, y in ys)
         end 
(nothing, nothing)
```
And also, some nested uses won't work, like this one from the first message above:
```julia
julia> Zygote.gradient(rand(3)) do xs
           sum([x^i+y for (i,x) in enumerate(xs), y in xs if i>2]) 
       end
```
And according to CI, the last iteration breaks `(((sin')')')(1.0)`, I don't know why.

PR #787, opened 2020-09-10
Fix up Euclidean distance at the origin

willtebbutt:
(Hopefully) fixes some Euclidean distance issues when the inputs are close together.

Fixes https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/166

@molet could you try this out and see whether or not it fixes your issue?

I've introduced `FiniteDifferences` as a test dep because Zygote's own finite differencing isn't sufficiently accurate in this case. I'm not completely sure why.

### DIFF
--- a/src/lib/distances.jl
+++ b/src/lib/distances.jl
@@ -50,7 +50,7 @@ end
   D = x .- y
   δ = sqrt(sum(abs2, D))
   function euclidean(Δ::Real)
-    x̄ = (Δ / (δ + eps(typeof(δ)))) .* D
+    x̄ = (Δ / max(δ, eps(typeof(δ)))) .* D
     return x̄, -x̄
   end
   return δ, euclidean
@@ -59,7 +59,7 @@ end
 @adjoint function colwise(s::Euclidean, x::AbstractMatrix, y::AbstractMatrix)
   d = colwise(s, x, y)
   return d, function (Δ::AbstractVector)
-    x̄ = (Δ ./ (d .+ eps(eltype(d))))' .* (x .- y)
+    x̄ = (Δ ./ max.(d, eps(eltype(d))))' .* (x .- y)
     return nothing, x̄, -x̄
   end
 end
@@ -68,7 +68,7 @@ end
   D, back = pullback((X, Y) -> pairwise(SqEuclidean(), X, Y; dims=dims), X, Y)
   D .= sqrt.(D)
   function pairwise_pullback(Δ)
-    return (nothing, back(Δ ./ (2 .* (D .+ 1000 * eps(eltype(D)))))...)
+    return (nothing, back(Δ ./ (2 .* max.(D, eps(eltype(D)))))...)
   end
   return D, pairwise_pullback
 end
@@ -77,7 +77,7 @@ end
   D, back = pullback(X -> pairwise(SqEuclidean(), X; dims = dims), X)
   D .= sqrt.(D)
   return D, function(Δ)
-    Δ = Δ ./ (2 .* D .+ eps(eltype(D)))
+    Δ = Δ ./ (2 .* max.(D, eps(eltype(D))))
     Δ[diagind(Δ)] .= 0
     return (nothing, first(back(Δ)))
   end
--- a/src/lib/distances.jl
+++ b/src/lib/distances.jl
@@ -65,12 +65,17 @@ end
 end
 
 @adjoint function pairwise(::Euclidean, X::AbstractMatrix, Y::AbstractMatrix; dims=2)
-  D, back = pullback((X, Y) -> pairwise(SqEuclidean(), X, Y; dims=dims), X, Y)
-  D .= sqrt.(D)
-  function pairwise_pullback(Δ)
-    return (nothing, back(Δ ./ (2 .* max.(D, eps(eltype(D)))))...)
+
+  # Modify the forwards-pass slightly to ensure stability on the reverse.
+  function _pairwise_euclidean(X, Y)
+    δ = eps(promote_type(eltype(X), eltype(Y)))^2
+    return sqrt.(max.(pairwise(SqEuclidean(), X, Y; dims=dims), δ))
   end
-  return D, pairwise_pullback
+  D, back = pullback(_pairwise_euclidean, X, Y)
+
+  return D, function(Δ)
+    return (nothing, back(Δ)...)
+  end
 end
 
 @adjoint function pairwise(::Euclidean, X::AbstractMatrix; dims=2)
--- a/src/lib/distances.jl
+++ b/src/lib/distances.jl
@@ -26,17 +26,8 @@ end
 
 ∇pairwise(s, x, y, f) =
   function(Δ)
-    # println("In SqEuclidean")
-    # display(Δ)
-    # println()
     x̄ = 2 .* (x * Diagonal(vec(sum(Δ; dims=2))) .- y * transpose(Δ))
     ȳ = 2 .* (y * Diagonal(vec(sum(Δ; dims=1))) .- x * Δ)
-    # println("x̄")
-    # display(x̄)
-    # println()
-    # println("ȳ")
-    # display(ȳ)
-    # println()
     return (nothing, f(x̄), f(ȳ))
   end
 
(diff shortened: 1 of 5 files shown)


src/lib/distances.jl
```diff
@@ -59,26 +59,25 @@ end
...
-  D, back = pullback(
-    (X, Y) -> pairwise(SqEuclidean(), X, Y; dims = dims),
-    X,
-    Y,
-  )
+  D, back = pullback((X, Y) -> pairwise(SqEuclidean(), X, Y; dims=dims), X, Y)
   D .= sqrt.(D)
-  return D, Δ -> (nothing, back(Δ ./ (2 .* D))...)
+  function pairwise_pullback(Δ)
+    return (nothing, back(Δ ./ (2 .* (D .+ 1000 * eps(eltype(D)))))...)
```
molet, line 71: why is there 1000?

src/lib/distances.jl
```diff
@@ -47,10 +47,10 @@ end
   end
 
 @adjoint function (::Euclidean)(x::AbstractVector, y::AbstractVector)
-  D = x.-y
+  D = x .- y
   δ = sqrt(sum(abs2, D))
   function euclidean(Δ::Real)
-    x̄ = (Δ / δ) .* D
+    x̄ = (Δ / (δ + eps(typeof(δ)))) .* D
```
molet, line 53: I'm wondering whether it would be better to apply epsilon only for the zero case.
Something like this is in my mind:
```julia
δ = sqrt(sum(abs2, D))
δ = max(δ, eps(typeof(δ)))
```

src/lib/distances.jl
willtebbutt: Seemed to make FiniteDifferences and this implementation line up best, but of course it could be that `FiniteDifferences` is also wrong near the origin...

src/lib/distances.jl
willtebbutt: Hmm yeah, that's also a good point. Will this not make a negligible difference once we're away from `δ == 0` though?

src/lib/distances.jl
willtebbutt: To be blunt, it was hard to get tests to pass otherwise. I'm really just not sure how to get good ground truth values to test against.

src/lib/distances.jl
molet: It will provide the correct value until `δ >= eps(typeof(δ)))` and yes it has a negligible effect for `δ < eps(typeof(δ)))`. My main concern with the current implementation is that for small but not negligible differences (compared to machine epsilon) `δ + eps(typeof(δ)))` can already lead to a significant deviation:

```julia
x = [1.0]; y = [1.0 + 1.0e-12]
D = x .- y
δ = sqrt(sum(abs2, D))

D ./ (max(δ, eps(typeof(δ))))
1-element Array{Float64,1}:
 -1.0

D ./ (δ + eps(typeof(δ)))
1-element Array{Float64,1}:
 -0.9997780244173141
```

molet:
> Fixes [JuliaGaussianProcesses/KernelFunctions.jl#166](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/166)

> @molet could you try this out and see whether or not it fixes your issue?

Yes, it will fix the issue above:

```julia
julia> x = rand(1, 10)
julia> k_se(θ) = TransformedKernel(SEKernel(), ScaleTransform(θ))
julia> k_exp(θ) = TransformedKernel(ExponentialKernel(), ScaleTransform(θ))
julia> θ = 2.0

julia> for k in [k_se, k_exp]
           obj(θ) = sum(kernelmatrix(k(θ), x, x[:, 1:1]; obsdim = 2))
           println((obj(θ), Zygote.gradient(obj, θ)))
       end
(8.579178644842253, (-1.1450487011851709,))
(7.600532275422401, (-0.9108241610417317,))
```

src/lib/distances.jl
willtebbutt: I see. Will make the change.

willtebbutt: @molet I've made the change. One of the tests is still failing and doing something weird. It's the only test currnetly enabled. Would you mind taking a look and seeing if I'm doing something obviously silly?

molet: Hm, your change LGTM - let me check the failing test.

molet: Ok, this is crazy: I have spent almost 2 hours to find why on earth the test is failing - but I still could not figure out.
My suggestion is to replace the current function with something like below as this one works well:
```julia
@adjoint function pairwise(::Euclidean, X::AbstractMatrix, Y::AbstractMatrix; dims=2)
    D, back = pullback((X, Y) -> sqrt.(max.(pairwise(SqEuclidean(), X, Y; dims=dims), eps(promote_type(eltype(X), eltype(Y)))^2)), X, Y)
    return D, function(Δ)
        return (nothing, back(Δ)...)
    end
end
```

willtebbutt: Thanks for battling through this -- this looks like a good solution to me.

willtebbutt:
@DhairyaLGandhi I think this looks good now. It's already gone through a round of review with @molet who found the problem in the first place and agrees this fixes it. I'm going to be unavailable for the next week or so, so would appreciate if you could handle this from here -- feel free to push to this branch directly if you feel that there are more changes necessary.

test/runtests.jl
```diff
@@ -2,51 +2,51 @@ using Zygote, Test
 using Zygote: gradient
 using CUDA: has_cuda
 
-@testset "Interface" begin
-  include("interface.jl")
-end
+# @testset "Interface" begin
```
molet, line 5: I guess these need to be uncommented?

src/lib/distances.jl
```diff
@@ -47,10 +47,10 @@ end
   end
 
 @adjoint function (::Euclidean)(x::AbstractVector, y::AbstractVector)
-  D = x.-y
+  D = x .- y
   δ = sqrt(sum(abs2, D))
   function euclidean(Δ::Real)
-    x̄ = (Δ / δ) .* D
+    x̄ = (Δ / max(δ, eps(typeof(δ)))) .* D
```
oxinabox, line 53: So the only time ` δ`  is less than `eps(typeof(δ))` is  is `δ==0`, since this is definition of `eps`.
and that can only occur if `D==[0, 0, 0,...]`.
and in that case we get that `x̄ ` iz also zero, of the same size as `D`.

So we can do:
```suggestion
    x̄ = ifelse(iszero(δ), D, (Δ / δ) .* D)
```

src/lib/distances.jl
```diff
@@ -59,26 +59,30 @@ end
 @adjoint function colwise(s::Euclidean, x::AbstractMatrix, y::AbstractMatrix)
   d = colwise(s, x, y)
   return d, function (Δ::AbstractVector)
-    x̄ = (Δ ./ d)' .* (x .- y)
+    x̄ = (Δ ./ max.(d, eps(eltype(d))))' .* (x .- y)
```
oxinabox, line 62: I wonder if we can do a similar trick here

oxinabox (approved):
If you can generalize https://github.com/FluxML/Zygote.jl/pull/787/files#r487137860
then that would be cool.
If not then i think it should be done there still.

Don't forget to uncomment out the tests before merging.

src/lib/distances.jl
molet: Something like this would work:
```julia
(Δ .* map(di->ifelse(iszero(di), di, inv(di)), d))' .* (x .- y)
```

src/lib/distances.jl
```diff
@@ -47,10 +47,10 @@ end
   end
 
 @adjoint function (::Euclidean)(x::AbstractVector, y::AbstractVector)
-  D = x.-y
+  D = x .- y
   δ = sqrt(sum(abs2, D))
   function euclidean(Δ::Real)
-    x̄ = (Δ / δ) .* D
+    x̄ = ifelse(iszero(δ), D, (Δ / δ) .* D)
```
devmotion, line 53: Isn't this mathematically incorrect if `x !== y`? E.g., in the 1D case the derivative of the Euclidean distance `x -> |x - a|` with respect to `x` is not defined for `x = a` since the limit from the left and right do not agree (they are -1 and 1, so also not close to 0).

I understand that this might be unsatisfying for practical applications, also due to floating point issues, but it seems mathematically actually the NaN value resulting from the previous implementation makes sense if `x !== y`.

oxinabox: we tend to use a subgradient-based convention that if gradient limit from the left has opposite sign to the gradient approaching from the right.
Or more generally: if 0 is an element of the subgradient,
then we say the gradient is zero.

This has the benificial effect of ensuring that the gradient is considered to be zero at local minima,
which from perspective of doing gradient based optimisation is what you want.

devmotion: I see :+1: Thanks for the explanation!

devmotion:
It seems there is no new release yet? Would it be possible to make one?

oxinabox:
Will is on holidays,  I guess he forgot to make one.
I will make one now

oxinabox:
Ah Dhairya  is ahead of me
https://github.com/FluxML/Zygote.jl/commit/ed366f32c0f520567526040d9f8acaf0d83613c3#commitcomment-42336366

PR #780, opened 2020-09-03
Move a bunch of no_grad to ChainRules

oxinabox:
this is the partner to https://github.com/JuliaDiff/ChainRules.jl/pull/252
It will fail til that is merged and tagged

What is left is:

- Types (because https://github.com/JuliaDiff/ChainRulesCore.jl/issues/213) (e.g. `Colon`, `OneTo` `Channel`)
- Things to which the derivative is `Zero()` not `DoesNotExist()` (e.g. `one`, `ones`, `zero`, `zeros`)
- Things that felt too magic: e.g. `Base.eval`

Should I bump patch version and tag a release?

PR #788, opened 2020-09-13
fix for rrules with kwargs and multiple zero partials

oxinabox:
This fixes https://github.com/JuliaDiff/ChainRulesCore.jl/issues/216
correcting a bug I introduced in #780 

cc @devmotion 

(I have confirmed that the second test that uses the kwargs fails on Zygore#master


oxinabox:
@willtebbutt can you see this merged and tagged ASAP?
It is breaking peoples code still

willtebbutt:
bors r+

Issue seems to have been fixed.

PR #793, opened 2020-09-16
fix error while broadcasting

gxyd:
Fixes #724

I'll go through the code-flow and see if the current solution is the best one.

**Test case**: I'll figure out how to write the test case for it on me own. I'll need sometime on that.

On this branch:
```jl
julia> Zygote.gradient(x -> sum(x .* (x .== maximum(x, dims=1))), rand(3,3))
([0.0 0.0 0.0; 1.0 1.0 1.0; 0.0 0.0 0.0],)

julia> Zygote.gradient(x -> sum(sum(x .== 0.5, dims=1)), rand(3,3))
(nothing,)

julia> Zygote.gradient(x -> sum(x .== 0.5), rand(3,3))
(nothing,)
```

Issue #724, opened 2020-07-01
Error with broadcasting ==


mcabbott:
```julia
julia> using Zygote

julia> Zygote.gradient(x -> sum(x .* (x .== maximum(x))), rand(3))
([1.0, 0.0, 0.0],)

julia> Zygote.gradient(x -> sum(x .* (x .== maximum(x, dims=1))), rand(3,3))
ERROR: MethodError: no method matching size(::Nothing, ::Int64)
Closest candidates are:
  size(::BitArray{1}, ::Integer) at bitarray.jl:103
  size(::Tuple, ::Integer) at tuple.jl:22
  size(::Number, ::Integer) at number.jl:63
  ...
Stacktrace:
 [1] (::Zygote.var"#1086#1087"{Nothing})(::Int64) at /Users/me/.julia/packages/Zygote/pmW1K/src/lib/broadcast.jl:51
 [2] ntuple at ./ntuple.jl:42 [inlined]
 [3] trim(::Array{Float64,2}, ::Nothing) at /Users/me/.julia/packages/Zygote/pmW1K/src/lib/broadcast.jl:51
 [4] unbroadcast at /Users/me/.julia/packages/Zygote/pmW1K/src/lib/broadcast.jl:53 [inlined]
```
Julia 1.4 and Zygote v0.5.1.

gxyd:
Hi,

Can I try to fix the issue, if noone is working on it? I don't know much Zygote but I wanna give a shot at fixing this error.

gxyd:
BTW, doing the same thing in PyTorch:
```
>>> def func(a):
            return sum(a * (a == torch.max(a, dim=0).values))

>>> x = torch.rand((3, 3), requires_grad=True)
>>> y = func(x)
>>> y.backward()
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-22-ab75bb780f4c> in <module>
----> 1 y.backward()

~/opt/anaconda3/envs/dl_learn/lib/python3.6/site-packages/torch/tensor.py in backward(self, gradient, retain_graph, create_graph)
    193                 products. Defaults to ``False``.
    194         """
--> 195         torch.autograd.backward(self, gradient, retain_graph, create_graph)
    196 
    197     def register_hook(self, hook):

[...]

PR #794, opened 2020-09-16
Correct primal of Matrix from UniformScaling & handle sum of boolean arrays

oxinabox:
Can't do AD on bools.

This was breaking code like:
```
using Distances, Zygote
function o(x)
           d = pairwise(Euclidean(), x, dims=2)
           return sum(iszero, d)
 end

gradient(o, rand(5,3))
````

since the sum was returning an `Array{Nothing}`.
Possibly we should post process those and normalize them into `nothing` more generally.
but to fix this particular case, just making sum of arrays of bools seems like a good start.

I would fix this in ChainRules but because Zygote defines `sum(::AbstractArray)` with a better definition than what ChainRules has right now, that will always take precidence.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -780,17 +780,17 @@ end
   end
 end
 
-@adjoint function Matrix(::UniformScaling, i::Integer, j::Integer)
-  return Matrix(I, i, j), Δ -> ((λ=tr(Δ),), nothing, nothing)
+@adjoint function Matrix(S::UniformScaling, i::Integer, j::Integer)
+  return Matrix(S, i, j), Δ -> ((λ=tr(Δ),), nothing, nothing)
 end
-@adjoint function Matrix(::UniformScaling, ij::NTuple{2, Integer})
-  return Matrix(I, ij), Δ -> ((λ=tr(Δ),), nothing)
+@adjoint function Matrix(S::UniformScaling, ij::NTuple{2, Integer})
+  return Matrix(S, ij), Δ -> ((λ=tr(Δ),), nothing)
 end
-@adjoint function Matrix{T}(::UniformScaling, i::Integer, j::Integer) where {T}
-  return Matrix{T}(I, i, j), Δ -> ((λ=tr(Δ),), nothing, nothing)
+@adjoint function Matrix{T}(S::UniformScaling, i::Integer, j::Integer) where {T}
+  return Matrix{T}(S, i, j), Δ -> ((λ=tr(Δ),), nothing, nothing)
 end
-@adjoint function Matrix{T}(::UniformScaling, ij::NTuple{2, Integer}) where {T}
-  return Matrix{T}(I, ij), Δ -> ((λ=tr(Δ),), nothing)
+@adjoint function Matrix{T}(S::UniformScaling, ij::NTuple{2, Integer}) where {T}
+  return Matrix{T}(S, ij), Δ -> ((λ=tr(Δ),), nothing)
 end
 @adjoint function +(A::AbstractMatrix, S::UniformScaling)
   return A + S, Δ->(Δ, (λ=tr(Δ),))
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -165,7 +165,7 @@ function unzip(tuples)
 end
 
 # Reverse iteration order when ∇map is applied to vector,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/lib/array.jl [hunk 1]
```diff
@@ -231,6 +230,10 @@ end
   end
 end
 
+@adjoint function sum(xs::AbstractArray{Bool}; dims = :)
```
molet, line 233: What is about other operations like `prod`?

oxinabox: Yes, this is not a general solution this just solves the `sum` case.

molet: I see, ok.

molet (approved):
LGTM.

molet:
However, there are some tests failing now.

oxinabox:
Wow the tests are failing for reasons that are not related to this PR,
except for the fact that this PR reveals them.

Zygote breaks the forward pass of `k*I`


```julia
julia> g(x) = Matrix(x*I, 2,2)

julia> g(2.5)
2×2 Array{Float64,2}:
 2.5  0.0
 0.0  2.5

julia> Zygote._pullback(g, 2.5)
(Bool[1 0; 0 1], ∂(g))
```

Which is a much more significant issue.

@mohamed82008 I will put fix in this PR if you could review

oxinabox:
Tests are now only failing because tests fail in general in Julia 1.5

DhairyaLGandhi:
Looks fine to me

mcabbott:
This touches integer promotion rules. Is it obvious we want to treat `Bool` specially? 

The test you are adding `@test gradient(x->sum(x .== 0.0), [1.2, 0.2, 0.0, -1.1, 100.0]) == (nothing,)` is I think for #724, the problem isn't sum, it's the broadcast. #793 is a possible fix, which as you say processes `Array{Nothing}` into `nothing`.

oxinabox:
> This touches integer promotion rules. Is it obvious we want to treat Bool specially?

I think so.
While it is not unusual to treat other integers as being actually embedded floats,
it is unusual to treat bools as.

> the problem isn't sum, it's the broadcast

It's both.
No matter where the array of Bools came from it should not had AD done to it.
But also things that output arrays of bools (like broadcast) should have be returning `nothing` to indicate not differentiable.

mcabbott:
Have you got another example which produces an array of Bools, besides the broadcast? 

In #724's example, the array of bools never reaches `sum`, so it does have to be dealt with earlier. 

Re promotion, I'm not really taking a position, maybe they should be special. Just observing that it's a philosophical step, not a tiny bugfix. I don't think they are a special case elsewhere.

oxinabox:
I would not say this is a tiny bug fix it is a feature enhancement.

But the other fix on the primal of Matrix from uniform scaling is a significant bug fix.
and I only found it because of the other change.

Which is itself an example of a time one might end up summing an array of bools, i guess.
Since our tests were. (though they were only doing so because of a bug this PR fixes)

```
julia> Matrix(I, 2,2)
2×2 Array{Bool,2}:
 1  0
 0  1

julia> sum(Matrix(I, 2,2))
2
```

mcabbott:
Indeed, well found, and there could be more bugs of this class, without sum. Maybe disabling the gradient for things like `exp.(Matrix(I, 2,2))` and `map(>(0), rand(10))` might unearth more.

oxinabox:
The real fix is to test things with ChainRulesTestUtils.jl
Which checks everything properly.

Anyway, I can split this PR if J have to.
Do I have to, or can this be merged?

DhairyaLGandhi:
I think it's fine. We should also merge the updates needed for julia 1.5

CarloLucibello:
> While it is not unusual to treat other integers as being actually embedded floats,
it is unusual to treat bools as.

I think this approach is the most reasonable one. 
I'm merging this as it is better to fix immediately this particular but likely common case

bors r+

oxinabox:
@CarloLucibello 
Not a big deal but ideally you would have approved this PR and then left it to be to trigger Bors to merge it.
Since I have merge rights.

See https://github.com/SciML/ColPrac

 > PRs by people with merge rights should not be merged by people other than the author (just approved).

In case I had some last things left to do before merging.

Idk if I would have caught it in my own final check before merge, but the PR seems to have mistakenly included a submodule from the dev folder.

PR #797, opened 2020-09-22
Fix Euclidean... again

willtebbutt:
The implementation that we recently implemented turns out not to be optimal from a numerical standpoint.

This new implementation actually seems to be pretty good, and passes more stringest tests in all cases.


willtebbutt:
@molet this is ready for review.

willtebbutt:
Note that the test failures are the usual Zygote#master failures that are being addressed in a separate PR.

DhairyaLGandhi:
usually I would put the helpers outside the adjoint, just because its cleaner to read. Keeping it together is good too :)

How about something like

```julia
@adjoint f(...) = pullback(_f, ...)
```

DhairyaLGandhi:
Have we been able to tighten the accuracy issues? Also, is the adjoint fast?

willtebbutt:
> usually I would put the helpers outside the adjoint, just because its cleaner to read. Keeping it together is good too :)

If you're okay keeping them together, I think I would prefer that. The point being that you avoid polluting `Zygote`'s set of functions with stuff from specific adjoints.

> Have we been able to tighten the accuracy issues? Also, is the adjoint fast?

IIRC this resolves the accuracy issues. It should be fast 🤷 . Will check.

willtebbutt:
Does indeed appear to solve the accuracy issue (note the tight constraints in the accuracy tests).

@DhairyaLGandhi this is probably ready for another round of review.

(Sorry for letting this drop for a while -- I've had a deadline last Friday that I've been working towards for the last month or so)

test/gradcheck.jl
```diff
@@ -1129,8 +1132,40 @@ end
...
+        # FiniteDifferences.jl instead.
+        Y = vcat(Xt, Xt)
+
+        Δ = randn(size(Y, 1), size(Y, 1))
+        Δ_fd = FiniteDifferences.j′vp(
+          central_fdm(5, 1), X -> pairwise(metric, X; dims=1), Δ, Y,
+        )
+        _, pb = Zygote.pullback(X -> pairwise(metric, X; dims=1), Y)
+
+        # This is impressively inaccurate, but at least it doesn't produce a NaN.
```
oxinabox, line 1166: what is this comment about (and the one above) still exist, isn't `1e-9` pretty good accuracy?

willtebbutt: 😂  yes, that should be going

CarloLucibello:
after mergeging this we shold consider porting those rules to Distances.jl

oxinabox:
> after mergeging this we shold consider porting those rules to Distances.jl

see this discussion
https://github.com/JuliaStats/Distances.jl/issues/172

CarloLucibello:
what should we do with this PR? Worth rebasing?

willtebbutt:
Was superceded by https://github.com/FluxML/Zygote.jl/pull/1307 , so I'm going to close. Thanks for flagging.

PR #799, opened 2020-09-24
Tweak compat

willtebbutt:
Addresses https://github.com/JuliaRegistries/General/pull/21772#issuecomment-698197723


CarloLucibello:
forwarddiff 0.10.0 dates back to 2018, so I think we are leaving plenty of compatibility here

bors r+

PR #801, opened 2020-09-28
Add adjoint for Zip

### LINKED ISSUE #221: Gradient doesn't work properly with `enumerate`, `zip`

DhairyaLGandhi:
Fixes #221


willtebbutt:
Is there a reason for this to be implemented here, rather than in `ChainRules`?

DhairyaLGandhi:
We already have a bunch of the infrastructure here already, but I agree we should talk about exporting `rrules`

DhairyaLGandhi:
The resolution error in CI is annoying, did anything change in the Reexport package?

src/lib/base.jl
```diff
@@ -137,6 +137,10 @@ end
     return getfield(p, i), pair_getfield
 end
 
+@adjoint function Base.Iterators.Zip(is)
+  Base.Iterators.Zip(is), Δ -> (unzip(Δ),)
+end
```
willtebbutt, line 142:
```suggestion
@adjoint function Base.Iterators.Zip(is)
  Zip_pullback(Δ) = (unzip(Δ),)
  return Base.Iterators.Zip(is), Zip_pullback
end
```
so that the `@code_warntype` is readable.

test/gradcheck.jl
```diff
@@ -180,6 +180,11 @@ end
   @test back([nothing]) == ([nothing for _ in 1:3], nothing)
 end
 
+@testset "Zip" begin
```
willtebbutt, line 183: I'm slightly nervous that there's only one test here, given that we've got the dependence on `_unzip` which needs the `Zip` to be non-empty and for `length` to be defined over the things being iterated over.

It would be good to add some additional tests on objects without a defined length and empty `Zip`s, since we've got no way to exclude them via dispatch.

willtebbutt (commented):
Going to hold off approving because I think that we might need more testing.

DhairyaLGandhi:
Testing is fair, will add a few. I suspect we would end up generalising `unzip` to some extent to support these cases more generally, which seems like a good thing

CarloLucibello:
the issue has been already fixed

PR #805, opened 2020-10-04
remove submodule

CarloLucibello:
accidentally introduced in #794 

@oxinabox is removing the folder enough here to get rid of the submodule?


oxinabox (approved):

oxinabox:
I think so

PR #807, opened 2020-10-14
Add stateful broadcast expressions back

DhairyaLGandhi:
Fixes https://github.com/FluxML/Flux.jl/issues/1209


CarloLucibello:
I'm worried about the performance impact of this. We will need proper benchmarking before merging this

ToucheSir:
#1209 reared its ugly head again in https://github.com/FluxML/Flux.jl/issues/1547. Is this PR still the intended solution for that?

DhairyaLGandhi:
Yeah, this would be the cleaner solution overall

src/lib/array.jl
```diff
@@ -168,12 +168,12 @@ end
...
-_tryreverse(m, backs, Δ) = backs, Δ
-function _tryreverse(m::typeof(map), backs, Δ::Union{AbstractVector, Tuple})
+_tryreverse(backs, Δ) = backs, Δ
+function _tryreverse(backs, Δ::Union{AbstractVector, Tuple})
   return reverse(backs), reverse(Δ)
 end
-_tryreverse(m, x) = x
-_tryreverse(m::typeof(map), x::Union{AbstractVector, Tuple}) = reverse(x)
+_tryreverse(x) = x
+# _tryreverse(x::Union{AbstractVector, Tuple}) = reverse(x)
```
darsnack, line 176: How come we don't reverse the mapped gradients anymore?

mcabbott:
It's not entirely clear that stateful functions should be allowed in broadcasting, since (as far as I know) the order of iteration is whatever the compiler thinks will be efficient, so even the forward results are uncertain. Maybe an error would be ideal.

If we do nevertheless decide to reverse the order for broadcasts, I also think it ought to take a less naive form. Making `f.(ones(3))` behave differently to `f.(ones(3,1))` or `g.(ones(3), 1)` seems pretty surprising.I guess it would take some digging in the broadcast machinery to do properly. And it needs to be done carefully not to kill performance in more common cases.

At present `map` calculates the gradient in reverse order, but only for vectors, not 1-column matrices, and not tuples. That's pretty odd, even if you assume that `map` is going to iterate in order.

DhairyaLGandhi:
We definitely need stateful expressions to return the correct results.

CarloLucibello:
closing as stale

PR #809, opened 2020-10-14
Fixes #808

### LINKED ISSUE #808: Broken methods(Base.show)
```julia
using Zygote
methods(Base.show)
```
is broken, causes
```julia
[97] Error showing value of type Base.MethodList:
ERROR: UndefVarError: S not defined
Stacktrace:
 [1] show(::IOContext{Base.GenericIOBuffer{Array{UInt8,1}}}, ::Type{
SYSTEM (REPL): showing an error caused an error
ERROR: UndefVarError: S not defined
Stacktrace:
 [1] show(::IOContext{REPL.Terminals.TTYTerminal}, ::Type{
SYSTEM (REPL): caught exception of type UndefVarError while trying to handle a nested exception; giving up
```

racinmat:
Fixes #808 
hopefully it's done right.
Could this PR be labeled by `hacktoberfest-accepted`? 
Thanks.

### DIFF
--- a/src/compiler/show.jl
+++ b/src/compiler/show.jl
@@ -10,4 +10,4 @@ end
 
 Base.show(io::IO, j::Pullback{S}) where S = print(io, "∂($(funcname(S.parameters[1])))")
 
-Base.show(io::IO, P::Type{<:Pullback{S}}) where S<:Tuple = print(io, "typeof(∂($(funcname((@isdefined S) ? S.parameters[1] : nothing))))")
+Base.show(io::IO, P::Type{<:Pullback{S}}) where S<:Tuple = print(io, "typeof(∂($(funcname(@isdefined(S) ? S.parameters[1] : nothing))))")
--- a/src/compiler/show.jl
+++ b/src/compiler/show.jl
@@ -10,4 +10,4 @@ end
 
 Base.show(io::IO, j::Pullback{S}) where S = print(io, "∂($(funcname(S.parameters[1])))")
 
-Base.show(io::IO, P::Type{<:Pullback{S}}) where S<:Tuple = print(io, "typeof(∂($(funcname(S.parameters[1]))))")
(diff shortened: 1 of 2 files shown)


src/compiler/show.jl
```diff
@@ -10,4 +10,4 @@ end
 
 Base.show(io::IO, j::Pullback{S}) where S = print(io, "∂($(funcname(S.parameters[1])))")
 
-Base.show(io::IO, P::Type{<:Pullback{S}}) where S<:Tuple = print(io, "typeof(∂($(funcname(S.parameters[1]))))")
+Base.show(io::IO, P::Type{<:Pullback{S}}) where S<:Tuple = print(io, "typeof(∂($(funcname((@isdefined S) ? S.parameters[1] : nothing))))")
```
DhairyaLGandhi, line 13:
```suggestion
Base.show(io::IO, P::Type{<:Pullback{S}}) where S<:Tuple = print(io, "typeof(∂($(funcname(@isdefined(S) ? S.parameters[1] : nothing))))")
```

DhairyaLGandhi (commented):
We should also test this

racinmat (commented):
Any idea for good test? Because I'm afraid this happens only when listing methods, not during calling the function.
Should I just test that `methods(Base.show)` does not crash or is there soemthin better we can do?

racinmat:
Testcase added, I couldn't figure out better way than this, I'm looking for ideas how to test it better.
Btw, could you please change the label to `hacktoberfest-accepted`.
My bad, I misspelled it in my PR.

DhairyaLGandhi:
Yeah perhaps not having it error would be enough. Thanks!

Could you add it in a test set please? Seems like a pretty standalone fix, so differentiating it in the tests

racinmat:
Ok, I changed tests, now it just tests that printed buffer is not empty

racinmat:
@DhairyaLGandhi

racinmat:
@CarloLucibello  is this mergeable or something more to do?

CarloLucibello:
thanks!

bors r+

PR #819, opened 2020-11-04
Fix UnionAll stackoverflow

DhairyaLGandhi:
MWE:

```Julia
c = RGB(0,1,1)
t = RGB(1,1,1)

gradient(colordiff, c, t)
ERROR: StackOverflowError:
Stacktrace:
 [1] unwrap_unionall at ./essentials.jl:234 [inlined]
 [2] _pullback(::Zygote.Context, ::typeof(Base.unwrap_unionall), ::Type{M}) at /Users/dhairyagandhi/Downloads/new_clones/Zygote.jl/src/compiler/interface2.jl:0
 [3] nameof at ./reflection.jl:211 [inlined]
 ... (the last 2 lines are repeated 43620 more times)
 [87244] _pullback(::Zygote.Context, ::typeof(nameof), ::Type{M}) at /Users/dhairyagandhi/Downloads/new_clones/Zygote.jl/src/
```

This pointed me to check why there was a stackoverflow, which really shouldn't happen. In some cases, it crashes Julia entirely.

Turns out `nameof` which `UnionAll` structs can cause this but rewriting the complete function like so differentiates as expected.

```Julia
function mynameof(x::UnionAll)
  while isa(x, UnionAll)
    x = x.body
  end
  x.name.name
end
```

Since `nameof` would not produce gradients anyway, I went ahead and gave it a consistent gradient and added the couple other annotations needed to get the Colors.jl examples functional again.

### DIFF
--- a/src/lib/base.jl
+++ b/src/lib/base.jl
@@ -138,3 +138,7 @@ end
     end
     return getfield(p, i), pair_getfield
 end
+
+@adjoint Base.nameof(x::UnionAll) = nameof(x), _ -> (nothing,)
+
+@nograd typeintersect
--- a/src/Zygote.jl
+++ b/src/Zygote.jl
@@ -51,6 +51,10 @@ include("profiler/Profile.jl")
   include("flux.jl")
 end
 
+@init @require Colors="5ae59095-9a9b-59fe-a467-6f913c188581" begin
+  @nograd Colors.ColorTypes._parameter_upper_bound
+end
+
 precompile() = include(joinpath(@__DIR__, "precompile.jl"))
 
 # helps to work around 265-y issues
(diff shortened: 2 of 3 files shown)


src/lib/base.jl
```diff
@@ -138,3 +138,7 @@ end
     end
     return getfield(p, i), pair_getfield
 end
+
+@adjoint Base.nameof(x::UnionAll) = nameof(x), _ -> (nothing,)
+
+@nograd typeintersect
```
CarloLucibello, line 144: Should these adjoints be moved to ChainRules?

DhairyaLGandhi: I don't see much precedence in ChainRules for these kinds of methods. Is that judgement correct @willtebbutt ?

Either way is fine, of course. We'll need them to emit `rrules` soon.

willtebbutt: We would very much like to have this in `ChainRules` -- it's actually precisely this kind of thing that we want to cover really thoroughly, there's just a lot of cover :)

DhairyaLGandhi:
I am going to merge this here for now, but let's work on making all the rules emit rrules as well

bors r+

PR #823, opened 2020-11-10
Better handling for nesting Params

DhairyaLGandhi:
This allows for differentiating through the Params code to be able to nest gradient calls with `Params` instead of the usual arguments.

```julia
nn = Chain(Dense(3,3), Dense(3,1))
ip = rand(Float32, 3,2)
gradient(ps) do
  y,b = pullback(ps) do
    sum(nn(ip))
  end
  _gs = b(y)
  sum(x -> sum(_gs[x]), ps)
end
```


src/compiler/interface.jl
```diff
@@ -135,7 +135,7 @@ end
 
 Base.show(io::IO, ps::Grads) = print(io, "Grads(...)")
 
-@forward Grads.grads Base.getindex, Base.haskey
+@forward Grads.grads Base.getindex, Base.haskey, Base.iterate, Base.keys
```
darsnack, line 138: Think this needs a rebase because these forwards will conflict with the current master.

src/lib/base.jl
```diff
@@ -44,6 +44,27 @@ end
...
+  end
+end
+@adjoint! function get(d::AbstractDict, k, default)
+  hk = Ref{Bool}()
+  val = if haskey(d, k)
+    hk[] = true
+    d[k]
+  else
+    hk[] = false
+    d[k] = default
```
CarloLucibello, line 59: `get` should not mutate, maybe you want to define the adjoiint for `get!`?

DhairyaLGandhi:
This resolves the conflicts and moves iteration over to the dictionary which more honestly captures gradients from globals. This can break the current assumption of algebra on `Grads` but that's alright since most cases access the params fields directly, and those that don't can be fixed up.

src/lib/base.jl
DhairyaLGandhi: We aren't mutating anything in the user defined objects, so it should be fine. `get!` would still mutate the gradient dictionary so its not any better.

CarloLucibello: you are defining an adjoint for `get(d::AbstractDict, k, default)` that mutates `d`, this is not fine at all

src/compiler/interface.jl
```diff
@@ -171,15 +172,15 @@ const ADictOrGrads = Union{AbstractDict, Grads}
 # Dictionary interface.
 # Don't use the IdDict directly since it may contain some spurious pairs.
 Base.haskey(gs::Grads, x) = x ∈ gs.params 
-Base.keys(gs::Grads) = gs.params
+# Base.keys(gs::Grads) = gs.params
 Base.values(gs::Grads) = (gs.grads[p] for p in gs.params)
```
CarloLucibello, line 176: Forwarding `keys` to gs.grads while relying on gs.params for the values is not good. Either we base the dictionary interface entirely on gs.params or we base it on gs.grads, can't have mixed stuff. 

The comments above
```
# Don't use the IdDict directly since it may contain some spurious pairs
```
suggests that we should think changes thoroughly or leave things as they are (which seems the best option to me)

CarloLucibello:
This needs some tests of the features it adds and for FluxML/Zygote.jl#941 if it actually fixes it

src/compiler/interface.jl
DhairyaLGandhi: What is the spurious stuff you refer to? It contains references to the objects that had gradients along the way.

DhairyaLGandhi:
#941 may well require extra handling. I'd still prefer to solve the underlying problems first.

src/lib/base.jl
DhairyaLGandhi: Ok, I see what you mean

src/compiler/interface.jl
ToucheSir: I think Carlo is referring to the existing comment at https://github.com/FluxML/Zygote.jl/pull/823/files#diff-7511b224d7f3ebb56465690de8e307422e3c9798a22bdd4e960d5c86ba6528aaR173. My understanding of that is that `Base.keys(Grads.grads)` may contain items not in `Base.keys(Grads.params)`. This seems like it should never happen though, so is the comment out of date or am I missing some scenario where it could?

CarloLucibello: the comment is not outdated
```julia
julia> using Flux

julia> m = Chain(Dense(2,2), x->relu.(x), BatchNorm(2)) 
Chain(Dense(2, 2), #1, BatchNorm(2))

julia> gs = gradient(() -> sum(m(rand(2,2))), Flux.params(m))
Grads(...)

julia> gs.grads
IdDict{Any, Any} with 8 entries:
  Float32[0.0, 0.0]                              => [0.0, 0.0]
  BatchNorm(2)                                   => RefValue{Any}((λ = nothing, β = nothing, γ = nothing, μ = nothing, σ² = nothing, ϵ = 0.0, momentum = nothing, affi…
  Float32[-0.63824 0.222623; -0.785237 0.536415] => [0.0 0.0; 0.0 0.0]
  :(Main.m)                                      => (layers = (nothing, nothing, RefValue{Any}((λ = nothing, β = nothing, γ = nothing, μ = nothing, σ² = nothing, ϵ = …
  Box([0.0; 0.0])                                => RefValue{Any}((contents = nothing,))
  Float32[0.0, 0.0]                              => [2.0, 2.0]
  Box([0.0; 0.0])                                => RefValue{Any}((contents = nothing,))
  Float32[1.0, 1.0]                              => [0.0, 0.0]
```

CarloLucibello: we have to keep the current dict interface based on `gs.params`

src/lib/array.jl
```diff
@@ -773,56 +770,47 @@ AbstractFFTs.ifft(x::Fill, dims...) = AbstractFFTs.ifft(collect(x), dims...)
...
 AbstractFFTs.brfft(x::Fill, d, dims...) = AbstractFFTs.brfft(collect(x), d, dims...)
-
 # the adjoint jacobian of an FFT with respect to its input is the reverse FFT of the
 # gradient of its inputs, but with different normalization factor
 @adjoint function fft(xs)
   return AbstractFFTs.fft(xs), function(Δ)
     return (AbstractFFTs.bfft(Δ),)
   end
 end
-
```
CarloLucibello, line 784: ? this empty lines shouldn't be removed

src/lib/base.jl
```diff
@@ -44,6 +44,27 @@ end
   end
 end
 
+@adjoint function Base._oidd_nextind(a, i)
```
CarloLucibello, line 47: do we need to define an adjoint for an internal function of Base?

DhairyaLGandhi:
Fixes https://github.com/FluxML/Zygote.jl/issues/941

CarloLucibello:
This is not fixing #941

ToucheSir:
Yeah, the most direct fix for this is to define `@adjoint`[`Base.push!(ps::Params, x...)`](https://github.com/FluxML/Zygote.jl/blob/f8191cafe1b99444061aa1a57d1d70038cd918c9/src/compiler/interface.jl#L106) because that's what [`params` calls](https://github.com/FluxML/Flux.jl/blob/master/src/functor.jl#L41) and what https://github.com/FluxML/Zygote.jl/pull/876 (implicitly/inadvertently?) eliminated.

ToucheSir:
Since this was mentioned in https://github.com/FluxML/Zygote.jl/pull/1035#issuecomment-886914094, here's what I'd like to see before a merge:
- Review comments addressed
- Tests for the issues it covers. Currently there isn't one for #941, so I'm a little unsure as to which issue(s) this is resolving (it obviously is resolving something, I just don't know the issue number(s)). In particular, it would be good to know if this resolves:
  - [ ]  #941 
  - [ ] #37 (or rather, the knock-on effect mentioned in https://github.com/FluxML/Zygote.jl/pull/992#issuecomment-867340277). Test cases in https://github.com/FluxML/Flux.jl/pull/1614#issuecomment-877946209
  - [ ] https://github.com/FluxML/Zygote.jl/pull/1025#issuecomment-878239541
  - [ ] https://github.com/FluxML/Flux.jl/pull/1614#issuecomment-877873705

DhairyaLGandhi:
Well, some of this is orthogonal. Handling `Params` in nested differentiation is more than just handling `push!` so I wouldn't block it on `push!`. `push!` was needed for construction, this is necessary for nesting and tracking parameters in higher order differentiation.

src/compiler/interface.jl
ToucheSir: Then per Carlo's point, `Base.values(gs::Grads)` should also forward to `.grads` as well. Having the 2 be differently sized is unexpected (i.e. potentially subtly breaking), and arguably breaking the contract of `keys` + `values`.

DhairyaLGandhi: Yes, I'll add that

src/lib/base.jl
DhairyaLGandhi: Unfortunately I couldn't see a different way at the time. I'm with you on internal functions. We ended up dropping some grads without it, which hopefully shouldn't be.

ToucheSir:
Coming back to this with more clarity about the whole push! story (i.e. the understanding that this is completely unrelated), I think the only changes left are a non-mutating `get` adjoint and having `Grads` forward `Base.values` to its inner `IdDict`? Most of the CI failures seem unrelated, so I presume they'll go away after a rebase and we can safely land this.

CarloLucibello:
closing as stale

PR #824, opened 2020-11-16
move NNlib rules out of Zygote

simeonschaub:
Partner to https://github.com/FluxML/NNlib.jl/pull/242. I talked a bit about how to go about releasing this there, would appreciate any feedback and suggestions.


src/lib/broadcast.jl
```diff
@@ -89,11 +88,6 @@ end
 
 @adjoint broadcasted(::typeof(identity), x::Numeric) = x, Δ -> (nothing, Δ)
 
-@adjoint function broadcasted(::typeof(σ), x::Numeric)
```
mcabbott, line 92: Should rules like this move?

If I understand right, these break broadcast fusion, but are still worthwhile because Zygote's generic broadcast rule is pretty slow. But that's specific to Zygote.

Or perhaps, they should move to be ZygoteRules rules located within NNlib.

test/gradcheck.jl
```diff
@@ -235,49 +200,6 @@ end
   @test gradient(xs -> sum(inv, [x^2 for x in xs]), ones(2)) == ([-2, -2],)
 end
 
-@testset "conv: spatial_rank=$spatial_rank" for spatial_rank in (1, 2, 3)
```
mcabbott, line 238: It might also be worth keeping NNlib as a test dependency, so that these tests can stay. Possibly moved around. But maybe they catch edge cases which would be easy to miss in re-writing.

src/lib/broadcast.jl
simeonschaub: Do you mean whether to keep them at all? I moved these to ChainRulesCore as well, I don't think we need ZygoteRules for these.

test/gradcheck.jl
simeonschaub: They are still in NNlib, so will be caught sooner or later. We could run NNlib tests as an integration test as part of CI, but I am not sure it's worth it, since these don't really test anything that isn't already tested another way.

src/lib/broadcast.jl
mcabbott: What I mean is that they might not be desirable for other AD packages besides Zygote. With a better handling of broadcasting in general, using just the scalar rule for `σ` might be more efficient than using this one. Perhaps they should be thought of as performance patches for Zygote specifically, not mathematical properties for general use.

test/gradcheck.jl
mcabbott: Oh I'm sorry I didn't scroll down far enough, they are indeed copied by your PR!

src/lib/broadcast.jl
simeonschaub: Ah, that's a good point! Maybe we still want to use ZygoteRules for these then.

simeonschaub: Addressed in FluxML/NNlib.jl#242

simeonschaub:
As discussed in https://github.com/FluxML/NNlib.jl/pull/242#issuecomment-730171436, I have now tagged this as a breaking release, since that seemed more straightforward. I can change that though, if so desired. Should be ready to go from my side.

CarloLucibello (approved):

CarloLucibello:
@simeonschaub merge?

simeonschaub:
Sure!

simeonschaub:
Sorry, I missed one `using NNlib` statement, but that should be fixed now.

CarloLucibello:
what's going on with bors?

simeonschaub:
Is bors fixed now or can we proceed without it?

simeonschaub:
Just so this doesn't get missed: We probably want to wait for #842 to be merged until we release this. (https://github.com/FluxML/Zygote.jl/pull/848#issuecomment-740644537)

willtebbutt:
@simeonschaub I'm a bit confused by this PR -- what aspect of it is breaking?

simeonschaub:
See the discussion in https://github.com/FluxML/NNlib.jl/pull/242. We need to ensure that nobody is using a Zygote version with this PR together with an older NNlib version. Since those packages now don't directly depend on each other, tagging both as breaking and requiring packages with a dependency to explicitly bump both versions simultaneously seemed like the most straightforward solution.

mcabbott:
Same story, right? If Zygote no longer provides some functionality, then code that uses it will break.

PR #825, opened 2020-11-16
use matmul rule from ChainRules

simeonschaub:
Any reason not to use ChainRules for these rules? This bit me, because I defined a custom rrule for multiplication of my custom matrix type and it wasn't hit because it was shadowed by this adjoint defined in Zygote. cc: @oxinabox


oxinabox (approved):

oxinabox:
Bump version?
I will merge if CI passes

oxinabox:
thanks!

PR #826, opened 2020-11-17
type-stable pullback for norm

simeonschaub:
This was accidentally promoting everything to Float64.


DhairyaLGandhi:
Thanks!

bors r+

PR #828, opened 2020-11-17
delete Matric Vector multiply rules that are now in ChainRules

oxinabox:
@simeonschaub  want to review?

this needs @gxyd 's PR to ChainRules to be merged first
https://github.com/JuliaDiff/ChainRules.jl/pull/305

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -26,7 +26,7 @@ ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 [compat]
 AbstractFFTs = "0.5"
 ArrayLayouts = "0.1, 0.2, 0.3, 0.4"
-ChainRules = "0.7.16"
+ChainRules = "0.7.33"
 DiffRules = "1.0"
 FillArrays = "0.8, 0.9, 0.10"
 ForwardDiff = "0.10"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.5.14"
+version = "0.5.15"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -317,21 +317,6 @@ end
 # LinAlg
 # ======
 
-# TODO: remove these once https://github.com/JuliaDiff/ChainRules.jl/pull/305 is merged
-@adjoint function(A::AbstractMatrix * x::AbstractVector)
-  return A * x, Δ::AbstractVector->(Δ * x', A' * Δ)
-end
-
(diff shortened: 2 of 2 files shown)


Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.5.12"
+version = "0.5.10"
```
oxinabox, line 3: Our last actual release was 0.5.9,
looks like we got a bit a head of ourselves.
Once this is merged i will tag a release

simeonschaub (approved):
Looks good to me. CI should of course be rerun once a new ChainRulesCore version is tagged.

Project.toml
CarloLucibello: This habit of changing the version number in each PR doesn't scale very well, creates too many conflicts. 
What about a separate PR (or commit straight to master) with the version bump and then the registrator tagging?
It's an easy process, can be done in the browser

simeonschaub: Another solution might be to just always register the new version right away after such a PR is merged.

CarloLucibello:
@oxinabox let's drop the version change and merge

oxinabox:
I am on leave.
Otherwise this would have been merged a week ago.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.5.14"
+version = "0.5.15"
```
CarloLucibello, line 3: really we should avoid this, open PRs will continuously have conflicts with master, really annoying

oxinabox: 🤷 Its just not that big a deal.
it takes 30 seconds to rebase.

The person it annoys is the PR author, and that is me.
And I am not annoyed.
If other people don't want to rebase to deal with it then they don't have to bump the version number in the PR.
I am not here telling anyone else to.

On a small PR like this one the expectation is that it would be merged quickly (and this would have been except I was on leave).
and in a big PR the time to rebase because the version number has changed is nothing compared to the time taken to create the PR.

Further more, rebasing pulls in the changes from master.
And it is possible to end-up with a senario that has two PRs can both pass tests on their own, but not pass tests if merged one after the other.
Rebasing prevents that.
(Some CI (definately Bors) also prevents that by testing on the result of the merge, rather than testing the actual code in the branch.)

CarloLucibello: I agree, just making sure we don't force this habit on contributors

CarloLucibello (approved):

DhairyaLGandhi:
Any performance considerations here? Otherwise lgtm

oxinabox:
> Any performance considerations here? Otherwise lgtm

Shouldn't be its for all intents and purposes the same code

oxinabox:
Gitlab failure is unrelated, some kind of trouble installing the package, i think i network blip at a guess

PR #833, opened 2020-11-20
Bug Fix: norm of complex array

ho-oto:
#826 made it impossible to take gradient of norm of complex array.
```julia
julia> gradient(x -> norm(x), randn(ComplexF64,2,2))
ERROR: Output is complex, so the gradient is not defined.
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] sensitivity(::Complex{Float64}) at /home/hnakano/.julia/packages/Zygote/JXGyI/src/compiler/interface.jl:49
 [3] gradient(::Function, ::Array{Complex{Float64},2}) at /home/hnakano/.julia/packages/Zygote/JXGyI/src/compiler/interface.jl:54
 [4] top-level scope at REPL[9]:1
```
This patch fixes it.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -437,7 +437,7 @@ end
 end
 
 function _pullback(cx::AContext, ::typeof(norm), x::AbstractArray, p::Real = 2)
-  fallback = (x, p) -> sum(abs.(x).^p .+ eps(0f0)) ^ (one(eltype(x)) / p) # avoid d(sqrt(x))/dx == Inf at 0
+  fallback = (x, p) -> sum(abs.(x).^p .+ eps(0f0)) ^ (real(one(eltype(x))) / p) # avoid d(sqrt(x))/dx == Inf at 0
   _pullback(cx, fallback, x, p)
 end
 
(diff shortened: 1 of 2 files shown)


CarloLucibello:
Thank @ho-oto, this looks fine, travis error is just a time-out. Could you add a test?

ho-oto:
I've added tests for `norm`.
BTW, now gradient of norm of `Matrix{Float16}` becomes `Matrix{Float32}` since we use `+ eps(0f0)`.
Should we fix it?

CarloLucibello:
Yes please

ho-oto:
The simplest solution is to remove `+ eps(0f0)` or to replace it by `+ eps(zero(real(eltype(x))))`.
But I cannot determine if these approaches are OK or not, because I don't understand why `+ eps(0f0)` is used now.

I cannot find the similar `eps` in `ChainRules.jl`.
https://github.com/JuliaDiff/ChainRules.jl/pull/226/files

How should I fix it?

ho-oto:
This bug was fixed by #847

PR #834, opened 2020-11-22
move to GitHub Actions CI

CarloLucibello:


CarloLucibello:
Ok this is done, the only question remaining is whether we want to see a green check and silently fail on nightly or not

CarloLucibello:
I'm merging this as it is, I want to be aware of nightly fails. If at some point we decide to have nightly fail silently instead, we just comment out the  `continue-on-error` lines

DhairyaLGandhi:
It's a lot of noise to see red crosses with every pr, its much more valuable to be aware of pr statuses based on the content of the PR, not the failing nightly. Please add the continue on error if that makes it so

PR #838, opened 2020-11-27
Adjoint for broadcasting `A .+-* Bool`

mcabbott:
This adds some broadcast gradients for `A .+ false` and friends, treating `Bool` at categorical.

Moved over from Flux as part of the move away from a special `Zeros()` type, but worthwhile in general.


CarloLucibello:
This arises a general question, do we want to allow slight modifications, mostly performance-oriented, of the forward (the primal? @oxinabox) in a gradient context? E.g. in this case  if `y = x .+ false`,   we are changing  the standard `x == y` but `x !== y` behavior.
I'm completely fine with this change, just wanted to make things explicit

mcabbott:
Good point. The precedent is that `identity.(x)` normally makes a copy, but:
```
@adjoint broadcasted(::typeof(identity), x::Numeric) = x, Δ -> (nothing, Δ)
```
Which I think is fine, since the difference only matters if you mutate the result, which is already forbidden.

DhairyaLGandhi:
Generally much better not to branch off of values in adjoints. Why do we need these adjoints to start off?

mcabbott:
As I said, and you know, it's motivated by simplifying bias-free layers. But, you can view it as an optimisation like https://github.com/JuliaDiff/ChainRules.jl/pull/267 if you prefer. 

While I agree that generally branch-free code is better, this doesn't stop for instance LinearAlgebra branching on `beta==0` all over the place. If you have real performance concerns then please let us know.

DhairyaLGandhi:
As it stands, we haven't decided to change it to a Bool which is why this confused me. In addition the newer PR to Flux addresses the issues that had been raised earlier, and would be compatible with things like https://github.com/FluxML/Flux.jl/pull/1405.

mcabbott:
Yea I still need to reply to that one. But this can happen independently, it's an optimisation which is worthwhile in general.

CarloLucibello:
obviously, this optimization is motivated by the perspective of having `false` has the Zero bias (which I support but may never happen), nonetheless, it may also be of independent utility so I think we should have it

DhairyaLGandhi:
Correct me if I'm wrong, but isn't that what zygote does already?

mcabbott:
> Correct me if I'm wrong, but isn't that what zygote does already?

Your claim is that the tests in this PR pass on the tagged version of Zygote, I think. If so, that is easy to check.

mcabbott:
Notice that bors got VectorizationBase v0.13.2 (which is what requires HWloc I think) which isn't quite the latest, 0.13.3 was only recently tagged. This is part of LoopVectorization's big 0.9 update. 

Earlier CI got LoopVectorization 0.8.X & passed. I've re-triggered it to see what happens.

Why does Zygote depend on LoopVectorization, BTW? It looks to be just in order to define one adjoint for `vmap`, is that a good idea?

CarloLucibello:
why bors is not picking up the correct versions?

> Why does Zygote depend on LoopVectorization, BTW? It looks to be just in order to define one adjoint for vmap, is that a good idea?

Probably not. Also the logic  of vmap's adjoint should be simplified, I think it is safe to assume it will only involve simple stateless functions

mcabbott:
The last run seems to have the same versions as I get locally, Hwloc v1.0.3, VectorizationBase v0.13.3, LoopVectorization v0.9.0. The server has CUDA right? Unlike my laptop...

mcabbott:
Bump?

CarloLucibello:
last try, then I'll merge manually

bors r+

PR #841, opened 2020-12-01
continue on error

CarloLucibello:


maleadt:
The reason for the timeouts is that the Gitlab CI has fewer resources nowadays, and there's certain packages (OrdinaryDiffEqGPU) whose tests take multiple hours (!)...

Issue #835, opened 2020-11-22
move Distances, AbstractFFTs and SpecialFunctions' rules out of Zygote


CarloLucibello:
to the respective repos, using ChainRulesCore

CarloLucibello:
@mcabbott @willtebbutt so we want to do all this by v0.6, migrating stuff to packages and ChainRules if package owners are not willing to carry the extra dependence (which could be the case for all of them, since they are all very lean)?

DhairyaLGandhi:
Wouldn't put it out for 0.6, but bump it for a bit later since this is a fair amount of work and can incur breakages which need to be accounted for beforehand

willtebbutt:
I wouldn't personally object to these all going into 0.6, provided that new patch releases to 0.5 aren't blocked while this is happening (i.e. the development happens on a separate 0.6-DEV branch). I guess it depends on what time scale you want to see 0.6 on.

Also, I think `SpecialFunctions` is being handled already. See [this PR](https://github.com/JuliaDiff/ChainRules.jl/pull/320) and [the corresponding change in the ChainRules code](https://github.com/JuliaDiff/ChainRules.jl/blob/2dab3ab127709daa61c8c958e4498f492fa1981b/src/ChainRules.jl#L63) (which actually might be a way to push making a breaking release down the line?)

CarloLucibello:
NNlib and LoopVectorization excision are already on master, so we could just do whatever is left to do for SpecialFunctions and tag v0.6, without the additional complexity of having to branch out a release-0.5 and start backporting things

mcabbott:
Does Zygote's forward mode use ChainRules? If so, then removing SpecialFunctions should be easy.

Distances.jl sounds unlikely in the short term. Which others might be worth trying to do quickly, and holding up 0.6? 

If it's just a short time we should just leave everything on master; a medium time maybe it's worth messing with branches; a long time then it should just be 0.7 someday, I think. Messing with Requires to make temporary solutions doesn't sound like great use of everyone's time.

willtebbutt:
@oxinabox what are your thoughts on the `AbstractFFTs` remark above?

oxinabox:
> Does Zygote's forward mode use ChainRules? If so, then removing SpecialFunctions should be easy.

Sadly it does not.

----

Re: AbstractFFTs.
We should ask @StevenGj what to do.
If ever there was a package that is is ok for ChainRules to directly depend on it is probably that one.
[...]

PR #842, opened 2020-12-01
Remove LoopVectorization.jl

mcabbott:
Zygote depends on LoopVectorization only in order to add one gradient definition, for `vmap`. That doesn't seem like a good idea. Surely the rule can live elsewhere, via ZygoteRules ~~or ChainRulesCore~~ which exist for this purpose. Seems to have been added along with `pmap` in #728. 

Marked as draft here until such a rule exists elsewhere. (It's possible that, after that, Zygote should keep the dep. & not load it, for a while, just all Pkg to bound versions.)

The rule might also be simplified, noted here https://github.com/FluxML/Zygote.jl/pull/838#issuecomment-735904764, as I don't think `vmap` can guarantee that the elements will be acted on in order. ~~Since `pmap` also can't guarantee to work in order, it also seems to have no need to reverse the order on the reverse pass.~~


CarloLucibello:
having a second look at the code, I now see we have
```
_tryreverse(m, backs, Δ) = backs, Δ
function _tryreverse(m::typeof(map), backs, Δ::Union{AbstractVector, Tuple})
  return reverse(backs), reverse(Δ)
end
_tryreverse(m, x) = x
```
so `vmap` and `pmap` are not affected

CarloLucibello:
Before doing this we should file a PR to LoopVectorization with the corresponding rule using ChainRulesCore.
We could also get rid of the pmap adjoint here and add it to ChainRules

mcabbott:
I think this is ready to be merged, with the plan being to make a breaking release with this and #824. These need not wait for the corresponding LoopVectorization / NNlib releases.

willtebbutt:
I'm a little confused as to why this is breaking -- it doesn't appear that the user will see any changes. Am I missing something?

mcabbott:
If anyone was using this gradient, and it no longer exists, that will break things right?

And if they were locked to LoopVectorization 0.8, Zygote 0.5, then they will stay broken, as the rule's new home will only be in LoopVectorization 0.9.x.

willtebbutt:
> And if they were locked to LoopVectorization 0.8, Zygote 0.5, then they will stay broken, as the rule's new home will only be in LoopVectorization 0.9.x.

I see. So the argument is essentially that changing a lower-bound on dependency corresponds to a breaking change?

mcabbott:
No, we are dropping a dependency, so can no longer bound it at all. 

But if someone was using it, their code will break.

willtebbutt:
Ahh understood. Would it not be possible to avoid the breaking release by using `Requires` and some static versioning stuff to avoid a breaking release? e.g. what we do with [`Distances.jl`](https://github.com/FluxML/Zygote.jl/blob/d1b06d7b60c1b79f3ce5ab8244843b9c373e0c6f/src/Zygote.jl#L40) and `StatsFuns.jl`?

simeonschaub:
One thing I proposed before was manually adding an upper bound on LoopVectorization/NNlib for all previous Zygote versions in the registry.

mcabbott:
Other plans are possible, we could add gradients for many packages here via Requires. 

But I thought the plan was to go the other way, ideally let many packages contain their own gradients via ChainRulesCore. Maybe #835 is meant to track this. https://github.com/JuliaMath/SpecialFunctions.jl/pull/238 is another step.

willtebbutt:
> But I thought the plan was to go the other way, ideally let many packages contain their own gradients via ChainRulesCore. Maybe #835 is meant to track this. JuliaMath/SpecialFunctions.jl#238 is another step.

Yes -- I'm very much on board with this plan, and am glad to see this kind of moving-stuff-out-of-Zygote work going on. My motivation for asking was to see if it's possible to avoid making a breaking change for a bit longer, since they tend to be quite disruptive. I'm completely on board with removing this stuff from this package (similarly for `Distances` and `StatsFuns`, when / if we can get the corresponding package authors on board), but IMHO the longer that we can avoid making a breaking release the better.

I can't think of an obvious solution though, and it looks like there's been quite a bit of thought put into this, so I now agree that this probably has to be breaking.

mcabbott:
OK, I agree it would be ideal to gang more of them together. I'm lazy to make special temporary versions with Requires just to solve this, though. 

We could look into killing the Distances & StatsFuns deps quickly too, I guess? If using ZygoteRules it would be just copy-paste, and then later it could be upgraded without breaking anything.

willtebbutt:
> We could look into killing the Distances & StatsFuns deps quickly too, I guess? If using ZygoteRules it would be just copy-paste, and then later it could be upgraded without breaking anything.

I that's what we would ideally be doing, but maintainers are still wary of having `ChainRulesCore` as a dependency. See e.g. [this issue](https://github.com/JuliaStats/Distances.jl/issues/172)

mcabbott:
OK, I hadn't see that. Then it doesn't look imminent. 

Another step to consider would be moving the code loaded via Requires from here to ChainRules. Maybe they won't be too keen either but better there than duplicated in each AD package depending on ChainRules, eventually, right? And that move sideways can be done without breaking, as Zygote depends directly on ChainRules.

willtebbutt:
Good point. I have mixed feelings about that, but can certainly see the benefit. @oxinabox what are your thoughts on the above?

PR #844, opened 2020-12-03
Dispatch on AContext

willtebbutt:
Some of the rules in lib/lib have concrete type constraints to `Context`, rather than the (abstract) `AContext`. This has caused me some problems while playing around with my own subtype of `AContext`.

@DhairyaLGandhi  do you reckon this is going to be okay? I can't forsee it causing problems, but I'm not sure if I'm missing something important.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.5.14"
+version = "0.5.15"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
--- a/src/lib/lib.jl
+++ b/src/lib/lib.jl
@@ -125,14 +125,14 @@ end
   val, back
 end
 
-function _pullback(cx::Context, ::typeof(literal_indexed_iterate), xs::Tuple, ::Val{i}) where i
+function _pullback(cx::AContext, ::typeof(literal_indexed_iterate), xs::Tuple, ::Val{i}) where i
   y, b = _pullback(cx, literal_getindex, xs, Val(i))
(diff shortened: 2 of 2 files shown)


DhairyaLGandhi:
Should be fine.

DhairyaLGandhi (approved):

willtebbutt:
@DhairyaLGandhi any idea what's going on with the gitlab builds? Doesn't appear to be something I've changed I don't think...

DhairyaLGandhi:
I'm not certain, maybe out of sync pkg servers?

willtebbutt:
Interesting. I noticed a thing recently where my docs CI couldn't find a new version of a package that was definitely merged into the general registry -- could be the same thing?

DhairyaLGandhi:
Possibly. Hard to say for sure, but we are running identical test suites on both. I might check the versions for loop vectorisation related packages.

CarloLucibello:
tests here are passing on my gpu machine, we could manually merge this

willtebbutt:
> tests here are passing on my gpu machine, we could manually merge this

I'm fine with this if @DhairyaLGandhi is

DhairyaLGandhi:
Could we do a run with `JULIA_PKG_SERVER=""` to test the theory?

Other than that, I think we should be fine to merge this once.

willtebbutt:
> Could we do a run with JULIA_PKG_SERVER="" to test the theory?

What would this involve? Is this a command line argument when you start up Julia, or something relating to CI?

willtebbutt:
Bump @DhairyaLGandhi

DhairyaLGandhi:
Oh, yeah, it's a env var you would set while starting Julia. But I think there were other ci runs that passed, so im going to try calling bors to hopefully it gets it in, else we can do it manually for now.

bors r+

PR #845, opened 2020-12-04
Add missing dispatch for pmap

ChrisRackauckas:
It's a bit difficult to work this into the other scheme because pmap is the only map that has this specific dispatch.


DhairyaLGandhi:
Needs a missing `using Distributed` and `addprocs`

ChrisRackauckas:
I assume you'll only want to addprocs on a separate test environment, which isn't setup for Zygote so for now I was just going to leave that be.

ChrisRackauckas:
Also, note that the `pmap` adjoint can stall, and that's a behavior I don't know how to address yet.

src/lib/array.jl
```diff
@@ -197,6 +197,26 @@ for (mapfunc,∇mapfunc) in [(:map,:∇map),(:pmap,:∇pmap),(:vmap,:∇vmap)]
...
 
+function ∇pmap(cx, f, wp, args...; kwargs...)
+  ys_and_backs = pmap((args...) -> _pullback(cx, f, args...), wp, args...; kwargs...)
+  if isempty(ys_and_backs)
+    ys_and_backs, _ -> nothing
+  else
+    ys, backs = unzip(ys_and_backs)
+    ys, function (Δ)
+      # Apply pullbacks in reverse order. Needed for correctness if `f` is stateful.
+      Δf_and_args_zipped = pmap((f, δ) -> f(δ), wp, _tryreverse(pmap, backs, Δ)...)
```
mcabbott, line 208: `_tryreverse` isn't doing anything, right? Nor is the comment true, I think.

ChrisRackauckas: Yeah the comment isn't true on pmap because it needs to be pure anyways because it distributes.

ChrisRackauckas:
This strategy of enclosing the context is not safe for distributed computing. I think we need to do something a lot more clever. That might be causing the issues.

ChrisRackauckas:
The bigger issue seems to be the deadlock as noted downstream: https://github.com/SciML/DiffEqFlux.jl/issues/452 . Until that's fixed I don't think we can call this done.

src/lib/array.jl
```diff
@@ -5,20 +5,20 @@ using Distributed: pmap
...
 
-@adjoint copy(x::AbstractArray) = copy(x), ȳ -> (ȳ,)
+@adjoint copy(x::AbstractArray) = copy(x), ȳ -> (ȳ,)
 
 @adjoint collect(x::Tuple) = collect(x), dy -> (Tuple(dy),)
 @adjoint collect(x::AbstractArray) = collect(x), dy -> (dy,)
 
 # Array Constructors
-@adjoint (::Type{T})(x::T) where T<:Array = T(x), ȳ -> (ȳ,)
+@adjoint (::Type{T})(x::T) where T<:Array = T(x), ȳ -> (ȳ,)
```
DhairyaLGandhi, line 21: need to get rid of these font changes please

ChrisRackauckas: I don't know if I can. Juno did it automatically, so if I try to save it at all it's going to do it.

DhairyaLGandhi:
Seems like we are hitting a try/ catch somewhere in pmap.

ChrisRackauckas:
That means it's missing the adjoint. I guess we'll need to have separate dispatches for whether there's a caching pool or not?

DhairyaLGandhi:
> That means it's missing the adjoint.

We already `@nograd CachingPool`

ChrisRackauckas:
`@adjoint function pmap(f, args...; kwargs...)` is missing an adjoint.

ChrisRackauckas:
What's the case that's not hitting the adjoints and doing try/catch?

DhairyaLGandhi:
Gotcha - yes, probably

DhairyaLGandhi:
Should work now, could you try with https://github.com/SciML/DiffEqFlux.jl/issues/452 ?

ChrisRackauckas:
```julia
using Distributed
using StochasticDiffEq
using Flux
using DiffEqBase.EnsembleAnalysis
using IterTools: ncycle
using Flux: @epochs
using DiffEqSensitivity
using Statistics
using DiffEqFlux


# Input parameters
T = 1.0f0
tspan = (0.0f0,T)
m=1.0
v=1.0
c=1.0
true_y0 = exp(T)*m + (exp(T)-1)*( sqrt(  2*v/(1-exp(-2*T)) )+1 )
true_Z = sqrt(2*v/(1-exp(-2*T)))
u0 = 7.5 # initial guess
Z = 1.6
p = [Z, u0]

function drift_nn!(du,u,p,t)
    du[1] = - u[1] - p[1] - c
end

function stoch_nn!(du,u,p,t)
    du[1] = p[1]
end

probSDE_nn = SDEProblem(drift_nn!,stoch_nn!,[p[2]],tspan,p)
ensemble_prob = gpu(EnsembleProblem(probSDE_nn))
@time sol = solve(ensemble_prob,EulerHeun(),saveat = T,EnsembleDistributed(), trajectories=1000,dt=0.05)

function loss_2(p)
    tmp_prob = remake(probSDE_nn,p=p,u0=[p[2]])
    ensemble_prob = gpu(EnsembleProblem(tmp_prob))
    sim = solve(ensemble_prob,EulerHeun(),saveat = T,EnsembleDistributed(),trajectories=1000, dt=0.05)
    arrsol = Array(sim)
    loss = (m - mean(arrsol,dims=3)[2])^2 + (v - var(arrsol,dims=3)[2])^2
end

p_init = deepcopy(p)

# train model on the same data num_cycle times
opt = ADAM(0.01)
@time res = DiffEqFlux.sciml_train(loss_2,p,opt, maxiters = 100)

res.minimizer != p_init
loss_2(res.minimizer) < loss_2(p_init)
```

Now works. Can we get a quick tag on this? We've been wanting to do distributed for awhile.

src/lib/array.jl
```diff
@@ -175,7 +175,7 @@ end
 _tryreverse(m, x) = x
 _tryreverse(m::typeof(map), x::Union{AbstractVector, Tuple}) = reverse(x)
 
-for (mapfunc,∇mapfunc) in [(:map,:∇map),(:pmap,:∇pmap)]
+for (mapfunc,∇mapfunc) in [(:map,:∇map),(:pmap, :∇pmap)]
```
DhairyaLGandhi, line 178:
```suggestion
for (mapfunc,∇mapfunc) in [(:map,:∇map),(:pmap,:∇pmap)]
```

PR #846, opened 2020-12-05
Dont error with nothing in backwards pass

### LINKED ISSUE #804: Unused comprehension throws "iterate(nothing)" method error
MWE:

```julia
using Zygote

function h(x)
    [2 for i in 1:2]
    sum(x)
end

gradient(h, zeros(2)) # ERROR: MethodError: no method matching iterate(::Nothing)
```

DhairyaLGandhi:
This still means we do a backwards pass, which might be more correct, although we may be allowed to not do any compute at all, but I want to see if doing this is correct in the first place

Fixes #804

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -174,6 +174,7 @@ function _tryreverse(m::typeof(map), backs, Δ::Union{AbstractVector, Tuple})
 end
 _tryreverse(m, x) = x
 _tryreverse(m::typeof(map), x::Union{AbstractVector, Tuple}) = reverse(x)
+_tryreverse(m::typeof(map), backs, Δ::Nothing) = backs, fill(0,size(backs))
 
 for (mapfunc,∇mapfunc) in [(:map,:∇map),(:pmap,:∇pmap),(:vmap,:∇vmap)]
   @eval function $∇mapfunc(cx, f, args...)
@@ -184,6 +185,7 @@ for (mapfunc,∇mapfunc) in [(:map,:∇map),(:pmap,:∇pmap),(:vmap,:∇vmap)]
       ys, backs = unzip(ys_and_backs)
       ys, function (Δ)
         # Apply pullbacks in reverse order. Needed for correctness if `f` is stateful.
+        tr = _tryreverse($mapfunc, backs, Δ)
         Δf_and_args_zipped = $mapfunc((f, δ) -> f(δ), _tryreverse($mapfunc, backs, Δ)...)
         Δf_and_args = unzip(_tryreverse($mapfunc, Δf_and_args_zipped))
         Δf = reduce(accum, Δf_and_args[1])
--- a/test/gradcheck.jl
+++ b/test/gradcheck.jl
@@ -1641,3 +1641,13 @@ end
     @test gradient(x -> sum(randexp(Random.default_rng(), Float32, (1,1))), 1) == (nothing,)
   end
 end
+
+@testset "iterate(::nothing) #804" begin
+  function h(x)
+    t = [i for i in x]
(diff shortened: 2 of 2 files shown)


DhairyaLGandhi:
Curious to hear from @oxinabox @willtebbutt about how evil this might be

oxinabox:
I don't see how this is ever hit.
I will need to look at this closer to see why `back(:: Nothing)=nothing` isn't catching that delta before it gets there.
This isn't the right place to fix it. I don't think.
I will look more closely on Monday

DhairyaLGandhi:
Which definition are you thinking of here exactly? I agree that this isn't the best place to fix this, but it is a very common code path that this occurs in.

It is happening because we are returning a nothing for something that doesn't contribute to the gradient (correctly) but the plumbing around it expects something in the same shape as the primal.

oxinabox:
> Which definition are you thinking of here exactly?

This defination:
https://github.com/FluxML/ZygoteRules.jl/blob/3ff2464eff6c131d54d543ccb242a4c12a58fbd3/src/adjoint.jl#L58

How can `Δ` be nothing, while hitting the defintion of `back(Δ)` that calls `_tryreverse`?
AFAICT, it should hit the `back(:: Nothing)=nothing` that ZygoteRules adds

src/lib/array.jl
```diff
@@ -174,6 +174,7 @@ function _tryreverse(m::typeof(map), backs, Δ::Union{AbstractVector, Tuple})
 end
 _tryreverse(m, x) = x
 _tryreverse(m::typeof(map), x::Union{AbstractVector, Tuple}) = reverse(x)
+_tryreverse(m::typeof(map), backs, Δ::Nothing) = backs, fill(0,size(backs))
```
oxinabox, line 177: It is also weird that we only add this for `map` not for `vmap` or `pmap`

oxinabox:
Ah i finally get it.
This is only use in
https://github.com/FluxML/Zygote.jl/pull/846/files#diff-a9e025ac90a30d27e7512546971c5d92ea7c3496ba759336ae6bf1cace6db4b2R203

for ` _pullback(cx::AContext, ::typeof(collect), g::Base.Generator)`
which uses `∇map` directly.
Should we just define the `back` function on `nothing` on that line then?
Or are we ensuring we support this having side-effects?
(and if so why do we *not* rely on that for normal `map`?)

src/lib/array.jl
willtebbutt: Also, is this meant to work when the objects being mapped over aren't subtypes of `Real`? If so, we'll need to `fill` with an element other than `0`.

mcabbott: `vmap` and `pmap` can't guarantee any order of iteration, so there seems little point in reversing them.

DhairyaLGandhi: True, we really don't need to do any compute here technically, but it seemed like the least breaking approach here.

Agree that we can fill with the output of the forward pass type, but the thing is also that we know that they don't have gradients so maybe it just needs to return `nothing`

DhairyaLGandhi:
I'm not exactly sure I understand what you're mentioning here, but I think there are more cases than just the MWE that can have code which doesn't participate in calculating gradients, so it would need to be handled more generally. Erroring here might be correct in the sense that "hey its not doing anything and can be removed" but I'm not sure if that is the up to Zygote to make assumptions on.

DhairyaLGandhi:
Happy to take feedback on this for sure.

CarloLucibello:
This should be superseded by #951

CarloLucibello:
closing as stale and no longer needed

Issue #331, opened 2019-09-18
Column vectors work but row vectors fail


Krastanov:
Here are a couple of examples that I thought were very simple, but do not work:

```
gradient(x->norm(x*[1, 1]),1.23) # works fine
gradient(x->norm(x*[1 1]),1.23) # ERROR: Mutating arrays is not supported
gradient(x->sum(x*[1 1]),1.23) # ERROR: Mutating arrays is not supported
gradient(x->(x*[1 1])[1],1.23) # ERROR: Mutating arrays is not supported
```

Same behavior with complex numbers, but different error message:

```
gradient(x->norm(x*[1im, 1]),1.23) # works fine
gradient(x->norm(x*[1im 1]),1.23) # ERROR: Compiling Tuple{typeof(Base.hvcat_fill),Array{Complex{Int64},2},Tuple{Complex{Int64},Int64}}: DimensionMismatch("dimensions must match")
gradient(x->abs(sum(x*[1im 1])),1.23) # same error
gradient(x->abs((x*[1im 1])[1]),1.23) # same error
```

On Julia 1.3-rc2 and Zygote master as of Sep 18th.

Krastanov:
In particular, `x -> (x*[1 1])[1]` raises "Mutating arrays is not supported", but I really do not see how that expression mutates arrays.

MikeInnes:
In practice that error comes up because you're trying to differentiate code in the standard library, which does mutate arrays under the hood. The fix is just to add whatever explicit gradient we're missing so that doesn't happen. In this case it looks like array-scalar multiplication is the issue, so we probably just need to have a method to handle that.

mcabbott:
Here’s a work-around: 
```
julia> y = [1 1]
1×2 Array{Int64,2}:
 1  1

julia> Zygote.gradient(x -> (x*y)[1], 1.23)
(1.0,)

julia> Zygote.gradient(x -> (x*hcat(1,1))[1], 1.23)
ERROR: Mutating arrays is not supported
```
I haven’t checked, but presumably nobody thought to write a gradient for this method of `hcat`, which is what `[1 1]` really calls.

sethaxen:
This was fixed by #847 and can be closed

PR #847, opened 2020-12-06
Remove adjoint for norm

### LINKED ISSUE #663: pullback of norm causes forward pass to be type unstable

sethaxen:
https://github.com/JuliaDiff/ChainRules.jl/pull/226 added `rrule`s for `norm` and friends. This PR removes Zygote's not-so-great adjoint so that ChainRules' version is caught.
The complex norm tests are kept, presumably because they're really checking that gradients of complex-to-real functions work.

Fixes #663, #331 (also fixed by #833)
Supersedes #666 (😈 ), #833

cc @ho-oto @dsweber2

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -26,7 +26,7 @@ ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 [compat]
 AbstractFFTs = "0.5"
 ArrayLayouts = "0.1, 0.2, 0.3, 0.4"
-ChainRules = "0.7.33"
+ChainRules = "0.7.34"
 DiffRules = "1.0"
 FillArrays = "0.8, 0.9, 0.10"
 ForwardDiff = "0.10"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.5.15"
+version = "0.5.16"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -421,11 +421,6 @@ end
   end
 end
 
-function _pullback(cx::AContext, ::typeof(norm), x::AbstractArray, p::Real = 2)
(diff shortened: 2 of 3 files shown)


CarloLucibello:
Could add some tests for the fixed issues?
And why not leaving that norm test also here. Do we have a general policy for the things we port to ChainRules? Having some redundancy in the tests doesn't seem so bad

sethaxen:
> Could add some tests for the fixed issues?
> And why not leaving that norm test also here. Do we have a general policy for the things we port to ChainRules? Having some redundancy in the tests doesn't seem so bad

I've left in that specific test and added a few more. I don't know if there is a policy. cc @oxinabox

oxinabox:
The policy we suggested in the original PR that added ChainRuled support was that we would leave the tests for the stuff that was removed initially with that PR, since those were by their nature very common operations we wanted to make sure didn't breaking, and that would serve as an extensive integration suite for Zygote's connection to ChainRules.
But that we should probably delete the tests of anything else that was later moved over.
I think that still makes sense.

 - ChainRules has more comprehensive and accurate tests because it had ChainRulesTestUtils.jl which uses a much higher accuracy finite differencing method than gradcheck.
 - deleting tests makes the tests faster, and they are kinda slow enough
 - code that doesn't exist doesn't need maintainance.
 - tests you can't fix in the package are annoying.

These factors for have to be balanced though.
I agree keeping some tests is good.
But I think we should now be deleting tests as we delete code.
The tests that will remain that the original ChainRules PR not deleting any should be enough.

DhairyaLGandhi:
I say keeping the tests is the least breaking way of moving forward, and explicitly catching bugs in zygote as introduced by other prs that aren't simply adjoint defs or change some Lowe level adjoints for example. The ci time is a compromise but worth to me

oxinabox:
Fair enough

sethaxen:
Okay, then this PR should be complete.

oxinabox (approved):

Issue #603, opened 2020-04-20
Use ChainRules types


oxinabox:
This issue is about swapping Zygote over to use ChainRule's types by default.
When #366  is merged  rules coming out of ChainRules will use its types like `Composite` and `AbstractZero`,
but thing created via Source Code Transform (SCT) will still use `NamedTuple` and `nothing`.

This is fine as they are mutually compatible because `accum` falls back to `+` and ChainRules overloads `+`.

The use of Base types in Zygote  tends to cause issues as its hard to add methods for `NamedTuple` and `nothing`, due to type piracy.
Especially things like defining linear operations on them (e.g. overloading things from LinearAlgebra), as well as defining addition `+`.
Particular discussion on `Composite` ans structured differential types is in #462

Related to: 
#454 #419 #329

oxinabox:
Discussion @MikeInnes  and I had today

> 
> **Lyndon White:ox:**
> I am now going through and fixing broken tests
> or rather fixing issues that the tests reveal
> One of the tests fails unless I define
> `Base.:*(::Int, ::Nothing) = nothing`
> which presumable means that somewhere Zygote (on perhaps inside a ZygoteRule?) there was something removing that nothing before it could hit
> if I just went full over to ChainRules types this would not be an issue since `AbstractZero` does define multiplcation.
> Does this mean there is some code somewhere that would be able to be deleted?
> 
> **Mike Innes**
> By default `@adjoint` rules ignore nothing entirely
> So I guess it’s just that we’re passing nothing to an rrule which tries to use it like a numeric gradient
> 
> **Lyndon White:ox:**
> yep
> 
> **Mike Innes**
> We can just add that dispatch to the CR rule wrapper function though, right?
> 
> **Lyndon White:ox:**
> Yeah we can
> What do you do when 1 input is nothing and other is non-nothing?
> and so on and so forth for nested tuples of input?
> 
> **Mike Innes**
> Since it’s an rrule there’s only one input
> 
> **Lyndon White:ox:**
> but that input could be a structure (inc tuple)
> 
> **Mike Innes**
> The tuple/struct functions can just carry on using the Zygote definitions for now
> 
> **Lyndon White:ox:**
> indeed, I am just curious
> 
> **Mike Innes**
> If you have an adjoint that takes a struct, all you usually do with the struct elements is forward them on to other gradient functions
> 
> **Lyndon White:ox:**
> That seems like it could mean missing out on some CSE.
> 
> **Mike Innes**
[...]

PR #848, opened 2020-12-07
fix Zygote on 1.6, fix #851

### LINKED ISSUE #851: Forward pass of getfield acts like getproperty in _pullback
MWE
```
julia> using Zygote

julia> struct Funky
    x
    y
end

julia> f = Funky(1, 1);

julia> function Base.getproperty(f::Funky, i::Symbol)
    return 2
end

julia> getproperty(f, :x)
2

julia> getfield(f, :x)
1

julia> y, pb = Zygote._pullback(getproperty, f, :x) # y = 2, that's ok
(2, Zygote.var"#1704#back#159"{Zygote.var"#back#158"{:x,Zygote.Context,Funky,Int64}}(Zygote.var"#back#158"{:x,Zygote.Context,Funky,Int64}(Zygote.Context(nothing), Funky(1, 1), 2)))

julia> y, pb = Zygote._pullback(getfield, f, :x) # y = 2 but should be 1
(2, Zygote.var"#1704#back#159"{Zygote.var"#back#158"{:x,Zygote.Context,Funky,Int64}}(Zygote.var"#back#158"{:x,Zygote.Context,Funky,Int64}(Zygote.Context(nothing), Funky(1, 1), 2)))

```

I suspect [these lines](https://github.com/FluxML/Zygote.jl/blob/master/src/lib/lib.jl#L230-L231) are the issue.

Because then the pullback for `getfield` [will just call `getproperty`](https://github.com/FluxML/Zygote.jl/blob/master/src/lib/lib.jl#L213).

Could we just define a similar `_pullback` for `getfield`?

simeonschaub:
Relies on ~~FluxML/ZygoteRules.jl#13~~ FluxML/ZygoteRules.jl#15. The test I marked as broken doesn't fail if run in an interpreter, so this might be a Julia bug. ~~Other than that, I still see failures related to LoopVectorization, might be due to llvmcall. @chriselrod, do you have any ideas?~~

fixes #851

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,13 +49,13 @@ jobs:
             ${{ runner.os }}-
       - uses: julia-actions/julia-buildpkg@v1
         # `allow-failure` not available yet https://github.com/actions/toolkit/issues/399
-        continue-on-error: ${{ matrix.version == 'nightly' }} # comment out to report nightly failures
+        #continue-on-error: ${{ matrix.version == 'nightly' }} # comment out to report nightly failures
       - uses: julia-actions/julia-runtest@v1
-        continue-on-error: ${{ matrix.version == 'nightly' }} # comment out to report nightly failures
+        #continue-on-error: ${{ matrix.version == 'nightly' }} # comment out to report nightly failures
       - uses: julia-actions/julia-processcoverage@v1
-        continue-on-error: ${{ matrix.version == 'nightly' }} # comment out to report nightly failures
+        #continue-on-error: ${{ matrix.version == 'nightly' }} # comment out to report nightly failures
       - uses: codecov/codecov-action@v1
-        continue-on-error: ${{ matrix.version == 'nightly' }} # comment out to report nightly failures
+        #continue-on-error: ${{ matrix.version == 'nightly' }} # comment out to report nightly failures
         with:
           file: lcov.info
   docs:
--- a/src/compiler/reverse.jl
+++ b/src/compiler/reverse.jl
@@ -26,12 +26,19 @@ end
 unwrapquote(x) = x
 unwrapquote(x::QuoteNode) = x.value
 
-is_literal_getproperty(ex) = iscall(ex, Base, :getproperty) && ex.args[3] isa Union{QuoteNode,Integer}
+is_getproperty(ex) = iscall(ex, Base, :getproperty)
 
 function instrument_getproperty!(ir, v, ex)
-  is_literal_getproperty(ex) ?
-    (ir[v] = xcall(Zygote, :literal_getproperty, ex.args[2], Val(unwrapquote(ex.args[3])))) :
+  if is_getproperty(ex)
+    if ex.args[3] isa Union{QuoteNode,Integer}
+      ir[v] = xcall(Zygote, :literal_getproperty, ex.args[2], Val(unwrapquote(ex.args[3])))
+    else
+      f = insert!(ir, v, :(Val($(ex.args[3]))))
+      ir[v] = xcall(Zygote, :literal_getproperty, ex.args[2], f)
+    end
+  else
     ex
+  end
 end
 
 is_literal_getfield(ex) =
--- a/src/compiler/reverse.jl
+++ b/src/compiler/reverse.jl
@@ -32,7 +32,9 @@ is_literal_getproperty(ex) =
 
 function instrument_getproperty!(ir, v, ex)
   is_literal_getproperty(ex) ?
-    (ir[v] = xcall(Zygote, :literal_getproperty, ex.args[2], Val(unwrapquote(ex.args[3])))) :
+    iscall(ex, Base, :getproperty) ?
+      (ir[v] = xcall(Zygote, :literal_getproperty, ex.args[2], Val(unwrapquote(ex.args[3])))) :
+      (ir[v] = xcall(Zygote, :literal_getproperty, ex.args[2], Val(unwrapquote(ex.args[3])), ex.args[1])) :
     ex
 end
 
--- a/src/compiler/reverse.jl
+++ b/src/compiler/reverse.jl
@@ -26,15 +26,21 @@ end
 unwrapquote(x) = x
 unwrapquote(x::QuoteNode) = x.value
 
-is_literal_getproperty(ex) =
-  (iscall(ex, Base, :getproperty) || iscall(ex, Core, :getfield) || iscall(ex, Base, :getfield)) &&
-  ex.args[3] isa Union{QuoteNode,Integer}
+is_literal_getproperty(ex) = iscall(ex, Base, :getproperty) && ex.args[3] isa Union{QuoteNode,Integer}
 
 function instrument_getproperty!(ir, v, ex)
   is_literal_getproperty(ex) ?
-    iscall(ex, Base, :getproperty) ?
-      (ir[v] = xcall(Zygote, :literal_getproperty, ex.args[2], Val(unwrapquote(ex.args[3])))) :
-      (ir[v] = xcall(Zygote, :literal_getproperty, ex.args[2], Val(unwrapquote(ex.args[3])), ex.args[1])) :
+    (ir[v] = xcall(Zygote, :literal_getproperty, ex.args[2], Val(unwrapquote(ex.args[3])))) :
+    ex
+end
+
+is_literal_getfield(ex) =
+  (iscall(ex, Core, :getfield) || iscall(ex, Base, :getfield)) &&
+  ex.args[3] isa Union{QuoteNode,Integer}
+
+function instrument_getfield!(ir, v, ex)
+  is_literal_getfield(ex) ?
+    (ir[v] = xcall(Zygote, :literal_getfield, ex.args[2], Val(unwrapquote(ex.args[3])))) :
     ex
 end
 
@@ -59,6 +65,7 @@ end
 
 function instrument_literals!(ir, v, ex)
   ex = instrument_getproperty!(ir, v, ex)
+  ex = instrument_getfield!(ir, v, ex)
   ex = instrument_getindex!(ir, v, ex)
   ex = instrument_iterate!(ir, v, ex)
 end
--- a/src/compiler/reverse.jl
+++ b/src/compiler/reverse.jl
@@ -28,6 +28,13 @@ unwrapquote(x::QuoteNode) = x.value
 
 is_getproperty(ex) = iscall(ex, Base, :getproperty)
 
+# The initial premise of literal_getproperty was in some ways inherently flawed, because for
+# getproperty it was intended that _pullback falls back to literal_getproperty, but we actually
+# want the opposite to happen, since Zygote should fall back to recursing into the getproperty
+# implementation by default. Users still want to define custom adjoints using only
+# literal_getproperty, though. We can't really have mutually recursive definitions here, so we
+# now always instrument getproperty as literal_getproperty, no matter whether the second
+# argument is a literal or not.
 function instrument_getproperty!(ir, v, ex)
   if is_getproperty(ex)
     if ex.args[3] isa Union{QuoteNode,Integer}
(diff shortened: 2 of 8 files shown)


simeonschaub:
Ok, besides LoopVectorization, the remaining failures should be fixed by https://github.com/JuliaLang/julia/pull/38764. @mcabbott What's the status of #842? With that merged, we would be pretty close to getting all tests to pass on 1.6.

mcabbott:
#842 is fine, I think, but it wants https://github.com/chriselrod/LoopVectorization.jl/pull/167. 

If the version of Zygote which has #842 is 0.6, then it need not wait --- anyone needing `vmap` should not upgrade until that's ready.

simeonschaub:
Ah, I see! We should coordinate that with #824 then, since that needs a breaking release as well.

simeonschaub:
Note that this doesn't need to hold up a 0.6 release, the changes in this PR should not be breaking.

simeonschaub:
OK, I was still missing some things, but I have now verified this on top of https://github.com/JuliaLang/julia/pull/38764 and all tests pass! :tada:

test/cuda.jl
```diff
@@ -1,5 +1,8 @@
 using CUDA
 
+# FIXME
+if VERSION < v"1.6-"
```
simeonschaub, line 4: I tried with CUDA and GPUCompiler master, but CUDA is still broken for me. Doesn't seem to be related to Zygote though.

simeonschaub: OK, I needed the master branch of LLVM.jl as well. Now it works

CarloLucibello:
Once the compat issues here are fixed, you should also comment out the `continue-on-error` lines in .github/workflow/ci.yml so that we can check any regression

simeonschaub:
What's still holding this up is that https://github.com/FluxML/ZygoteRules.jl/pull/13 hasn't been released yet, but I know @oxinabox had some doubts about the approach there, so maybe we don't want to rush this.

oxinabox:
> What's still holding this up is that FluxML/ZygoteRules.jl#13 hasn't been released yet, but I know @oxinabox had some doubts about the approach there, so maybe we don't want to rush this.

Yeah, i think that it is confusing and inconsistent with all other uses of adjoint.
Futher it will make #603  harder because now some inputs don't have sensitivities in the pullback.
which the current code in https://github.com/FluxML/ZygoteRules.jl/pull/10 depends on.

simeonschaub:
>  I don't understand the logic of that being less likely to break things than what I proposed in #851

My thought process was that someone might have defined an adjoint for `literal_getproperty` with the intention of overloading `getfield`, but the more I think about it, the more I think that case might now be broken with my solution as well, so I would be in favor of introducing a `literal_getfield` instead.

oxinabox:
According to [JuliaHub code search](https://juliahub.com/ui/CodeSearch?q=literal_getproperty&u=all&t=all
) the things to test/check  are:
 - https://github.com/cossio/ZygoteStructArrays.jl
 - https://github.com/SciML/Quadrature.jl
 - https://github.com/SciML/RecursiveArrayTools.jl
 - https://github.com/SciML/DiffEqBase.jl
 - https://github.com/avik-pal/RayTracer.jl

I think it should only matter if they were directly using `getfield` in code (sincde using `.` would be come `getproperty`).
But i think the case that it matters for is broken enough that they would have openned an issue.

simeonschaub:
OK, I updated this PR accordingly and tested that this still works. Since master is now at 0.6 anyways, so I think this should be fine.

src/compiler/reverse.jl
```diff
@@ -26,13 +26,28 @@ end
...
 
 function instrument_getproperty!(ir, v, ex)
-  is_literal_getproperty(ex) ?
-    (ir[v] = xcall(Zygote, :literal_getproperty, ex.args[2], Val(unwrapquote(ex.args[3])))) :
+  if is_getproperty(ex)
+    if ex.args[3] isa Union{QuoteNode,Integer}
+      ir[v] = xcall(Zygote, :literal_getproperty, ex.args[2], Val(unwrapquote(ex.args[3])))
+    else
+      f = insert!(ir, v, :(Val($(ex.args[3]))))
+      ir[v] = xcall(Zygote, :literal_getproperty, ex.args[2], f)
```
simeonschaub, line 37: This is what the fallback pullback definition for `getproperty` was doing anyways and it doesn't look like anyone is defining adjoints for `getproperty` explicitely: https://juliahub.com/ui/RepoSearch?q=%40adjoint%20%28Base%5C.%29%3Fgetproperty&r=true

test/chainrules.jl
```diff
@@ -160,8 +160,15 @@ using Zygote, Test, ChainRules
 
         @test (1,) == h(1)
 
-        a3, pb3 = Zygote.pullback(h, 1)
-        @test ((1,),) == pb3(1)
+        if VERSION >= v"1.6-"
+            @test_broken begin
+                a3, pb3 = Zygote.pullback(h, 1)
+                ((1,),) == pb3(1)
+            end
```
oxinabox, line 167: what is broken here?

simeonschaub: Really not sure. That's the one I mentioned in the description, which doesn't occur when run with JuliaInterpreter. I didn't have time yet to really look into it, but this might be a Julia bug.

oxinabox: Nested AD is hard

simeonschaub: Yeah. Diffractor.jl to the rescue! :)

CarloLucibello:
what is left to do to make CI happy?

simeonschaub:
> what is left to do to make CI happy?

Merging FluxML/ZygoteRules.jl#15 and tagging a new ZygoteRules should make tests pass. I don't know whether all the CUDA fixes are released yet, but IIUC, those tests should only get run by bors.

test/chainrules.jl
oxinabox: Not entirely.
But a but.
The code will be *less* gross and a bit faster.
But the problem is just legit hard.

simeonschaub:
Ok, this should be good to go now. It wouldn't hurt to have another set of eyes look this over, but not sure who would be best to do this.

CarloLucibello:
Maybe @oxinabox or @willtebbutt could approve this?

.github/workflows/ci.yml
```diff
@@ -49,13 +49,13 @@ jobs:
             ${{ runner.os }}-
       - uses: julia-actions/julia-buildpkg@v1
         # `allow-failure` not available yet https://github.com/actions/toolkit/issues/399
-        continue-on-error: ${{ matrix.version == 'nightly' }} # comment out to report nightly failures
+        #continue-on-error: ${{ matrix.version == 'nightly' }} # comment out to report nightly failures
```
willtebbutt, line 52: Is it the intention to keep these commented out after this is merged?

src/compiler/reverse.jl
```diff
@@ -26,13 +26,28 @@ end
...
+    end
+  else
+    ex
+  end
+end
+
+is_literal_getfield(ex) =
```
oxinabox, line 44: to match `get_property` this should just be called `is_getfield`

src/compiler/reverse.jl
```diff
@@ -26,13 +26,28 @@ end
...
+  (iscall(ex, Core, :getfield) || iscall(ex, Base, :getfield)) &&
+  ex.args[3] isa Union{QuoteNode,Integer}
+
+function instrument_getfield!(ir, v, ex)
+  is_literal_getfield(ex) ?
+    (ir[v] = xcall(Zygote, :literal_getfield, ex.args[2], Val(unwrapquote(ex.args[3])))) :
     ex
```
oxinabox, line 51: I think the multiline trinary is strictly harder to read than `if-else`
but morei mportantly we should be consistent withthe `literal_getproperty` above

src/forward/lib.jl
```diff
@@ -62,14 +62,20 @@ if VERSION >= v"1.4.0-DEV.304"
     _pushforward((first(args), tail(tail(dargs))...), Core._apply, f, args...)
 end
 
-using ..Zygote: literal_getproperty, literal_getindex
+using ..Zygote: literal_getproperty, literal_getfield, literal_getindex
 
-_pushforward(dargs, ::typeof(getproperty), x, f) =
-  _pushforward(dargs, literal_getproperty, x, Val(f))
+_pushforward(dargs, ::typeof(literal_getproperty), x::NamedTuple, ::Val{f}) where {f} =
```
willtebbutt, line 67: Is `f` some arbitrary set of indices in this context?

willtebbutt (commented):
Broadly LGTM, just a couple of queries I would like addressed before I approve.

src/forward/lib.jl
DhairyaLGandhi: Presumably it should always be a key from the NT

src/compiler/reverse.jl
oxinabox: It is and inconsistent to have literall getproperty do the lifting things to `Val` in the instrumentation call, but to have `literal_getfield` define a fallback method.
We should do them both the same.
and maybe we should use a little `@eval` loop

```
for getfunc in (:getfield, :getproperty) 
    function $(Symbol(:instrument_, getfunc, :!)(ir, v, ex)
        if $(Symbol(:is_, getfunc) 
```

src/compiler/reverse.jl
simeonschaub: They are checking slightly different things though. `is_literal_getfield` also checks if the second argument is either a symbol or int literal.

src/compiler/reverse.jl
DhairyaLGandhi: Why do we need to refactor like that? Seems like doing it this way makes the intention muddier and still removes the possibility where an adjoint over `getproperty` might be needed in the future.

src/lib/lib.jl
```diff
@@ -207,35 +207,39 @@ end
...
   end
   unwrap(val), back
 end
 
-_pullback(cx::AContext, ::typeof(getproperty), x, f::Symbol) =
-  _pullback(cx, literal_getproperty, x, Val(f))
-
 _pullback(cx::AContext, ::typeof(getfield), x, f::Symbol) =
-  _pullback(cx, literal_getproperty, x, Val(f))
+  _pullback(cx, literal_getfield, x, Val(f))
```
oxinabox, line 229: See above about having this fallback but deleting the ones above is inconsistent

src/compiler/reverse.jl
simeonschaub: I agree. I just tried to keep consistency with the code base, but I think I can make an exception here

src/lib/lib.jl
```diff
@@ -207,35 +207,39 @@ end
 
 @generated nt_nothing(x) = Expr(:tuple, [:($f=nothing) for f in fieldnames(x)]...)
 
-@generated pair(::Val{k}, v) where k = :($k = v,)
+@generated pair(::Val{k}, v, _=nothing) where k = :($k = v,)
+@generated pair(::Val{k}, v, ::NamedTuple{keys}) where {k,keys} = k isa Int ? :($(getfield(keys, k)) = v,) : :($k = v,)
 
-@adjoint function literal_getproperty(x, ::Val{f}) where f
-  val = getproperty(x, f)
+@adjoint function literal_getfield(x, ::Val{f}) where f
```
DhairyaLGandhi, line 213: Might be cleaner to have separate adjoints for literal_getproperty and literal_getfield

oxinabox (approved):
LGTM barring on thing with the inconsistency.
Once that is fixed feel free to merge.

.github/workflows/ci.yml
simeonschaub: I think it would make sense to keep them commented out until something breaks again. That way we know once things are broken again on master.

DhairyaLGandhi: At least for this PR we should have this commented out.

Nightly and 1.6 would have diverged a bit now though

src/forward/lib.jl
simeonschaub: For `getproperty` `f` can only be a symbol. The second argument of `getfield` can also be an integer for `NamedTuples` though.

willtebbutt: Hmm okay. Is there a more descriptive naming convention that we could adopt other than `f`? If something is called `f`, my immediate reaction is to assume it represents a function of some kind. (I realise I'm nit picking here)

oxinabox: > For getproperty f can only be a symbol. The second argument of getfield can also be an integer for NamedTuples though.

This is not true.
`getproperty` can also be a integer on tuples
```
julia> x = (10,20,30)
(10, 20, 30)

julia> getproperty(x, 2)
20
```

Infact you can go weirder if you want
```
julia> struct Foo end

julia> Base.getproperty(::Foo, i::String) = i^3

julia> Foo()."a b c"
"a b ca b ca b c"
```

src/compiler/reverse.jl
simeonschaub: I don't really have a strong opinion here. For `getproperty` this change was necessary, because we want to allow people to define custom adjoints using `literal_getproperty`, but recurse into `getproperty` otherwise. This should happen no matter whether the second argument is a literal or not, so we always need to instrument the IR accordingly. For `getfield`, this way is fine, since it's not a generic function and no user should define an adjoint for `literal_getfield`.

src/forward/lib.jl
simeonschaub: Sorry, I only meant in the case of `NamedTuples`

src/compiler/reverse.jl
oxinabox: Do we not want to do this for if it is `getfield` via a variable or an expression also though?

src/lib/lib.jl
simeonschaub: Could you explain what you mean here? We now don't define a default adjoint for `literal_getproperty`, but instead fall back to recursing into `getproperty`

src/compiler/reverse.jl
oxinabox: How so?

simeonschaub: Are you asking why users shouldn't define an adjoint for `literal_getfield`?

src/compiler/reverse.jl
simeonschaub: We could, but not sure we want to. Let's discuss that in the other thread: https://github.com/FluxML/Zygote.jl/pull/848#discussion_r542334717

src/compiler/reverse.jl
oxinabox: oh sorry no it was github muddying the order of text, I was asking what Dhairya ment but you comment made it clear

Now I mostly get it.
We should include comments in the code that explain that users should not be able to overload `literal_getfield` but that we do need to allow that for `literal_getproperty`, so that that is not lost.

simeonschaub: The general problem here is is that design with `literal_getproperty` in some ways was inherently flawed, because for `getproperty` it was intended that `_pullback` falls back to `literal_getproperty`, but we actually want the opposite to happen, since Zygote should fall back to recursing into the `getproperty` implementation by default.

simeonschaub: >  We should include comments in the code that explain that users should not be able to overload literal_getfield but that we do need to allow that for literal_getproperty, so that that is not lost.

Ok, I can add some comments that explain this a bit better.

simeonschaub:
@oxinabox Did I address your comments?

src/forward/lib.jl
simeonschaub: We could call it `property_name`?

src/lib/lib.jl
simeonschaub: I think this was addressed in the above discussion?

src/forward/lib.jl
willtebbutt: That would work for me.

simeonschaub: Ok, should be addressed now

oxinabox (approved):

oxinabox:
You did looks great

PR #852, opened 2020-12-11
WIP: Fix getfield bug

### LINKED ISSUE #851: Forward pass of getfield acts like getproperty in _pullback

mzgubic:
Closes #851 

Will add tests

### DIFF
--- a/src/lib/lib.jl
+++ b/src/lib/lib.jl
@@ -209,7 +209,7 @@ end
 
 @generated pair(::Val{k}, v) where k = :($k = v,)
 
-@adjoint function literal_getproperty(x, ::Val{f}) where f
+@adjoint function literal_getproperty(x, ::Val{f}, getproperty=getproperty) where f
   val = getproperty(x, f)
   function back(Δ)
     accum_param(__context__, val, Δ) === nothing && return
@@ -228,7 +228,7 @@ _pullback(cx::AContext, ::typeof(getproperty), x, f::Symbol) =
   _pullback(cx, literal_getproperty, x, Val(f))
 
 _pullback(cx::AContext, ::typeof(getfield), x, f::Symbol) =
-  _pullback(cx, literal_getproperty, x, Val(f))
+  _pullback(cx, literal_getproperty, x, Val(f), getfield)
 
 _pullback(cx::AContext, ::typeof(literal_getindex), x::NamedTuple, ::Val{f}) where f =
   _pullback(cx, literal_getproperty, x, Val(f))


src/lib/lib.jl
```diff
@@ -209,7 +209,7 @@ end
 
 @generated pair(::Val{k}, v) where k = :($k = v,)
 
-@adjoint function literal_getproperty(x, ::Val{f}) where f
+@adjoint function literal_getproperty(x, ::Val{f}, getproperty=getproperty) where f
```
oxinabox, line 212: I feel this is going to lead to confusion since getproperty doesn't take a `getproperty` argument.
I thikn better is to declare a `literal_getfield` that matches `literal_getproperty`

```julia
for getfun in (:getproperty, :getfield)
    @eval @adjoint function $(Symbol(:literal_, $getfun)(x, ::Val{f}) where f
        val = $getfun(x,f)
    ...
end
```

src/lib/lib.jl
```diff
@@ -228,7 +228,7 @@ _pullback(cx::AContext, ::typeof(getproperty), x, f::Symbol) =
   _pullback(cx, literal_getproperty, x, Val(f))
 
 _pullback(cx::AContext, ::typeof(getfield), x, f::Symbol) =
-  _pullback(cx, literal_getproperty, x, Val(f))
+  _pullback(cx, literal_getproperty, x, Val(f), getfield)
```
oxinabox, line 231: and to match my suggestion above
```suggestion
  _pullback(cx, literal_getfield, x, Val(f))
```

src/lib/lib.jl
simeonschaub: This was added in https://github.com/FluxML/ZygoteRules.jl/pull/13. My thinking was that it should be more backwards compatible this way, but maybe we should go with your proposal instead.

simeonschaub:
Was this adapted from #848? The question then is, how we want to proceed with releases. Should we release another 0.5.x with this fix, or do we want to just release 0.6 as soon as possible, in which case, we might as well just merge all of #848 (unless anyone has concerns about changes in there).

simeonschaub:
Also note that forward mode has the same issues, so that should probably be fixed as well.

mzgubic:
Hey, thanks for reviewing. It looks like you've taken this into account in #848 now, am I correct? In which case we can close this?

simeonschaub:
Yes, I think closing in favor of #848 makes sense, since other than fixing #851, #848 only really marks two tests as broken in 1.6, so I don't think splitting these changes up is worthwhile. (I also missed this before, but note that this PR on its own isn't enough to fix #851, since `src/compiler/reverse.jl` wold still need to be changed to actually emit `literal_getfield`) Thanks for this anyways, and sorry I didn't communicate my changes better in #848!

Issue #850, opened 2020-12-11
`isderiving` fails one step deep


ChrisRackauckas:
```julia
using Zygote
isderiving() = false
Zygote.@adjoint isderiving() = true, _ -> nothing

struct Tester
    cpu_offload::Float64
end

function Tester()
    @show isderiving()
    cpu_offload = isderiving() ? 0.0 : 0.2
    Tester(cpu_offload)
end

function f(p)
  sum(Tester().cpu_offload .* p)
end

p = [1.0]
f()
dp = Zygote.gradient(f,p) # doesn't reshow `isderiving` ?
@test dp == [0.0] # false: it used cpu_offload == 0.2
```

cossio:
Was this closed in https://github.com/FluxML/Zygote.jl/pull/853?

DhairyaLGandhi:
Yup, you prolly want to use `isderiving(x)` instead of the no-arg method for higher order

PR #853, opened 2020-12-12
Add isderiving utility

DhairyaLGandhi:
cc @ChrisRackauckas who wanted a utility like this

### DIFF
--- a/src/lib/utils.jl
+++ b/src/lib/utils.jl
@@ -133,4 +133,6 @@ Check whether the current function call is happening while taking the derivative
     (nothing,)
 """
 isderiving() = false
+isderiving(x) = false
 @adjoint isderiving() = true, _ -> nothing
+@adjoint isderiving(x) = true, x -> (nothing,)
--- a/src/lib/utils.jl
+++ b/src/lib/utils.jl
@@ -114,6 +114,7 @@ hessian(f, x::AbstractArray) = forward_jacobian(x -> gradient(f, x)[1], x)[2]
 
 """
    isderiving()
+   isderiving(x)
 
 Check whether the current function call is happening while taking the derivative.
 
--- a/src/lib/utils.jl
+++ b/src/lib/utils.jl
@@ -111,3 +111,26 @@ a real.
      1  0
 """
 hessian(f, x::AbstractArray) = forward_jacobian(x -> gradient(f, x)[1], x)[2]
+
+"""
+   isderiving()
+
+Check whether the current function call is happening while taking the derivative.
+
+
+    julia> function f(x)
+             @show isderiving()
+           end
+    
+    f (generic function with 1 method)
+    
+    julia> f(3)
+    isderiving() = false
+    false
+    
+    julia> gradient(f, 4)
+    isderiving() = true
+    (nothing,)
+"""
+isderiving() = false
+@adjoint isderiving() = true, _ -> nothing
--- a/test/tools.jl
+++ b/test/tools.jl
@@ -27,7 +27,7 @@ end
 @testset "isderiving" begin
 
   function f(x)
-    if isderiving(x)
+    if Zygote.isderiving(x)
       x^2
     else
       2x^2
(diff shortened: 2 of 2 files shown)


CarloLucibello:
I would also consider the name `isdifferentiating`

ChrisRackauckas:
https://github.com/FluxML/Zygote.jl/issues/850 is an example of how this fails.

mcabbott:
Agree this needs a better name. As verbs, "deriving" and "differentiating" mean quite different things. Could be `inderivative()` or `Zygote.isactive()`? Or since this is specific to reverse-mode, maybe `isreversemode()`, `haspullback()`?

ChrisRackauckas:
This should definitely include tests, with #850 as an example.

DhairyaLGandhi:
Added #850 as a test, as well as a second order derivative example.

How about `ispullback`?

mcabbott:
It's a little weird that `Zygote.pullback(f, x)` is the function that *returns* what I'd call the pullback, the closure. What this function is testing is whether you are inside the former, not the latter. Which Zygote also calls the adjoint. 

`ispullback` sounds to me like it ought to tell you about its argument, like `iseven`. And `inpullback` sound like you are testing whether you are on the backward pass.

Maybe `withgrad()`, `withadjoint()` are possibilities?

oxinabox:
this is another one like `ignore` that should be in ChainRules/ChainRulesCore.
but I won't block this over that

src/lib/utils.jl
```diff
@@ -111,3 +111,28 @@ a real.
      1  0
 """
 hessian(f, x::AbstractArray) = forward_jacobian(x -> gradient(f, x)[1], x)[2]
+
+"""
+   isderiving()
```
DhairyaLGandhi, line 116:
```suggestion
   isderiving()
   isderiving(x)
```

DhairyaLGandhi:
Beyond the name, this shows that #850 can be had with this design, so we should be okay

CarloLucibello:
So is this supposed to give no error?
```julia
  function f(x, y)
    @assert !Zygote.isderiving(x)
    @assert Zygote.isderiving(y)
    x + y
  end

gradient(y -> f(1, y), 1)
```

src/lib/utils.jl
```diff
@@ -111,3 +111,29 @@ a real.
...
+   isderiving(x)
+
+Check whether the current function call is happening while taking the derivative.
+
+
+    julia> function f(x)
```
CarloLucibello, line 122: this should be inside a ```julia-repl  block I think

src/lib/utils.jl
```diff
@@ -111,3 +111,29 @@ a real.
      1  0
 """
 hessian(f, x::AbstractArray) = forward_jacobian(x -> gradient(f, x)[1], x)[2]
+
+"""
+   isderiving()
+   isderiving(x)
+
+Check whether the current function call is happening while taking the derivative.
```
CarloLucibello, line 119: We should specify what is the argument `x`

DhairyaLGandhi:
Pretty sure that errors.

src/lib/utils.jl
DhairyaLGandhi: This is common practice if the input can be generic - adding an `Any` adds no value

CarloLucibello: I'm not talking about the type, I mean when one should use `isderiving()` and when `isderiving(x)`, the docstring only mentions a current function call without any reference to x

DhairyaLGandhi:
bors try

I want to check buildkite doing its thing

DhairyaLGandhi:
bump @ChrisRackauckas is we need to change something here

DhairyaLGandhi:
Let have this in I think

bors r+

CarloLucibello:
@DhairyaLGandhi why was this merged without review and a follow-up to the naming discussion?

DhairyaLGandhi:
Please do suggest a better name if you some thoughts around it

CarloLucibello:
also there a few unaddressed comments

CarloLucibello:
Could you file a follow-up PR addressing them?

PR #855, opened 2020-12-15
Cleaner pmap adjoint

DhairyaLGandhi:
cc @ChrisRackauckas 

We would need a simple MWE, which should ideally hold something in a local cache maintained by the workers, without serialising them

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -1,7 +1,7 @@
 using Random, FillArrays, AbstractFFTs
 using FillArrays: AbstractFill, getindex_value
 using Base.Broadcast: broadcasted, broadcast_shape
-using Distributed: pmap
+using Distributed: pmap, CachingPool, workers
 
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
https://github.com/FluxML/Zygote.jl/pull/845 was updated and works correctly. This will stall on the construction of caching pools.

DhairyaLGandhi:
Cool, I'll close this then

DhairyaLGandhi:
Seems like the same diff, we need to make ci catch the stall

ChrisRackauckas:
It just needs to add the caching pull to the call. I'll add that.

DhairyaLGandhi:
To confirm, #845 doesn't stall with the call to CachingPool, right?

ChrisRackauckas:
Yes, you just have to `@nograd` it and the `workers` call, which both shouldn't have Zygote touching.

PR #856, opened 2020-12-18
Make PackageCompiler happy

DhairyaLGandhi:


DhairyaLGandhi:
Moving  the `using` statements out of the files shouldn't be necessary but seem like Requires doesn't run it in the right module (I get an error saying `IRTools` is not a dep of Zygote on calling `Zygote.refresh()`, when everything else works, so yeah)

Other than that, this is good to go

bors r+

Issue #863, opened 2020-12-28
getindex(::Fill, idx)


willtebbutt:
As pointed out [here](https://github.com/JuliaDiff/ChainRules.jl/pull/335) by @mcabbott , the following happens:
```julia
julia> using FillArrays, Zygote

julia> Zygote.gradient(x -> x[1], Fill(3, 2))[1]
2-element Array{Int64,1}:
 1
 0
```
The type is innappropriate -- the answer should be
```julia
(value=1, axes=nothing)
```
or at the very least another `Fill`.

DhairyaLGandhi:
Hmm, the output seems consistent since you have gradients defined for some elements and zeros for others.

Do we want to say DNE for the remainder of the elements? Either way this would materialize the array iiuc

willtebbutt:
The point is not that the output is inconsistent (I agree that the output is consistent with what you would get if you provided an `Array`), it's that the optimal thing to do would be to return `(value=1, axes=nothing)` vs what currently happens, which largely removes the benefit of using a `Fill` if you plan to do AD with it using `Zygote`, since the memory (and hence time) requirements are O(length_of_fill), rather than O(1).

(Part) of a solution is to stop getting in the way of Zygote automatically deriving the correct adjoint by restricting [this rrule](https://github.com/FluxML/Zygote.jl/blob/4fdb6919273854f79300b491197f97a8d71028e0/src/lib/array.jl#L31) to something less general than `AbstractArray`.

mcabbott:
I guess cases like this are the motivation for treating `Fill` as being one number:
```julia
julia> gradient(y -> sum(sqrt, gradient(x -> sum(x.^2), y)[1]), reshape(1:6,2,3))[1]
2×3 Matrix{Float64}:
 0.707107  0.408248  0.316228
 0.5       0.353553  0.288675
```
The gradient of the inner `sum` is a `Fill`, which gets fed into another function which Zygote is differentiating, and not materialising that array would save some allocations, as you say.

But what seems a little tricky is that, if the returned gradient escapes from Zygote, the returning the one-number version seems like it violates the fact that `Fill <: AbstractArray`. If I type this, I would be pretty surprised to be given back something not `==` the result from using dense `fill` instead:
```julia
julia> gradient(x -> sum(cumsum(x, dims=2)), Fill(pi,2,7))[1]
2×7 Matrix{Float64}:
 7.0  6.0  5.0  4.0  3.0  2.0  1.0
[...]

PR #864, opened 2020-12-28
WIP: cache-less contexts

willtebbutt:
EDIT: it looks like I mis-diagnosed what was causing performance issues -- it was to do with the stability of `literal_getproperty` rather than being related to this. Consequently, I'm stuck for a good example where this fixes something that was broken. Sorry for getting your hopes up @oxinabox !

This is a proposal to introduce a specialised `AContext` type that sacrifices the ability to handle
1. globals
2. tasks
 
in exchange for substantially reduced overhead in certain circumstances. This gives users and rule-writers the ability to tell Zygote not to worry about either globals or tasks if they're not using them and, moreover, is done in (I believe) a safe way in that if such things _are_ encountered, the code errors.

In terms of performance improvements, with the standard `Context`, the following performance is obtained:
```julia
using BenchmarkTools, StaticArrays, Zygote
a = SMatrix{2, 2}(randn(2, 2));

julia> @benchmark Zygote._pullback($(Zygote.Context()), x -> x[1], $a)
BenchmarkTools.Trial:
  memory estimate:  2.11 KiB
  allocs estimate:  26
  --------------
  minimum time:     4.780 μs (0.00% GC)
  median time:      4.951 μs (0.00% GC)
  mean time:        5.368 μs (2.72% GC)
  maximum time:     504.230 μs (98.55% GC)
  --------------
  samples:          10000
  evals/sample:     7

out, pb = Zygote._pullback(Zygote.Context(), x -> x[1], a);

julia> @benchmark $pb($out)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     0.037 ns (0.00% GC)
  median time:      0.039 ns (0.00% GC)
  mean time:        0.039 ns (0.00% GC)
  maximum time:     0.047 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000
```
while the new `NoContext` context yields:
```julia
julia> @benchmark Zygote._pullback(Zygote.NoContext(), x -> x[1], $a)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     0.036 ns (0.00% GC)
  median time:      0.039 ns (0.00% GC)
  mean time:        0.039 ns (0.00% GC)
  maximum time:     0.046 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> out, pb = Zygote._pullback(Zygote.NoContext(), x -> x[1], a);

julia> @benchmark $pb($out)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     0.036 ns (0.00% GC)
  median time:      0.039 ns (0.00% GC)
  mean time:        0.039 ns (0.00% GC)
  maximum time:     0.055 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000
```
As you can see, the overhead is completely eliminated, which is due to the loss of the cache.

Note that to provide the above example I've temporarily modified the `getindex` implementation because it was the easiest way to provide a MWE. The actual reason I investigated the use of this type was for some work on my [TemporalGPs](https://github.com/willtebbutt/TemporalGPs.jl/) package, where I've managed to get Zygote to derive zero-allocation implementations of some code where it's _really_ crucial that it does so (on the 0.4-DEV branch) -- I had previously hand-written rules for these sections of code but they were a complete nightmare to maintain. The point being that you don't _just_ see performance impovements for `getindex`, it's just one example.

I _believe_ that the main source of performance improvement is that `accum_param` function is always type stable for `NoContext`, whereas for `Context` is it only type-stable if `isbitstype(x)`.

@DhairyaLGandhi do you a) think that this is a good thing to have, and b) what sort of testing do you think we would need?

edit: I should say that I do think that we need to restrict the types to which `getindex` applies for the sake of #863 , but I don't want to address that in this PR.

edit2: this doesn't require the addition of StaticArrays -- the dep would be removed prior merging something like this.

### DIFF
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -11,6 +11,27 @@ Context() = Context(nothing)
 
 cache(cx::Context) = cx.cache === nothing ? (cx.cache = IdDict()) : cx.cache
 
+"""
+  NoContext <: Zygote.AContext
+
+The `NoContext` context assumes that there is no global context, and errors if any part of a
+function that is being differentiated interacts with either globals or tasks.
+
+The benefit of using such a construct is the elimination of overhead associated with the
+`cache` `IdDict` in a standard `Context`, which can be prohibitively large in certain
+settings.
+"""
+struct NoContext <: Zygote.AContext end
+
+# There is no cache, so there are never any fields.
+cache(cx::NoContext) = (cache_fields=nothing)
+
+# The is no cache, so this context never has any fields.
+Base.haskey(cx::NoContext, x) = false
+
+accum_param(::NoContext, x, Δ) = Δ
+
+
 struct Pullback{S,T}
   t::T
 end
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.6.0"
+version = "0.6.1"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
@@ -17,6 +17,7 @@ NaNMath = "77ba4419-2d1f-58cd-9bb1-8ffee604a2e3"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -28,9 +28,13 @@ end
 @adjoint (::Type{T})(sz) where {T<:Zeros} = T(sz), Δ->(nothing,)
 @adjoint (::Type{T})(sz) where {T<:Ones} = T(sz), Δ->(nothing,)
 
-@adjoint getindex(x::AbstractArray, inds...) = x[inds...], ∇getindex(x, inds)
+@adjoint function getindex(x::Array, inds...)
+  return x[inds...], ∇getindex(x, inds)
+end
 
-@adjoint view(x::AbstractArray, inds...) = view(x, inds...), ∇getindex(x, inds)
+@adjoint function view(x::AbstractArray, inds...)
+  return view(x, inds...), ∇getindex(x, inds)
+end
 
 ∇getindex(x::AbstractArray, inds) = dy -> begin
   if inds isa  NTuple{<:Any, Integer}


src/compiler/interface.jl
```diff
@@ -11,6 +11,27 @@ Context() = Context(nothing)
...
+`cache` `IdDict` in a standard `Context`, which can be prohibitively large in certain
+settings.
+"""
+struct NoContext <: Zygote.AContext end
+
+# There is no cache, so there are never any fields.
+cache(cx::NoContext) = (cache_fields=nothing)
+
+# The is no cache, so this context never has any fields.
+Base.haskey(cx::NoContext, x) = false
```
oxinabox, line 30: Should this be `false` or should it error?
I guess `false` is safe, because it would have already errored when it tried to set a value in the cache?

src/compiler/interface.jl
```diff
@@ -11,6 +11,27 @@ Context() = Context(nothing)
...
+function that is being differentiated interacts with either globals or tasks.
+
+The benefit of using such a construct is the elimination of overhead associated with the
+`cache` `IdDict` in a standard `Context`, which can be prohibitively large in certain
+settings.
+"""
+struct NoContext <: Zygote.AContext end
+
+# There is no cache, so there are never any fields.
+cache(cx::NoContext) = (cache_fields=nothing)
```
oxinabox, line 27: Should this error?
Or return a custom type that has setindex overloaded, to give a helpful exception that says you can't use `NoContext`

oxinabox:
I think this is a fantastic idea.

Zygote has to support ADing in the presense of globals, so that people messing around in the REPL still get sensible results.
and when you are just messing around in the REPL you don't super care about performance anyway.

But once you are serious you write code that uses functions and has no globals.
And you do care about performance.
And so you can use this `NoContext` and get a speed-up.
(assuming also not ADing though `Task`s, which is fairly rare)


### Follw up ideas
Two things to think about that I don't think should block this but that we could think about doing in folow up PRs.
Lets keep this PR as small and self-contained as we can.

1. Can we rewrite how tasks and such are done so they don't use the cache?
Looking at the code
https://github.com/FluxML/Zygote.jl/blob/4fdb6919273854f79300b491197f97a8d71028e0/src/lib/base.jl#L68-L76
I think maybe we can replace the two uses of `cache(__context__)[t] ` with just `local backtask` declared during the primal pass (outside of any task) that just gets assigned in the `t.code` and `fetch`ed in the pullback

2. Can we do this automatically?
Like can we get to the end of the forward-pass, look in the Context,
see that the the cache has never been used, and then somehow changes something so that `accum_param` is type stable.
Maybe due to constant folding?

src/compiler/interface.jl
willtebbutt: I'm not at all sure whether the current implementation of this stuff is the most logical way to do things -- it's a hack that appears to work.

For example, I'm not even really sure what the extent of the use of `accum_param` is. It appears to be used in the `_pullback` definitions for
1. `unwrap`.
2. `literal_getindex` for `NTuple`s,
3. `getindex` for `NTuple`s
4. `literal_getfield`

I'm not at all clear on the purpose of `unwrap`, so I'm not really sure what I'm breaking there.

willtebbutt:
> Can we do this automatically?

This would be interesting. You would probably still want to have some kind of mechanism to ensure that you've not accidentally used a global though.

oxinabox:
I think doing this automatically is very hard.

3. Can we write some type-stable special cases of `accum_param`?
For a ton of cases the result type shoud just agree with the type of `Δ`, `accum_param(ctx, x, Δ:T)::T`
Like if `Δ` is an `Array` then the result is an `Array`.
(if it is a special array it might not be)

DhairyaLGandhi:
I like the thought behind it, `accum_*` functions are typically need to go over all the available parameters but if we can guarantee somehow that we hit type stable pathways, that would take care of a lot of overhead.

How would we expose this to users too? Typically, that's not something users mess with, but I can understand if there are cases where something like this is useful.

willtebbutt:
> I like the thought behind it, accum_* functions are typically need to go over all the available parameters but if we can guarantee somehow that we hit type stable pathways, that would take care of a lot of overhead.

Yeah, that's what I assumed. Is it clear what sort of code this would benefit most from? I realised that almost all of the stuff that I am doing actually does involve bits-types, so this wasn't a problem.

> How would we expose this to users too? Typically, that's not something users mess with, but I can understand if there are cases where something like this is useful.

This is a good question. Perhaps we could just expose another method of `pullback` and `gradient` that allow the user to specify an `AContext`? Maybe as a kwarg?

PR #875, opened 2021-01-07
Remove eigen and eigvals adjoints for Symmetric/Hermitian

sethaxen:
https://github.com/JuliaDiff/ChainRules.jl/pull/323 added rrules for `eigen` and `eigvals` for Symmetric/Hermitian matrices. This PR removes those rules from Zygote. Once the old tests pass, I'll remove them, since the rrules are more thoroughly tested.

Depends on https://github.com/JuliaRegistries/General/pull/27508 is merged.

### DIFF
--- a/test/gradcheck.jl
+++ b/test/gradcheck.jl
@@ -775,6 +775,49 @@ function _gradtest_hermsym(f, ST, A)
   end
 end
 
+@testset "eigen(::RealHermSymComplexHerm)" begin
+  MTs = (Symmetric{Float64}, Hermitian{Float64}, Hermitian{ComplexF64})
+  rng, N = MersenneTwister(123), 7
+  @testset "eigen(::$MT)" for MT in MTs
+    T = eltype(MT)
+    ST = _hermsymtype(MT)
+
+    A = ST(randn(rng, T, N, N))
+    U = eigvecs(A)
+
+    @test _gradtest_hermsym(ST, A) do (A)
+      d, U = eigen(A)
+      return U * Diagonal(exp.(d)) * U'
+    end
+
+    y = Zygote.pullback(eigen, A)[1]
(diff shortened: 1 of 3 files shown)


sethaxen:
Currently this fails because Zygote passes a `Composite{Any}` to the pullback for `eigen`, while ChainRules expects a `Composite{<:Eigen}`. @oxinabox what's the recommended way to handle this? Wait on #603 to remove the adjoint from Zygote, or change the pullback for `eigen` to take a `Composite{Any}`?

oxinabox:
Accept `Composite{Any`
The primal type param here is just being used for extra checking

sethaxen:
Okay, the test suite passed without the rule (except on Julia 1.3, which seems to have a CUDA issue), and the tests have now been removed, so this is ready for review.

test/gradcheck.jl
```diff
@@ -775,49 +775,6 @@ function _gradtest_hermsym(f, ST, A)
   end
 end
 
-@testset "eigen(::RealHermSymComplexHerm)" begin
```
oxinabox, line 778: @DhairyaLGandhi  is keen to leave tests inplace after functionality is moved.

oxinabox (approved):

DhairyaLGandhi:
Thanks, this looks good!

DhairyaLGandhi:
It'd be nice to keep tests if it's not too much, so we can test changes in Zygote more thoroughly as well

sethaxen:
Okay, I added the tests back. This should now be ready to merge.

DhairyaLGandhi:
The failure on 1.3 is curious but not related to this PR, so wouldn't block it. Thanks!

DhairyaLGandhi (approved):

PR #876, opened 2021-01-07
Add push! adjoint for vector of arrays

DhairyaLGandhi:


Project.toml
```diff
@@ -22,7 +22,7 @@ ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 
 [compat]
 AbstractFFTs = "0.5, 1.0"
-ChainRules = "0.7.34"
+ChainRules = "0.7.45"
```
CarloLucibello, line 25:
```suggestion
ChainRules = "0.7.35"
```

CarloLucibello:
this PR has no tests and it's doing a few unrelated things

ChrisRackauckas:
For the MWE https://github.com/FluxML/Zygote.jl/issues/37#issuecomment-771532104 it still doesn't work but it seems that's because `pop!` doesn't catch `pop!(::BinaryHeap)` which might need a similar adjoint to the one here.

DhairyaLGandhi:
Maybe DataStructures.jl needs an AD-ready file somewhere in their source, so the adjoints are defined there, instead of here?

Issue #879, opened 2021-01-11
Scalar indexing with vcat(::CuArray, ::Number)


DhairyaLGandhi:
Simpler MWE for https://github.com/SciML/DiffEqFlux.jl/issues/401

```julia
julia> r = rand(Float32, 3) |> gpu;

julia> gradient((x,y) -> sum(vcat(x,y)), r, 5)
ERROR: scalar getindex is disallowed
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] assertscalar(::String) at /home/dhairyalgandhi/.julia/packages/GPUArrays/WV76E/src/host/indexing.jl:41
 [3] getindex at /home/dhairyalgandhi/.julia/packages/GPUArrays/WV76E/src/host/indexing.jl:96 [inlined]
 [4] pull_block_vert at /home/dhairyalgandhi/Zygote.jl/src/lib/array.jl:100 [inlined]
 [5] #437 at /home/dhairyalgandhi/Zygote.jl/src/lib/array.jl:105 [inlined]
 [6] iterate at ./generator.jl:47 [inlined]
 [7] collect_to!(::Array{CuArray{Float32,1},1}, ::Base.Generator{Base.OneTo{Int64},Zygote.var"#437#439"{CuArray{Float32,1},Tuple{CuArray{Float32,1},Int64},Array{Int64,1}}}, ::Int64, ::Int64) at ./array.jl:732
 [8] collect_to_with_first! at ./array.jl:710 [inlined]
 [9] _collect at ./array.jl:704 [inlined]
 [10] collect_similar at ./array.jl:628 [inlined]
 [11] map at ./abstractarray.jl:2162 [inlined]
 [12] #436 at /home/dhairyalgandhi/Zygote.jl/src/lib/array.jl:105 [inlined]
 [13] #2425#back at /home/dhairyalgandhi/.julia/packages/ZygoteRules/OjfTt/src/adjoint.jl:59 [inlined]
 [14] #11 at ./REPL[29]:1 [inlined]
 [15] (::Zygote.var"#41#42"{typeof(∂(#11))})(::Float32) at /home/dhairyalgandhi/Zygote.jl/src/compiler/interface.jl:40
 [16] gradient(::Function, ::CuArray{Float32,1}, ::Vararg{Any,N} where N) at /home/dhairyalgandhi/Zygote.jl/src/compiler/interface.jl:49
 [17] top-level scope at REPL[29]:1
```

[...]

PR #880, opened 2021-01-11
Add scalar method for vcat with number

DhairyaLGandhi:
Fixes https://github.com/SciML/DiffEqFlux.jl/issues/401, https://github.com/FluxML/Zygote.jl/issues/879


DhairyaLGandhi:
not sure what happened to build kite, @maleadt would you know?

maleadt:
You're using an ancient base branch that doesn't have a pipeline yaml yet.

DhairyaLGandhi:
Oof, maybe it's from an out of date clone. Sorry for bothering, and thank you

maleadt:
Usage of `allowscalar` seems fine here. Do know that there might be issues with heavy use of that, e.g. it's currently not a thread-safe feature of GPUArrays.jl. We can always fix that, though.

src/lib/broadcast.jl
```diff
@@ -239,4 +239,6 @@ end
     Base.convert(T, xs), Δ -> (nothing, Base.convert(Array, Δ),)
   end
 
+
+  pull_block_vert(sz, Δ::CuArray, A::Number) = CUDA.@allowscalar Δ[sz:sz]
```
DhairyaLGandhi, line 243:
```suggestion
  pull_block_vert(sz, Δ::CuArray, A::Number) = CUDA.@allowscalar Δ[sz]
```

src/lib/broadcast.jl
```diff
@@ -239,4 +239,6 @@ end
     Base.convert(T, xs), Δ -> (nothing, Base.convert(Array, Δ),)
   end
 
+
```
DhairyaLGandhi, line 242:
```suggestion
```

ChrisRackauckas:
@DhairyaLGandhi revive this?

DhairyaLGandhi:
@ChrisRackauckas does this look fine?

ChrisRackauckas:
👍

PR #882, opened 2021-01-13
Remove Symmetric/Hermitian matrix function rules

sethaxen:
https://github.com/JuliaDiff/ChainRules.jl/pull/193 (which is about to be merged) adds to ChainRules the `rrule`s for  matrix functions (i.e. power series functions) of `Symmetric` and `Hermitian` matrices. This PR removes the corresponding adjoints from Zygote. Tests pass locally but won't here until a new CR release is registered. This PR is blocked until then.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -574,7 +574,7 @@ end
 # so we call it manually. This can be removed when the generic rule for exp is moved to
 # ChainRules
 function _pullback(::AContext, ::typeof(exp), A::LinearAlgebra.RealHermSymComplexHerm)
-@adoint exp(A::LinearAlgebra.RealHermSymComplexHerm)
+@adjoint exp(A::LinearAlgebra.RealHermSymComplexHerm)
     Y, back = chain_rrule(exp, A)
     return Y, Δ -> (back(Δ)[2],)
 end
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -673,39 +673,6 @@ function _pullback(cx::AContext,
   return _pullback(cx, (A, p) -> _apply_series_func(f, A, p), A, p)
 end
 
-for func in (:exp, :log, :cos, :sin, :tan, :cosh, :sinh, :tanh, :acos, :asin, :atan, :acosh, :asinh, :atanh, :sqrt)
-  @eval begin
-    function _pullback(cx::AContext,
-                       f::typeof($func),
-                       A::LinearAlgebra.RealHermSymComplexHerm)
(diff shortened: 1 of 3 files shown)


src/lib/array.jl
```diff
@@ -570,6 +570,13 @@ end
   return (Ā,)
 end
 
+# The adjoint for exp(::AbstractArray) intercepts ChainRules' rrule for exp(::Hermitian),
+# so we call it manually. This can be removed when the generic rule for exp is moved to
+# ChainRules
+function _pullback(::AContext, ::typeof(exp), A::LinearAlgebra.RealHermSymComplexHerm)
+    return chain_rrule(exp, A)
```
sethaxen, line 577: This seems to be the only special-casing in this file, so if anyone knows a better way to work around the generality of the current `exp` adjoint, let me know.

sethaxen:
This PR is ready for review

src/lib/array.jl
DhairyaLGandhi: Do we need to do it via `_pullback`? Seems like a regular `@adjoint` should do the trick too

DhairyaLGandhi (commented):
Thanks for looking into it!

src/lib/array.jl
sethaxen: We don't need to do it via `_pullback`, it just let me use `chain_rrule` directly. The alternative is this:

```suggestion
@adoint exp(A::LinearAlgebra.RealHermSymComplexHerm)
    Y, back = chain_rrule(exp, A)
    return Y, Δ -> (back(Δ)[2],)
```

DhairyaLGandhi: Yeah that's what I was thinking, thanks

src/lib/array.jl
```diff
@@ -570,6 +570,15 @@ end
   return (Ā,)
 end
 
+# The adjoint for exp(::AbstractArray) intercepts ChainRules' rrule for exp(::Hermitian),
+# so we call it manually. This can be removed when the generic rule for exp is moved to
+# ChainRules
+function _pullback(::AContext, ::typeof(exp), A::LinearAlgebra.RealHermSymComplexHerm)
+@adoint exp(A::LinearAlgebra.RealHermSymComplexHerm)
```
DhairyaLGandhi, line 577:
```suggestion
@adjoint exp(A::LinearAlgebra.RealHermSymComplexHerm)
```

sethaxen:
Okay @DhairyaLGandhi, tests pass again with the `@adjoint`.

DhairyaLGandhi:
Thanks!

bors r+

PR #886, opened 2021-01-18
add gradient checkpointing

### LINKED ISSUE #884: Gradient checkpointing?
How to do gradient checkpointing? Is it implemented in Zygote or some other package in the ecosystem?
I tried
```julia
using Zygote, ChainRulesCore

function checkpointed(f, args...)
    f(args...)
end

function ChainRulesCore.rrule(::typeof(checkpointed), f, xs...)
    y = f(xs...)
    function pullback_checkpointed(Δy)
        y, pb =Zygote.pullback(f, xs...) 
        return pb(Δy)
    end
    y, pullback_checkpointed
end
```
but it does not work:
```julia
y, pb = Zygote.pullback(checkpointed, sin, pi/2)
pb(1.0)]
()
```

jw3126:
Close #884.


src/lib/grad.jl
```diff
@@ -14,3 +14,24 @@ macro which(ex)
...
+
+Use gradient checkpointing on the call `f(xs...)`. This means that
+`checkpointed(f, xs...) === f(xs...)`, but when computing the derivative
+intermediate results from the forward pass of `f` will not be stored. Instead the forward
+pass will be repeated, when computing the derivative.
+This this saves memory at the cost of inreasing exectution time.
+"""
+checkpointed(f, xs...) = f(xs...)
+
+function Zygote._pullback(ctx::Zygote.AContext, ::typeof(checkpointed), f, xs...)
```
jw3126, line 30: Implementation taken from @willtebbutt in #884

test/runtests.jl
```diff
@@ -2,6 +2,52 @@ using Zygote, Test
...
+    f = CountCalls()
+    g = CountCalls()
+    h = CountCalls()
+    function doit(x)
+        x2 = f(x)
+        Zygote.checkpointed(h∘g, x2)
+    end
+    y,pb = Zygote.pullback(doit, 4.0)
+    @test f.ncalls === g.ncalls === h.ncalls === 1
+    @show pb(1.0)
```
jw3126, line 45: This call fails:
```
  BoundsError: attempt to access (nothing, 1.0)
    at index [3]
  Stacktrace:
   [1] getindex at ./tuple.jl:24 [inlined]
   [2] gradindex(::Tuple{Nothing,Float64}, ::Int64) at /home/jan/.julia/dev/Zygote/src/compiler/re
verse.jl:12
   [3] doit at /home/jan/.julia/dev/Zygote/test/runtests.jl:41 [inlined]
   [4] (::typeof(∂(λ)))(::Float64) at /home/jan/.julia/dev/Zygote/src/compiler/interface2.jl:0
   [5] (::Zygote.var"#41#42"{typeof(∂(λ))})(::Float64) at /home/jan/.julia/dev/Zygote/src/compiler
/interface.jl:40
   [6] top-level scope at show.jl:641
   [7] top-level scope at /home/jan/.julia/dev/Zygote/test/runtests.jl:45
   [8] top-level scope at /home/jan/julia/usr/share/julia/stdlib/v1.5/Test/src/Test.jl:1115
   [9] top-level scope at /home/jan/.julia/dev/Zygote/test/runtests.jl:7
   [10] include(::String) at ./client.jl:457
   [11] top-level scope at none:6
   [12] eval(::Module, ::Any) at ./boot.jl:331
   [13] exec_options(::Base.JLOptions) at ./client.jl:272
   [14] _start() at ./client.jl:506
```
what am I doing wrong?

src/lib/grad.jl
jw3126: Is `lib/grad.jl`the file where this should live? Should I export?

test/runtests.jl
```diff
@@ -2,6 +2,52 @@ using Zygote, Test
 using Zygote: gradient
 using CUDA: has_cuda
 
+using Zygote
+@testset "gradient checkpointing" begin
```
jw3126, line 6: In which file should these tests live? They are here right now only for quicker development iteration.

src/lib/grad.jl
```diff
@@ -14,3 +14,24 @@ macro which(ex)
   @capture(ex, f_(args__)) || error("Zygote.@which f(args...)")
   :(InteractiveUtils.@which adjoint(Context(), $(esc(f)), $(esc.(args)...)))
 end
+
+"""
+
+    checkpointed(f, xs...)
```
jw3126, line 20: Naming? `checkpointed` or `checkpoint` come to my mind.

src/lib/grad.jl
```diff
@@ -14,3 +14,24 @@ macro which(ex)
...
+
+Use gradient checkpointing on the call `f(xs...)`. This means that
+`checkpointed(f, xs...) === f(xs...)`, but when computing the derivative
+intermediate results from the forward pass of `f` will not be stored. Instead the forward
+pass will be repeated, when computing the derivative.
+This this saves memory at the cost of inreasing exectution time.
```
willtebbutt, line 26:
```suggestion
This this saves memory at the cost of increasing exectution time.
```

src/lib/grad.jl
willtebbutt: @DhairyaLGandhi would you want to export this functionality?

test/runtests.jl
willtebbutt: @DhairyaLGandhi ?

test/runtests.jl
willtebbutt: Ah, sorry, the code I linked in the issue has a bug. Currently, the pullback isn't returning a cotangent for `checkpointed` itself. It should just be nothing, so returning something like
```julia
(nothing, pb(Δy)...)
```
should be sufficient.

willtebbutt (commented):
Thanks for opening. I've asked for Dhairya's input on a few things.

DhairyaLGandhi (commented):
Sorry I missed this PR will take a deeper look

test/runtests.jl
DhairyaLGandhi: This would be good for `features.jl`

src/lib/grad.jl
DhairyaLGandhi: Hmm, I think it's a bit generic sounding to export. It's not obvious which context it is expected to run in.

DhairyaLGandhi:
This would need a better docstring to suggest not using it with yield-ing constructs like FGenerators.jl etc. Thoughts @willtebbutt

willtebbutt:
> This would need a better docstring to suggest not using it with yield-ing constructs like FGenerators.jl etc. Thoughts @willtebbutt

Could you explain this a little? It's not immediately obvious what sort of thing you've got in mind / what sort of problem it would cause.

DhairyaLGandhi:
To be clear, it's probably not a big deal, but for non pure functions you might end up getting different results for the same input. For example, check out the examples in https://github.com/JuliaFolds/FGenerators.jl

DhairyaLGandhi:
Two forward passes may not produce the same code and can therefore cause incorrect gradients

willtebbutt:
Ohhhhh yeah, I get you. Yeah, that should definitely be placed in a big disclaimer in the docstring. Anything involving random number generation is also a great example of this (and has caught me out a couple of times when doing similar things to this!)

jw3126:
Any remaining blockers to this PR from getting merged?

willtebbutt:
Seems fine to me, just requires a patch bump and @DhairyaLGandhi 's approval (since I contributed to the code, I shouldn't approve)

DhairyaLGandhi (approved):

src/lib/grad.jl
```diff
@@ -14,3 +14,27 @@ macro which(ex)
...
+    checkpointed(f, xs...)
+
+Use gradient checkpointing on the call `f(xs...)`. This means that
+`checkpointed(f, xs...) === f(xs...)`, but when computing the derivative
+intermediate results from the forward pass of `f` will not be stored. Instead the forward
+pass will be repeated, when computing the derivative.
+This saves memory at the cost of increasing exectution time.
+
+!!! Warning:
+If `f` is not a pure function, `checkpointed` will likely give wrong results.
```
mcabbott, line 29:
```suggestion
!!! warning
    If `f` is not a pure function, `checkpointed` will likely give wrong results.
```
Right now this produces
```
help?> checkpointed
search: checkpointed

   checkpointed(f, xs...)

  Use gradient checkpointing on the call f(xs...). This means that checkpointed(f, xs...) ===
  f(xs...), but when computing the derivative intermediate results from the forward pass of f will
  not be stored. Instead the forward pass will be repeated, when computing the derivative. This
  saves memory at the cost of increasing exectution time.

  !!! warning: If f is not a pure function, checkpointed will likely give wrong results.
```
while with the suggestion it should produce (with the label in colour)
```
  not be stored. Instead the forward pass will be repeated, when computing the derivative. This
  saves memory at the cost of increasing exectution time.

  │ Warning
  │
  │  If f is not a pure function, checkpointed will likely give wrong results.

julia> 
```
which I think becomes a pretty box in HTML docs. 

Nice PR, btw!

mcabbott:
Zygote doesn't seem to have a NEWS file, but should this be mentioned in the docs? Perhaps in https://github.com/FluxML/Zygote.jl/blob/master/docs/src/utils.md even though that's not perfect.

DhairyaLGandhi:
It can go in https://github.com/FluxML/fluxml.github.io/pull/77 and in the Flux NEWS.md

mcabbott:
CI failure on 1.3 is (I think) because the resolver is confused -- it adds Reexport 1.0 as a full-time dep, and then tries to add CUDA as a test dep, an old version which supports Julia 1.3, which doesn't work. Short-term plan is just to ignore it I guess.

DhairyaLGandhi:
Thanks!

bors r+

PR #888, opened 2021-01-20
Remove rules for matrix exponential

sethaxen:
https://github.com/JuliaDiff/ChainRules.jl/pull/351 added rules for the dense matrix exponential to ChainRules. This PR removes the corresponding adjoint from Zygote.


DhairyaLGandhi:
Looks good, could we also fix the passing test on nightly?

Issue #865, opened 2020-12-30
How to write Jacobian and Hessian functions stripped of ForwardDiff


tholdem:
ForwardDiff doesn't seem to support FFT or complex-type, but Zygote.gradient does. Per this discussion,
https://discourse.julialang.org/t/forwarddiff-and-zygote-cannot-automatically-differentiate-ad-function-from-c-n-to-r-that-uses-fft/52440/4?u=tholdem
I was made aware of that Zygote.hessian uses ForwardDiff. This prevents Zygote to AD the Hessian of my objective function involving FFT and complex-numbers. I don't have much experience with these things and I'm new to Julia. So I don't really know if I can pull off writing a Hessian function myself using Zygote only. I was wondering 1. why would Zygote call ForwardDiff instead of its own functions? 2. if there is more guidance or resource on how to write a Hessian AD function that calls Zygote only that supports complex-numbers and FFT? Thank you so much for your help.

For example, [this code](https://github.com/FluxML/Zygote.jl/pull/747/files) doesn't work because Zygote doesn't support mutating arrays, and it's unclear whether this code handles complex-numbers.
```
function jacobian(f,x)
    y,back  = Zygote.pullback(f,x)
    k  = length(y)
    n  = length(x)
    J  = Matrix{eltype(y)}(undef,k,n)
    e_i = fill!(similar(y), 0)
    @inbounds for i = 1:k
        e_i[i] = oneunit(eltype(x))
        J[i,:] = back(e_i)[1]
        e_i[i] = zero(eltype(x))
    end
    (J,)
end

hessian(f, x) = jacobian(x -> gradient(f, x)[1], x)
```
There seems to be a workaround for mutating arrays, https://github.com/rakeshvar/Zygote-Mutating-Arrays-WorkAround.jl, but it is much harder to find the gradient of the mutating steps in the code above so it would be very daunting to find a workaround myself.

mcabbott:
It would be better if the gradient in #747 used `reduce(hcat, ...)` instead of writing into a matrix, as then it might behave well inside second derivatives. (Although its re-use of `e_i` might also need some trickery.)

However, if re-writing the `hessian` function to use that not the ForwardDiff version, the jacobian would be on the outside, so should not itself be differentiated. Maybe something else is going wrong?

[...]

Issue #564, opened 2020-03-28
RFC: Jacobian of multiple return functions and structured Jacobian etc.?


Roger-luo:
I meet this problem while fixing @willtebbutt 's request in https://github.com/JuliaDiff/FiniteDifferences.jl/pull/54 since I hope we could have the same API in both packages and it seems this problem was not discussed before, I'll post what I'm thinking here

My previous implementation of the jacobian (as well as its original pytorch version) function won't work for multiple return value functions, and I actually never considered it. Now I start wondering which is better

## Option 1: non-structured

the most straightforward way is to convert all the output to a large vector. This is what's implemented in FiniteDifference, but it's not implemented in Zygote.

but this is definitely not something good in more serious calculation since this will cause a lot of intermediate allocations and it lost the original structure of the values

## Option 2: structured

It's more complicated but I feel it will benefit a lot of things

**Definition** *Adjoint type*: every type `T` has an adjoint type `∇(T)`, which is the type of its gradient, and an adjoint type `J(T, R)` which is its jacobian type, where `R` is its return type.

e.g for naive cases, it's usually the original type itself, the adjoint type of `Array` is `Array` and `Matrix`, the adjoint of `Numer` is `Numer` and `Matrix`, `Tuple` is `Tuple` and `Tuple` of `Matrix`. But for `struct`s, it's a `NamedTuple` and in my own case while using Flux, I actually had to define my own type of a model as `(index::Int, grad::Array)` to avoid type instability caused by auto generated gradient `Tuple{Nothing, Matrix}` and `Tuple{Matrix, Nothing}` so far this is just what Zygote is doing current.

first, as we all agree, the Jacobian of multivariant function `f(xs...)::Union{Array, Numer}` is a `Tuple`, where each element is the Jacobian matrix of each `x`, thus we could expand this definition a bit, the Jacobian of the multi-return function `f(x) -> ys...` is a tuple of each `y`'s jacobian, e.g

the jacobian of

```julia
f(x::Array)::Tuple{Array, Array}
```

is

```julia
Tuple{Matrix, Matrix}
```

now, a bit more general, the jacobian type of  a multi return value functions `f(xs...) -> ys...` where each `y` can be `AbstractArray` or `Numer` will have a double layered `Tuple` (also make sense if you take `Tuple` as some sort of array), e.g the jacobian of

```julia
f(x1, x2)::Tuple{Array, Array, Array}
```

is

```julia
[...]

PR #890, opened 2021-01-27
Add `jacobian`, at last?

### LINKED ISSUE #51: How to calculate Jacobians (or just derivatives of vector functions) with Zygote?

### LINKED ISSUE #98: Jacobian and Hessian

### LINKED ISSUE #413: Jacobian in Zygote with Params

### LINKED ISSUE #891: Second derivative of univariate functions: hessian() only takes arrays
Unlike gradient, hessian only has an (f, x::AbstractArray) method, so it won't take Floats. One has to pass an array of length one, and use a wrapper (anonymous) function to extract the value. Example:

```julia
using Zygote

f(x) = (sin(x)/x)

g = x -> Zygote.gradient(f, x)
g(2.0) # OK

h = x -> Zygote.hessian(f, x)
h(2.0) # no method matching..

h = x -> Zygote.hessian(x-> f(x[1]), x)
h([2.0]) # OK
```

Btw, the `f''()` notation fails for my actual function with a _Can't differentiate foreigncall expression_ error

mcabbott:
This adds a Jacobian function. 

Compared to #747 this one:
* has tests
* accepts multiple arguments
* shouldn't fail if `back` returns `nothing`
* inserts `vec` a few more places
* has a method which accepts implicit `Params`, and returns `Grads`. (This was the only part I actually needed in real life.)
* now works on the GPU too!

Compared to #414 this one:
* always inserts `vec`, never makes higher-dimensional arrays
* runs on current Zygote
* has tests.

Compared to #235 this one:
* doesn't try to provide numerical jacobian
* doesn't alter testing infrastructure
* doesn't provide `jacobian!`, nor any code for structured matrices.

This does not address #564's concerns about functions which return a tuple of arrays. Only functions returning an array (or a scalar) are permitted. Similar considerations might give sensible jacobians when the argument of a function is a tuple, or some other struct, but for now these are handled by putting up a giant warning sign.

Nothing in the file `utils.jl` seems to have any tests at all. So I also added tests for `hessian`. 

And, while I was there, made `hessian` actually accept a real number like its docstring promises. (Hence this closes #891.) And, made a version that is reverse-over-reverse, using this `jacobian`, which works less well of course but may as well exist to test things. (See for example #865.) Ideally there would be a pure-Zygote version using its own forward mode, but I didn't write that.

Fixes #51, fixes #98, fixes #413. Closes #747.


DhairyaLGandhi:
I'd prefer something that is somewhat more generic rather than using if blocks for type checking. Not a big deal, but the approach in #747 seems better suited to be extended using our existing tools.

mcabbott:
> approach in #747 

Can you say what differences you see in approach? I tried other things and we can discuss them, but circled back to pretty small adjustments. 

> seems better suited to be extended using our existing tools

Which tools are you referring to?

src/lib/utils.jl
```diff
@@ -102,12 +102,167 @@ end
...
+  out = map(args) do x
+    T = promote_type(eltype(x), eltype(y))
+    dx = x isa AbstractArray ? similar(x, T, length(y), length(x)) :
+      x isa Number ? similar(y, T, length(y)) :
+      nothing
+  end
+  delta = fill!(similar(y), 0)
+  for k in LinearIndices(y)
+    delta[k] = 1
+    grads = back(delta)
```
mcabbott, line 204: I guess this won't be GPU-friendly. And re-using `delta` is probably a premature optimisation, `y` is assumed to be small for this to be worth using at all. Would this be GPU-safe?
```suggestion
  delta = Diagonal(fill!(similar(y), 1))
  for k in LinearIndices(y)
    grads = back(delta[:,k])
```

mcabbott: Done. But CI says my GPU tests fail:

https://buildkite.com/julialang/zygote-dot-jl/builds/58#01e54944-c523-46d1-9aa9-3cd90bc7d96b

src/compiler/interface.jl
```diff
@@ -42,8 +42,17 @@ end
...
 
+"""
+    gradient(f, args...)
+
+Returns a tuple containing `∂f/∂x` for each argument `x`,
+the derivative (for scalar x) or the gradient.
+
+`f(x)` must be a real number, see [`jacobian`](@ref) for array output.
+"""
 function gradient(f, args...)
```
CarloLucibello, line 56: omg, we didn't have a docstring for gradient...

src/Zygote.jl
```diff
@@ -12,7 +12,7 @@ using MacroTools, Requires
 using MacroTools: @forward
 
 import Distributed: pmap, CachingPool, workers
-export Params, gradient, pullback, pushforward, @code_adjoint
+export Params, gradient, jacobian, pullback, pushforward, @code_adjoint
```
CarloLucibello, line 15: we could export hessian as well if we feel confident enough

src/lib/utils.jl
```diff
@@ -102,12 +102,170 @@ end
 """
     hessian(f, x)
 
-Construct the Hessian of `f`, where `x` is a real or real array and `f(x)` is
-a real.
+Construct the Hessian `∂²f/∂x∂x`, where `x` is a real number or an array,
+and `f(x)` is a real number.
```
CarloLucibello, line 106: should this say something on how it handles higher-order arrays?

mcabbott: Something like the output being the shape of `vec(x) .* vec(x)'`? I did put an example marked "uses linear indexing" but could add more words too.

test/runtests.jl
```diff
@@ -2,15 +2,26 @@ using Zygote, Test
 using Zygote: gradient
 using CUDA: has_cuda
 
+if has_cuda()
```
mcabbott, line 5: I moved these tests up, so that CI would fail faster on them. They can go back to being last if anyone feels strongly about it. These are the only optional tests, so getting early feedback about whether they are going to be run or not, locally, doesn't seem like a terrible idea.

src/lib/utils.jl
```diff
@@ -102,12 +102,170 @@ end
...
+hessian_reverse(f, x::Number) = gradient(x -> gradient(f, x)[1], x)[1]
+
 
-    julia> hessian(((a, b),) -> a*b, [2, 3])
-    2×2 Array{Int64,2}:
-     0  1
-     1  0
 """
-hessian(f, x::AbstractArray) = forward_jacobian(x -> gradient(f, x)[1], x)[2]
+    jacobian(f, args...)
```
mcabbott, line 149: Not really sure `utils.jl` is the right place for this. Maybe `grad.jl`?

src/lib/utils.jl
```diff
@@ -102,12 +102,170 @@ end
...
+
+```jldoctest
+julia> jacobian((a,s) -> a.^length(s), [1,2,3], "str")
+([3 0 0; 0 12 0; 0 0 27], nothing)
+
+julia> jacobian((a,t) -> sum(a .* t[1]) + t[2], [1,2,3], (4,5))
+([4 4 4], nothing)
+
+julia> gradient((a,t) -> sum(a .* t[1]) + t[2], [1,2,3], (4,5))  # gradient undersands the tuple
+([4, 4, 4], (6, 1))
```
mcabbott, line 191: Many doctests fail, with things like "UndefVarError: gradient not defined", not sure why?

https://github.com/FluxML/Zygote.jl/pull/890/checks?check_run_id=1780288837

src/Zygote.jl
mcabbott: That would be fine by me, too.

src/lib/utils.jl
```diff
@@ -102,12 +102,170 @@ end
...
+
+_gradcopy!(dst::AbstractArray, src::AbstractArray{<:Number}) = copyto!(dst, src)
+_gradcopy!(dst::AbstractArray, src::Number) = copyto!(dst, src)
+_gradcopy!(dst::AbstractArray, src::Nothing) = dst .= 0
+_gradcopy!(dst::AbstractArray, src::AbstractArray) = copyto!(dst, g isa Number ? g : 0 for g in src) # e.g. Union{Nothing,Float64}
+
+"""
+    jacobian(loss, ::Params)
+
+Like `gradient` with implicit parameters, this method takes a zero-argument function
```
CarloLucibello, line 233:
```suggestion
Like [`gradient`](@ref) with implicit parameters, this method takes a zero-argument function
```

src/lib/utils.jl
CarloLucibello: yes, `vec(y) .* vec(x)'` would make the ordering very clear

src/lib/utils.jl
CarloLucibello: `highorder.jl`?

CarloLucibello:
looks very good, had only minor comments

src/lib/utils.jl
mcabbott: Done.

src/lib/utils.jl
mcabbott: Would prefer not to make too many files. The current organisation seems a little random, `grad.jl` just has `@nograd`... but by name is where I would have looked. 

I would only call the Hessian higher-order, not the Jacobian. But it does seem like a good idea for them to live together.

src/lib/utils.jl
mcabbott: Fixed

src/lib/utils.jl
```diff
@@ -102,12 +102,174 @@ end
...
+_eyelike(y::Vector) = Matrix{eltype(y)}(I, length(y), length(y))
+function _eyelike(y::AbstractVector) # version which works on GPU
+  out = fill!(similar(y, length(y), length(y)), 0)
+  out[LinearAlgebra.diagind(out)] .= 1
+  out
+end
+
+_gradcopy!(dst::AbstractArray, src::AbstractArray{<:Number}) = copyto!(dst, src)
+_gradcopy!(dst::AbstractArray, src::Number) = copyto!(dst, src)
+_gradcopy!(dst::AbstractArray, src::Nothing) = dst .= 0
```
DhairyaLGandhi, line 231: This doesn't seem correct

DhairyaLGandhi (commented):
Why do we need to explicitly copy?

src/lib/utils.jl
mcabbott: See what you think of the current wording, updated in https://github.com/FluxML/Zygote.jl/pull/890/commits/e320b0902e8faa7322da8e52d94bf9dfb21be2ab. This one is a square matrix, but Jacobian has `y`'s rows.

src/lib/utils.jl
mcabbott: Which one? If you delete any of those lines, tests will fail. 

(Actually, if you delete the first, I presume GPU tests will fail, and CPU will just be slow).

mcabbott: > Why do we need to explicitly copy?

Because the jacobian calls the pullback repeatedly, and then needs to combine all those rows into one matrix. This PR follows precisely the approach of #747, in that it makes the output matrix before calling the pullback.

src/lib/utils.jl
```diff
@@ -102,12 +102,170 @@ end
...
 
-    julia> hessian(((a, b),) -> a*b, [2, 3])
-    2×2 Array{Int64,2}:
-     0  1
-     1  0
 """
-hessian(f, x::AbstractArray) = forward_jacobian(x -> gradient(f, x)[1], x)[2]
+    jacobian(f, args...)
+
+For each array `a ∈ args` this returns a matrix with `Ja[k,i] = ∂y[k]/∂a[i]`
```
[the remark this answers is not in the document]
mcabbott: This one I think I prefer to leave.

CarloLucibello:
Julia 1.3 tests failing due to CUDA compatibility issues is annoying. Maybe we should drop < 1.5 support

mcabbott:
I believe that if we move tests to 1.4 they will get the new resolver, maybe that's the minimal step?

src/compiler/interface.jl
```diff
@@ -42,8 +42,17 @@ end
...
+sensitivity(y::AbstractArray) = error("output an array, so the gradient is not defined. Perhaps you wanted jacobian.")
 sensitivity(y) = error("Output should be scalar; gradients are not defined for output $(repr(y))")
 
+"""
+    gradient(f, args...)
+
+Returns a tuple containing `∂f/∂x` for each argument `x`,
+the derivative (for scalar x) or the gradient.
+
+`f(x)` must be a real number, see [`jacobian`](@ref) for array output.
```
CarloLucibello, line 54:
```suggestion
`f(args...)` must be a real number, see [`jacobian`](@ref) for array output.
```

mcabbott:
> This is not quite true, we do deal with non real output in several cases in sciml

MWE?

This docstring just matches the errors:
```
julia> gradient(identity, [1,2])
ERROR: Output should be scalar; gradients are not defined for output [1, 2]

julia> gradient(identity, 1+im)
ERROR: Output is complex, so the gradient is not defined.
```

DhairyaLGandhi:
https://github.com/SciML/DiffEqSensitivity.jl/blob/e04f8a1a300ed2dbc2e0e8da520bb3ef6d9548a2/src/concrete_solve.jl#L279 outputs a vector in the forward pass.

mcabbott:
> https://github.com/SciML/DiffEqSensitivity.jl/blob/e04f8a1a300ed2dbc2e0e8da520bb3ef6d9548a2/src/concrete_solve.jl#L279 outputs a vector in the forward pass.

Sure. But I meant an example for `gradient`, clearly.

src/Zygote.jl
CarloLucibello:
```suggestion
export Params, gradient, jacobian, hessian, pullback, pushforward, @code_adjoint
```
... seems consistent

src/lib/grad.jl
```diff
@@ -38,3 +38,178 @@ function Zygote._pullback(ctx::Zygote.AContext, ::typeof(checkpointed), f, xs...
...
+ 1.0  0.0
+
+julia> Zygote.hessian(x -> sum(x.^3), [1 2; 3 4])  # uses linear indexing of x
+4×4 Array{$Int,2}:
+ 6   0   0   0
+ 0  18   0   0
+ 0   0  12   0
+ 0   0   0  24
+
+julia> Zygote.hessian(sin, pi/2)
```
mcabbott, line 68: If exported, it need not be qualified. Goes with https://github.com/FluxML/Zygote.jl/pull/890#discussion_r567702074
```suggestion
julia> hessian(x -> x[1]*x[2], randn(2))
2×2 Array{Float64,2}:
 0.0  1.0
 1.0  0.0

julia> hessian(x -> sum(x.^3), [1 2; 3 4])  # uses linear indexing of x
4×4 Array{$Int,2}:
 6   0   0   0
 0  18   0   0
 0   0  12   0
 0   0   0  24

julia> hessian(sin, pi/2)
```

DhairyaLGandhi:
I mean, I've shown you it working in the ecosystem

mcabbott:
> I mean, I've shown you it working in the ecosystem

Sorry, what does "it" refer to here? 

The file you linked to does not seem to contain the function `gradient`. It contains many other functions, and I totally agree that other functions can return vector arguments. In fact, in my example, there is also such a function!

CarloLucibello (approved):

PR #892, opened 2021-01-28
Store the primal value in Grads

mcabbott:
This alters the `Grads` struct to also keep the forward value. Occasionally this will save you from having to write `y,b=pullback(...); b(1.0)` (or is it `b(1f0)` today?). It's already calculated, there seems no reason not to return it, in case it's useful. Should  be non-breaking. 

For `gradient(f, args...)` with explicit arguments, no change. There too it would sometimes be nice to have the primal available, but since it just returns a tuple there's no super-easy way to add it.

### DIFF
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -123,12 +123,13 @@ function copy!(x::AbstractVector, ps::Params)
 end
 
 
-struct Grads
+struct Grads{T}
   grads::IdDict{Any,Any}
   params::Params
+  value::T
 end
 
-Base.show(io::IO, ps::Grads) = print(io, "Grads(...)")
+Base.show(io::IO, gs::Grads) = print(io, "Grads([...], value = ", gs.value, ")")
 
 @forward Grads.grads Base.getindex, Base.haskey
 
@@ -170,7 +171,7 @@ function pullback(f, ps::Params)
       cache(cx)[p] = nothing
     end
     back(Δ)
-    Grads(cx.cache, ps) # TODO make a copy
+    Grads(cx.cache, ps, y) # TODO make a copy
   end
 end
 
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -122,12 +122,22 @@ function copy!(x::AbstractVector, ps::Params)
   ps
 end
 
+"""
+    gradient(() -> loss(), ::Params) -> Grads
+
+Gradient with implicit parameters. Returns a container, from which
+`grads[W]` extracts the gradient with respect to `W`.
+This contains as `grads.value` the value of the function.
+"""
+function gradient end
 
 struct Grads{T}
   grads::IdDict{Any,Any}
   params::Params
   value::T
 end
+
+Grads(gs, ps) = Grads(gs, ps, nothing)
 
 Base.show(io::IO, gs::Grads) = print(io, "Grads([...], value = ", gs.value, ")")
 
--- a/test/features.jl
+++ b/test/features.jl
@@ -296,6 +296,7 @@ let
     p′[1]
   end
   @test θ̄[p][1] == 1
+  @test θ̄.value == 1
 end
 
 @test gradient(2) do x


DhairyaLGandhi:
This muddies the meaning of the struct and the pullback seems to be simple enough and explicit. 

Cc @oxinabox, @willtebbutt for their thoughts

oxinabox:
I am not convinced that this is a good thing to do.
Please tell me more of the motivation

`Grads` is gradent.
If you want the primal then it should be `PrimalAndGrads` (matching swifts `ValueAndGrad` iirc)

The extra pointers are cheap, but they are not free.


`Zygote.pullback` exists and is useful.
It should be called `Zygote.primal_and_pullback` but that is another topic.
A convience wrapper around it that using the basis vector `Zygote.primal_and_gradient` would be good also.
(again IIRC matches to a thing Swift has?)

mcabbott:
We can rename it to `PrimalAndGrads` but that's a much bigger step.

It is very, very, close to free. It's one `Float64` or whatever as part of the struct. A struct which contains a dictionary (typed Any) and whatever Params is. I don't think there should be any new pointers. The implicit params story is not a tight-inner-loop gadget.

The main motivation is that dropping down from `g = gradient(loss(m), params(m))` to dealing with `pullback` seems like stepping down to a lower-level interface, where suddenly you need to give this function a name and produce the appropriate input for it and call it to get the gradient... and that all feels like boilerplate, just to get the one number it threw away.

DhairyaLGandhi:
> Is there a non-breaking way to keep the primal when calling gradient

`pullback`

DhairyaLGandhi:
> What I don't like is the asymmetry between implicit and explicit parameters

Can you elaborate? Btw we don't *have* to use the implicit mode at all

mcabbott:
>> when calling `gradient(f, xs...)`

> `Zygote.pullback`

Yes, thank you, I have noticed. 

> Can you elaborate?

After this PR, there is a convenient way to extract `y` without dropping down to lower-level functions, for implicit gradients. There is no such way provided for the explicit case. That's the asymmetry.

src/compiler/interface.jl
```diff
@@ -122,13 +122,24 @@ function copy!(x::AbstractVector, ps::Params)
   ps
 end
 
+"""
+    gradient(() -> loss(), ::Params) -> Grads
+
+Gradient with implicit parameters. Returns a container, from which
+`grads[W]` extracts the gradient with respect to `W`.
+This contains as `grads.value` the value of the function.
```
CarloLucibello, line 130: `value` referred to gradient can be confusing. Can we think of a better name? `output`? 
or maybe is fine already

CarloLucibello:
This adds some convenience with so little effort and essentially no side-effects, I think we should have it. 

> It would need to be some struct which acts like a tuple, instead of like a dictionary.

Asymmetry is a pity indeed. So you suggest the explicit gradient returning a tuple-like struct that also has a `value` field? That would be essentially non-breaking, except for the code relying on the fact that `grads isa Tuple` obviously

mcabbott:
I think we should not mess with explicit-arguments `gradient`. It would be too weird to return some strange thing which isn't a tuple. And that function can be used in pretty hot loops, `@btime gradient(sin, pi/2)` is 1.5ns, so we should be cautious about adding extra baggage.

But the implicit one is already a custom struct. "some convenience with so little effort and essentially no side-effects" is precisely my take, too.

src/compiler/interface.jl
mcabbott: Could be `output`, or `primal`, is that an official term or one we made up in Julia? Maybe `forward` is another option?

`value` is what ForwardDiff uses, but I guess it's usage is in fact a little different. I mean the elements apart from the value are its partials, numbers of the same units/shape/etc. While the other parts here are the actual gradients (and params), completely different objects.

DhairyaLGandhi:
I think grads is not the place to introduce this. I agree with @oxinabox about the intent.

The assymetry that you mention seems to mean that you want a separate struct that can map some params, and transform to primal, which this doesn't solve anyway.

DhairyaLGandhi:
The custom-nesd of struct is not a convincing argument to me. Structs don't imply slowness or have performance related contracts associated with them

jonathan-laurent:
I do not have an opinion on the proposed way to achieve this, but I agree that there should be an easy way to access both the loss and gradient at the same time. This is needed in most ML pipelines and making two separate calls is wasteful.

Ideally, such a function should be available from Flux without the user needing to import Zygote explicitly.

Currently, I am using the following in my own codebases:
```julia
function gradloss(f, args...)
  val, back = Zygote.pullback(f, args...)
  grad = back(Zygote.sensitivity(val))
  return val, grad
end
```

ToucheSir:
For reference, [here](https://www.tensorflow.org/swift/api_docs/Functions#valuewithgradientat:in:) is the thing Swift has (had?) and [here's](https://github.com/tensorflow/swift-models/blob/main/Examples/Custom-CIFAR10/main.swift) an example of it in use.

DhairyaLGandhi:
What @jonathan-laurent has suggested is the way to go for such a utility. Prolly calling with `val_and_grad` is something folks are willing to accept?

ToucheSir:
I'm less invested in the bikeshedding than replacing any docs than replacing any occurrence of `gradient` in the docs with this where applicable :)

PR #893, opened 2021-01-31
Bump lower bound on Requires version

mohamed82008:
`Requires.@include` is used in Zygote but it is not available before v1.1 so I am bumping the lower bound version on Requires.


DhairyaLGandhi:
Since we use the `@include` macro, that seems like a cogent thing to lower bound on. Thanks +1

DhairyaLGandhi (approved):
bors r+

DhairyaLGandhi:
Will release a patch once its merged

PR #894, opened 2021-02-01
drop julia 1.3 support

CarloLucibello:
We are having package version resolution issues when testing on julia 1.3, see https://github.com/FluxML/Zygote.jl/pull/890. 
Since we cannot do CI on julia 1.3, let's drop support entirely

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.3' # Replace this with the minimum Julia version that your package supports.
+          - '1.4' # Replace this with the minimum Julia version that your package supports.
           - '1'   # automatically expands to the latest stable 1.x release of Julia
           - 'nightly'
         os:
--- a/Project.toml
+++ b/Project.toml
@@ -32,7 +32,7 @@ NaNMath = "0.3"
 Requires = "0.5, 1.0"
 SpecialFunctions = "0.10, 1.0"
 ZygoteRules = "0.2.1"
-julia = "1.3"
+julia = "1.4"
 
 [extras]
 CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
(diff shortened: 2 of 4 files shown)


DhairyaLGandhi:
I don't think #890 is a sufficient reason to do that, besides, the issue is with cuda needing to do another compat release.

CarloLucibello:
the problem is not #890, any PR  now has failing CI on 1.3. If cuda.jl has some planned backport addressing the compact issue fine, we can postpone dropping 1.3, I don't see it coming though

DhairyaLGandhi:
It's easy, and considerate to offer a compat bump to their `compat` branch. It would be merged and released quickly. We have been good about that.

CarloLucibello:
Of course. What should I bump there?

CarloLucibello:
Actually the `compat` branch of CUDA.jl is julia >= 1.5 only. We have to merge this

DhairyaLGandhi:
It was AbstractFFTs, and maybe GPUArrays iirc, please check the actual failure for the resolver error.

DhairyaLGandhi:
I am able to add CUDA v1.3.3 with latest Zygote just fine on julia 1.3.0

CarloLucibello:
probably it conflicts with abstractffts in the test environment. Try running `] test`

CarloLucibello:
we cannot add a specific package version when testing a specific julia version, the package solver has to figure it out

DhairyaLGandhi:
The resolver does figure it out :)

This is using the local clone of CUDA#master
image

This is with the resolver
image

This is the test (its still going on, but I've run this enough times just in the past week to know that it will eventually pass all the CUDA tests also)
image

CarloLucibello:
I think the 1.3 resolver is not smart enough to add packages in the right order: if you don't add CUDA explicitly before testing, test should fail. Could you check? I don't know a workaround. Maybe heavily mess with the ci workflow

DhairyaLGandhi:
Nope that would be alright too
image


We don't need to heavily mess with the CI, I'll come back to this in a couple days, since its not too breaking. The users shouldn't face this.

CarloLucibello:
> We don't need to heavily mess with the CI, I'll come back to this in a couple days, since its not too breaking. The users shouldn't face this.

Looks like it didn't take a couple of days and we still have broken CI. Can we merge this?

DhairyaLGandhi:
I've made the compat changes on GPUArrays/ CUDA for the AbstractFFT support and on General as well.

CarloLucibello:
> I've made the compat changes on GPUArrays/ CUDA for the AbstractFFT support and on General as well.

What do you mean? where are these changes? please use crossrefs

DhairyaLGandhi:
Retroactive compat bound fixes are at https://github.com/JuliaRegistries/General/pull/30394

For the future releases: https://github.com/JuliaGPU/CUDA.jl/pull/721 and https://github.com/JuliaGPU/GPUArrays.jl/tree/dg/compat

PR #896, opened 2021-02-03
Gradients for `eachcol` & friends

mcabbott:
This is an optimisation to work around `getindex`'s gradient. In a few small tests it was twice as fast.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -331,6 +331,27 @@ end
   end
 end
 
+@adjoint eachrow(x::AbstractVecOrMat) = collect(eachrow(x)), dys -> ∇eachslice(dys, x, 1)
+@adjoint eachcol(x::AbstractVecOrMat) = collect(eachcol(x)), dys -> ∇eachslice(dys, x, 2)
+@adjoint eachslice(x::AbstractArray; dims::Integer) =
+  collect(eachslice(x; dims=dims)), dys -> ∇eachslice(dys, x, dims)
+
+function ∇eachslice(dys, x::AbstractArray, dim::Integer) where {TX}
+  i1 = findfirst(dy -> dy isa AbstractArray, dys)
+  i1 === nothing && return (zero(x),) # all slices get nothing
+  T = promote_type(eltype(dys[i1]), eltype(x))
+  dx = similar(x, T)
+  for i in axes(x, dim)
+    if dys[i] isa AbstractArray
+      copyto!(selectdim(dx,dim,i), dys[i])
+    else
+      selectdim(dx,dim,i) .= 0
+    end
+  end
+  (dx,)
+end
+
+
 # LinearAlgebra
 # =============
 
--- a/test/gradcheck.jl
+++ b/test/gradcheck.jl
@@ -190,6 +190,22 @@ end
   @test gradient(g, ones(3)) == ([1,0,0],)
 end
 
+@testset "eachcol" begin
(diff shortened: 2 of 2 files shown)


src/lib/array.jl
```diff
@@ -331,6 +331,27 @@ end
   end
 end
 
+@adjoint eachrow(x::AbstractVecOrMat) = collect(eachrow(x)), dys -> ∇eachslice(dys, x, 1)
+@adjoint eachcol(x::AbstractVecOrMat) = collect(eachcol(x)), dys -> ∇eachslice(dys, x, 2)
```
DhairyaLGandhi, line 335: Why do we need to collect? Seems like it's not used in the backwards pass

DhairyaLGandhi (commented):
Thanks, I think this is generally good idea to support

src/lib/array.jl
mcabbott: Without it, the gradient is (sometimes) some weird NamedTuple containing fields of the generator, and I got tired of trying to sort that out.

For a square matrix, it's one allocation of length N. The reverse pass is at least N such vector allocations. Master has, in addition, N matrix allocations.

DhairyaLGandhi: I'm aware of what master does, but this would break any laziness contract of the function which code user code could need.

mcabbott: I'm doubtful, but maybe you have an example in mind? 

Obviously one could make a pathalogical `@assert x isa Generator` but I mean plausibly useful code. 

Note that anything which broadcasts already collects.

DhairyaLGandhi: What cases do you see the gradient for the generator in? Maybe handling that would solve more cases than without.

One could think of several bad examples, of things like lazy arrays or "infinite" arrays that are generated lazily.

mcabbott: When last I tried, Zygote + LazyArrays didn't work well at all. But maybe things have improved, if you have an example in mind, please share.

The obvious use case is `reduce(hcat, map(fun, eachcol(x)))` which is a faster `mapslices`, and would now have a (fairly) fast gradient. The lazy version `reduce(hcat, fun(xi) for xi in eachcol(x))` is much slower, without mentioning Zygote.

DhairyaLGandhi: Last I tried it needed a couple basic adjoints around `apply` and that helped with a large chunk of it.

test/gradcheck.jl
```diff
@@ -190,6 +190,22 @@ end
...
+    @test gradtest(x -> map(norm, eachcol(x)), (3,4))
+    @test gradtest(x -> map(norm, eachrow(x)), (3,4))
+    @test gradtest(x -> map(norm, eachslice(x, dims=3)), (3,4,5))
+
+    # some slices may have gradient nothing
+    @test gradient(x -> sum(y -> rand()>0.5 ? 0 : first(y), eachcol(x)), rand(3,10))[1] isa Matrix
+
+    # strange errors
+    @test_skip gradient(x -> sum(norm, eachcol(x)), [1 2 3; 4 5 6])[1] isa Matrix
+    @test_skip gradient(x -> sum(norm, eachcol(x)), rand(3,400))[1] isa Matrix
```
CarloLucibello, line 206: these are broken or erratic?

mcabbott: This one is #897, thus not broken on 1.5 I think.

PR #901, opened 2021-02-18
remove Tracker

CarloLucibello:
In an effort to remove Requires dependence at some point, this remove Tracker, since its been essentially deprecated for quite some time. 

Also removes a very outdated Flux MLP example.


CarloLucibello:
closing as maybe we want to have a weak dependency instead of removing

PR #902, opened 2021-02-18
add map/broadcast/algebra/iteration/dict interface for Grads

CarloLucibello:
Fix https://github.com/FluxML/Flux.jl/issues/707

### DIFF
--- a/docs/src/utils.md
+++ b/docs/src/utils.md
@@ -24,10 +24,10 @@ Zygote.checkpointed
 
 `Params` and `Grads` can be copied to and from arrays using the `copy!` function.
 
-### Operations with Grads
+## Working with Grads
 
 Map, broadcast, and iteration are supported for the dictionary-like `Grads` objects.
-These operations preserve the keys and transform the values.
+These operations are value based and preserve the keys.
 
 ```julia
 using Zygote, Test
@@ -47,4 +47,8 @@ map(x -> clamp.(x, -0.1, 0.1), gs)
 
 # clip gradients in-place
 foreach(x -> clamp!(x, -0.1, 0.1), gs)
+
+for (p, g) in pairs(gs)
+  # do something with parameter `p` and corresponding gradient `g`
+end
 ```
--- a/docs/src/utils.md
+++ b/docs/src/utils.md
@@ -38,7 +38,7 @@ gs1 = gradient(() -> sum(tanh.(w .* x1 .+ b)), Params([w, b]))
 gs2 = gradient(() -> sum(tanh.(w .* x2 .+ b)), Params([w, b])) 
 
 # accumulate gradients
-gs = gs .+ gs
+gs = gs1 .+ gs2
 @test gs[w] ≈ gs1[w] + gs2[w] 
 @test gs[b] ≈ gs1[b] + gs2[b] 
 
--- a/docs/src/utils.md
+++ b/docs/src/utils.md
@@ -26,23 +26,25 @@ Zygote.checkpointed
 
 ### Operations with Grads
 
-Map and broadcast are supported for the dictionary-like `Grads` object.
+Map, broadcast, and iteration are supported for the dictionary-like `Grads` objects.
+These operations preserve the keys and transform the values.
+
 ```julia
 using Zygote, Test
 
-w = rand(2)
-x1 = rand(2)
-x2 = rand(2)
-b = rand(2)
+w, x1, x2, b = rand(2), rand(2), rand(2), rand(2)
 
-gs1 = gradient(() -> sum(w .* x1 .+ b), Params([w])) 
-gs2 = gradient(() -> sum(w .* x2  .+ b), Params([w])) 
+gs1 = gradient(() -> sum(tanh.(w .* x1 .+ b)), Params([w, b])) 
+gs2 = gradient(() -> sum(tanh.(w .* x2 .+ b)), Params([w, b])) 
 
 # accumulate gradients
 gs = gs .+ gs
 @test gs[w] ≈ gs1[w] + gs2[w] 
 @test gs[b] ≈ gs1[b] + gs2[b] 
 
+# clip gradients
+map(x -> clamp.(x, -0.1, 0.1), gs)
+
 # clip gradients in-place
-map!(x -> clamp!(x, -0.1, 0.1), gs, gs)
+foreach(x -> clamp!(x, -0.1, 0.1), gs)
 ```
--- a/docs/src/utils.md
+++ b/docs/src/utils.md
@@ -23,3 +23,26 @@ Zygote.checkpointed
 ```
 
 `Params` and `Grads` can be copied to and from arrays using the `copy!` function.
+
+### Operations with Grads
+
+Map and broadcast are supported for the dictionary-like `Grads` object.
+```julia
+using Zygote, Test
+
+w = rand(2)
+x1 = rand(2)
+x2 = rand(2)
+b = rand(2)
+
+gs1 = gradient(() -> sum(w .* x1 .+ b), Params([w])) 
+gs2 = gradient(() -> sum(w .* x2  .+ b), Params([w])) 
+
+# accumulate gradients
+gs = gs .+ gs
+@test gs[w] ≈ gs1[w] + gs2[w] 
+@test gs[b] ≈ gs1[b] + gs2[b] 
+
+# clip gradients in-place
+map!(x -> clamp!(x, -0.1, 0.1), gs, gs)
+```
--- a/docs/src/utils.md
+++ b/docs/src/utils.md
@@ -54,4 +54,9 @@ foreach(x -> clamp!(x, -0.1, 0.1), gs)
 for (p, g) in pairs(gs)
   # do something with parameter `p` and corresponding gradient `g`
 end
+
+# note that gradients must be w.r.t. to the same parameter key set
+gs3 = gradient(() -> sum(tanh.(w .* x2)), Params([w]))
+# gs3 does not have the key b
+@test_throws ArgumentError gs1 .+ gs3
 ```
--- a/docs/src/utils.md
+++ b/docs/src/utils.md
@@ -42,6 +42,9 @@ gs = gs1 .+ gs2
 @test gs[w] ≈ gs1[w] + gs2[w] 
 @test gs[b] ≈ gs1[b] + gs2[b] 
 
+# gradients and dictionaries interact nicely
+gs .+= Dict(p => randn(size(p)) for p in keys(gs)) 
+
 # clip gradients
 map(x -> clamp.(x, -0.1, 0.1), gs)
 
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -151,7 +151,7 @@ Base.pairs(gs::Grads) =  (p => gs.grads[p] for p in gs.params)
 
 function Base.iterate(gs::Grads, state...)
   res = iterate(gs.params, state...)
-  res === nothing && return nothing
+  isnothing(res) && return nothing
   p, next_state = res
   return gs[p], next_state
 end
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -140,13 +140,14 @@ end
 
 Base.show(io::IO, ps::Grads) = print(io, "Grads(...)")
 
-@forward Grads.grads  Base.haskey, Base.setindex!
+@forward Grads.grads  Base.setindex!
 @forward Grads.params  Base.length
 
 const ADictOrGrads = Union{AbstractDict, Grads}
 
 # Dictionary interface.
-# Don't use the IdDict directly since it may contain some spurious pairs. 
+# Don't use the IdDict directly since it may contain some spurious pairs.
+Base.haskey(gs::Grads, x) = x ∈ gs.params 
 Base.keys(gs::Grads) = gs.params
 Base.values(gs::Grads) = (gs.grads[p] for p in gs.params)
 
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -143,6 +143,9 @@ Base.show(io::IO, ps::Grads) = print(io, "Grads(...)")
 @forward Grads.grads  Base.haskey, Base.setindex!
 @forward Grads.params  Base.length
 
+const ADictOrGrads = Union{AbstractDict, Grads}
+
+
 # Dictionary interface.
 # Don't use the IdDict directly since it may contain some spurious pairs. 
 Base.keys(gs::Grads) = gs.params
@@ -185,7 +188,7 @@ function copy!(x::AbstractVector,  gs::Grads)
   x
 end
 
-broadcasted(f, gss::Grads...) = map(f, gss...)
+broadcasted(f, gs::Grads, gss::ADictOrGrads...) = map(f, gs, gss...)
 
 for op in (:*, :/)
   @eval broadcasted(::typeof($op), a::Number, gs::Grads) = map(x -> $op(a, x), gs)
@@ -201,13 +204,14 @@ function materialize!(gs1::Grads, gs2::Grads)
   return gs1
 end
 
-function Base.map(f, gs1::Grads, gss::Grads...)
+
+function Base.map(f, gs1::Grads, gss::ADictOrGrads...)
   gsout = Grads(IdDict{Any,Any}(), Params(gs1.params))
   return map!(f, gsout, gs1, gss...)
 end
 
-function Base.map!(f, gsout::Grads, gss::Grads...)
-  all(issetequal(gsout.params, gs.params) for gs in gss) || 
+function Base.map!(f, gsout::Grads, gss::ADictOrGrads...)
+  all(issetequal(gsout.params, keys(gs)) for gs in gss) || 
     throw(ArgumentError("map! expects Grads objects with the same Params."))
   for p in gsout.params
     gsout[p] = f((_getformap(gs, p) for gs in gss)...) 
@@ -215,7 +219,7 @@ function Base.map!(f, gsout::Grads, gss::Grads...)
   return gsout
 end
 
-function _getformap(gs::Grads, p)
+function _getformap(gs, p)
   g = gs[p]
   isnothing(g) ? fill!(similar(p), 0) : g 
 end
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -146,8 +146,7 @@ Base.show(io::IO, ps::Grads) = print(io, "Grads(...)")
 # Dictionary interface.
 # Don't use the IdDict directly since it may contain some spurious pairs. 
 Base.keys(gs::Grads) = gs.params
-Base.values(gs::Grads) =  (gs.grads[p] for p in gs.params)
-Base.pairs(gs::Grads) =  (p => gs.grads[p] for p in gs.params)
+Base.values(gs::Grads) = (gs.grads[p] for p in gs.params)
 
 function Base.iterate(gs::Grads, state...)
   res = iterate(gs.params, state...)
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -145,7 +145,6 @@ Base.show(io::IO, ps::Grads) = print(io, "Grads(...)")
 
 const ADictOrGrads = Union{AbstractDict, Grads}
 
-
 # Dictionary interface.
 # Don't use the IdDict directly since it may contain some spurious pairs. 
 Base.keys(gs::Grads) = gs.params
@@ -190,10 +189,8 @@ end
 
 broadcasted(f, gs::Grads, gss::ADictOrGrads...) = map(f, gs, gss...)
 
-for op in (:*, :/)
-  @eval broadcasted(::typeof($op), a::Number, gs::Grads) = map(x -> $op(a, x), gs)
-  @eval broadcasted(::typeof($op), gs::Grads, a::Number) = map(x -> $op(x, a), gs)
-end
+broadcasted(f, a::Numeric, gs::Grads) = map(x -> f(a, x), gs)
+broadcasted(f, gs::Grads, a::Numeric) = map(x -> f(x, a), gs)
 
 function materialize!(gs1::Grads, gs2::Grads)
   issetequal(gs1.params, gs2.params) || 
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -2,6 +2,7 @@ using InteractiveUtils
 using InteractiveUtils: typesof
 using Core: Typeof
 import Base: copy!
+import Base.Broadcast: broadcasted
 
 mutable struct Context <: AContext
   cache::Union{IdDict{Any,Any},Nothing}
@@ -139,7 +140,8 @@ end
 
 Base.show(io::IO, ps::Grads) = print(io, "Grads(...)")
 
-@forward Grads.grads Base.getindex, Base.haskey
+@forward Grads.grads  Base.haskey
+@forward Grads.params  Base.length
 
 function Base.getindex(gs::Grads, x)
   isbits(x) && error("Only reference types can be differentiated with `Params`.")
@@ -169,6 +171,39 @@ function copy!(x::AbstractVector,  gs::Grads)
       i += length(p)
   end
   x
+end
+
+# # Gradient Algebra: unary 
+# for op in (:+, :-)
+#   @eval broadcasted(::typeof($op), gs::Grads) = map($op, gs)
+# end
+
+# # Gradient Algebra: binary 
+# for op in (:+, :-)
+#   @eval broadcasted(::typeof($op), gs1::Grads, gs2::Grads) = map($op, gs1, gs2)
+# end
+
+broadcasted(f, gss::Grads...) = map(f, gss...)
+broadcasted(f, gs1::Grads, xs...) = map(f, gs1, xs...)
+broadcasted(::typeof(*), a::Number, gs::Grads) = map(*, gs, a)
+
+function Base.map(f, gs1::Grads, gss::Grads...)
+  @assert all(issetequal(gs1.params, gs.params) for gs in gss)
+  grads = IdDict{Any,Any}()
+  ps = Params(gs1.params)
+  for p in ps
+    grads[p] = f(gs1[p], (gs[p] for gs in gss)...) 
+  end
+  return Grads(grads, ps)
+end
+
+function Base.map(f, gs1::Grads, xs...)
+  grads = IdDict{Any,Any}()
+  ps = Params(gs1.params)
+  for p in ps
+    grads[p] = f(gs1[p], xs...) 
+  end
+  return Grads(grads, ps)
 end
 
 function pullback(f, ps::Params)
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -143,6 +143,19 @@ Base.show(io::IO, ps::Grads) = print(io, "Grads(...)")
 @forward Grads.grads  Base.haskey, Base.setindex!
 @forward Grads.params  Base.length
 
+# Dictionary interface.
+# Don't use the IdDict dircetly since it sometimes contains some spurios pairs. 
+Base.keys(gs::Grads) = gs.params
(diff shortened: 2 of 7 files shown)


src/compiler/interface.jl
```diff
@@ -171,6 +173,39 @@ function copy!(x::AbstractVector,  gs::Grads)
...
+  @assert all(issetequal(gs1.params, gs.params) for gs in gss)
+  grads = IdDict{Any,Any}()
+  ps = Params(gs1.params)
+  for p in ps
+    grads[p] = f(gs1[p], (gs[p] for gs in gss)...) 
+  end
+  return Grads(grads, ps)
+end
+
+function Base.map(f, gs1::Grads, xs...)
```
CarloLucibello, line 200: I'm not sure this is ok to do, I wonder if not specifying the type of `xs` could lead to ambiguity errors

darsnack: Yeah this seems unclear when `xs` contains some `Grads` with other types too.

src/compiler/interface.jl
```diff
@@ -171,6 +173,39 @@ function copy!(x::AbstractVector,  gs::Grads)
...
+
+broadcasted(f, gss::Grads...) = map(f, gss...)
+broadcasted(f, gs1::Grads, xs...) = map(f, gs1, xs...)
+broadcasted(::typeof(*), a::Number, gs::Grads) = map(*, gs, a)
+
+function Base.map(f, gs1::Grads, gss::Grads...)
+  @assert all(issetequal(gs1.params, gs.params) for gs in gss)
+  grads = IdDict{Any,Any}()
+  ps = Params(gs1.params)
+  for p in ps
```
darsnack, line 194:
```suggestion
  for p in keys(gs1)
```

src/compiler/interface.jl
```diff
@@ -171,6 +173,39 @@ function copy!(x::AbstractVector,  gs::Grads)
...
+# end
+
+# # Gradient Algebra: binary 
+# for op in (:+, :-)
+#   @eval broadcasted(::typeof($op), gs1::Grads, gs2::Grads) = map($op, gs1, gs2)
+# end
+
+broadcasted(f, gss::Grads...) = map(f, gss...)
+broadcasted(f, gs1::Grads, xs...) = map(f, gs1, xs...)
+broadcasted(::typeof(*), a::Number, gs::Grads) = map(*, gs, a)
```
darsnack, line 188: These seem useful but I'm worried about the ambiguity and the asymmetry. Since `Grads` behaves like a `Dict` and not an `AbstractArray`, it's unclear what the correct broadcasting semantics are.

darsnack (commented):
Looks good, thanks for taking care of this! My comment is that since `Grads` behaves like a `Dict`, I think we should limit this to operations involving only `Grads` (where we can verify the key set equality completely).

src/compiler/interface.jl
CarloLucibello: I'd like to support at least `0.1 .* gs1  .+ 0.9 .* gs2`, can remove more general broadcast

src/compiler/interface.jl
CarloLucibello: I need `ps` to construct the output Grads object

CarloLucibello:
I'm trying to figure out how to handle `nothing`

CarloLucibello:
> I'm trying to figure out how to handle nothing

I decided to materialize `nothing` into a zeros array when involved in `map`

src/compiler/interface.jl
```diff
@@ -171,6 +173,47 @@ function copy!(x::AbstractVector,  gs::Grads)
   x
 end
 
+broadcasted(f, gss::Grads...) = map(f, gss...)
+
+for op in (:*, :/)
+  @eval broadcasted(::typeof($op), a::Number, gs::Grads) = _mapscalar($op, a, gs)
+  @eval broadcasted(::typeof($op), gs::Grads, a::Number) = _mapscalar($op, gs, a)
```
mcabbott, line 180: It might be simpler to route these to `map(x->op.(a,x), gs)` etc?

docs/src/utils.md
```diff
@@ -23,3 +23,26 @@ Zygote.checkpointed
...
+gs1 = gradient(() -> sum(w .* x1 .+ b), Params([w])) 
+gs2 = gradient(() -> sum(w .* x2  .+ b), Params([w])) 
+
+# accumulate gradients
+gs = gs .+ gs
+@test gs[w] ≈ gs1[w] + gs2[w] 
+@test gs[b] ≈ gs1[b] + gs2[b] 
+
+# clip gradients in-place
+map!(x -> clamp!(x, -0.1, 0.1), gs, gs)
```
mcabbott, line 47: Ideally this would be `foreach(x -> clamp!(x, -0.1, 0.1), gs)`, maybe that should be allowed?

docs/src/utils.md
```diff
@@ -23,3 +23,26 @@ Zygote.checkpointed
...
+### Operations with Grads
+
+Map and broadcast are supported for the dictionary-like `Grads` object.
+```julia
+using Zygote, Test
+
+w = rand(2)
+x1 = rand(2)
+x2 = rand(2)
+b = rand(2)
```
mcabbott, line 36: In the interest of less scrolling:
```suggestion
w, x1, x2, b = rand(2), rand(2), rand(2), rand(2);
```

docs/src/utils.md
```diff
@@ -23,3 +23,26 @@ Zygote.checkpointed
...
+
+gs1 = gradient(() -> sum(w .* x1 .+ b), Params([w])) 
+gs2 = gradient(() -> sum(w .* x2  .+ b), Params([w]))
```
mcabbott, line 39: I think?
```suggestion
gs1 = gradient(() -> sum(w .* x1 .+ b), Params([w, b])) 
gs2 = gradient(() -> sum(w .* x2 .+ b), Params([w, b])) 
```

src/compiler/interface.jl
```diff
@@ -171,6 +173,31 @@ function copy!(x::AbstractVector,  gs::Grads)
...
+  @eval broadcasted(::typeof($op), gs::Grads, a::Number) = map(x -> $op.(x, a), gs)
+end
+
+function Base.map(f, gs1::Grads, gss::Grads...)
+  gsout = Grads(IdDict{Any,Any}(), Params(gs1.params))
+  return map!(f, gsout, gs1, gss...)
+end
+
+function Base.map!(f, gsout::Grads, gss::Grads...)
+  @assert all(issetequal(gsout.params, gs.params) for gs in gss)
```
mcabbott, line 189:
```suggestion
  all(issetequal(gsout.params, gs.params) for gs in gss) || throw(ArgumentError("map! expected several Grads objects for the same Params"))
```

docs/src/utils.md
CarloLucibello: we should define iteration at this point. 
Julia's `Dict`s iterate over pairs. 
While dictionaries here https://github.com/andyferris/Dictionaries.jl iterate over values. 
Seems more convenient for us to iterate over values, 
and use `pairs(gs)` if one wants pairs.

DhairyaLGandhi: Following base conventions is important to keep the same semantics for when you have arbitrary code samples to contend with. Yeah, Julia should potentially have a method to iterate over values, but seems weird to have a dictionary when you need the values and not the keys.

darsnack: You can already iterate `Grads` with `pairs(gs)`. Given that we are using `IdDict`, it makes sense to stick with those semantics.

`Params` has an `order` field, does that get respected by `pairs(gs)` (I don't think so).

src/compiler/interface.jl
```diff
@@ -171,6 +173,31 @@ function copy!(x::AbstractVector,  gs::Grads)
...
+  for p in gsout.params
+    gsout[p] = f((_getformap(gs, p) for gs in gss)...) 
+  end
+  return gsout
+end
+
+function _getformap(gs::Grads, p)
+  g = gs[p]
+  g === nothing ? fill!(similar(p), 0) : g
```
darsnack, line 198:
```suggestion
  isnothing(g) ? fill!(similar(p), 0) : g 
```

docs/src/utils.md
darsnack: The reason for iterating over pairs is that ordering is not guaranteed. Those iteration semantics prevent accidentally iterating two dictionaries thinking their ordering is the same.

I think the same applies here. If two `Grads` have the same `Params` in different orders, should we be able to add them? I think so.

test/interface.jl
```diff
@@ -55,4 +57,54 @@
...
+
+    @test .- gs3 isa Grads
+    @test gs3 .- gs4 isa Grads 
+    @test .+ gs3 isa Grads
+    @test gs3 .+ gs4 isa Grads 
+    @test 2 .* gs3 isa Grads 
+    @test gs3 .* 2 isa Grads 
+    @test gs3 ./ 2 isa Grads  
+    @test (gs3 .+ gs4)[w] ≈ gs3[w] .+ gs4[w]
+    @test (gs3 .+ gs4)[b] ≈ gs4[b]
```
darsnack, line 92: Make sure in-place broadcasting works (not sure if it should currently).
```suggestion
    @test (gs3 .+ gs4)[b] ≈ gs4[b]
    gs5 .= gs3 + gs4
    @test gs5[w] ≈ gs3[w] .+ gs4[w]
    @test gs5[b] ≈ gs3[b]
```

docs/src/utils.md
mcabbott: `Grads` doesn't subtype anything, so it doesn't seem too crazy for it to support a limited range of operations we think useful. It's just a container to hold particular results.

If it was made iterable, writing `for (p,g) in grads; ...` might sometimes be tidier than `for p in params; grads[p] ...`. It does expose a bit more, in that you're explicitly using the fact that it's really a grads-and-params struct. 

Its `Params` does provide an order. Although as you say, the same parameters wrapped in a different `Params` struct could have a different order. So `grads1 .+ 0.1 .* grads2` would probably still want to disobey this order and go by the keys.

test/interface.jl
CarloLucibello: which method should I overload for broadcasted assignment?

docs/src/utils.md
darsnack: You should already be able to do `for (p, g) in grads` (I was doing this yesterday when playing around with alternate implementations of this PR).

test/interface.jl
darsnack: I think you will want to define `copyto!(dest::Grads, bc::Broadcasted)` (reference: https://docs.julialang.org/en/v1/manual/interfaces/#extending-in-place-broadcast-2)

docs/src/utils.md
darsnack: Ah never mind, I see from the latest commits that maybe this includes extraneous pairs somehow?

CarloLucibello: yes, this is what I remember from inspecting Grads objects in the past, altough i'm not 100% sure

darsnack: This is better anyways since I am not sure the ordering of `grads.params` was being respected before.

mcabbott: I didn't think it was iterable at all:
```
julia> g0 = Zygote.Grads(IdDict(), Zygote.Params())
Grads(...)

julia> (g0...,)
ERROR: MethodError: no method matching iterate(::Zygote.Grads)
```

darsnack: Woah you're right. I must have been mistakenly operating on the `IdDict` I created? I was surprised to see it worked and assumed it happened by some Base magic since `haskey` was forwarded. Guess I was wrong!

test/interface.jl
CarloLucibello: I didn't understand copyto! quite well, also we don't have broadcasted styles, so I implemented a solution with materialize!

docs/src/utils.md
CarloLucibello: I think the value based iteration is the most natural here, since it's consistent with how I implemented map and is battle tested in Dictionaries.jl. I'll mark this conversion as resolved since it's on an outdated commit, but we can continue the discussion in the main thread

test/interface.jl
darsnack: I think you want specifically `copyto!(dest::Grads, bc::Broadcasted{Nothing})` since we don't specialize on the style. I'm okay with `materialize!` since it doesn't require "indexing" the `Broadcasted` wrapper. We just want to make sure this is safe (I don't know, maybe @mcabbott)?

DhairyaLGandhi: https://docs.julialang.org/en/v1/manual/interfaces/#man-interfaces-broadcasting

Specifically says copyto is the interface the internal machinery uses, so that's the best one forward. Also yes, in place broadcasting can cause corner cases (and incorrect gradients). You could broadcast setindex directly.

CarloLucibello:
ok, I think I'm happy with the design and implementation, should be ready

src/compiler/interface.jl
```diff
@@ -171,6 +186,41 @@ function copy!(x::AbstractVector,  gs::Grads)
...
 end
 
+broadcasted(f, gss::Grads...) = map(f, gss...)
+
+for op in (:*, :/)
+  @eval broadcasted(::typeof($op), a::Number, gs::Grads) = map(x -> $op(a, x), gs)
+  @eval broadcasted(::typeof($op), gs::Grads, a::Number) = map(x -> $op(x, a), gs)
+end
+
+function materialize!(gs1::Grads, gs2::Grads)
```
DhairyaLGandhi, line 196: Why are we overloading this function? Broadcasting on dictionaries is reserved so messing with it seems odd

cc @oxinabox

CarloLucibello: Grads is not a dictionary

oxinabox: This seems fine and good to me.
Grads isn't a dictionary.
I also say dictionaries should get with it and define broadcast, there  are about two options and they just need to pick one.

It makes a lot of sense for Grads objects to be broadcast assignable inplace.
Which is what this function does, no? It makes `x .= y` work.

Means we can do gradient clipping via 
```julia
generalized_clamp(x::Number, low, high) = clamp(x, low, high)
generalized_clamp(x, low, high) = generalized_clamp.(x, low, high)

g .= generalized_clamp.(g, -300, 300)
```

This function is noncontentioius. Other things might be. but `materialize` isn't

oxinabox:
There is an argument to be made that `map` should get the key and the value,
and that broadcast should just get the value.
 
Broadcast just seeing value would make sense to me.
e.g. `g .= 2 .* (a .+ b)`
Because one reasonably will want to broadcast against operations on numbers (which have no key).

But also access to keys is useful.
When writing an earlier comment i wanted to suggest the following as a usecase and relized i couldn't:
Consider learning rate scaling. Which used to be a big deal for deep belief networks.
Where Gaussian Layers should have 0.01 times the learning rate used for Bernoulli layers.

```julia
scaled_g = map(g) do k, v
    if is_gaussian_layer(k)
        0.01*v
    else
        v
    end
end
```

This wouldn't quite match what @darsnack proposed here https://github.com/FluxML/Flux.jl/issues/707#issuecomment-780033296
but i think this distinction might make sense.

Although: `NamedTuples` (and `ChainRules.Composite`s) do just `map` with values.
So this wouldn't be consistent with that.
But OTOH, if this does workout: map with pairs, and broadcast with values, then this could be a good place to test it, and it could resolve the stalemate that `Base.AbstrctDict` arguments are having.
(and that resolution would also result in `map(pairs(Dict)) do k,v` working, which feels right to me.)

CarloLucibello:
> Consider learning rate scaling. Which used to be a big deal for deep belief networks.
Where Gaussian Layers should have 0.01 times the learning rate used for Bernoulli layers.

Consider that keys are parameters, not layers, so I don't think you can detect if `is_gaussian_layer` in your example. 

The use cases I can think of, such as clamping, make only use of values, so value-based map and iteration seem more convenient, and in any case `pairs` are just a few keystrokes away.  Another point in favor of values is that 
https://github.com/andyferris/Dictionaries.jl
took that stance, and that seems a really well-thought library. 

I'm not strongly against pairs iteration and map though, just seems less convenient to me but I'm sure it would work just fine.

src/compiler/interface.jl
```diff
@@ -139,7 +140,21 @@ end
...
 
-@forward Grads.grads Base.getindex, Base.haskey
+@forward Grads.grads  Base.haskey, Base.setindex!
+@forward Grads.params  Base.length
+
+# Dictionary interface.
+# Don't use the IdDict directly since it may contain some spurious pairs. 
+Base.keys(gs::Grads) = gs.params
+Base.values(gs::Grads) =  (gs.grads[p] for p in gs.params)
+Base.pairs(gs::Grads) =  (p => gs.grads[p] for p in gs.params)
```
oxinabox, line 150: `Base.pairs` returns a `AbstractDict` subtype.
I think Base is wrong for doing that. Because you then can't call `map` on it.
But it is surprising that we diverge from this.

oxinabox:
Ok, I am sold, I trust you have thought this through.

src/compiler/interface.jl
```diff
@@ -171,6 +186,41 @@ function copy!(x::AbstractVector,  gs::Grads)
...
+    throw(ArgumentError("map! expects Grads objects with the same Params."))
+  for p in gsout.params
+    gsout[p] = f((_getformap(gs, p) for gs in gss)...) 
+  end
+  return gsout
+end
+
+function _getformap(gs::Grads, p)
+  g = gs[p]
+  isnothing(g) ? fill!(similar(p), 0) : g
```
oxinabox, line 221: Aside: this will likely not be needed one we finish #603 
because the output will be a subtype of `AbstractZero`, rather than `nothing`
Which should be number/array-like enough for all normal use-cases.

darsnack:
I feel like it would be surprising if `map` and broadcast did different things (though there is no rule that they should be the same). In this case, I believe the proposed `map(f, pairs(grads))` is the clearest version of the intent of the code. `map(f, grads)` operating on pairs would become confusing to me when we go to multiple grads: `map(f, gs1, gs2, ...)`. You could make sure the key sets are equal and only have one key at the start, but that's too confusing imo.

The main issue for me though is that the ordering of the pairs is not guaranteed across `Grads`. As a user, if two `Grads` has the same parameters, then I expect the values associated parameter A to be matched regardless of storage order. This is hard to guarantee in a way that's communicated to the user. For example, for `gs1 .+ gs2`, we could just pick the order of the first argument. But for `gs1 .+ gs2 .- [rand(size(p)) for p in params(m)]`, how does one know what order the generator should be in? Broadcasting on values allows us to play nicer with the rest of the broadcasted expression, but I think you lose the ordering guarantee which makes things murky.

One thing I will suggest is that maybe we extend the current binary operation broadcast between a number and a `Grads` to be any binary function? Would be surprising that I can do `2 .* grads` but not `min.(grads, 1e-4)` for example.

darsnack:
One other option is to allow broadcasting not just with other `Grads` but any `Pair`. We can just check for key set equality like we do. This would allow for broadcasting with regular arrays in a sensible way. But we can save this for another PR if we want.

src/compiler/interface.jl
darsnack: And in-place is wanted for the [original issue](https://github.com/FluxML/Flux.jl/issues/707) where you need to accumulate many gradients. In-place would let you do it without allocating too many copies of the model.

src/compiler/interface.jl
CarloLucibello: turns out we don't have to define `pairs` explicitly, Base defines it generically from keys and values

CarloLucibello:
> Would be surprising that I can do 2 .* grads but not min.(grads, 1e-4) for example.

This a bit tricky, because `min` would need two levels of broadcasting two work, kind of `min..`.  While `2 * grads[p]` is well defined, `min(grads[p], 1e-4)` is not. 

> One other option is to allow broadcasting not just with other Grads but any Pair. We can just check for key set equality like we do. This would allow for broadcasting with regular arrays in a sensible way. But we can save this for another PR if we want.

handling `Pairs` seems fine, I'll try to implement it

darsnack:
> This a bit tricky, because min would need two levels of broadcasting two work, kind of min... While 2 * grads[p] is well defined, min(grads[p], 1e-4) is not.

My bad, that was a poor example. I just meant that if `f(x::Number, y::Number)` is defined, it would be surprising if `2 .* grads` worked but `f.(grads, 2)` did not. I think if you just change the binary op def to:
```julia
broadcasted(f, a::Number, gs::Grads) = map(x -> f(a, x), gs)
broadcasted(f, gs::Grads, a::Number) = map(x -> f(x, a), gs)
```
it would work. Probably people only care about `*` but this is easy enough to add safely.

CarloLucibello:
> broadcasted(f, a::Number, gs::Grads) = map(x -> f(a, x), gs)
broadcasted(f, gs::Grads, a::Number) = map(x -> f(x, a), gs)

@darsnack should we generalize this to `Union{Number, AbstractArray}`?

src/compiler/interface.jl
```diff
@@ -185,7 +188,7 @@ function copy!(x::AbstractVector,  gs::Grads)
   x
 end
 
-broadcasted(f, gss::Grads...) = map(f, gss...)
+broadcasted(f, gs::Grads, gss::ADictOrGrads...) = map(f, gs, gss...)
```
darsnack, line 191: Would want this
```suggestion
broadcasted(f, gs::ADictOrGrads, gss::ADictOrGrads...) = map(f, gs, gss...)
```

But that would overload broadcasting for the case where `gs` and `gss` are just `Dict`s (no `Grads`). Don't know a quick way around it but unfortunate that the first arg must be a `Grad`.

darsnack:
> should we generalize this to Union{Number, AbstractArray}?

We could. It seems unlikely that every parameter in the `Grads` has the same shape, but I think semantically this is safe to do. I think now that we have the ability to support different shapes through broadcasting with `AbstractDict`, I don't have any gripes. We just want to make clear in the docs that if you need to broadcast with a vector of arrays (cause the shapes are different), you need to use `[get_array(p) for p in keys(grads)]`.

src/compiler/interface.jl
CarloLucibello: yep, I don't know either, unless we start adding a bunch of definitions like
```julia
broadcasted(f, gs1::AbstractDict, gs2::Grads, gss::ADictOrGrads...) = map(f, gs1, gs2, gss...)
```
and the corresponding ones for `map`, but this seems pretty annoying and I would leave it to future PRs

darsnack: Is `broadcasted(f, gss::Vararg{Union{AbstractDict, Grads}})` different from `broadcasted(f, gss::Vararg{<:Union{AbstractDict, Grads}})`? I think the first forces the union type. Let me double check.

CarloLucibello:
>  you need to use  [get_array(p) for p in keys(grads)].

actually,  this case is problematic, since if `a =  [get_array(p) for p in keys(grads)]` then `gs .+ a` would imply `gs[p] + a` which is wrong. Given this scenario, we better disallow broadcast with arrays. 
That usage is achieved with `Dict(p => get_array(p) for p in keys(grads))`.

darsnack:
I totally agree and just forgot to write the `p => ` my comment.

CarloLucibello:
In the end, I allowed broadcasted binary operations involved numerical types. 

I think we are done. One little regret is that we now support
```julia
gs .+ Dict(w => rand(2), b => nothing)
```
but it would nice to relax the `issetequal` assumption on the keys and allow
```julia
gs .+ Dict(w => rand(2))
```
Doesn't seem easy to do with current `map` implementation ... material for future PRs

docs/src/utils.md
```diff
@@ -23,3 +23,35 @@ Zygote.checkpointed
...
+
+# clip gradients
+map(x -> clamp.(x, -0.1, 0.1), gs)
+
+# clip gradients in-place
+foreach(x -> clamp!(x, -0.1, 0.1), gs)
+
+for (p, g) in pairs(gs)
+  # do something with parameter `p` and corresponding gradient `g`
+end
```
darsnack, line 56:
```suggestion
end

# note that gradients must be w.r.t. to the same parameter key set
gs3 = gradient(() -> sum(tanh.(w .* x2)), Params([w]))
# gs3 does not have the key b
@test_throws ArgumentError gs1 .+ gs3
```

src/compiler/interface.jl
```diff
@@ -139,7 +140,22 @@ end
 
 Base.show(io::IO, ps::Grads) = print(io, "Grads(...)")
 
-@forward Grads.grads Base.getindex, Base.haskey
+@forward Grads.grads  Base.haskey, Base.setindex!
```
darsnack, line 143: Is it correct to forward `haskey` here (thinking of the comment where the `IdDict` contained spurious entries)? Can we just define `Base.haskey(gs::Grads, p) = p ∈ keys(gs)` instead?

src/compiler/interface.jl
```diff
@@ -171,6 +187,40 @@ function copy!(x::AbstractVector,  gs::Grads)
...
+    throw(ArgumentError("map! expects Grads objects with the same Params."))
+  for p in gsout.params
+    gsout[p] = f((_getformap(gs, p) for gs in gss)...) 
+  end
+  return gsout
+end
+
+function _getformap(gs, p)
+  g = gs[p]
+  isnothing(g) ? fill!(similar(p), 0) : g
```
darsnack, line 221:
```suggestion
  isnothing(g) ? fill!(similar(p), zero(g)) : g 
```
unless the promoted type is inferred?

test/interface.jl
```diff
@@ -55,4 +57,86 @@
...
+    @test 2 .* gs3 isa Grads 
+    @test gs3 .* 2 isa Grads 
+    @test gs3 ./ 2 isa Grads  
+    @test (gs3 .+ gs4)[w] ≈ gs3[w] .+ gs4[w]
+    @test (gs3 .+ gs4)[b] ≈ gs4[b] 
+    
+    @test gs3 .+ Dict(w => similar(w), b => similar(b)) isa Grads
+    gs3 .+= Dict(p => randn(size(p)) for p in keys(gs3))
+    @test gs3 isa Grads 
+  end
```
darsnack, line 100:
```suggestion

  @test_throws ArgumentError gs1 .+ gs4
  end
```

darsnack (approved):
Looks really great! Thanks for the effort.

src/compiler/interface.jl
CarloLucibello: good catch

src/compiler/interface.jl
CarloLucibello: `fill!` doesn't promote

darsnack: Shouldn't the type returned by `fill!` match the type of `g` for stability?

darsnack (approved):

src/compiler/interface.jl
DhairyaLGandhi: It should I think. Could it be a footgun performance wise?

CarloLucibello: I think you guys are doing multiple mistakes here. `fill!` return type depends only on the first argument, so `typeof(p)`. `fill!` is called only when `g == nothing`, so deriving its return type from `g` is meaningless. The branching based on a nothing check is normally optimized away. The function is probably type-unstable, but the cause is upstream, in Grads containing an `IdDict{Any}`.

andyferris:
I’m not sure I can comment specifically on `Grad` but these are really interesting discussions. 

I particularly agree with:

> In this case, I believe the proposed `map(f, pairs(grads))` is the clearest version of the intent of the code. `map(f, grads)` operating on pairs would become confusing to me when we go to multiple grads: `map(f, gs1, gs2, ...)`.

This is a really good point. The other nice one in der default `values` iteration is `map(f, keys(d), d, d2, ...)` works well for when you want the key, while on the other hand `f` is a bit tortured under `map(f, pairs(d1), pairs(d2), ...)`.  

As to differences between `map` and `broadcast` is that in my personal interpretation `map` is about matching up on iteration and `broadcast` is about (more flexibly) matching up on indices. I think it would be confusing if the elements they got differed, though.

PR #909, opened 2021-02-23
WIP: improve inference for getproperty

simeonschaub:
This has regressed quite a bit due to #848. With this PR, we should be able to get back the same performance as before, assuming there is no custom implementation or pullback for `getproperty`. Still need to add tests.

### DIFF
--- a/src/Zygote.jl
+++ b/src/Zygote.jl
@@ -33,6 +33,7 @@ include("compiler/show.jl")
 
 include("lib/grad.jl")
 include("lib/lib.jl")
+include("lib/literal_getproperty.jl")
 include("lib/number.jl")
 include("lib/base.jl")
 include("lib/array.jl")
--- a/src/lib/literal_getproperty.jl
+++ b/src/lib/literal_getproperty.jl
@@ -1,0 +1,82 @@
+# Mostly copied over from Cassette in `src/overdub.jl`
+# Return `Reflection` for signature `sigtypes` and `world`, if possible. Otherwise, return `nothing`.
+function reflect(@nospecialize(sigtypes::Tuple), world::UInt = typemax(UInt))
+    if length(sigtypes) > 2 && sigtypes[1] === typeof(invoke)
+        @assert sigtypes[3] <: Type{<:Tuple}
+        sigtypes = (sigtypes[2], sigtypes[3].parameters[1].parameters...)
+    end
+    # This works around a subtyping bug. Basically, callers can deconstruct upstream
+    # `UnionAll` types in such a way that results in a type with free type variables, in
+    # which case subtyping can just break.
+    #
+    # God help you if you try to use a type parameter here (e.g. `::Type{S} where S<:Tuple`)
+    # instead of this nutty workaround, because the compiler can just rewrite `S` into
+    # whatever it thinks is "type equal" to the actual provided value. In other words, if
+    # `S` is defined as e.g. `f(::Type{S}) where S`, and you call `f(T)`, you should NOT
+    # assume that `S === T`. If you did, SHAME ON YOU. It doesn't matter that such an
+    # assumption holds true for essentially all other kinds of values. I haven't counted in
+    # a while, but I'm pretty sure I have ~40+ hellish years of Julia experience, and this
+    # still catches me every time. Who even uses this crazy language?
+    S = Tuple{map(s -> Core.Compiler.has_free_typevars(s) ? typeof(s.parameters[1]) : s, sigtypes)...}
+    (S.parameters[1]::DataType).name.module === Core.Compiler && return nothing
+    _methods = Base._methods_by_ftype(S, -1, world)
+    method_index = 0
+    for i in 1:length(_methods)
+        if _methods[i][1] === S
+            method_index = i
+            break
+        end
+    end
+    method_index === 0 && return nothing
+    type_signature, raw_static_params, method = _methods[method_index]
+    method_instance = Core.Compiler.specialize_method(method, type_signature, raw_static_params, false)
+    method_signature = method.sig
+    static_params = Any[raw_static_params...]
+    return method_instance, method_signature, static_params
+end
+
+
+# ugly hack to make differentiating `getproperty` infer a lot better
+@generated function _pullback(cx::AContext, ::typeof(literal_getproperty), x, ::Val{f}) where f
+    sig(x) = Tuple{x, typeof(f)}
+    rrule_sig(x) = Tuple{typeof(getproperty), x, typeof(f)}
+    pb_sig(x) = Tuple{cx, typeof(getproperty), x, typeof(f)}
+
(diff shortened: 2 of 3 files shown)


willtebbutt:
This is cool.

> assuming there is no custom implementation or pullback for getproperty

Could you elaborate a bit on this? Why is this necessary? I'm sure it could be ascertained from the PR itself, but I could use a helping hand 😂

DhairyaLGandhi:
I am still going through the pr, it'd be good to get an idea if the generated function is too low level for Julia ir so as to break with newer releases.

simeonschaub:
> > assuming there is no custom implementation or pullback for getproperty
> 
> Could you elaborate a bit on this? Why is this necessary? I'm sure it could be ascertained from the PR itself, but I could use a helping hand 😂

Zygote's previous assumption was that `getproperty` could always be treated just like `getfield`, unless a custom adjoint was defined. This is obviously wrong if `getproperty` is overloaded, but it of course opens up a lot more opportunities for optimization. 😄 In the case where a user has overloaded `getproperty`, there is not really a way around recursing into the actual definition, but this PR special-cases the case where `getproperty` just calls the fallback definition, so we can basically retain the previous behavior.

willtebbutt:
Ohhh I see. This makes sense. Will comment on the PR itself.

src/lib/lib.jl
```diff
@@ -210,6 +210,42 @@ end
...
+@generated function _pullback(cx::AContext, ::typeof(literal_getproperty), x, ::Val{f}) where f
+    sig(x) = Tuple{x, typeof(f)}
+    pb_sig(x) = Tuple{cx, typeof(getproperty), x, typeof(f)}
+
+    # either `getproperty` has a custom implementation or `_pullback(cx, ::getproperty, x, f)`
+    # is overloaded directly
+    is_getfield_fallback = which(_pullback, pb_sig(x)) == which(_pullback, pb_sig(Any)) && 
+        which(getproperty, sig(x)) == which(getproperty, sig(Any))
+
+    if is_getfield_fallback
```
willtebbutt, line 223: Another high-level question, what's the reason that this entire case can't simply be something like
```julia
# we've ensured that we're fine to short-circuit and call `getfield` directly,
return :(Zygote._pullback(cx, literal_getfield, x, f))
```
rather than pulling up the IR for that method and pasting it here?

simeonschaub:
> I am still going through the pr, it'd be good to get an idea if the generated function is too low level for Julia ir so as to break with newer releases.

It is definitely fairly low-level, so it is not impossible this might cause problems with future Julia versions. It is basically the same mechanism `IRTools` and `Cassette` already rely upon, so while I'd definitely prefer a higher-level solution, I think it is a necessary evil if we want the inference improvements. At least in my testing, more aggressive inline annotations didn't really help.

src/lib/lib.jl
simeonschaub: The main reason is that we can't (at least that I know of) manually attach backedges to regular expressions, so we need to return a CodeInfo object.  We could return a CodeInfo that just contains that call, but if we already go to lengths like this, we might as well just directly inline the whole thing.

willtebbutt: This also makes sense to me.

src/lib/lib.jl
```diff
@@ -210,6 +210,42 @@ end
...
+        # just copy pullback of `literal_getfield`
+        _sig = Tuple{cx, typeof(literal_getfield), x, Val{f}}
+        ci = code_lowered(_pullback, _sig)[1]
+
+        Meta.partially_inline!(
```
willtebbutt, line 228: Could you add a comment here explaining the intent behind this line?

simeonschaub: Ok, will do

src/lib/lib.jl
```diff
@@ -210,6 +210,42 @@ end
...
+            ci.code, Any[_pullback, Core.SlotNumber(2), literal_getfield],
+            Tuple{typeof(_pullback), _sig.parameters...}, Any[_sig.parameters...],
+            0, 0, :propagate,
+        )
+
+        # backedge for `getproperty`
```
willtebbutt, line 234: Possibly a silly suggesion, but maybe it would be a good idea to link to the docs regarding backedges? e.g. [here](https://docs.julialang.org/en/v1/devdocs/ast/#MethodInstance) They're definitely one of the less well-known features, and this might this kind of code more accessible to the uninitiated.

pevnak:
This is awesome. I have been tracking the regression. I was originally suspicious to Julia 1.6. and reported here
https://github.com/JuliaLang/julia/issues/39586
which many people confirmed. Bud digging deeper, it seems to me that the problem is the getproperty

This was my MWE
```
using Flux
using BenchmarkTools
x = randn(Float32, 10, 10);
m = Dense(10,10);
ps = Flux.params(m);
f(w,b,x) = w * x .+ b
w, b = m.W, m.b

@btime gradient(() -> sum(m(x)), ps)
@btime gradient(() -> sum(f(w,b,x)), ps)
```
which on Zygote 0.5.17 with Flux 11.3 runs as
```
julia> @btime gradient(() -> sum(m(x)), ps)
  5.785 μs (47 allocations: 4.81 KiB)
Grads(...)

julia> @btime gradient(() -> sum(f(w,b,x)), ps)
  6.145 μs (52 allocations: 4.72 KiB)
Grads(...)
```

But on Zygote@0.6.3, Flux v0.11.6 it is fairly slow
```
julia> @btime gradient(() -> sum(m(x)), ps)
  25.046 μs (102 allocations: 7.39 KiB)
Grads(...)

julia> @btime gradient(() -> sum(f(w,b,x)), ps)
  6.247 μs (58 allocations: 4.91 KiB)
Grads(...)
```

and testing this PR with Flux v0.11.5 gives
```
julia> @btime gradient(() -> sum(m(x)), ps)
  5.895 μs (54 allocations: 5.02 KiB)
Grads(...)

julia> @btime gradient(() -> sum(f(w,b,x)), ps)
  6.361 μs (58 allocations: 4.91 KiB)
Grads(...)
```

So this is awesome. Can it be merged and tagged?

simeonschaub:
Yes, sorry, I had other stuff to do and didn't come around to finishing this. I will try to finish this up some time later today.

willtebbutt:
@simeonschaub how's this going? :)

simeonschaub:
Please don't merge this yet, it currently breaks custom adjoints for `getproperty`.

SimonMandlik:
Is there any chance of this being merged in the near future? :)

simeonschaub:
Yes, feel free to bug me about it from time to time, which probably increases the chances of me eventually finishing it. :smile: I will try to get to it soon again but every time I work on it, I discover some new issue...

DhairyaLGandhi:
bump :)

willtebbutt:
@simeonschaub bump :)

DhairyaLGandhi:
I see they're going

simeonschaub:
Sure, please go ahead!

PR #923, opened 2021-03-22
Changing Distances adjoints to ChainRules syntax

theogf:
Using @adjoint disable the overloading of `rrule` see : https://github.com/JuliaDiff/ChainRulesCore.jl/issues/239#issuecomment-713560336
It makes it impossible to define new rules for different metrics.
I am aware that the long-term plan is to move out the rules to the respective packages but in the mean-time it would be nice for it to be solved (it blocks our development in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/208).


DhairyaLGandhi:
Lgtm for the most part.

DhairyaLGandhi:
`@adjoint` does take precedence over `rrule`, so you could also define it via the macro to change it later. Although I would ask what the use case for that would be?

theogf:
Sorry, I made a mistake: trying JuliaGaussianProcesses/KernelFunctions.jl#208 with this branch does not seem to solve the problem.

Also precompiling this patch returns a surprising warning: 
```julia
┌ Warning: Error requiring `Distances` from `Zygote`
│   exception =
│    LoadError: error in method definition: function ChainRulesCore.rrule must be explicitly imported to be extended
...
```

DhairyaLGandhi:
Yeah, so we usually move the adjoints over, but since ChainRules is a dep, you could add the import

willtebbutt:
@theogf I know you've resolved this on your end now, but would you be up for pushing this PR through anyway? Seems like something that we should be doing throughout `Zygote` when the opportunity arises.

devmotion:
It seems the adjoint definitions have to be updated to the ChainRules-syntax? In particular, I guess `nothing` should be replaced with the ChainRules-equivalents.

theogf:
I officially don't know how to import rrule correctly via Requires...

theogf:
I am not sure the failing checks are related to this PR...

willtebbutt:
Doesn't look related to me. @theogf I reckon this is basically good to go after a patch bump. I'll bors it after that unless @DhairyaLGandhi has any objections?

DhairyaLGandhi (approved):

DhairyaLGandhi:
There was a merge conflict, but lgtm

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.6.6"
+version = "0.6.7"
```
CarloLucibello, line 3:
```suggestion
version = "0.6.10"
```

src/lib/distances.jl
```diff
@@ -43,28 +44,28 @@ end
...
   d = colwise(s, x, y)
   return d, function (Δ::AbstractVector)
     x̄ = (Δ ./ max.(d, eps(eltype(d))))' .* (x .- y)
-    return nothing, x̄, -x̄
+    return NO_FIELDS, NO_FIELDS, x̄, -x̄
   end
 end
 
-@adjoint function pairwise(::Euclidean, X::AbstractMatrix, Y::AbstractMatrix; dims=2)
+function rrule(::typeof(pairwise), ::Euclidean, X::AbstractMatrix, Y::AbstractMatrix; dims=2)
```
devmotion, line 68: This definition calls `Zygote.pullback` to compute the pullback of a more stable modification of the forward pass. This seems unfortunate for everyone that would like to use the ChainRules definitions with another AD backend. But on the other hand, the definitions here are type piracy anyway and problematic in such cases where Zygote is loaded (possibly by some other dependency).

Ref: https://github.com/JuliaDiff/ChainRules.jl/issues/338, https://github.com/JuliaDiff/ChainRulesCore.jl/issues/68

ToucheSir: @theogf up to revive this using the pattern mentioned in https://github.com/JuliaDiff/ChainRules.jl/issues/338#issuecomment-884772928?

theogf: Done! Hopefully I understood correctly how it worked

src/lib/distances.jl
```diff
@@ -1,22 +1,23 @@
 using .Distances
+import .ChainRules: NO_FIELDS, rrule
```
devmotion, line 2: IIRC `NO_FIELDS` was deprecated a long time ago. Instead one should use just `NoTangent`.

src/lib/distances.jl
```diff
@@ -1,22 +1,23 @@
...
 
-@adjoint function (::SqEuclidean)(x::AbstractVector, y::AbstractVector)
+function rrule(::SqEuclidean, x::AbstractVector, y::AbstractVector)
```
devmotion, line 4: I think it's not a good idea to add these type piracies to Zygote. I assume it is better to move it to e.g. ChainRules or Distances (IIRC the main developers did not want to add rules...) right away. Otherwise I fear it will be not only surprising that Zygote adds these rules (and one has to load Zygote to get these chainrules definitions) but also more annoying if one moves it to another package later on (since one has to be careful to avoid method redefinitions etc.).

A temporary workaround could also be to define the rules in Zygote, but only for `ZygoteRuleConfig`.

theogf: Sure, but the problem is that if I start to add these rules in ChainRules, will we be able to test them?

devmotion: In ChainRules? There you would just test them with ChainRulesTestUtils. Actually, this should be done regardless of where they are defined to ensure that the rules are implemented properly, return the correct types, have the correct signature, are type stable etc.

mcabbott: Xref https://github.com/JuliaArrays/FillArrays.jl/pull/153 maybe. I don't think CR.jl wants to provide rules for, and hence depend on, other packages. If they don't want to depend on CRC.jl, then the other option is to make tiny glue packages for the rules. Or to wait for Pkg to build in this ability. 

But setting up rules using ZygoteRuleConfig here can't hurt, and would make it easier to move them elsewhere later.

ToucheSir: Would ChainRules be willing to take on Distances.jl as a dependency? I assumed the only path for an external package outside of upstreaming to Distances itself would be something like DistributionsAD.jl.

In the short-medium term interest of removing as many (ideally all) `adjoint`s from Zygote itself, using `ZygoteRuleConfig` is a great option.

devmotion: Please not something like DistributionsAD :smile: It's a complete mess, full of type piracy, and to a large extent it contains workarounds and hacks that should have been added to the respective AD backends or other upstream packages such as Distributions or PDMats. The long term plan is to get rid of it as much as possible, and already all CR things were moved to Distributions. While it speeded up initial development to accumulate all the stuff in the package instead of fixing other packages, we now have to pay the price for it and spend a significant amount of time with cleaning it.

ToucheSir: Not saying it's a good idea :P, more that it's the only option outside of hosting rules in the AD backend if upstream is not willing to accept PRs.

devmotion: :stuck_out_tongue: 

IMO defining the rules for ZygoteRuleConfig only is probably the easiest and cleanest option if one wants to avoid (repeating) these discussions about whether it rather belongs to ChainRules/Distances/etc. here.

mzgubic: >Would ChainRules be willing to take on Distances.jl as a dependency?

The policy is that ChainRules provides rules for functionality in Julia Base. Packages should depend on ChainRulesCore (which is super light and has no dependencies other than Julia Base) to define rules for their functionality. @oxinabox would have to approve exceptions, but I don't think I can remember any, apart from two in the early days that have since been moved out.

I agree that the use of ZygoteRuleConfig is the best solution for now. If we decide in the future where we want to put them it would be made very easy as well.

src/lib/distances.jl
```diff
@@ -1,22 +1,23 @@
 using .Distances
+import .ChainRules: NoTangent, rrule, rrule_via_ad
 
-@adjoint function (::SqEuclidean)(x::AbstractVector, y::AbstractVector)
+function rrule(::SqEuclidean, x::AbstractVector, y::AbstractVector)
```
devmotion, line 4:
```suggestion
function rrule(::ZygoteRuleConfig, ::SqEuclidean, x::AbstractVector, y::AbstractVector)
```

src/lib/distances.jl
```diff
@@ -1,22 +1,23 @@
...
   function sqeuclidean(Δ::Real)
     x̄ = (2 * Δ) .* δ
-    return x̄, -x̄
+    return NoTangent(), x̄, -x̄
   end
   return sum(abs2, δ), sqeuclidean
 end
 
-@adjoint function colwise(s::SqEuclidean, x::AbstractMatrix, y::AbstractMatrix)
+function rrule(::typeof(colwise), s::SqEuclidean, x::AbstractMatrix, y::AbstractMatrix)
```
devmotion, line 13:
```suggestion
function rrule(::ZygoteRuleConfig, ::typeof(colwise), s::SqEuclidean, x::AbstractMatrix, y::AbstractMatrix)
```

src/lib/distances.jl
```diff
@@ -1,22 +1,23 @@
...
   return colwise(s, x, y), function (Δ::AbstractVector)
     x̄ = 2 .* Δ' .* (x .- y)
-    return nothing, x̄, -x̄
+    return NoTangent(), NoTangent(), x̄, -x̄
   end
 end
 
-@adjoint function pairwise(s::SqEuclidean, x::AbstractMatrix, y::AbstractMatrix; dims::Int=2)
+function rrule(::typeof(pairwise), s::SqEuclidean, x::AbstractMatrix, y::AbstractMatrix; dims::Int=2)
```
devmotion, line 20:
```suggestion
function rrule(::ZygoteRuleConfig, ::typeof(pairwise), s::SqEuclidean, x::AbstractMatrix, y::AbstractMatrix; dims::Int=2)
```

src/lib/distances.jl
```diff
@@ -28,10 +29,10 @@ end
   function(Δ)
     x̄ = 2 .* (x * Diagonal(vec(sum(Δ; dims=2))) .- y * transpose(Δ))
     ȳ = 2 .* (y * Diagonal(vec(sum(Δ; dims=1))) .- x * Δ)
-    return (nothing, f(x̄), f(ȳ))
+    return NoTangent(), NoTangent(), f(x̄), f(ȳ)
   end
 
-@adjoint function pairwise(s::SqEuclidean, x::AbstractMatrix; dims::Int=2)
+function rrule(::typeof(pairwise), s::SqEuclidean, x::AbstractMatrix; dims::Int=2)
```
devmotion, line 35:
```suggestion
function rrule(::ZygoteRuleConfig, ::typeof(pairwise), s::SqEuclidean, x::AbstractMatrix; dims::Int=2)
```

src/lib/distances.jl
```diff
@@ -43,47 +44,47 @@ end
   function(Δ)
     d1 = Diagonal(vec(sum(Δ; dims=1)))
     d2 = Diagonal(vec(sum(Δ; dims=2)))
-    return (nothing, x * (2 .* (d1 .+ d2 .- Δ .- transpose(Δ))) |> f)
+    return NoTangent(), NoTangent(), x * (2 .* (d1 .+ d2 .- Δ .- transpose(Δ))) |> f
   end
 
-@adjoint function (::Euclidean)(x::AbstractVector, y::AbstractVector)
+function rrule(::Euclidean, x::AbstractVector, y::AbstractVector)
```
devmotion, line 50:
```suggestion
function rrule(::ZygoteRuleConfig, ::Euclidean, x::AbstractVector, y::AbstractVector)
```

src/lib/distances.jl
```diff
@@ -43,47 +44,47 @@ end
...
   function euclidean(Δ::Real)
     x̄ = ifelse(iszero(δ), D, (Δ / δ) .* D)
-    return x̄, -x̄
+    return NoTangent(), x̄, -x̄
   end
   return δ, euclidean
 end
 
-@adjoint function colwise(s::Euclidean, x::AbstractMatrix, y::AbstractMatrix)
+function rrule(::typeof(colwise), s::Euclidean, x::AbstractMatrix, y::AbstractMatrix)
```
devmotion, line 60:
```suggestion
function rrule(::ZygoteRuleConfig, ::typeof(colwise), s::Euclidean, x::AbstractMatrix, y::AbstractMatrix)
```

devmotion (commented):
I think it would also be good to add some tests with ChainRulesTestUtils.test_rrule.

CarloLucibello:
needs a rebase, otherwise seems a sensible change

devmotion:
I wonder if possibly maintainers of Distances who didn't want to add a dependency on ChainRulesCore might be fine with adding it as a weak dependency?

src/lib/distances.jl
```diff
@@ -75,7 +75,7 @@ function rrule(config::ZygoteRuleConfig, ::typeof(pairwise), dist::Euclidean, X:
     return _sqrt_if_positive.(D2, δ)
   end
   D, back = rrule_via_ad(config, _pairwise_euclidean, SqEuclidean(dist.thresh), X, Y)
-  pairwise_Euclidean_rrule(Δ) = (NoTangent(), back(Δ)...)
+  pairwise_Euclidean_rrule = back
   return D, pairwise_Euclidean_rrule
```
mcabbott, line 79: Maybe clearer without these lines, stop at `D, back = rrule_via_ad(config, ...)`? Or just `return rrule_via_ad(config, ...)`.

theogf: Sure! I only wanted to check that this was the issue 😊

ToucheSir:
CI is complaining about a `\Delta` not being defined, but it seems like it is?

Issue #922, opened 2021-03-22
accumulation of gradients


mzgubic:
The issue occurs when the following are all true:

- We have a struct with two differentiable fields
- We need to accumulate the gradients with respect to both fields
- The gradients that need to be accumulated both originate from an rrule

An example is svd factorisation:

```julia
using Zygote
using LinearAlgebra
using FiniteDifferences
using Test

# function to test AD against finite differences. Can't wait for https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/114
function test_ad(test_function, Δoutput, inputs...; atol=1e-7, rtol=1e-7)
    # Verify that the forwards-pass produces the correct answer.
    output, pb = Zygote.pullback(test_function, inputs...)
    @test output ≈ test_function(inputs...)

    # Compute the adjoints using AD and FiniteDifferences.
    dW_ad = pb(Δoutput)
    dW_fd = FiniteDifferences.j′vp(central_fdm(5, 1), test_function, Δoutput, inputs...)

    # Compare AD and FiniteDifferences results.
    @testset "$(typeof(test_function)) argument $n" for n in eachindex(inputs)
        @test dW_ad[n] ≈ dW_fd[n] atol=atol rtol=rtol
    end
end

function two_svds(X::StridedMatrix{<:Union{Real, Complex}})
    return svd(X).U * svd(X).V'
end

function one_svd(X::StridedMatrix{<:Union{Real, Complex}})
    F = svd(X)
    return F.U * F.V'
end

Δoutput = randn(3,2)
X = randn(3,2)

test_ad(two_svds, Δoutput, X) # works

test_ad(one_svd, Δoutput, X) # fails

  Expression: ≈(dW_ad[n], dW_fd[n], atol = atol, rtol = rtol)
   Evaluated: [0.10086994903629046 0.017821972969422787; -0.025950275563480844 -0.20012541965315742; 0.1827211721197507 0.16897788019829582] ≈ [-1.5134220464481536 -1.9386355487247564; 0.4546889986237118 0.8662734137317394; 0.9222144395826724 1.4004778288991626] (atol=1.0e-7, rtol=1.0e-7)

```

I think the reason is that the accumulation of gradients is not defined for `literal_getproperty`, just `literal_getfield`
https://github.com/FluxML/Zygote.jl/blob/890b6f57303fb983a815aec5fea92620c93a4bd3/src/lib/lib.jl#L213-L226

Will submit a PR later today

[...]

PR #924, opened 2021-03-23
Throw error when trying to accum two NamedTuples with different keys

mzgubic:
Would have helped here:
https://github.com/FluxML/Zygote.jl/issues/922#issuecomment-804334846

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.6.4"
+version = "0.6.5"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
--- a/src/lib/lib.jl
+++ b/src/lib/lib.jl
@@ -16,9 +16,10 @@ accum(x::Tuple, y::Tuple) = accum.(x, y)
 accum(x::AbstractArray, y::AbstractArray) = accum.(x, y)
 
 @generated function accum(x::NamedTuple, y::NamedTuple)
-  # Zygote assumes that the NamedTuples will have the same keys
-  fieldnames(x) === fieldnames(y) || throw(ArgumentError("$x and $y keys must be the same"))
-  grad(x) = x in fieldnames(y) ? :(y.$x) : :nothing
+  # assumes that y has no keys apart from those also in x
+  fieldnames(y) ⊆ fieldnames(x) || throw(ArgumentError("$y keys must be a subset of $x keys"))
+
+  grad(field) = field in fieldnames(y) ? :(y.$field) : :nothing
   Expr(:tuple, [:($f=accum(x.$f, $(grad(f)))) for f in fieldnames(x)]...)
 end
 
--- a/src/lib/lib.jl
+++ b/src/lib/lib.jl
@@ -16,6 +16,8 @@ accum(x::Tuple, y::Tuple) = accum.(x, y)
 accum(x::AbstractArray, y::AbstractArray) = accum.(x, y)
 
 @generated function accum(x::NamedTuple, y::NamedTuple)
+  # Zygote assumes that the NamedTuples will have the same keys
+  fieldnames(x) === fieldnames(y) || throw(ArgumentError("$x and $y keys must be the same"))
   grad(x) = x in fieldnames(y) ? :(y.$x) : :nothing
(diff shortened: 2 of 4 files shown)


mzgubic:
@willtebbutt mind taking a look?

DhairyaLGandhi:
Should we not have it so that keys of y are a subset of keys of x?

mzgubic:
Quite possibly, I don't know what the precise assumptions are here. My understanding was that Zygote internally uses `NamedTuple`s with explicit `nothing` for all zero gradients? (Unlike ChainRules where `Composite` only needs non zero derivatives explicitly, and the others are zero implicitly)

willtebbutt:
> Playing devil's advocate a bit here, but only to take the conversation forward. Should we not have it so that keys of y are a subset of keys of x?

Ah, yeah, based on the current functionality, I think you're correct @DhairyaLGandhi .

@mzgubic I think this looks good modulo the subset stuff (I'm just assuming you'll write some tests). While you're here, would you mind renaming the argument of `grad` to something other than `x`? This argument being called `x` makes the rest of the code hard to read.

mzgubic:
Oh, when you say assumptions you just mean what this function does, rather than the broader Zygote assumptions about NamedTuples as gradients for structs? In that case I totally agree.

willtebbutt (approved):
LGTM. @DhairyaLGandhi should also check he's happy though.

mzgubic:
Great, thanks, just FYI I don't have merge rights so one of you is going to have to merge it

DhairyaLGandhi:
Looks fine, thanks!

bors r+

marius311:
Looks like this PR broke CI for my package CMBLensing.jl https://github.com/marius311/CMBLensing.jl/runs/2282205515

Glancing over this and linked issues, my understanding is Zygote was returning an incorrect result before, and now this errors (fwiw, previously I wasn't finding any incorrect gradients)

More importaintly, would one of you guys be able to offer some hint as to how to fix whatever my package presumably was doing wrong before? I'm not expecting anyone to go read my source code, but just wondering if you could suggest the general thing I should be looking for?

DhairyaLGandhi:
Well it wasn't incorrect in the sense that things that shouldn't have gradients, never had gradients to begin with.

Generally you want to check for cases where you were generating namedtuples in custom adjoints.

mzgubic:
From the error message
```julia
ArgumentError:
NamedTuple{(:op, :recompute_function, :parameters),Tuple{Diagonal{Complex{Float64},CMBLensing.BaseField{CMBLensing.Basis2Prod{CMBLensing.𝐄𝐁,Fourier},ProjLambert{Float64,Array{Float64,1},Array{Float64,2}},Complex{Float64},Array{Complex{Float64},3}}},Nothing,Nothing}}
keys must be a subset of 
NamedTuple{(:L,),Tuple{NamedTuple{(:op, :recompute_function, :parameters),Tuple{Diagonal{Complex{Float64},CMBLensing.BaseField{CMBLensing.Basis2Prod{CMBLensing.𝐄𝐁,Fourier},ProjLambert{Float64,Array{Float64,1},Array{Float64,2}},Complex{Float64},Array{Complex{Float64},3}}},Nothing,Nothing}}}}
keys
```

It looks like the accumulation is between a `NamedTuple` and a nested `NamedTuple`. It looks like previously the contents of the non-nested `NamedTuple` were ignored, which may have given the correct result if the values were zero.

marius311:
Thanks a ton for the replies @DhairyaLGandhi @mzgubic, ended up finding the mistake, it was definitely me returning an incorrect adjoint, and your tips were super helpful in tracking it down!

mzgubic:
Great! Is that the one?https://github.com/marius311/CMBLensing.jl/pull/55/commits/efb618d3f9de83e45a1e8f48af482a7848df6d2a

marius311:
Yep, good sleuthing :) If you're curious / confused whats going on there, its this: https://discourse.julialang.org/t/how-to-deal-with-zygote-sometimes-pirating-its-own-adjoints-with-worse-ones/32579/2?u=marius311

mzgubic:
Oh, I thought it was just that `logdet` has two arguments and needs a pullback of a function with two arguments (rather than a one argument closure as before). Well, it works now, so we don't have to dig in deeper :)

marius311:
No, you're exactly right as to what the mistake was c.f. this PR, although the underlying reason there's a lot of explicit Zygote.pullback calls in that file has to do with that forum post.

PR #925, opened 2021-03-23
add test for literal getproperty overwrite

mzgubic:
Another small thing that came from #922 , namely https://github.com/FluxML/Zygote.jl/issues/922#issuecomment-804128905.

Writing a separate @adjoint for `literal_getproperty`, which shouldn't be done, passes all tests at the moment. This test will fail if a custom adjoint is written for `literal_getproperty`.

@simeonschaub this is from our discussion yesterday

Issue #861, opened 2020-12-23
accum NamedTuple sometimes silently drops derivatives


willtebbutt:
This works correctly:
```julia
julia> Zygote.accum((a=nothing, b=4), (a=4,))
(a = 4, b = 4)
```

This drops the derivative w.r.t. `b`:
```julia
julia> Zygote.accum((a=4, ), (a=nothing, b=4))
(a = 4,)
```
Now this is definitely surprising, and I _think_ it's a bug. The only situation in which it's intended behaviour is if it's the case that the derivative w.r.t. some struct should always be a `NamedTuple` containing _all_ of the associated fields, since this would preclude this kind of problem.

Note that ChainRule types don't suffer from this issue:
```julia
julia> Composite{Any}(; a=4) + Composite{Any}(; a=Zero(), b=4)
Composite{Any}(a = 4, b = 4)
```
so we can probably nick the `ChainRulesCore` implementation of `+` for `Composite`s and deploy it for `accum` with `NamedTuple`s.

willtebbutt:
Although looking at what `Zygote` does by default, perhaps it _is_ assumed that all fields will always be present and I just missed the memo on that one?
```julia
julia> Zygote.gradient(x -> sin(x.a), (a=4,b=4))
((a = -0.6536436208636119, b = nothing),)
```
(Had the gradient not had a field `b`, it would clearly be the case that `Zygote` doesn't make this assumption).

oxinabox:
>  The only situation in which it's intended behaviour is if it's the case that the derivative w.r.t. some struct should always be a NamedTuple containing all of the associated fields, since this would preclude this kind of problem.

This is the case. As I understand it.
Zygote requires all structural derivatives to be NamedTuples with the same fields (in the same order?)  as the primal type.
What ChainRules calls the _canonical_ form.

As such `Zygote.accum((a=nothing, b=4), (a=4,))`
and `Zygote.accum((a=4, ), (a=nothing, b=4))`
should both error.
Since it means something went wrong somewhere to give you a bad dertivative.
Probably an incorrect rule.

willtebbutt:
> Since it means something went wrong somewhere to give you a bad dertivative.
> Probably an incorrect rule.

This is exactly what happened. I returned a subset of the required fields from a particular hand-writen `_pullback`.

mzgubic:
Closed by https://github.com/FluxML/Zygote.jl/pull/924

PR #926, opened 2021-03-24
Canonicalize the Composite before changing it to a NamedTuple

### LINKED ISSUE #922: accumulation of gradients

mzgubic:
Closes #922.

The issue occurs when the following are all true:
1) We have a struct with two differentiable fields
2) We need to accumulate the gradients with respect to both fields
3) The gradients that need to be accumulated both originate from an rrule

In that case, the two gradients that originate from an rrule are `Composite`s with disjoint fields. When they are transformed to a `NamedTuple`, the Zygote internal representation of derivatives w.r.t. structs, these two `NamedTuple`s have disjoint sets of keys, and are not `accum`ulated correctly.

By adding `canonicalize(x)`, the `Composite` gets explicit `Zero()` fields, which means the resulting `NamedTuple`s have the complete set of fields. 

~~- blocked by https://github.com/JuliaDiff/ChainRules.jl/pull/390 (tests only)~~
~~- need to change compat once the above is merged~~


mzgubic:
Looks like we either need https://github.com/JuliaDiff/ChainRulesCore.jl/issues/321 or use 
```julia
    maybe_canonicalized = isabstracttype(P) ? x : canonicalize(x)
    xp = map(wrap_chainrules_output, maybe_canonicalized)
```
instead. It seems to be able to do type inference alright.

DhairyaLGandhi:
Seems a bit hairy. Could you give me an idea of what `canonicalize` does concretely?

mzgubic:
Yeah I'd prefer the issue but there might be reasons not to do that (I can't think of any, but there might be).

To understand `canonicalize` one needs to know that `Composite`s, which are the differential types for structs, only store non-zero gradients explicitly. For example, if I have a `MyStruct` with two fields, `a` and `b`, then the `rrule` for `getproperty(mystruct, a)` can return `Composite{MyStruct}(;a=ȳ)` rather than `Composite{MyStruct}(;a=ȳ, b=Zero())`.

The zeros for other fields are implicit, meaning it is possible to sum `Composites` easily:
```julia
julia> Composite{MyStruct}(;a=ȳ) + Composite{MyStruct}(;b=ȳ2)
Composite{MyStruct}(;a=ȳ, b=ȳ2)
```

`canonicalize` will simply make these implicit `Zero()`s explicit, i.e. 
```julia
julia> canonicalize(Composite{MyStruct}(;a=ȳ))
Composite{MyStruct}(;a=ȳ, b=Zero())
```

This relies on knowing the primal type (`MyStruct`) that `Composite` stores the derivative for in order to access the fields. Here, the issue is that in higher-order AD sometimes the primal type is unknown*, and in that case it is not possible to know which fields should explicitly be set to zero.

*I _think_ we can get rid of this once Zygote uses ChainRules differential types internally

oxinabox:
I am pretty sure we should do this.
Since otherwise the NamedTuple that comes out of a ChainRules rule will not ensc have the same fields as one from Zygote working it out on its own.
And that will lead to badness.

I commented on this here: https://github.com/FluxML/Zygote.jl/issues/861#issuecomment-750480710

oxinabox (approved):

mzgubic:
I've added ChainRulesCore as a direct dependency to be able to specify a compat bound. I don't have merge rights but there is nothing I want to add, so please go ahead once you think it's ready

oxinabox:
the GPU tests for o Cholesky weem flaky

See if they work for bores

Bors r+

mzgubic:
Same test failed in a [different PR which only added another test](https://github.com/FluxML/Zygote.jl/pull/925). Not sure what to think about that, but I don't think it's related to either of the PRs?

mzgubic:
Could you try again please?

PR #929, opened 2021-03-25
remove cufunc

CarloLucibello:
It has been removed on CUDA master.


CarloLucibello:
CUDA 3.0 is out, so we should hurry up with this. 

Am I doing this right here @DhairyaLGandhi ? The `@eval` is needed?

CarloLucibello:
Buildkite shouldtest CUDA 3.0 on julia 1.6 and CUDA 2.4 on julia 1.5, hopefully if anything is wrong it should tell us

PR #935, opened 2021-03-29
Update Array printing in doctests for 1.6

ToucheSir:
These are the only failures I've seen on the latest CI output, so hopefully it unblocks the docs site deployment.


DhairyaLGandhi:
Seems fine. Julia can't find Documenter in the tests, but we can tackle that later

bors r+

DhairyaLGandhi:
Hmm, the cholesky adjoint seems to have become a bit moody. Maybe ease it's tolerance @willtebbutt ?


bors r+

willtebbutt:
Sounds reasonable to me.

PR #936, opened 2021-03-29
Fix docs deployment

darsnack:
I'm not sure when/how this affected things, but the current `docs/make.jl` activates the v1.6 environment instead of the `docs/Project.toml` environment. This means that Documenter is not available to the CI.

I minimized the changes to just what's necessary to get things working, but there's generally a bunch of weird path related stuff happening with the environments in this package. I'd like to clean it up @DhairyaLGandhi @CarloLucibello (not sure if it is necessary for Zygote.jl for some reason, but I don't see why it should be).


DhairyaLGandhi:
I think this is basically the thing to do.

bors r+

PR #947, opened 2021-04-17
nograd typejoin

DhairyaLGandhi:
Should fix the issue @jonathan-laurent

### DIFF
--- a/src/lib/base.jl
+++ b/src/lib/base.jl
@@ -139,4 +139,4 @@ end
 
 @adjoint Base.nameof(x::UnionAll) = nameof(x), _ -> (nothing,)
 
-@nograd typeintersect
+@nograd typeintersect, typejoin


willtebbutt:
@DhairyaLGandhi would you consider doing this using `@non_differentiable` inside `ChainRules` instead of adding it here?

DhairyaLGandhi:
I guess that's fine, but I was hoping to make a release with this to unblock AlphaZero

willtebbutt:
> I guess that's fine, but I was hoping to make a release with this to unblock AlphaZero

We make a patch release with every non-breaking change in ChainRules, so you'll have a release to bind to :)

DhairyaLGandhi:
Fixed in https://github.com/JuliaDiff/ChainRules.jl/pull/398

PR #949, opened 2021-04-19
Define custom adjoints for LogExpFunctions instead of StatsFuns

devmotion:
Since StatsFuns 0.9.8, StatsFuns just reexports the log/exp functions defined in LogExpFunctions. The main motivation for this change is that users and downstream packages can avoid the Rmath dependency by depending on LogExpFunctions only. Thus the custom adjoints should be defined for LogExpFunctions and loaded if LogExpFunctions is available.

I am not completely sure what the best way is to ensure that the derivatives are still defined on older versions of StatsFuns that do not depend on LogExpFunctions:
- should one keep a `@requires` block for StatsFuns and load the definitions if `!isdefined(StatsFuns, :LogExpFunctions)`?
- should one add a compat entry for StatsFuns (not a dependency!) to ensure that StatsFuns depends on LogExpFunctions?
- or can one assume that users have installed/will install the latest version of StatsFuns?

(Ideally, LogExpFunctions - or a glue package - would define the differentials with ChainRulesCore but this is a temporary solution for Zygote users)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -41,8 +41,8 @@ CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
 Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
 FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
 FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
-StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
+LogExpFunctions = "2ab3a3ac-af41-5b50-aa03-7779005ae688"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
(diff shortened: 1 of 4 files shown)


DhairyaLGandhi:
The old versions can maybe cap Zygote to ensure that they have the adjoints defined?

devmotion:
You mean, one could add an upper bound for Zygote in the registry for older StatsFuns versions? Does this work even though StatsFuns does not depend on Zygote?

CarloLucibello:
> should one add a compat entry for StatsFuns (not a dependency!) to ensure that StatsFuns depends on LogExpFunctions?

I didn't know you could add a compact for a non-dependence, but it looks like we can so we should. 

As you said, this an ok temp solution until the rules are added to LogExpFunctions. We should really get rid of all those `@requires` for Zygote 0.7

devmotion:
Bump :slightly_smiling_face:

PR #951, opened 2021-04-20
Handle nothing in map

### LINKED ISSUE #804: Unused comprehension throws "iterate(nothing)" method error

0:
I ran into #804, and made an attempt to fix it before I noticed #846.

Unlike #846, which still runs `map`, this patch fixes #804 by propagating `Δ === nothing` through the `∇map` pullback more directly.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -203,6 +203,7 @@ for (mapfunc,∇mapfunc) in [(:map,:∇map),(:pmap,:∇pmap)]
     else
       ys, backs = unzip(ys_and_backs)
       ys, function (Δ)
+        isnothing(Δ) && return nothing
         # Apply pullbacks in reverse order. Needed for correctness if `f` is stateful.
         Δf_and_args_zipped = $mapfunc((f, δ) -> f(δ), _tryreverse($mapfunc, backs, Δ)...)
         Δf_and_args = unzip(_tryreverse($mapfunc, Δf_and_args_zipped))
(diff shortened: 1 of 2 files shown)


DhairyaLGandhi:
Cc @oxinabox

oxinabox:
This is much less weird to me.
Let's do this

oxinabox (approved):

DhairyaLGandhi:
We'd return that tuple generically, so there might be some unforeseen consequences to consider.

0:
That's a good point. I was focused on the [`_pullback` for `collect`](https://github.com/FluxML/Zygote.jl/blob/2d5edf44ad7191fcd5d5816d8ab161dff8e14765/src/lib/array.jl#L236-L242) and neglected the general case. For example, the patch currently results in the following:
```julia
julia> Zygote.∇map(Zygote.Context(), +, [1.0], [1.0])[2](1.0)
(nothing, [1.0], [1.0])

julia> Zygote.∇map(Zygote.Context(), +, [1.0], [1.0])[2](nothing)
(nothing, nothing)
```

It also doesn't work for the empty array case, because that's handled in the other branch:
```julia
julia> gradient([]) do xs; [x for x in xs]; 0.0; end
ERROR: MethodError: no method matching iterate(::Nothing)
```

---

Since `∇map` [already returns](https://github.com/FluxML/Zygote.jl/blob/2d5edf44ad7191fcd5d5816d8ab161dff8e14765/src/lib/array.jl#L202) `_ -> nothing` when `args` are empty, callers should presumably be ready for this. Would it therefore be better to have
```julia
isnothing(Δ) && return nothing
```
instead of returning a tuple, and simply make sure that it's handled correctly at [line 239](https://github.com/FluxML/Zygote.jl/blob/2d5edf44ad7191fcd5d5816d8ab161dff8e14765/src/lib/array.jl#L239)? That might be the least weird option.

CarloLucibello:
> Would it therefore be better to have

that seems a good solution, also what was discussed [here](https://github.com/FluxML/Zygote.jl/pull/846#issuecomment-739435872)

oxinabox (approved):

DhairyaLGandhi:
Were the failing cases tested for @0 ?

Issue #897, opened 2021-02-03
Two errors on Julia 1.6


mcabbott:
These two probably unrelated things work fine on 1.5, but give errors on 1.6 or master. 

First, only for long enough generators:
```julia
julia> using Zygote, LinearAlgebra

julia> gradient(x -> sum(norm, collect(eachcol(x))), rand(3,4))[1]
3×4 Matrix{Float64}:
 0.0980459  0.916691  0.397663  0.548254
 0.78843    0.216233  0.900898  0.0887359
 0.60726    0.336038  0.173914  0.831591

julia> gradient(x -> sum(norm, collect(eachcol(x))), rand(3,400))[1] # only size changed
ERROR: Can't differentiate loopinfo expression
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:33
  [2] macro expansion
    @ ./simdloop.jl:79 [inlined]
  [3] Pullback
    @ ./reduce.jl:245 [inlined]
  [4] (::typeof(∂(mapreduce_impl)))(Δ::Float64)
    @ Zygote ~/.julia/dev/Zygote/src/compiler/interface2.jl:0
  [5] Pullback
    @ ./reduce.jl:259 [inlined]
  [6] (::typeof(∂(mapreduce_impl)))(Δ::Float64)
    @ Zygote ~/.julia/dev/Zygote/src/compiler/interface2.jl:0
  [7] Pullback
    @ ./reduce.jl:417 [inlined]
  [8] (::typeof(∂(_mapreduce)))(Δ::Float64)
    @ Zygote ~/.julia/dev/Zygote/src/compiler/interface2.jl:0
  [9] Pullback
    @ ./reducedim.jl:330 [inlined]
 [10] Pullback
    @ ./reducedim.jl:322 [inlined]
 [11] (::typeof(∂(mapreduce)))(Δ::Float64)
    @ Zygote ~/.julia/dev/Zygote/src/compiler/interface2.jl:0
 [12] Pullback
    @ ./reducedim.jl:890 [inlined]
 [13] (::typeof(∂(#_sum#717)))(Δ::Float64)
    @ Zygote ~/.julia/dev/Zygote/src/compiler/interface2.jl:0
 [14] Pullback
    @ ./reducedim.jl:890 [inlined]
 [15] (::typeof(∂(_sum)))(Δ::Float64)
    @ Zygote ~/.julia/dev/Zygote/src/compiler/interface2.jl:0
 [16] Pullback
[...]

PR #955, opened 2021-04-26
fix adjoint for sum

simeonschaub:
addresses the second part of #897

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -270,14 +270,9 @@ end
   sum(xs, dims = dims), Δ -> (nothing,)
 end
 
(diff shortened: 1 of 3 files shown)


CarloLucibello (approved):

DhairyaLGandhi (commented):
bors r+

CarloLucibello:
I don't think bors failure is related https://buildkite.com/julialang/zygote-dot-jl/builds/214#78e9850d-b908-44e5-badb-1004d9e21153

In the failing cholesky test we should probably have `≈`  instead of `==`

cpfiffer:
Could we bump this one? It'd be helpful for a project I'm on.

Issue #957, opened 2021-04-27
`Base.Fix1` and `Base.Fix2` are much slower than closures


devmotion:
`Base.Fix1` and `Base.Fix2` are much slower than closures. This was observed in KernelFunctions where the issue was fixed by reverting back closures. However, I think it would be good to fix this issue more properly since e.g. Distributions makes use of `Base.Fix1` etc. and hence probably is affected by the same performance issues.

A simple example (Julia 1.6.1, Zygote 0.6.10):
```julia
julia> using Zygote, BenchmarkTools

julia> x = rand(100);

julia> @btime Zygote.gradient($(x -> prod(y -> y + 1, x)), $x);
  10.373 μs (343 allocations: 13.27 KiB)

julia> @btime Zygote.gradient($(x -> prod(Base.Fix1(+, 1), x)), $x);
  342.138 μs (3608 allocations: 138.08 KiB)

julia> @btime Zygote.gradient($(x -> prod(Base.Fix2(+, 1), x)), $x);
  352.705 μs (3607 allocations: 138.06 KiB)
```

DhairyaLGandhi:
Would the correct course of action be to add a shortcut path here? Probably seems like so.

devmotion:
Yes, that was my idea as well but I was not sure how such shortcuts are typically implemented in Zygote. Just calling the pullback of the corresponding closure and then extracting and recombining the results?

DhairyaLGandhi:
We could potentially overload the pullback directly via an adjoint or `pullback`. Both should be equivalent in this case.

devmotion:
Is it not necessary to return a NamedTuple for the `Fix1` etc. structs?

DhairyaLGandhi:
We can still return the appropriate named tuples. We can return the struct back as well, if we can accumulate the fields appropriately but I don't think that would be the case here.

devmotion:
I tried different things (it's not completely clear to me if one should use `pullback`, `_pullback`, or `@adjoint`) but I didn't manage to get it down to the same speed and allocations as the closure. The best I achieved was:
```julia
julia> using Zygote, BenchmarkTools

julia> x = rand(100);

julia> function Zygote._pullback(cx::Zygote.AContext, g::Base.Fix1, y)
           f = g.f
           x = g.x
           return Zygote._pullback(cx, z -> f(x, z), y)
       end

julia> function Zygote._pullback(cx::Zygote.AContext, g::Base.Fix2, y)
           f = g.f
[...]

PR #958, opened 2021-04-27
Improve performance of `Base.Fix1` and `Base.Fix2`

devmotion:
This PR addresses https://github.com/FluxML/Zygote.jl/issues/957.

Currently (copied from the issue above):
```julia
julia> using Zygote, BenchmarkTools

julia> x = rand(100);

julia> @btime Zygote.gradient($(x -> prod(y -> y + 1, x)), $x);
  10.373 μs (343 allocations: 13.27 KiB)

julia> @btime Zygote.gradient($(x -> prod(Base.Fix1(+, 1), x)), $x);
  342.138 μs (3608 allocations: 138.08 KiB)

julia> @btime Zygote.gradient($(x -> prod(Base.Fix2(+, 1), x)), $x);
  352.705 μs (3607 allocations: 138.06 KiB)
```

With this PR:
```julia
julia> using Zygote, BenchmarkTools

julia> x = rand(100);

julia> @btime Zygote.gradient($(x -> prod(y -> y + 1, x)), $x);
  10.033 μs (343 allocations: 13.27 KiB)

julia> @btime Zygote.gradient($(x -> prod(Base.Fix1(+, 1), x)), $x);
  12.150 μs (453 allocations: 27.73 KiB)

julia> @btime Zygote.gradient($(x -> prod(Base.Fix2(+, 1), x)), $x);
  12.223 μs (453 allocations: 27.73 KiB)
```

My main questions are:
- Why is it still slower and causes many more allocations?
- Why are `Base.Fix1` and `Base.Fix2` so slow in the first place and don't just work?

### DIFF
--- a/test/gradcheck.jl
+++ b/test/gradcheck.jl
@@ -1672,3 +1672,18 @@ end
     gradient(x->norm(x*[1im, 1]), 1.23)
     gradient(x->norm(x*[1im 1]), 1.23)
 end
+
+@testset "Fix1 and Fix2" begin
+    @test gradcheck(x -> prod(Base.Fix1(+, 1), x), randn(100))
+    @test gradcheck(x -> prod(Base.Fix2(+, 1), x), randn(100))
+
+    # compile once and check the execution times compared with a closure
+    # https://github.com/FluxML/Zygote.jl/issues/957
+    x = randn(100)
+    gradient(x -> prod(y -> y + 1, x), x)
+    t = @elapsed(gradient(x -> prod(y -> y + 1, x), x))
+    gradient(x -> prod(Base.Fix1(+, 1), x), x)
+    @test @elapsed(gradient(x -> prod(Base.Fix1(+, 1), x), x)) < 2 * t
+    gradient(x -> prod(Base.Fix1(+, 1), x), x)
+    @test @elapsed(gradient(x -> prod(Base.Fix2(+, 1), x), x)) < 2 * t
+end
--- a/test/gradcheck.jl
+++ b/test/gradcheck.jl
@@ -4,6 +4,7 @@ using Zygote: gradient
 using Base.Broadcast: broadcast_shape
 using Distributed: pmap, CachingPool, workers
 import FiniteDifferences
+using BenchmarkTools
 
 function ngradient(f, xs::AbstractArray...)
   grads = zero.(xs)
@@ -1677,13 +1678,12 @@ end
     @test gradcheck(x -> prod(Base.Fix1(+, 1), x), randn(100))
     @test gradcheck(x -> prod(Base.Fix2(+, 1), x), randn(100))
 
-    # compile once and check the execution times compared with a closure
+    # check the execution times compared with a closure
     # https://github.com/FluxML/Zygote.jl/issues/957
     x = randn(100)
-    gradient(x -> prod(y -> y + 1, x), x)
-    t = @elapsed(gradient(x -> prod(y -> y + 1, x), x))
-    gradient(x -> prod(Base.Fix1(+, 1), x), x)
-    @test @elapsed(gradient(x -> prod(Base.Fix1(+, 1), x), x)) < 2 * t
-    gradient(x -> prod(Base.Fix1(+, 1), x), x)
-    @test @elapsed(gradient(x -> prod(Base.Fix2(+, 1), x), x)) < 2 * t
+    tclosure = @belapsed(gradient($(x -> prod(y -> y + 1, x)), $x))
+    tfix1 = @belapsed(gradient($(x -> prod(Base.Fix1(+, 1), x)), $x))
+    tfix2 = @belapsed(gradient($(x -> prod(Base.Fix2(+, 1), x)), $x))
+    @test tfix1 < 2 * tclosure
+    @test tfix2 < 2 * tclosure
 end
--- a/test/gradcheck.jl
+++ b/test/gradcheck.jl
@@ -4,7 +4,6 @@ using Zygote: gradient
 using Base.Broadcast: broadcast_shape
 using Distributed: pmap, CachingPool, workers
 import FiniteDifferences
-using BenchmarkTools
 
 function ngradient(f, xs::AbstractArray...)
   grads = zero.(xs)
@@ -1719,6 +1718,7 @@ end
     @test gradcheck(x -> prod(Base.Fix1(+, 1), x), randn(100))
     @test gradcheck(x -> prod(Base.Fix2(+, 1), x), randn(100))
 
+#= regression tests are not included to reduce CI times
     # check the execution times compared with a closure
     # https://github.com/FluxML/Zygote.jl/issues/957
     x = randn(100)
@@ -1727,4 +1727,5 @@ end
     tfix2 = @belapsed(gradient($(x -> prod(Base.Fix2(+, 1), x)), $x))
     @test tfix1 < 2 * tclosure
     @test tfix2 < 2 * tclosure
+=#
 end
--- a/Project.toml
+++ b/Project.toml
@@ -38,14 +38,12 @@ ZygoteRules = "0.2.1"
 julia = "1.3"
 
 [extras]
-BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
 Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
 FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
 FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
 LogExpFunctions = "2ab3a3ac-af41-5b50-aa03-7779005ae688"
-StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["BenchmarkTools", "CUDA", "Distances", "FFTW", "FiniteDifferences", "LogExpFunctions", "Test"]
+test = ["CUDA", "Distances", "FFTW", "FiniteDifferences", "LogExpFunctions", "Test"]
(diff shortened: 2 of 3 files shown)


test/gradcheck.jl
```diff
@@ -1672,3 +1672,18 @@ end
...
+    @test gradcheck(x -> prod(Base.Fix2(+, 1), x), randn(100))
+
+    # compile once and check the execution times compared with a closure
+    # https://github.com/FluxML/Zygote.jl/issues/957
+    x = randn(100)
+    gradient(x -> prod(y -> y + 1, x), x)
+    t = @elapsed(gradient(x -> prod(y -> y + 1, x), x))
+    gradient(x -> prod(Base.Fix1(+, 1), x), x)
+    @test @elapsed(gradient(x -> prod(Base.Fix1(+, 1), x), x)) < 2 * t
+    gradient(x -> prod(Base.Fix1(+, 1), x), x)
```
devmotion, line 1687:
```suggestion
    gradient(x -> prod(Base.Fix2(+, 1), x), x)
```

CarloLucibello:
I don't know the answer to any of those questions, but I like this simple fix to a huge performance hit.

CarloLucibello (approved):

CarloLucibello:
Ready to merge?

test/gradcheck.jl
```diff
@@ -1672,3 +1672,18 @@ end
...
+    @test gradcheck(x -> prod(Base.Fix1(+, 1), x), randn(100))
+    @test gradcheck(x -> prod(Base.Fix2(+, 1), x), randn(100))
+
+    # compile once and check the execution times compared with a closure
+    # https://github.com/FluxML/Zygote.jl/issues/957
+    x = randn(100)
+    gradient(x -> prod(y -> y + 1, x), x)
+    t = @elapsed(gradient(x -> prod(y -> y + 1, x), x))
+    gradient(x -> prod(Base.Fix1(+, 1), x), x)
+    @test @elapsed(gradient(x -> prod(Base.Fix1(+, 1), x), x)) < 2 * t
```
DhairyaLGandhi, line 1686: Not sure if this is a robust test generally.

DhairyaLGandhi (approved):
Barring a more reliable test, we should be good.

test/gradcheck.jl
devmotion: Similar tests exist e.g. in RuntimeGeneratedFunctions and ChainRulesCore, but there `@belapsed` and `@ballocated` are used. Do you think it is better to use BenchmarkTools and/or to check allocations?

mcabbott:
For me the `@elapsed` times are backwards, and hence the timing test will pass right now. 
```
julia> x = rand(100);

julia> @btime Zygote.gradient($(x -> prod(y -> y + 1, x)), $x);
  8.833 μs (343 allocations: 13.27 KiB)

julia> @btime Zygote.gradient($(x -> prod(Base.Fix1(+, 1), x)), $x);
  266.625 μs (3608 allocations: 138.08 KiB)

julia> t = @elapsed(gradient(x -> prod(y -> y + 1, x), x))  # not the first run
0.214923125

julia> tfix = @elapsed(gradient(x -> prod(Base.Fix1(+, 1), x), x))
0.029850208

julia> tfix < 2t  # is the test
true

(@v1.6) pkg> st Zygote  # similar on master
  [e88e6eb3] Zygote v0.6.10

julia> versioninfo()
Julia Version 1.6.0
Commit f9720dc2eb (2021-03-24 12:55 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin19.6.0) 
  CPU: Apple M1
```
Likewise allocations:
```
julia> @ballocated(gradient(x -> prod(y -> y + 1, x), $x))
13584

julia> @ballocated(gradient(x -> prod(Base.Fix1(+, 1), x), $x))
141392

julia> @allocated(gradient(x -> prod(y -> y + 1, x), x))
53513519

julia> @allocated(gradient(x -> prod(Base.Fix1(+, 1), x), x))
4279117```

**Comment by devmotion** [comment]
With this PR I obtain:

```julia
julia> using Zygote, BenchmarkTools

julia> const x = rand(100);

julia> @btime Zygote.gradient($(x -> prod(y -> y + 1, x)), $x);
  10.151 μs (343 allocations: 13.27 KiB)

julia> @btime Zygote.gradient($(x -> prod(Base.Fix2(+, 1), x)), $x);
  12.284 μs (453 allocations: 27.73 KiB)

julia> @elapsed(gradient(x -> prod(y -> y + 1, x), x))
0.338051284

julia> @elapsed(gradient(x -> prod(Base.Fix2(+, 1), x), x))
0.064180138

julia> @ballocated(gradient($(x -> prod(y -> y + 1, x)), $x))
13584

julia> @ballocated(gradient($(x -> prod(Base.Fix2(+, 1), x)), $x))
28400

julia> @allocated(gradient(x -> prod(y -> y + 1, x), x))
53453284

julia> @allocated(gradient(x -> prod(Base.Fix2(+, 1), x), x))
3991125

(Zygote) pkg> st
     Project Zygote v0.6.11
      Status `~/.julia/dev/Zygote/Project.toml`
  [621f4979] AbstractFFTs v1.0.1
  [6e4b80f9] BenchmarkTools v0.7.0
  [082447d4] ChainRules v0.7.61
  [d360d2e6] ChainRulesCore v0.9.41
  [b552c78f] DiffRules v1.0.2
  [1a297f60] FillArrays v0.11.7
  [f6369f11] ForwardDiff v0.10.18
  [7869d1d1] IRTools v0.4.2
  [1914dd2f] MacroTools v0.5.6
  [77ba4419] NaNMath v0.3.5
  [ae029012] Requires v1.1.3
  [276daf66] SpecialFunctions v1.3.0
  [700de1a5] ZygoteRules v0.2.1
  [8ba89e20] Distributed
  [b77e0a4c] InteractiveUtils
  [37e2e46d] LinearAlgebra
  [9a3f8284] Random
  [10745b16] Statistics

julia> versioninfo()
Julia Version 1.6.1
Commit 6aaedecc44 (2021-04-23 05:59 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, skylake)
Environment:
  JULIA_NUM_THREADS = 4
```

The differences seem to be caused between `@elapsed` and `@btime` and `@allocated` and `@ballocated` seem to be caused by the fact more time and allocations have to be spent for constructing the closure:
```julia
julia> f(x) = prod(y -> y + 1, x)

julia> g(x) = prod(Base.Fix2(+, 1), x)

julia> @elapsed(gradient(f, x))
0.0001255

julia> @elapsed(gradient(g, x))
0.000140757

julia> @allocated(gradient(f, x))
13568

julia> @allocated(gradient(g, x))
28384
```

In any case, probably better to use `@ballocated` and `@btime` here.

mcabbott:
> In any case, probably better to use `@ballocated` and `@btime` here.

With the downside that these make tests take longer, and they already take quite a while. My vote would probably be just to omit the test.

devmotion:
I don't know, the main point here is to fix the time it takes to compute the gradient. There's no problem with the value of the gradient, it's computed just fine by Zygote automatically, but it is very slow. So the only way to check for the problem that is fixed by the PR seems to be to add tests with `@belapsed`.

devmotion:
> With the downside that these make tests take longer, and they already take quite a while.

If the tests take too long, maybe it would be helpful to split them into different groups (similar to e.g. https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/test/runtests.jl).

mcabbott:
Right. But there are lots of potential performance problems, and it's hard to test them all. We don't have a comprehensive set of benchmarks to test for regressions, but even if we did, they would probably be too expensive to run on every commit. Julia itself has such a set, but they aren't included in the normal tests, I think for similar reasons -- they take ages, can't be done accurately on random CI hardware.

devmotion:
Sure, one can't test everything, but this is a PR that is really only about major performance issues. These issues were discovered in KernelFunctions only accidentally but affect many other packages and users. The rewriting of Fix1 and Fix2 in this PR should not be needed at all and in principle it is completely stupid to add these rules - but it fixes the performance problem to some extent. So I think it should definitely be tested to ensure that these problems don't reappear. Such tests with BenchmarkTools work reliably e.g. in RuntimeGeneratedFunctions.

mcabbott:
Yes, I hear you. If someone else thinks it's OK to start using BenchmarkTools, they can merge it.

test/gradcheck.jl
```diff
@@ -1707,3 +1708,17 @@ end
...
+    @test gradcheck(x -> prod(Base.Fix2(+, 1), x), randn(100))
+
+    # check the execution times compared with a closure
+    # https://github.com/FluxML/Zygote.jl/issues/957
+    x = randn(100)
+    tclosure = @belapsed(gradient($(x -> prod(y -> y + 1, x)), $x))
+    tfix1 = @belapsed(gradient($(x -> prod(Base.Fix1(+, 1), x)), $x))
+    tfix2 = @belapsed(gradient($(x -> prod(Base.Fix2(+, 1), x)), $x))
+    @test tfix1 < 2 * tclosure
+    @test tfix2 < 2 * tclosure
```
CarloLucibello, line 1723:
```suggestion
```
We can't really add these benchmarks to tests in a systematic way, even when fixing major performance problems, long testing time makes development very annoying. I suggest we just leave these benchmarks out.

mcabbott: It would not be crazy to start a collection of benchmarks somewhere, perhaps to be run manually before tagging?

DhairyaLGandhi:
Adding BenchmarkTools isn't going to help our already length CI times, so I would omit for now. We can try to add a benchmark repo separately (FluxBench already exists) where we unclog CI here and track performance too.

devmotion:
I disabled the regression tests since I think it would be good to release this fix and I don't want to block the PR.

ChrisRackauckas:
>Right. But there are lots of potential performance problems, and it's hard to test them all. We don't have a comprehensive set of benchmarks to test for regressions, but even if we did, they would probably be too expensive to run on every commit. Julia itself has such a set, but they aren't included in the normal tests, I think for similar reasons -- they take ages, can't be done accurately on random CI hardware.

Base is doing something that isn't relevant here. It's testing things like the speed of `sin`, which need to be extremely accurate. When you're trying to figure out regressions in something that takes 10ns, you need a ton of samples. Even though it's faster, it's slower because it's harder. Even having an email client open on Nanosoldier will make those tests fail because they are so close to the metal. As things get larger, it's simpler and simpler to just take one or two calls and get a representative sample.

PR #959, opened 2021-04-27
Function to compute the diagonal of the Hessian

mcabbott:
This adds `diaghessian` which in the simplest case is equal to `(diag(hessian(f,x)),)`, computed more efficiently but otherwise in the same spirit as `hessian`.

### DIFF
--- a/src/lib/forward.jl
+++ b/src/lib/forward.jl
@@ -81,6 +81,53 @@ function forward_diag(f, x::AbstractArray)
   end
 end
 
+# and another approach: all forward, directly 2nd order
+
+seed_2nd(x::Real, ::Val) = Dual(Dual(x, true), true)
+
+function seed_2nd(xs, ::Val{N}, offset = 0) where N
+  map(xs, reshape(1:length(xs), size(xs))) do x, i
+    b = ntuple(j -> j+offset == i, Val(N))
+    Dual(Dual(x, b), b)
+  end
+end
+function seed_2nd!(xplus, xs, ::Val{N}, offset) where N
+  map!(xplus, xs, reshape(1:length(xs), size(xs))) do x, i
+    b = ntuple(j -> j+offset == i, Val(N))
+    Dual(Dual(x, b), b)
+  end
+end
+
+function extract_2nd!(out, fx::ForwardDiff.Dual{T,V,N}, offset) where {T,V,N}
+  for j in 1:min(N, length(out)-offset)
+    out[j+offset] = fx.partials.values[j].partials.values[j]
+  end
+end
+
+function forward_2nd(f, x::AbstractArray{T}, ::Val{N}) where {N,T}
+  xplus = seed_2nd(x, Val(N), 0)
+  fx = f(xplus)
+  out = similar(x, ForwardDiff.valtype(ForwardDiff.valtype(typeof(fx))))
+  extract_2nd!(out, fx, 0)
+  offset = 0
+  while offset + N < length(x)
+    offset += N
+    fx = f(seed_2nd!(xplus, x, Val(N), offset))
+    extract_2nd!(out, fx, offset)
+  end
+  return fx.value.value, out
+end
+
+function forward_2nd(f, x::AbstractArray)
+  # if length(x) < ForwardDiff.DEFAULT_CHUNK_THRESHOLD
+  #   forward_2nd(f, x, Val(length(x)))
+  # else
+  #   forward_2nd(f, x, Val(ForwardDiff.DEFAULT_CHUNK_THRESHOLD ÷ 2))
+    forward_2nd(f, x, Val(3))
+  # end
+end
+
+
 """
     forwarddiff(f, x) -> f(x)
 
--- a/src/lib/forward.jl
+++ b/src/lib/forward.jl
@@ -81,53 +81,6 @@ function forward_diag(f, x::AbstractArray)
   end
 end
 
-# and another approach: all forward, directly 2nd order
-
-seed_2nd(x::Real, ::Val) = Dual(Dual(x, true), true)
-
-function seed_2nd(xs, ::Val{N}, offset = 0) where N
-  map(xs, reshape(1:length(xs), size(xs))) do x, i
-    b = ntuple(j -> j+offset == i, Val(N))
-    Dual(Dual(x, b), b)
-  end
-end
-function seed_2nd!(xplus, xs, ::Val{N}, offset) where N
-  map!(xplus, xs, reshape(1:length(xs), size(xs))) do x, i
-    b = ntuple(j -> j+offset == i, Val(N))
-    Dual(Dual(x, b), b)
-  end
-end
-
-function extract_2nd!(out, fx::ForwardDiff.Dual{T,V,N}, offset) where {T,V,N}
-  for j in 1:min(N, length(out)-offset)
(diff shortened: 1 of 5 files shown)


mcabbott:
Times:
```
julia> using Zygote, LinearAlgebra, BenchmarkTools

julia> x1 = rand(100); y1 = rand(100);

julia> f4(x,y=x) = (dot(x,y))^3; # no broadcasting

julia> @btime diaghessian(f4, $x1)[1]'  # first commit
  82.791 μs (325 allocations: 437.61 KiB)
1×100 Adjoint{Float64,Array{Float64,1}}:
 8300.89  8648.26  8215.4  8336.26  8134.61  …  8077.96  7953.14  7942.76  8154.08  7983.89

julia> @btime diaghessian(f4, $x1)[1]'  # with in-place improvements
  68.833 μs (216 allocations: 301.72 KiB)
1×100 Adjoint{Float64,Array{Float64,1}}:
 8300.89  8648.26  8215.4  8336.26  8134.61  …  8077.96  7953.14  7942.76  8154.08  7983.89

julia> @btime diaghessian(f4, $x1)[1]'  # with let block
  30.875 μs (72 allocations: 292.72 KiB)
1×100 Adjoint{Float64,Array{Float64,1}}:
 8300.89  8648.26  8215.4  8336.26  8134.61  …  8077.96  7953.14  7942.76  8154.08  7983.89

julia> @btime diaghessian(f4, $x1)[1]'  # let block, but without in-place stuff (current PR)
  33.666 μs (130 allocations: 385.25 KiB)
1×100 Adjoint{Float64,Array{Float64,1}}:
 8300.89  8648.26  8215.4  8336.26  8134.61  …  8077.96  7953.14  7942.76  8154.08  7983.89


# full Hessian

julia> @btime diag(hessian(f4, $x1))'
  75.375 μs (121 allocations: 547.34 KiB)
1×100 Adjoint{Float64,Array{Float64,1}}:
 8300.89  8648.26  8215.4  8336.26  8134.61  …  8077.96  7953.14  7942.76  8154.08  7983.89


# two-argument version

julia> @btime diaghessian(f4, $x1, $y1)[1]'  # first commit
  109.042 μs (394 allocations: 418.36 KiB)
1×100 Adjoint{Float64,Array{Float64,1}}:
 3.74909  11.2471  96.1063  13.5062  110.238  12.9313  …  0.000699977  67.4358  7.28111  28.6905

julia> @btime diaghessian(f4, $x1, $y1)[1]'  # with _splice moved inwards
  1.904 ms (4930 allocations: 659.95 KiB)
1×100 Adjoint{Float64,Array{Float64,1}}:
 3.74909  11.2471  96.1063  13.5062  110.238  12.9313  …  0.000699977  67.4358  7.28111  28.6905

julia> @btime diaghessian(f4, $x1, $y1)[1]'  with let block
  21.833 μs (124 allocations: 400.08 KiB)
1×100 Adjoint{Float64,Array{Float64,1}}:
 1.93126  52.1384  0.953772  0.110112  58.7963  …  14.5704  119.024  143.133  109.438  53.6648


# broadcasting is slow!

julia> f1(x) = sum(x ./ x');

julia> @btime diaghessian(f1, $x1)[1]'
  1.915 ms (297 allocations: 27.07 MiB)
1×100 Adjoint{Float64,Array{Float64,1}}:
 6939.84  4.54447e6  792.682  30588.7  132.979  …  833.735  31314.1  2053.83  251.575  194.202

julia> @btime diag(hessian(f1, $x1))'
  1.984 ms (346 allocations: 27.32 MiB)
1×100 Adjoint{Float64,Array{Float64,1}}:
 6939.84  4.54447e6  792.682  30588.7  132.979  …  833.735  31314.1  2053.83  251.575  194.202
```

CarloLucibello:
Need to clean out all those comments, otherwise looks good. I'm surprised the improvement is not so large compared to `diag(hessian(...))`. What was legitimate to expect, a 100x speedup with dim. 100 inputs?

mcabbott:
Without having thought too hard I hoped for one less power of N, but this isn't what happens unfortunately:

Screenshot 2021-04-30 at 08 37 18

mcabbott:
Buildkite failure looks unrelated, doesn't happen every time, probably just tolerances?

https://buildkite.com/julialang/zygote-dot-jl/builds/221#95b561a8-f208-44b8-9ca5-693735de6791/230-359

gradcheck.jl:558  Expression: `(Zygote.pullback((X->(f(X, Y);)), X))[1] == cholesky(X * X' + I) \ Y`

src/lib/forward.jl
```diff
@@ -101,3 +136,9 @@ forwarddiff(f, x) = f(x)
   y, J = forward_jacobian(f, x)
   return y, ȳ -> (nothing, reshape_scalar(x, J*vec_scalar(ȳ)))
 end
+
+# Second derivatives
+@adjoint ForwardDiff.derivative(f, x) = pullback(forwarddiff, x -> ForwardDiff.derivative(f, x), x)
```
DhairyaLGandhi, line 141: These seem like they would affect a lot more than the intended functions.

src/lib/grad.jl
```diff
@@ -213,3 +215,52 @@ function jacobian(f, pars::Params)
...
+julia> hessian(xy -> atan(xy[1], xy[2]), [1, 2])  # full Hessian is not diagonal
+2×2 Matrix{Float64}:
+ -0.16  -0.12
+ -0.12   0.16
+```
+"""
+function diaghessian(f, args...)
+  ntuple(length(args)) do n
+    let x = args[n], valn = Val(n)  # let Val improves speed, sometimes
+      if x isa AbstractArray
```
DhairyaLGandhi, line 257: Is the performance difference coming from choosing different modes for numbers vs arrays? If so it might be a good opportunity to beef up our forward mode to get simd speedups

src/lib/forward.jl
mcabbott: I guess these aren't in the end used by the final version. Can add tests, or move to another PR if you prefer.

src/lib/grad.jl
mcabbott: If you see room to make forwarddiff more SIMD-friendly, that would be great, please make a PR. I didn't try to change that.

mcabbott:
> If performance isn't substantially better

For now about 3x, which I'd call substantial. If you know an algorithmically faster way, speak up.

src/lib/grad.jl
DhairyaLGandhi: Using Zygote.Forward is one easy way.

mcabbott: Oh cool, I didn't think anyone was working on that.

DhairyaLGandhi:
If not algorithmically faster, avoiding branches could help. My main concern is around breaking assumptions using forward diff and zygote in incompatible ways.

mcabbott:
> If not algorithmically faster, avoiding branches could help. 

If you can see a way, then by all means. There's a chance that ForwardDiff's own jacobian implementation has tricks which the one here doesn't use, and could copy.

> My main concern is around breaking assumptions using forward diff and zygote in incompatible ways, which I guess you've confirmed we can remove.

Yup, this is a fair complaint, and https://github.com/FluxML/Zygote.jl/pull/959/commits/85fb41603193573032991cc3ee1b4817472bea5a has removed it for separate discussion. 

Done.

Issue #905, opened 2021-02-20
Inplace accumulation of gradients?


jw3126:
From discourse https://discourse.julialang.org/t/zygote-gradient-accumulation/55654

I have a densenet inspired architecture implemented in pytorch and ported it to julia. Sadly I get out of memory errors now. 
Here is a MWE, where Julia memory consumption is more then 5x compared to pytorch:
On my laptop (GeForce RTX 2060 5.9GB) Julia throws out of memory error, while pytorch does not. Observe that the pytorch tensor is 5x bigger!

# Zygote
```julia
using Zygote
using CUDA

function net(x1)
    x2  = x1
    x3  = x1 + x2
    x4  = x1 + x2 + x3
    x5  = x1 + x2 + x3 + x4
    x6  = x1 + x2 + x3 + x4 + x5
    x7  = x1 + x2 + x3 + x4 + x5 + x6
    x8  = x1 + x2 + x3 + x4 + x5 + x6 + x7
    x9  = x1 + x2 + x3 + x4 + x5 + x6 + x7 + x8
    x10 = x1 + x2 + x3 + x4 + x5 + x6 + x7 + x8 + x9
    x11 = x1 + x2 + x3 + x4 + x5 + x6 + x7 + x8 + x9 + x10
    x12 = x1 + x2 + x3 + x4 + x5 + x6 + x7 + x8 + x9 + x10 + x11
    #x13 = x1 + x2 + x3 + x4 + x5 + x6 + x7 + x8 + x9 + x10 + x11 + x12
    return x12
end

function loss(x)
    sum(abs2, net(x))
end

x = CUDA.randn(128,128,128,20)
CUDA.@time Zygote.gradient(loss, x)
```

# pytorch
```python
import torch

def net(x1):
    x2  = x1
    x3  = x1 + x2
    x4  = x1 + x2 + x3
    x5  = x1 + x2 + x3 + x4
    x6  = x1 + x2 + x3 + x4 + x5
    x7  = x1 + x2 + x3 + x4 + x5 + x6
    x8  = x1 + x2 + x3 + x4 + x5 + x6 + x7
    x9  = x1 + x2 + x3 + x4 + x5 + x6 + x7 + x8
    x10 = x1 + x2 + x3 + x4 + x5 + x6 + x7 + x8 + x9
    x11 = x1 + x2 + x3 + x4 + x5 + x6 + x7 + x8 + x9 + x10
[...]

Issue #644, opened 2020-05-13
How to speed up pullbacks when iterating over arrays?


sethaxen:
Consider the following example from [base Julia](https://github.com/JuliaLang/julia/blob/v1.4.1/base/math.jl#L119-L128):

```julia
evalpoly(x, p::AbstractVector) = _evalpoly(x, p)

function _evalpoly(x, p)
    N = length(p)
    ex = p[end]
    for i in N-1:-1:1
        ex = muladd(x, ex, p[i])
    end
    ex
end
```

This function is simple and looks at each element of `p` once. Ideally, its gradient would be fast. However:

```julia
julia> using Zygote, BenchmarkTools

julia> x, p = rand(), randn(10000);

julia> @btime evalpoly(x, p);
  17.323 μs (1 allocation: 16 bytes)

julia> @btime Zygote.gradient(evalpoly, x, p);
  679.433 ms (490088 allocations: 1.51 GiB)
```

To work around this, I'm adding a custom rule for `evalpoly` to ChainRules.jl (https://github.com/JuliaDiff/ChainRules.jl/pull/190), that will speed things up dramatically once https://github.com/FluxML/Zygote.jl/pull/366 is merged:

```julia
julia> @btime Zygote.gradient(evalpoly, x, p); #10,000x faster than before!!
  43.338 μs (21 allocations: 156.81 KiB)
```

But this is a band-aid. How can we improve this in the general case? Is the main problem likely the adjoint for `getindex` (e.g. https://github.com/FluxML/Zygote.jl/issues/365)? But even with a great adjoint for `getindex`, Zygote won't know that each element is only used once and therefore it could do something fast like allocate a single gradient vector and fill it efficiently.

For reference, jax is able to get comparable performance to the custom rule without a rule (see https://github.com/google/jax/issues/3047#issuecomment-627764538).

shoyer:
JAX solves this problem in two ways:

- The XLA compiler can convert functionally pure accumulation at a single index into in-place updates (assuming the original array values are not reused).
- JAX has a special autodiff primitive `lax.scan` for loops with a statically known number of iterations. This simplifies memory allocation for the backwards pass.

mcabbott:
Could this be done as some `foldl` type operation? That's one way to convey that you are iterating once; I don't know whether that information is at present used well by Zygote but it could surely be. Is this what `lax.scan` is doing?

[...]

Issue #365, opened 2019-10-07
getindex performance


willtebbutt:
Backproping through `getindex` is pretty slow at the minute, due to the need to make copies of the entire array on the reverse-pass, and to fill it with zeros. The current implementation basically has two steps

1. Fill array with zeros and place non-zero entries where the original array was accessed.
2. Create new array by calling `accum` on the thing constructed in 1 and whatever the reverse-mode sensitivity already is.

Point 1 can be dealt with through the careful use of structurally-sparse arrays (i.e. a new `AbstractArray` type that is zero everywhere except for certain blocks or whatever). Point 2 is hard to deal handle without in-place accumulation.

@MikeInnes have you thought about doing in-place accumulation at all? How do you feel about it? Is it a thing that you might be convinced to enable somehow?

Drvi:
Fwiw, I tried to fix this for Tracker: https://github.com/FluxML/Flux.jl/pull/589/files, never actually finished it...

MikeInnes:
Yes, spot on with the analysis. Some kind of dynamic analysis of sparsity might be the way to do it, though that also makes things like future type inferability difficult even in principle.

In-place accumulation was enabled on #61, where it's a natural fit. It's possible that doing just accumulation without actually supporting mutation generally might be significantly simpler, though I suspect it may not be. Will have to consider it some more.

Doing things in-place is a bit problematic with Zygote's semantics, since the gradient of an `Int` array can be a `Float` etc. Automatically converting int->float is easy but then complex numbers are an issue, etc. We could do fancier tricks here to make that fast in the common case, but it's increasingly inelegant.

The other option is to just recommend `Zygote.Buffer` here, since that does already have its accumulation done in-place.

willtebbutt:
Okay, thanks for the info. My particular problem can be solved with a custom adjoint, so I'll do that for now. Will leave this open in case anyone fancies tackling it in the general case.

sdaxen:
`getindex` is also not fully compatible with complex sensitivities. See #376.

PR #944, opened 2021-04-12
Nested AD for `getindex`

axsk:
This incorporates parts of #77 and fixes #820 by implementing the adjoint for the adjoint, i.e. nested derivative, for `getindex`, thus facilitating higher order AD.

I cannot asses which parts of #77 that I left out are further necessary, but this works for `getindex`  and other methods depending on it (e.g. `sum`).

PR #962, opened 2021-05-04
RFC: more efficient `∇getindex` 

mcabbott:
This is a small improvement to the gradient for `getindex` in the easiest case, by returning a very simple one-nonzero-element array. 

~~It also changes how `accum` works: adding two of these will produce an `Array`, but accumulating the next one may as well mutate that. Maybe it is safe to do that in general, I am not 100% sure? Maybe CI will tell us?~~ [It is not, and it didn't, see below]

Note that it will never produce a SparseArray. ~~My take is that if you call getindex once then this `OneElement` is fine, and if you call it on every element in the array, then accumulating in an Array is optimal; in-between cases where you call it on 1% of the elements seem likely to be very rare, and sparse arrays add all sorts of complication.~~ [This won't apply anymore]

And, are there `CuArray` concerns? `∇getindex` is careful about making a similar zero to write into, but this change is only for scalar indexing, which is (usually) disallowed anyway. 

This doesn't give a huge speedup, but it does save a lot of memory. I'm not too sure what the "10089 allocations" here actually are, maybe there is some further trick to remove them? (Might also be useful for someone to time this on a different computer too -- these are M1 mac, which has unusual memory, and Julia 1.7 native.)
```julia
f4(x) = sum([x[i]^2 for i in eachindex(x)])
@btime gradient(f4, $(collect(1:1000)))
#  1.997 ms (12829 allocations: 16.05 MiB)  originally
#  1.502 ms (11087 allocations: 8.46 MiB)   with OneElement alone
#  1.520 ms (11830 allocations: 8.31 MiB)   with accum(x::DenseArray,...) alone
#    992.917 μs (10089 allocations: 738.44 KiB)  this PR

@btime gradient(x -> sum(x * x'), $(rand(10,1000)));  # no getindex
#  79.625 μs (30 allocations: 237.06 KiB)  original
#  79.334 μs (28 allocations: 158.86 KiB)  with accum
```
On the example from https://github.com/FluxML/Zygote.jl/issues/644 -- discussion there had other comparable proposals:
```julia
function _evalpoly(x, p)
    N = length(p)
    ex = p[end]
    for i in N-1:-1:1
        ex = muladd(x, ex, p[i])
    end
    ex
end
x, p = rand(), randn(10000);

@btime _evalpoly(x, p);
# 12.416 μs (1 allocation: 16 bytes)

@btime Zygote.gradient(_evalpoly, x, p);
# 170.356 ms (730104 allocations: 1.53 GiB)     original
# 139.957 ms (710106 allocations: 800.45 MiB)   with accum(x::DenseArray,...) alone
# 136.468 ms (740104 allocations: 801.90 MiB)   with OneElement alone
#  87.931 ms (720108 allocations: 38.35 MiB)    with both
#  59.191 ms (720108 allocations: 38.35 MiB)    with accum(..., y::OneElement) too
```
Xref also #365 (getindex) and #905 (accumulation).

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -32,8 +32,10 @@ end
 
 @adjoint view(x::AbstractArray, inds...) = view(x, inds...), ∇getindex(x, inds)
 
-∇getindex(x::AbstractArray, inds) = dy -> begin
-  if inds isa  NTuple{<:Any, Integer}
+∇getindex(x::AbstractArray{T,N}, inds) where {T,N} = dy -> begin
+  if inds isa NTuple{N,Int} && T <: Number
+    dx = OneElement(dy, inds, axes(x))
+  elseif inds isa NTuple{<:Any, Integer}
     dx = _zero(x, typeof(dy))
     dx[inds...] = dy
   else
@@ -43,6 +45,16 @@ end
   end
   return (dx, map(_->nothing, inds)...)
 end
+
+struct OneElement{T,N,I,A} <: AbstractArray{T,N}
+  val::T
+  index::I
+  axes::A
+  OneElement(x::T, i::I, a::A) where {T,I<:NTuple{N,Int},A} where {N} = new{T,N,I,A}(x, i, a)
+end
+Base.size(A::OneElement) = map(length, A.axes)
+Base.axes(A::OneElement) = A.axes
+Base.getindex(A::OneElement{T,N}, i::Vararg{Int,N}) where {T,N} = ifelse(i==A.index, A.val, zero(T))
 
 _zero(xs::AbstractArray{<:Number}, T::Type{Nothing}) = fill!(similar(xs), zero(eltype(xs)))
 _zero(xs::AbstractArray{<:Number}, T) = fill!(similar(xs, T), false)
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -46,15 +46,20 @@ end
   return (dx, map(_->nothing, inds)...)
 end
 
+"""
+    OneElement(val, ind, axes) <: AbstractArray
+
+Extremely simple `struct` used for the gradient of scalar `getindex`.
+"""
 struct OneElement{T,N,I,A} <: AbstractArray{T,N}
   val::T
-  index::I
+  ind::I
   axes::A
-  OneElement(x::T, i::I, a::A) where {T,I<:NTuple{N,Int},A} where {N} = new{T,N,I,A}(x, i, a)
+  OneElement(val::T, ind::I, axes::A) where {T<:Number, I<:NTuple{N,Int}, A} where {N} = new{T,N,I,A}(val, ind, axes)
 end
 Base.size(A::OneElement) = map(length, A.axes)
 Base.axes(A::OneElement) = A.axes
-Base.getindex(A::OneElement{T,N}, i::Vararg{Int,N}) where {T,N} = ifelse(i==A.index, A.val, zero(T))
+Base.getindex(A::OneElement{T,N}, i::Vararg{Int,N}) where {T,N} = ifelse(i==A.ind, A.val, zero(T))
 
 
 _zero(xs::AbstractArray{<:Number}, T::Type{Nothing}) = fill!(similar(xs), zero(eltype(xs)))
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -56,6 +56,8 @@ Base.size(A::OneElement) = map(length, A.axes)
 Base.axes(A::OneElement) = A.axes
 Base.getindex(A::OneElement{T,N}, i::Vararg{Int,N}) where {T,N} = ifelse(i==A.index, A.val, zero(T))
 
+accum(x::Array, y::OneElement) = (@inbounds x[y.index...] = accum(x[y.index...], y.val); x)
+
 _zero(xs::AbstractArray{<:Number}, T::Type{Nothing}) = fill!(similar(xs), zero(eltype(xs)))
 _zero(xs::AbstractArray{<:Number}, T) = fill!(similar(xs, T), false)
 _zero(xs::AbstractArray, T) = fill!(similar(xs, Union{Nothing, T}), nothing)
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -56,7 +56,6 @@ Base.size(A::OneElement) = map(length, A.axes)
 Base.axes(A::OneElement) = A.axes
 Base.getindex(A::OneElement{T,N}, i::Vararg{Int,N}) where {T,N} = ifelse(i==A.index, A.val, zero(T))
 
-accum(x::Array, y::OneElement) = (@inbounds x[y.index...] = accum(x[y.index...], y.val); x)
 
 _zero(xs::AbstractArray{<:Number}, T::Type{Nothing}) = fill!(similar(xs), zero(eltype(xs)))
 _zero(xs::AbstractArray{<:Number}, T) = fill!(similar(xs, T), false)
--- a/src/lib/lib.jl
+++ b/src/lib/lib.jl
@@ -14,6 +14,7 @@ accum(x, y, zs...) = accum(accum(x, y), zs...)
 
 accum(x::Tuple, y::Tuple) = accum.(x, y)
 accum(x::AbstractArray, y::AbstractArray) = accum.(x, y)
+accum(x::DenseArray, y::AbstractArray) = x .= accum.(x, y)
 
 @generated function accum(x::NamedTuple, y::NamedTuple)
   # assumes that y has no keys apart from those also in x
--- a/src/lib/lib.jl
+++ b/src/lib/lib.jl
@@ -15,6 +15,8 @@ accum(x, y, zs...) = accum(accum(x, y), zs...)
 accum(x::Tuple, y::Tuple) = accum.(x, y)
 accum(x::AbstractArray, y::AbstractArray) = accum.(x, y)
 accum(x::DenseArray, y::AbstractArray) = x .= accum.(x, y)
+# work around bug fixed in https://github.com/JuliaLang/julia/pull/39859
+accum(x::DenseVector, y::AbstractArray) = x .= accum.(x, vec(y))
 
 @generated function accum(x::NamedTuple, y::NamedTuple)
   # assumes that y has no keys apart from those also in x
--- a/src/lib/lib.jl
+++ b/src/lib/lib.jl
@@ -12,11 +12,8 @@ accum(x, y) =
 
 accum(x, y, zs...) = accum(accum(x, y), zs...)
 
-accum(x::Tuple, y::Tuple) = accum.(x, y)
-accum(x::AbstractArray, y::AbstractArray) = accum.(x, y)
-accum(x::DenseArray, y::AbstractArray) = x .= accum.(x, y)
-# work around bug fixed in https://github.com/JuliaLang/julia/pull/39859
-accum(x::DenseVector, y::AbstractArray) = x .= accum.(x, vec(y))
+accum(x::Tuple, ys::Tuple...) = accum.(x, ys...)
+accum(x::AbstractArray, ys::AbstractArray...) = accum.(x, ys...)
 
 @generated function accum(x::NamedTuple, y::NamedTuple)
   # assumes that y has no keys apart from those also in x
--- a/test/features.jl
+++ b/test/features.jl
@@ -481,3 +481,22 @@ end
   Zygote.gradient(loss_adjoint,[1.0])
   @test x[1] == x[2]
 end
+
+@testset "accumulation" begin
+  # from https://github.com/FluxML/Zygote.jl/issues/905
+  function net(x1)
+    x2  = x1
+    x3  = x1 + x2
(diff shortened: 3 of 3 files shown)


oxinabox:
in the longer term this will be resolved once we support ChainRules inplacable thunks.
Til then this might be a good idea.

> It also changes how accum works: adding two of these will produce an Array, but accumulating the next one may as well mutate that. Maybe it is safe to do that in general, I am not 100% sure? Maybe CI will tell us?

We can define a custom rule for `accum` that changes the primal so that it doesn't mutate if you are doing nested AD.

oxinabox:
On my computer
```
julia> versioninfo()
Julia Version 1.6.2-pre.0
Commit dd122918ce* (2021-04-23 21:21 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin20.3.0)
  CPU: Intel(R) Core(TM) i7-8559U CPU @ 2.70GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, skylake)
```

here are the benchmarks
```
@btime gradient(f4, $(collect(1:1000)))
  2.027 ms (7068 allocations: 15.77 MiB)    current master
  90.018 μs (2073 allocations: 307.52 KiB)  this PR

@btime gradient(x -> sum(x * x'), $(rand(10,1000)));  # no getindex
  108.439 μs (29 allocations: 237.16 KiB)   current master
  105.164 μs (27 allocations: 158.95 KiB)   this PR

julia> @btime _evalpoly(x, p);
  9.589 μs (1 allocation: 16 bytes)

julia> @btime Zygote.gradient(_evalpoly, x, p);
  942.210 ms (680106 allocations: 1.52 GiB) current master
  67.668 ms (640110 allocations: 29.20 MiB) this PR
```

Notable differences from yours in the top post:
All allocations are lower (especially in the collect example)
but the allocations are much lower still for this PR.

This PR makes the collect example waaaay faster than what you saw: an order of magnitude faster.

The current maseter timing fro the eval poly example is waaay worse for me.

----

This is great. We should have been doing this for ages.
This is going to be the single biggest quality of life improvement in quite a while.

mcabbott:
Those times look great, thanks. I should try on 1.6 + rosetta too, maybe there are other bugs involved.

> We can define a custom rule for accum that changes the primal so that it doesn't mutate if you are doing nested AD.

Good point, hadn't thought about nested AD at all. There appear to be no tests, so perhaps that can be looked at in another PR. Nested `getindex` specifically might also be easier in this case, xref #944 which I have not read closely.

DhairyaLGandhi:
Is that correct in general? Aso, could we replicate this with `only` instead?

src/lib/lib.jl
```diff
@@ -14,6 +14,7 @@ accum(x, y, zs...) = accum(accum(x, y), zs...)
 
 accum(x::Tuple, y::Tuple) = accum.(x, y)
 accum(x::AbstractArray, y::AbstractArray) = accum.(x, y)
+accum(x::DenseArray, y::AbstractArray) = x .= accum.(x, y)
```
oxinabox, line 17: ChainRulesCore has a more complicated set of rules for working out if this is allowed.
Since it is not clear if all `Dense` subtypes are going to be mutable
Which we can get by doing
```suggestion
accum(x::DenseArray, y::AbstractArray) = ChainRulesCore.add!!(x, y)
```

https://github.com/JuliaDiff/ChainRulesCore.jl/blob/5f1613ffe4f46287984f2fc2733bf4ec8a8d65c9/src/accumulation.jl#L27-L63

mcabbott: No strong opinions, this was the low-complication way of catching Array and CuArray. 
```
julia> subtypes(DenseArray)
6-element Vector{Any}:
 Array
 Base.CodeUnits
 Base.Experimental.Const
 Random.UnsafeView
 SharedArrays.SharedArray
 SuiteSparse.CHOLMOD.Dense

julia> Base.Experimental.Const(rand(2)) isa DenseArray
true
```

oxinabox: Lets leave it as `DenseArray` for now and worry about it later

oxinabox:
> Is that correct in general? 

Is what correct in general?
This looks like a correct inplementation of a n-dimentional one-hot array.
And inplace accumulation of gradients is safe and correct in general due to the face that AD is logically linear (which means each differential is only ever used once so safe to mutate since it is not going to be used anywhere ekse at the point you are calling `accum` on it, at least during the pullback (less safe if it is something the user is inputting though, since then it might be used more than once.)

> Aso, could we replicate this with only instead?

Which bit?

willtebbutt:
The other thing: should we be concerned about race conditions?

@DhairyaLGandhi and @MikeInnes know more about this than I do (I know basically nothing), but I would imagine that if you have parallel code somewhere, this modification could be problematic? My impression is that Zygote currently works fine with threads / parallelism generally -- would this change that?

oxinabox:
> Should we be concerned about race conditions?

I don't think so.
From what I can see `Task` related things never call `accum` directly (most pullbacks never call `accum`). They return something that gets accumulated outside of that.
I suspect also there is something about structure of AD programs that also mean we don't have to worry about it.
Parallelism in reverse must follow parallelism in forward but backwards, which should mean that primals are free from race conditions in forward must have differentials that are free from race conditions in reverse (thoguh that might not apply since we are turning a copy into a mutation).

I think we can assume it is safe and wait for evidence to the contrary (though debugging race conditions *sucks*)

src/lib/lib.jl
mcabbott, line 17: Test error might be relevant:
```
Error During Test at /var/lib/buildkite-agent/builds/p6000-gpuci3-julia-csail-mit-edu/julialang/zygote-dot-jl/test/cuda.jl:27
  | Got exception outside of a @test
  | DimensionMismatch("cannot broadcast array to have fewer dimensions")

[8] accum(x::Vector{Float32}, y::LinearAlgebra.Adjoint{Float32, Matrix{Float32}})
  | @ Zygote /var/lib/buildkite-agent/builds/p6000-gpuci3-julia-csail-mit-edu/julialang/zygote-dot-jl/src/lib/lib.jl:17
```
https://buildkite.com/julialang/zygote-dot-jl/builds/223#94c4e405-b3e9-4cff-8150-689a90dcf254

This seems to only happen on CuArrays. Maybe this is a bug that got fixed in Base, I forget?

oxinabox: Gross. I feel like someone is playing loose and fast with their differential's types here.
I am guessing that somewhere a differnetial that should have been an `Adjoint{Float32, Vector{Float32}})` was instead 1x2  `Matrix{Float32}`.
And thus what should have happend was taking the adjoint of the adjoint of a vector  should have given us a `Vector`
(Since https://github.com/julialang/julia/issues/4774)
Anyway, if my theory is right then we should be able to trackdown and fix that root cause.

Otherwise we can special case this and call `accum(x::DenseVector, y::AbstractArray) = x .= accum.(x, vec(y))`
I think that code would always give the right answer or error?
as a hack, leaving a TODO linking to an issue to sort it out later?

mcabbott: Yea 4774 is a pain sometimes. I think everything that can be, needs to be robust to this difference. I suppose CUDA hasn't caught up with Base here.

mcabbott:
It turns out that #905 is the way to produce sort of bug I was a bit worried about. Something is doing `Δ -> (Δ, Δ, Δ)`, which results in `accum(x, y, zs...) = accum(accum(x, y), zs...)` running with `x === zs[1]`, which leads to:
```
julia> gradient(x -> sum(abs, x + x + x), ones(2))
([4.0, 4.0],)

julia> ForwardDiff.gradient(x -> sum(abs, x + x + x), ones(2))
2-element Vector{Float64}:
 3.0
 3.0
```

mcabbott:
Yes that might be enough. I don't know the literature (obviously!) but it's hard to then picture it going wrong. Would there be cases where this hurts, as well as ones where it helps overall?

You could surely hack in a test to check for aliasing, during testing. ChainRules could easily do that too. I'm not actually sure where the offending rule is, this one (for `A+B`) doesn't seem to be the culprit:

https://github.com/FluxML/Zygote.jl/blob/master/src/lib/array.jl#L742

Just making the update vararg like `x .= accum.(x, ys...)` in fact avoids the problem here, but obviously won't in general. Doing that without mutation seems like a good idea, saves a lot of memory in #905. Maybe less in real cases, but e.g. `gradient(x -> x[1] + x[2] + x[3], ones(3))` does hit that method.

mcabbott:
OK, I think I see how to ensure broadcasting takes preventative copies. For times on this page it only costs 10% off the fastest unsafe way, but in https://github.com/FluxML/Zygote.jl/issues/905#issuecomment-835054679 it saves no memory, and is slower than master.

mcabbott:
Re this PR as it stands, one comment is that it does a bit less to preserve the types of arrays, e.g.:
```julia
julia> using StaticArrays

julia> gradient(x -> x[1], SA[1,2,3])[1]
3-element Zygote.OneElement{Int64, 1, Tuple{Int64}, Tuple{SOneTo{3}}} with indices SOneTo(3):
 1
 0
 0

julia> gradient(x -> x[1] + x[2], SA[1,2,3])[1]
3-element SizedVector{3, Int64, Vector{Int64}} with indices SOneTo(3):
 1
 1
 0
```
compared to master:
```julia
julia> gradient(x -> x[1] + x[2], SA[1,2,3])[1]   # better than a SizedVector on similar case
3-element MVector{3, Int64} with indices SOneTo(3):
 1
 1
 0

julia> gradient(x -> sum(x .+ x'), SA[1,2,3])[1]  # but many other cases don't preserve this
3×1 Matrix{Int64}:
 6
 6
 6

julia> gradient(x -> sum(abs, x .+ x'), SA[1,2,3])[1]  # and many simple things don't work at all
ERROR: MethodError: no method matching _mapfoldl(::typeof(identity), ::typeof(+), ::Tuple{Int64, Int64}, ::StaticArrays._InitialValue, ::Size{(3, 3)}, ::SMatrix{3, 3, Int64, 9})
```
I wondered a bit whether OneElement should capture the type of the array, but the method of similar I think you'd want to use when turning back into a dense array doesn't seem to exist, surprisingly?
```julia
julia> similar(typeof([1,2]), axes([1,2]))
2-element Vector{Int64}:
 12
 57

julia> similar(typeof(SA[1,2]), axes(SA[1,2]))
2-element MVector{2, Int64} with indices SOneTo(2):
 4585456624
          3

julia> similar(typeof([1,2]), Float64, axes([1,2]))  # with eltype
ERROR: MethodError: no method matching similar(::Type{Vector{Int64}}, ::Type{Float64}, ::Tuple{Base.OneTo{Int64}})

julia> similar([1,2], Float64, axes([1,2]))  # with an instance it's fine
2-element Vector{Float64}:
 0.0
 2.2655161774e-314
```

oxinabox (approved):

oxinabox:
Are you ready for this to be merged?
Given all existing tests pass, and new ones, this feels safe.

src/lib/array.jl
```diff
@@ -44,6 +46,17 @@ end
...
 
+struct OneElement{T,N,I,A} <: AbstractArray{T,N}
+  val::T
+  index::I
+  axes::A
+  OneElement(x::T, i::I, a::A) where {T,I<:NTuple{N,Int},A} where {N} = new{T,N,I,A}(x, i, a)
+end
+Base.size(A::OneElement) = map(length, A.axes)
+Base.axes(A::OneElement) = A.axes
+Base.getindex(A::OneElement{T,N}, i::Vararg{Int,N}) where {T,N} = ifelse(i==A.index, A.val, zero(T))
```
mcabbott, line 57: Perhaps this should have a brief docstring:
```suggestion
"""
    OneElement(val, ind, axes) <: AbstractArray

Extremely simple `struct` used for the gradient of scalar `getindex`.
"""
struct OneElement{T,N,I,A} <: AbstractArray{T,N}
  val::T
  ind::I
  axes::A
  OneElement(val::T, ind::I, axes::A) where {T<:Number, I<:NTuple{N,Int}, A} where {N} = new{T,N,I,A}(val, ind, axes)
end
Base.size(A::OneElement) = map(length, A.axes)
Base.axes(A::OneElement) = A.axes
Base.getindex(A::OneElement{T,N}, i::Vararg{Int,N}) where {T,N} = ifelse(i==A.ind, A.val, zero(T))
```

mcabbott (commented):
Yes, let's do this.

Failure on master ought to be fixed by https://github.com/FluxML/IRTools.jl/pull/86

(Minor aesthetic changes in the review.)

src/lib/array.jl
```diff
@@ -44,6 +46,17 @@ end
...
+
```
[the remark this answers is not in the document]
oxinabox: Can make a follow up PR for it later?
Needs tests, i would rather get this in sooner than have more back and forth.
and waiting for CI

mcabbott:
Oh I see. I think we both hit the merge button at the same time, yours succeeded & mine gave a confusing error... which I tried to rebase to fix... but this is merged & can be closed.

DhairyaLGandhi:
Most of the work for static arrays is pretty orthogonal to the concerns of accumulation. It seems odd to make future work harder.

Issue #599, opened 2020-04-14
gradient of array splat should be an array


CarloLucibello:
The last example, behaves differently from the first two 
```julia
julia> gradient(w -> sum(w), [1,1])  # ok, gradient is array
([1, 1],)   

julia> gradient(w -> sum([w[1], w[2]]), [1,1]) # ok, gradient is array
([1, 1],)

julia> gradient(w -> sum([w...]), [1,1]) # NOT OK, gradient is tuple
((1, 1),)
```
One of the problems with returning a tuple, is that it breaks Flux's `update!` function, which is expecting an array

AzamatB:
See [#489](https://github.com/FluxML/Zygote.jl/pull/489#issuecomment-581983991), where @MikeInnes described the approach for fixing this.

mcabbott:
BTW this is not fixed, although ProjectTo on final results hides the problem in the example above. If the gradient needs to be passed to another pullback expecting an array, you will get errors. 

To avoid the projection & see the problem, you can do this:
```
julia> pullback(w -> sum([w...]), [1,1])[2](1.0)
((1.0, 1.0),)

julia> pullback(w -> sum([w...]), [1 2; 3 4])[2](1.0)
((1.0, 1.0, 1.0, 1.0),)
```

PR #965, opened 2021-05-06
Use `clamptype` mechanism to project onto cotangent space

### LINKED ISSUE #431: Gradient of inv and logdet UpperTriangular matrix is not UpperTriangular
For sparse Gaussian processes, one often wants to parametrize the variational distribution using its Cholesky factors, which form an upper-triangular matrix and are trainable.

However, when taking gradients, Zygote instead computes the gradient with respect to all entries in the UpperTriangular matrix, including ones that are set to zero by virtue of it being UpperTriangular. This is related to #163.

For the moment, a workaround suitable for training models is to simply call `UpperTriangular(x)` on the upper-triangular matrix `x` before using it.

### LINKED ISSUE #917: Trying to differentiate a function with complex and real inputs results in error
I want to evaluate the hessian of a $\mathbb C^n \times \mathbb R^m \to \mathbb R$  function.
The actual function has several inputs, some of which are real and some of which are complex. To generate the hessian, I create a wrapper function with one real vector as argument and create the complex and real arguments for the actual function from this.

I get the following error, though:

```
julia> Zygote.hessian(fun, rand(9))
ERROR: MethodError: no method matching extract(::Array{Complex{ForwardDiff.Dual{Nothing,Float64,12}},1})
Closest candidates are:
  extract(::ForwardDiff.Dual) at /home/constantin/.julia/packages/Zygote/KpME9/src/lib/forward.jl:12
  extract(::AbstractArray{ForwardDiff.Dual{T,V,N},N1} where N1) where {T, V, N} at /home/constantin/.julia/packages/Zygote/KpME9/src/lib/forward.jl:14
Stacktrace:
 [1] forward_jacobian(::Zygote.var"#1086#1087"{typeof(Ropt3)}, ::Array{Float64,1}, ::Val{12}) at /home/constantin/.julia/packages/Zygote/KpME9/src/lib/forward.jl:23
 [2] forward_jacobian(::Function, ::Array{Float64,1}) at /home/constantin/.julia/packages/Zygote/KpME9/src/lib/forward.jl:40
 [3] hessian(::Function, ::Array{Float64,1}) at /home/constantin/.julia/packages/Zygote/KpME9/src/lib/utils.jl:113
 [4] top-level scope at REPL[15]:1
 [5] run_repl(::REPL.AbstractREPL, ::Any) at /tmp/portage/dev-lang/julia-1.5.3/work/julia-1.5.3/usr/share/julia/stdlib/v1.5/REPL/src/REPL.jl:288
```

MWE
```julia
using Zygote

function fun(v)
    c = v[1:3] + v[4:6]*im
    r = v[7:9]
    sum(r .* abs2.(c)) # This would be calling my actual function depending on r and c
end

Zygote.hessian(fun, rand(9))
```

mcabbott:
This adds a mechanism to enforce that gradients stay within a cotangent space defined by the original variable. And uses this to enforce that real numbers have real gradients, and that some LinearAlgebra structured matrix types are respected.

This is a conservative in that, rather than trying to have a watertight theory of what cotangent types are allowed, it only acts on types it knows how to fix, and lets everything else pass through, as it does today. Perhaps it could learn how to fix some gradients which are NamedTuples / Composite, but later. It should also be easy for packages to extend this for their own types, if desired.

I believe there is broad agreement that we do want to respect these constraints. The argument against an explicit clamp like this is that ideally rules would always respect them and so it would not be needed. That's a nice asperation but a world in which the rule for (say) `atan(x,y)` gets re-written to understand what to do about `x::Real, y::Complex` seems far away. When the rule *does* get it right, this clamp should be free, as it's only about types. Individual `clamptype` methods can, if we wish, contain `@debug` statements to help track down problems. It is wasteful of course to carve a `Diagonal` out of a full Matrix, but better to do it immediately than to let the problem propagate. Again, it's a step forwards, not a watertight re-design.

Still WIP, ~~it hooks onto Zygote's own rules, but not yet onto those defined in ChainRules.~~, some bugs.

Closes https://github.com/FluxML/Zygote.jl/issues/342, closes https://github.com/FluxML/Zygote.jl/issues/402. Fixes #917, fixes #431.

`@testset "clamped gradients"` shows some examples of what works.
 
Needs https://github.com/FluxML/ZygoteRules.jl/pull/16, thus CI won't pass. Locally, the remaining issues are things like matrix `exp` of Hermitian (where I believe the output has not changed), and FFT tests (many of which relied on real input giving a complex gradient).

### DIFF
--- a/src/lib/clamp.jl
+++ b/src/lib/clamp.jl
@@ -7,30 +7,31 @@ import ZygoteRules: clamptype
 # Bool, Real, Complex
 
 clamptype(::Type{Bool}, dx) = nothing
-clamptype(::Type{Bool}, dx::Complex) = nothing  # ambiguity
 clamptype(::Type{<:AbstractArray{Bool}}, dx::AbstractArray) = nothing
-# clamptype(::Type{<:BitArray}, dx::AbstractArray) = nothing
 
 clamptype(::Type{<:Real}, dx::Complex) = real(dx)
 clamptype(::Type{<:AbstractArray{<:Real}}, dx::AbstractArray{<:Complex}) = real(dx)
+
+clamptype(::Type{Bool}, dx::Complex) = nothing  # ambiguity
+clamptype(::Type{<:AbstractArray{Bool}}, dx::AbstractArray{<:Complex}) = nothing
 
 # LinearAlgebra's matrix types
 
 for Wrap in [:Diagonal, :UpperTriangular, :LowerTriangular]
   @eval begin
-    clamptype(::Type{<:$Wrap}, dx::$Wrap) where {T,PT} = dx
-    clamptype(::Type{<:$Wrap{T,PT}}, dx::AbstractMatrix) where {T,PT} = clamptype(PT, $Wrap(dx))
-    clamptype(::Type{<:$Wrap{<:Real}}, dx::$Wrap{<:Complex}) = $Wrap(real(parent(dx)))
-    # clamptype(::Type{<:$Wrap{Bool}}, dx::$Wrap{<:Complex}) = nothing  # ambiguity
+    clamptype(::Type{<:$Wrap{T,PT}}, dx::$Wrap) where {T,PT} = 
+      clamptype(PT, dx)
+    clamptype(::Type{<:$Wrap{T,PT}}, dx::AbstractMatrix) where {T,PT} = 
+      clamptype(PT, $Wrap(dx))
   end
 end
 
 for (trans, Wrap) in [(transpose, :Symmetric), (Base.adjoint, :Hermitian)]
   @eval begin
-    clamptype(::Type{<:$Wrap}, dx::$Wrap) where {T,PT} = dx
-    clamptype(::Type{<:$Wrap{T,PT}}, dx::AbstractMatrix) where {T,PT} = clamptype(PT, $Wrap(_twofold($trans, dx)))
-    clamptype(::Type{<:$Wrap{<:Real}}, dx::$Wrap{<:Complex}) = $Wrap(real(parent(dx)))
-    # clamptype(::Type{<:$Wrap{Bool}}, dx::$Wrap{<:Complex}) = nothing  # ambiguity
+    clamptype(::Type{<:$Wrap{T,PT}}, dx::$Wrap) where {T,PT} = 
+      clamptype(PT, dx)
+    clamptype(::Type{<:$Wrap{T,PT}}, dx::AbstractMatrix) where {T,PT} = 
+      clamptype(PT, $Wrap(_twofold($trans, dx)))
   end
 end
 
--- a/src/lib/clamp.jl
+++ b/src/lib/clamp.jl
@@ -1,7 +1,4 @@
 
-using LinearAlgebra: Diagonal, UpperTriangular, UnitUpperTriangular, LowerTriangular, UnitLowerTriangular
-using LinearAlgebra: AdjointAbsVec, TransposeAbsVec, AdjOrTransAbsVec
-
 import ZygoteRules: clamptype
 # This sees a tuple of argument types, and can modify the resulting tuple of tangents
 
@@ -26,6 +23,9 @@ clamptype(Ts::Tuple, dxs::Tuple{}) = (@error "mismatch!" Ts; ())
 
 clamptype(::Type{<:Real}, dx::Complex) = real(dx)
 clamptype(::Type{<:AbstractArray{<:Real}}, dx::AbstractArray) = real(dx)
+
+using LinearAlgebra: Diagonal, UpperTriangular, UnitUpperTriangular, LowerTriangular, UnitLowerTriangular
+using LinearAlgebra: AdjointAbsVec, TransposeAbsVec, AdjOrTransAbsVec
 
 # LinearAlgebra's matrix types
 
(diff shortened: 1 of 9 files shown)


willtebbutt:
Modification of my comment on the ZygoteRules PR -- I agree it's better to aggregate words in the same place!

I generally really like this, but I wonder whether there's a better name availlable.

This is a bit verbose, but something like project_onto_cotangent_space but might be more informative?

My other question is whether there are any situations in which type information is insufficient to make this work? For example, to know how to clamp a `SparseMatrixCSC` you need to know the precise sparsity pattern, which is only available at runtime, so this seems like an example where you need the object itself. I'm not proposing to actually implement stuff for `SparseMatrixCSC`, just saying that perhaps the API should change to require the arguments to the rrule be passed to `clamptype`, rather than just their type.

I made some remarks on the other PR about implementing this here vs in ChainRules. It appears that this has been considered, and for the sake of obtaining something that works well, it probably makes sense to implement this here for now, rather than in ChainRules. We can think about transfering this to somewhere in ChainRules once ChainRules types are utilised fully in Zygote.

mcabbott:
One reason for here not ChainRules is that it can work today, as a step forward. The tests which work now work without interacting with ChainRules adjoints. Some are caused by broadcasting, and I didn't think there was proposal that ChainRules would handle the entire broadcast (even if it did the scalars). And I'd like to establish the precedent that this is what the correct behaviour is. 

If rules upstream get better at hitting the passthrough not the clamp, that’s great too.  If the entire thing gets re-organised some day before Zygote is sent to the glue factory, that's also OK. 

Mike's objection to the original version was that passing the object means it has to stay in memory, but the type is lighter. (#342) Maybe that's the problem opaque closures will solve, but not today, if I understand right? We could introduce another hook by which types define what object they want passed forwards instead of their type... at the cost of a bit more complexity, define an extensible `passthis(x) = typeof(x)` and use it in `argTs = map(typeof, ($(argnames...),))` here (in the ZygoteRules bit).   

Are there other examples where this might be desirable? It doesn't seem obvious to me that zeros of a `SparseMatrixCSC` should automatically be regarded as structural, see e.g. this discussion https://github.com/FluxML/Zygote.jl/issues/163#issuecomment-486343665. It was suggested that some `TakeSparsitySeriously` wrapper could opt-in to such things (which would still need a mechanism).

Not set on the name, Zygote's flavour is to make things compact so I try to fit in, but perhaps there's a better choice.

src/lib/clamp.jl
```diff
@@ -0,0 +1,64 @@
...
+for (trans, Wrap) in [(transpose, :Symmetric), (Base.adjoint, :Hermitian)]
+  @eval begin
+    clamptype(::Type{<:$Wrap{T,PT}}, dx::$Wrap) where {T,PT} = 
+      clamptype(PT, dx)
+    clamptype(::Type{<:$Wrap{T,PT}}, dx::AbstractMatrix) where {T,PT} = 
+      clamptype(PT, $Wrap(_twofold($trans, dx)))
+  end
+end
+
+_twofold(trans, dx) = (dx .+ trans(dx)) ./ 2
```
mcabbott, line 51: Notice this:
```julia
using Zygote # this PR
using LinearAlgebra, FiniteDifferences, ForwardDiff, ChainRules

gradient(x -> Symmetric([x 3x; 5x 7x])[1,2], pi)   # (3,)
gradient(A -> A[1,2], Symmetric(rand(2,2)))        # ([0.0 0.5; 0.5 0.0],)
gradient(A -> Symmetric(A)[1,2], rand(2,2))        # ([0.0 1.0; 0.0 0.0],)

grad(central_fdm(5, 1), A -> A[1,2], Symmetric(rand(2,2))) # ([0.0 1.0; 1.0 0.0],) # weird? 
grad(central_fdm(5, 1), A -> Symmetric(A)[1,2], rand(2,2)) # ([0.0 1.0; 0.0 0.0],) # fine
```
If you use `_twofold(trans, dx) = dx .+ trans(dx) .- Diagonal(dx)`, then the Zygote results will double, and the 1st & 3rd will be quite obviously wrong. And this isn't a projection operator.

Is the FiniteDifferences result the right thing for a tangent not cotangent? Or is it a bug? As far as I can tell it's doing something like this:
```julia
ve, re = to_vec(Symmetric([1 3; 5 7])) # ve == [1,3,3,7]
re([1,-42,3,7]) == [1 3; 3 7]

re2(v) = Symmetric(reshape(v,2,2))
v2 = rand(1:999, 4); re(v2) == re2(v2)
re2(ForwardDiff.gradient(v -> re2(v)[1,2], ve)) == [0 1; 1 0]
```
Looking in ChainRules, there seem to be few tests using this behaviour of FiniteDifferences. There is a rule for `Matrix`, which applies `(dx .+ dx' .- Diagonal(dx)` to match:
```julia
rrule(Matrix, Symmetric(rand(2,2)))[2]([1 3; 5 7])[2]    == [1 8; 8 7] 
pullback(Matrix, Symmetric(rand(3,3)))[2]([1 3; 5 7])[1] == [1 4; 4 7] # with this PR
```
Defined here https://github.com/JuliaDiff/ChainRules.jl/blob/4e3164a3a48d4da35e0112d30be7ea9dbdaf3920/src/rulesets/LinearAlgebra/symmetric.jl#L71 where `_symmetric_back` is also the gradient of `Symmetric` where it makes more sense IMO. (Originally from Zygote, I think.)

mcabbott:
> whether there are any situations in which type information is insufficient to make this work?

One example is https://github.com/FluxML/Zygote.jl/issues/599 (restoring arrays from splats), for which this PR can provide a solution very similar to https://github.com/FluxML/Zygote.jl/pull/489. But to recover a splatted matrix from a tuple, you need the size not just the type. (Discussion of options: https://github.com/FluxML/Zygote.jl/pull/489#issuecomment-581983991, this is 1.)

PR #61, opened 2019-01-24
Handle mutation

MikeInnes:
WIP; currently has a significant performance impact, as my main goal is to get more complex programs compiling before optimising things.

PR #966, opened 2021-05-07
Utilize ChainRulesCore thunks

oschulz:
**Note: Requires FluxML/ZygoteRules.jl#17**

Currently, Zygote always unthunks `ChainRuleCore` thunks, which is wasteful and may also lead to trouble in cases why a thunks just can't be run for the given types/contents.

With

```julia
using ChainRulesCore, Zygote
foo(a, b) = a * b
function ChainRulesCore.rrule(::typeof(foo), a, b)
    y = foo(a, b)
    function foo_pullback(Ȳ)
        ∂a = @thunk (@info "Thunk ∂a"; return Ȳ * b')
        ∂b = @thunk (@info "Thunk ∂b"; return a' * Ȳ)
        return (NO_FIELDS, ∂a, ∂b)
    end
    return y, foo_pullback
end
a = rand(4,3); b = rand(3,2); Ȳ = rand(4,2);
Zygote.pullback(foo, a, b)[2](Ȳ)
let a = a; Zygote.pullback(b -> foo(a, b), b)[2](Ȳ) end
```

we obviously get

```
julia> Zygote.pullback(foo, a, b)[2](Ȳ)
[ Info: Thunk ∂a
[ Info: Thunk ∂b
([...], [...])
```

but we currently also get

```
julia> let a = a; Zygote.pullback(b -> foo(a, b), b)[2](Ȳ) end
[ Info: Thunk ∂a
[ Info: Thunk ∂b
([...],)
```

so Zygote also executes both thunks if we only require the pullback in respect to `b`.

With this PR, we should get

```
julia> let a = a; Zygote.pullback(b -> foo(a, b), b)[2](Ȳ) end
[ Info: Thunk ∂b
([0.5793294513282393 0.5442572290286626; 0.748469566554706 0.8319235475298901; 0.609222600261931 0.6201995902507516],)
```

instead.

**Note: Requires FluxML/ZygoteRules.jl#17**

CC @oxinabox , @mzgubic

### DIFF
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -271,7 +271,9 @@ function ChainRulesCore.rrule_via_ad(config::ZygoteRuleConfig, f_args...; kwargs
         _pullback(config.context, f_args...)
     end
 
-    ad_pullback(Δ) = zygote2differential(pb(wrap_chainrules_output(Δ)), f_args)
+    ad_pullback(Δ) = zygote2differential(
+        pb(wrap_chainrules_output(unthunk_tangent(Δ))),
+        f_args)
     return y, ad_pullback
 end
 
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -1,5 +1,5 @@
 @inline unthunk_tangent(x::AbstractThunk) = wrap_chainrules_output(unthunk(x))
-@inline unthunk_tangent(x::AbstractArray{<:AbstractThunk}) = wrap_chainrules_output(map(unthunk, x))
+@inline unthunk_tangent(x::AbstractArray{<:AbstractThunk}) = map(unthunk_tangent, x)
 
 struct ZygoteRuleConfig{CTX<:AContext} <: RuleConfig{Union{HasReverseMode,NoForwardsMode}}
   context::CTX
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -1,6 +1,10 @@
 @inline unthunk_tangent(x::AbstractThunk) = wrap_chainrules_output(unthunk(x))
 @inline unthunk_tangent(x::AbstractArray{<:AbstractThunk}) = map(unthunk_tangent, x)
 unthunk_tangent(d::IdDict) = IdDict([unthunk_tangent(k) => unthunk_tangent(v) for (k, v) in d])
+function ChainRulesCore.rrule(::typeof(unthunk_tangent), d::IdDict)
+  unthunk_iddict_pullback(_) = (NoTangent(), ChainRulesCore.@not_implemented "unthunking IdDict")
+  return d, unthunk_iddict_pullback
+end
 @non_differentiable unthunk_tangent(::IdDict)
 
 
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -1,5 +1,7 @@
 @inline unthunk_tangent(x::AbstractThunk) = wrap_chainrules_output(unthunk(x))
 @inline unthunk_tangent(x::AbstractArray{<:AbstractThunk}) = map(unthunk_tangent, x)
+unthunk_tangent(d::IdDict) = IdDict([unthunk_tangent(k) => unthunk_tangent(v) for (k, v) in d])
+
 
 struct ZygoteRuleConfig{CTX<:AContext} <: RuleConfig{Union{HasReverseMode,NoForwardsMode}}
   context::CTX
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -1,6 +1,7 @@
 @inline unthunk_tangent(x::AbstractThunk) = wrap_chainrules_output(unthunk(x))
 @inline unthunk_tangent(x::AbstractArray{<:AbstractThunk}) = map(unthunk_tangent, x)
 unthunk_tangent(d::IdDict) = IdDict([unthunk_tangent(k) => unthunk_tangent(v) for (k, v) in d])
+@non_differentiable unthunk_tangent(::IdDict)
 
 
 struct ZygoteRuleConfig{CTX<:AContext} <: RuleConfig{Union{HasReverseMode,NoForwardsMode}}
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -1,5 +1,5 @@
-@inline unthunk_tangent(x::AbstractThunk) = unthunk(x)
-@inline unthunk_tangent(x::AbstractArray{<:AbstractThunk}) = map(unthunk, x)
+@inline unthunk_tangent(x::AbstractThunk) = wrap_chainrules_output(unthunk(x))
+@inline unthunk_tangent(x::AbstractArray{<:AbstractThunk}) = wrap_chainrules_output(map(unthunk, x))
 
 struct ZygoteRuleConfig{CTX<:AContext} <: RuleConfig{Union{HasReverseMode,NoForwardsMode}}
   context::CTX
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -1,5 +1,10 @@
+# ToDo: Move some of this to ZygoteRules, or move unthunk_tangent for Tuple and NamedTuple from
+# Zygote rules here?
+function unthunk_tangent end
 @inline unthunk_tangent(x::AbstractThunk) = wrap_chainrules_output(unthunk(x))
-@inline unthunk_tangent(x::AbstractArray{<:AbstractThunk}) = map(unthunk_tangent, x)
+@inline unthunk_tangent(x::NTuple{N,<:Number}) where N = x
+@inline unthunk_tangent(x::AbstractArray{<:Number,N}) where N = x
+@inline unthunk_tangent(x::AbstractArray) = map(unthunk_tangent, x)
 unthunk_tangent(d::IdDict) = IdDict([unthunk_tangent(k) => unthunk_tangent(v) for (k, v) in d])
 @non_differentiable unthunk_tangent(::IdDict)
 
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -1,10 +1,6 @@
 @inline unthunk_tangent(x::AbstractThunk) = wrap_chainrules_output(unthunk(x))
 @inline unthunk_tangent(x::AbstractArray{<:AbstractThunk}) = map(unthunk_tangent, x)
 unthunk_tangent(d::IdDict) = IdDict([unthunk_tangent(k) => unthunk_tangent(v) for (k, v) in d])
-function ChainRulesCore.rrule(::typeof(unthunk_tangent), d::IdDict)
-  unthunk_iddict_pullback(_) = (NoTangent(), ChainRulesCore.@not_implemented "unthunking IdDict")
-  return d, unthunk_iddict_pullback
-end
 @non_differentiable unthunk_tangent(::IdDict)
 
 
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -1,3 +1,6 @@
+@inline unthunk_tangent(x::AbstractThunk) = unthunk(x)
+@inline unthunk_tangent(x::AbstractArray{<:AbstractThunk}) = map(unthunk, x)
+
 struct ZygoteRuleConfig{CTX<:AContext} <: RuleConfig{Union{HasReverseMode,NoForwardsMode}}
   context::CTX
 end
@@ -107,7 +110,6 @@ is_kwfunc(k, ::Type{<:NamedTuple}, f, args...) = k===Core.kwftype(f)
 Convert `x` from the differentials types ChainRules uses to the format Zygote uses internally.
 """
 @inline wrap_chainrules_output(x) = x
-@inline wrap_chainrules_output(x::AbstractThunk) = wrap_chainrules_output(unthunk(x))  # For now we are just not going to deal with thunks
 @inline wrap_chainrules_output(x::Tuple) = map(wrap_chainrules_output, x)
 # Zygote convention: even if many AbstractZero partials (i.e. multi-input function), make just 1 nothing.
 @inline wrap_chainrules_output(x::Tuple{Vararg{ChainRules.AbstractZero}}) = nothing
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -474,7 +474,7 @@ function pullback(f, ps::Params)
       cache(cx)[p] = nothing
     end
     back(Δ)
-    Grads(IdDict([k => unthunk_tangent(v) for (k, v) in cx.cache]), ps) # TODO make a copy
+    Grads(unthunk_tangent(cx.cache), ps) # TODO make a copy
   end
 end
 
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -357,6 +357,9 @@ function copy!(x::AbstractVector, ps::Params)
   x
 end
 
+_maybe_unthunk(x::AbstractThunk) = unthunk(x)
+_maybe_unthunk(x) = x
+
 """
     Grads(...)
 
@@ -391,7 +394,7 @@ end
 
 function Base.getindex(gs::Grads, x)
   isbits(x) && error("Only reference types can be differentiated with `Params`.")
-  return gs.grads[x]
+  return _maybe_unthunk(gs.grads[x])
 end
 
 """
@@ -474,7 +477,7 @@ function pullback(f, ps::Params)
       cache(cx)[p] = nothing
     end
     back(Δ)
-    Grads(unthunk_tangent(cx.cache), ps) # TODO make a copy
+    Grads(_maybe_unthunk(cx.cache), ps)
   end
 end
 
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -37,7 +37,13 @@ end
 _pullback(f, args...) = _pullback(Context(), f, args...)
 
 tailmemaybe(::Nothing) = nothing
-tailmemaybe(x::Tuple) = Base.tail(x)
+tailmemaybe(x::Tuple) = unthunk_tangent(Base.tail(x))
+
+# unthunking is essentially an identity operation on a lazy value, but
+# `@adjoint unthunk_tangent(x) = unthunk_tangent(x), ȳ -> (ȳ,)` is not enough to make
+# nested AD work, so define
+@adjoint tailmemaybe(xs::Tuple) = tailmemaybe(xs), x̄s -> ((nothing, x̄s...),)
+
 
 """
     pullback(f, args...)
@@ -468,7 +474,7 @@ function pullback(f, ps::Params)
       cache(cx)[p] = nothing
     end
     back(Δ)
-    Grads(cx.cache, ps) # TODO make a copy
+    Grads(IdDict([k => unthunk_tangent(v) for (k, v) in cx.cache]), ps) # TODO make a copy
   end
 end
 
--- a/src/compiler/reverse.jl
+++ b/src/compiler/reverse.jl
@@ -3,6 +3,18 @@ using IRTools: IR, Variable, Pipe, xcall, var, prewalk, postwalk,
   insertafter!, finish, expand!, prune!, substitute!, substitute,
   block, block!, branch!, return!, stmt, meta
 
+
+# TODO: Temporary, to be removed when ChainRulesCore rrules are required to
+# support thunks as an input and all instances of _adjoint_keepthunks in
+# Zygote have been replaces by rrules:
+macro _adjoint_keepthunks(ex)
+  ZygoteRules.gradm(ex, false, true)
+end
+macro _adjoint_keepthunks!(ex)
+  ZygoteRules.gradm(ex, true, true)
+end
+
+
 @inline tuple_va(N, xs) = xs
 @inline tuple_va(N, x, xs...) = (x, tuple_va(N, xs...)...)
 @inline tuple_va(::Val{N}, ::Nothing) where N = ntuple(_ -> nothing, Val(N))
--- a/Project.toml
+++ b/Project.toml
@@ -38,8 +38,8 @@ ZygoteTrackerExt = "Tracker"
 
 [compat]
 AbstractFFTs = "1.3.1"
-ChainRules = "1.44.1"
-ChainRulesCore = "1.9"
+ChainRules = "1.72.2"
+ChainRulesCore = "1.25"
 ChainRulesTestUtils = "1"
 Colors = "0.12, 0.13"
 DiffRules = "1.4"
--- a/Project.toml
+++ b/Project.toml
@@ -74,7 +74,7 @@ Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 
 [targets]
-test = ["ChainRulesTestUtils", "Conda", "CUDA", "Distances", "FFTW", "FiniteDifferences", "PythonCall", "Test", "Pkg"]
+test = ["ChainRulesTestUtils", "Conda", "CUDA", "Distances", "FFTW", "FiniteDifferences", "PythonCall", "Test"]
 
 [sources]
 ChainRulesCore = {url = "https://github.com/pxl-th/ChainRulesCore.jl.git", rev = "pxl-th/thunks"}
--- a/Project.toml
+++ b/Project.toml
@@ -39,7 +39,7 @@ ZygoteTrackerExt = "Tracker"
 [compat]
 AbstractFFTs = "1.3.1"
 ChainRules = "1.72.2"
-ChainRulesCore = "1.25"
+ChainRulesCore = "1.25.1"
 ChainRulesTestUtils = "1"
 Colors = "0.12, 0.13"
 DiffRules = "1.4"
@@ -71,10 +71,6 @@ FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
 PythonCall = "6099a3de-0909-46bc-b1f4-468b9a2dfc0d"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
-Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 
 [targets]
 test = ["ChainRulesTestUtils", "Conda", "CUDA", "Distances", "FFTW", "FiniteDifferences", "PythonCall", "Test"]
-
-[sources]
-ChainRulesCore = {url = "https://github.com/pxl-th/ChainRulesCore.jl.git", rev = "pxl-th/thunks"}
--- a/Project.toml
+++ b/Project.toml
@@ -75,3 +75,6 @@ Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 
 [targets]
 test = ["ChainRulesTestUtils", "Conda", "CUDA", "Distances", "FFTW", "FiniteDifferences", "PythonCall", "Test", "Pkg"]
+
+[sources]
+ChainRulesCore = {url = "https://github.com/pxl-th/ChainRulesCore.jl.git", rev = "pxl-th/thunks"}
--- a/Project.toml
+++ b/Project.toml
@@ -71,6 +71,7 @@ FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
 PythonCall = "6099a3de-0909-46bc-b1f4-468b9a2dfc0d"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
+Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 
 [targets]
-test = ["ChainRulesTestUtils", "Conda", "CUDA", "Distances", "FFTW", "FiniteDifferences", "PythonCall", "Test"]
+test = ["ChainRulesTestUtils", "Conda", "CUDA", "Distances", "FFTW", "FiniteDifferences", "PythonCall", "Test", "Pkg"]
--- a/src/Zygote.jl
+++ b/src/Zygote.jl
@@ -45,9 +45,6 @@ include("lib/utils.jl")
 include("lib/range.jl")
 include("lib/logexpfunctions.jl")
 
-_usethunks() = true
-ChainRulesCore.rrule(::typeof(_usethunks)) = false, () -> (NoTangent(),)
-
 # we need to define this late, so that the genfuncs see lib.jl
 # Move using statements out of this file to help with sysimage building
 using IRTools: varargs!, inlineable!, pis!, slots!
@@ -86,10 +83,6 @@ using PrecompileTools
 # see https://github.com/SciML/DiffEqFlux.jl/issues/783
 @static if VERSION < v"1.8" || VERSION >= v"1.8.5"
   @compile_workload precompile()
-end
-
-function __init__()
-    ChainRulesCore.UTILIZE_THUNKS[] = _usethunks
 end
 
 end # module
--- a/src/Zygote.jl
+++ b/src/Zygote.jl
@@ -45,6 +45,9 @@ include("lib/utils.jl")
 include("lib/range.jl")
 include("lib/logexpfunctions.jl")
 
+_usethunks() = true
+ChainRulesCore.rrule(::typeof(_usethunks)) = false, () -> (NoTangent(),)
+
 # we need to define this late, so that the genfuncs see lib.jl
 # Move using statements out of this file to help with sysimage building
 using IRTools: varargs!, inlineable!, pis!, slots!
@@ -83,6 +86,10 @@ using PrecompileTools
 # see https://github.com/SciML/DiffEqFlux.jl/issues/783
 @static if VERSION < v"1.8" || VERSION >= v"1.8.5"
   @compile_workload precompile()
+end
+
+function __init__()
+    ChainRulesCore.UTILIZE_THUNKS[] = _usethunks
 end
 
 end # module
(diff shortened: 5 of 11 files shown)


src/compiler/interface.jl
```diff
@@ -250,7 +250,7 @@ function pullback(f, ps::Params)
       cache(cx)[p] = nothing
     end
     back(Δ)
-    Grads(cx.cache, ps) # TODO make a copy
+    Grads(IdDict(k => _unthunk(v) for (k, v) in cx.cache), ps) # TODO make a copy
```
oxinabox, line 253: this is the bit that it is struggled to handle in nested AD I think.
Possibly try (just for now)
```suggestion
    Grads(IdDict([k => _unthunk(v) for (k, v) in cx.cache]), ps) # TODO make a copy
```

oschulz: Thanks, I'll do a quick check regarding type inference of the one vs. the other suggested version.

oschulz:
I think it may actually be the unthunk in `tailmemaybe(x::Tuple)` that results in that error, not the one in `Grads(IdDict(...))`.

oschulz:
It looks like for this to work, we'll need to overload a lot of functions like `x'`, `Adjoint`, `transpose`, `Transpose`, `diag`, `Diagonal`, `permutedims`, `permutedims!`, `Ref`, `Vector`, `Matrix`, `Array`, `collect`, `convert` to unthunk thunks, since they can now appear as `Ȳ` in pullback functions. This list is probably far from exhaustive ...

oschulz:
Ok, with FluxML/ZygoteRules.jl#17 this passes the Zygote test suit on my system now (incl. CUDA tests).

In both the simple example above and more complex cases that I've tested only the required thunks are unthunked. Hard to give guarantees on this, obviously.

The `@adjoint_keepthunks` in FluxML/ZygoteRules.jl#17 provides a way for adjoint authors to declare whether their pullback code is prepared to handle thunks or not.

Maybe we could add a similar mechanism to ChainRulesCore (an `rrule_keepthunks` or so) to provide a soft transition to (hopefully) more and more rules that support thunks in the ecosystem?

oxinabox:
> @oxinabox Could we move all the @adjoint_keepthunk methods to rrules? Since some of the existing adjoints touch the accum_param?

Looking at them, I think they either are things that should just move to ChainRules.jl
or they are Zygote internals that can never the less be written using `rrule`s

The only one I am unsure about is:
https://github.com/FluxML/Zygote.jl/pull/966/files#diff-cd0210083ce3136f79bee6ebca2bcca77f41a14f11b5a7a65ea1cc54803164c3R103-R110
which I don't understand what it is doing.


----


I think @DhairyaLGandhi  needs to make the call as to if to do the parts of the 3 step plan that involve changes to Zygote/Zygote rules.
I am gently in favour of them. 
On the one hand they seem to work, and that is really useful, and fixing things sooner rather than later is better: Better an egg in the hand than two in the bush.
On the other hand it is more complexity to maintain in Zygote; and I dislike adding features to ZygoteRules when we are trying to stop using it.

I am strongly infavor of making changes in ChainRules and ChainRulesCore to facilitate this. Either as part of 3 step plan or as part of going straight to keeping thunks is done via `rrule` and `@adjoint` always removes them.


----

> Maybe we could offer a function unthunking in ChainRulesCore so that rrules can return y, unthunking(back) for rrules that need to unthunk.

I think better to leave that to the AD package.
It is a simple enough function, I would rather a little code duplication than expand the API surface in a way that might be wrong/made redundant. (I was reading https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction recently, I think it makes some good points)

oschulz:
> dislike adding features to ZygoteRules when we are trying to stop using it

We wouldn't really add a feature to ZygoteRules though, right? We'd just make `@adjoint` unthunk in general. And `Zygote.@_adjoint_keepthunks` would be a purely internal and temporary thing in Zygote.

oschulz:
> >Maybe we could offer a function unthunking in ChainRulesCore so that rrules can return y, unthunking(back) for rrules that need to unthunk.
> I think better to leave that to the AD package.

Ah, no, I meant to make life easier on rule writers (who would only depend on `ChainRulesCore` not an AD package). I assume there will be quite a few rules that need to always unthunk, and this could be a tool to keep their code more concise.

oschulz:
> I am, for the same reasons, happy for now to just leave it for the rule authors. It is still a very short function.

oschulz:
> but why is it that we need a separate macro to define thunk aware adjoints? Generally, I see this more as an implementation detail which shouldn't leak to the API

Yes, hence my changed proposal to do this as a Zygote-internal `@_adjoint_keepthunks`.

We do need two macros (one of them non-public/API-stable) for a soft transition, since we need a lot of non-unthunking Zygote-internal adjoints right now to have any benefit from thunks, but the ecosystem will break if existing adjoints suddenly get thunks as an input. That's why I'd like to do this in steps.

oschulz:
Ok I've removed `@adjoint_keepthunks` from ZygoteRules and added an internal `@_adjoint_keepthunks` to Zygote itself instead. I've also added some comments to label it as temporary.

`Zygote.@_adjoint_keepthunks` still uses the extended `ZygoteRules.gradm(ex, mut, keepthunks)` introduced by FluxML/ZygoteRules.jl#17. `gradm` is a fairly complicated bit of code and the difference between thunking and unthunking is minimal, so I didn't want to basically duplicate it in Zygote. But it's not an exported function, so we can remove the `keepthunks` argument from `gradm` later as a non-breaking change. FluxML/ZygoteRules.jl#17 is necessary in any case, since we do need to make `@adjoint` unthunk for any of this to work.

src/compiler/reverse.jl
```diff
@@ -3,6 +3,18 @@ using IRTools: IR, Variable, Pipe, xcall, var, prewalk, postwalk,
   insertafter!, finish, expand!, prune!, substitute!, substitute,
   block, block!, branch!, return!, stmt, meta
 
+
+# TODO: Temporary, to be removed when ChainRulesCore rrules are required to
```
mzgubic, line 7: would be good to link to the issue/PR here

oxinabox:
```suggestion
# TODO: Temporary, to be removed when https://github.com/JuliaDiff/ChainRules.jl/issues/408 is done
# and ChainRulesCore rrules are required to
```

oschulz:
Here's a little demo on how this will work one we can disable the unthunking of `rrule` input (step 3 in the proposal above).

With

```julia
# Requires
# * https://github.com/FluxML/ZygoteRules.jl/pull/17
# * https://github.com/FluxML/Zygote.jl/pull/966

using Zygote, LinearAlgebra, ChainRulesCore, ChainRulesCore
using ChainRules: CommutativeMulNumber


# Make thunks in log more readable:
Base.show(io::IO, x::Thunk) = print(io, typeof(x).name.name)
Base.show(io::IO, x::InplaceableThunk) = print(io, typeof(x).name.name)


# Disable unthunking of rrule input (to become default behavior in the future):
function Zygote.wrap_chainrules_input(x)
    @info "wrap_chainrules_input($(x))"
    return x
end


# Same rrule as in ChainRules, just add some logging:
function ChainRulesCore.rrule(
    ::typeof(*),
    A::AbstractVecOrMat{<:CommutativeMulNumber},
    B::AbstractVecOrMat{<:CommutativeMulNumber},
)
    @info "rrule for $(A) * $(B)"
    function times_pullback(Ȳ)
        @info "pullback of $(A) * $(B) for $(Ȳ)"
        return (
            NO_FIELDS,
            InplaceableThunk(
                @thunk((@info("!left thunk.val of pullback of $(A) * $(B) for $(Ȳ) = $(Ȳ) * $(B)"); Ȳ * B')),
                X̄ -> (@info("!left thunk.add! of pullback of $(A) * $(B) for $(Ȳ) = $(Ȳ) * $(B)");mul!(X̄, Ȳ, B', true, true))
            ),
            InplaceableThunk(
                @thunk((@info("!right thunk.val of pullback of $(A) * $(B) for $(Ȳ) = $(A') * $(Ȳ)"); A' * Ȳ)),
                X̄ -> (@info("!right thunk.add! of pullback of $(A) * $(B) for $(Ȳ) = $(A') * $(Ȳ)");mul!(X̄, A', Ȳ, true, true))
            )
        )
    end
    return A * B, times_pullback
end


A, B, C, D = [fill(i,1,1) for i in 2:5]
```

We get:

```
julia> Zygote.gradient(X -> sum(A * B * C * X), D)
[ Info: rrule for [2] * [3]
[ Info: rrule for [6] * [4]
[ Info: rrule for [24] * [5]
[ Info: wrap_chainrules_input(1×1 Fill{Int64}: entries equal to 1)
[ Info: pullback of [24] * [5] for 1×1 Fill{Int64}: entries equal to 1
[ Info: wrap_chainrules_input(InplaceableThunk)
[ Info: pullback of [6] * [4] for InplaceableThunk
[ Info: wrap_chainrules_input(InplaceableThunk)
[ Info: pullback of [2] * [3] for InplaceableThunk
[ Info: !right thunk.val of pullback of [24] * [5] for 1×1 Fill{Int64}: entries equal to 1 = [24] * 1×1 Fill{Int64}: entries equal to 1
([24],)

julia> Zygote.gradient(X -> sum(X * B * C * D), A)
[ Info: rrule for [2] * [3]
[ Info: rrule for [6] * [4]
[ Info: rrule for [24] * [5]
[ Info: wrap_chainrules_input(1×1 Fill{Int64}: entries equal to 1)
[ Info: pullback of [24] * [5] for 1×1 Fill{Int64}: entries equal to 1
[ Info: wrap_chainrules_input(InplaceableThunk)
[ Info: pullback of [6] * [4] for InplaceableThunk
[ Info: wrap_chainrules_input(InplaceableThunk)
[ Info: pullback of [2] * [3] for InplaceableThunk
[ Info: !left thunk.val of pullback of [2] * [3] for InplaceableThunk = InplaceableThunk * [3]
[ Info: !left thunk.val of pullback of [6] * [4] for InplaceableThunk = InplaceableThunk * [4]
[ Info: !left thunk.val of pullback of [24] * [5] for 1×1 Fill{Int64}: entries equal to 1 = 1×1 Fill{Int64}: entries equal to 1 * [5]
([60],)
```

This should save a lot of computation time in applications that don't diff in respect to every argument along the computational graph. Standard ML applications usually want a gradients for almost everything, of course (since almost everything is a free parameter), but applications like fitting, MCMC, "scientific" ML and the like tend to be more selective in which gradient(s) they need.


If we re-enable unthunking of rrule input (as is part of this PR, to keep compatibility until "step 3")

```julia
function Zygote.wrap_chainrules_input(x)
    @info "wrap_chainrules_input($(x))"
    return x
end
```

things aren't quite that nice if we diff in respect to `D`

```
julia> Zygote.gradient(X -> sum(A * B * C * X), D)
[ Info: rrule for [2] * [3]
[ Info: rrule for [6] * [4]
[ Info: rrule for [24] * [5]
[ Info: wrap_chainrules_input(1×1 Fill{Int64}: entries equal to 1)
[ Info: pullback of [24] * [5] for 1×1 Fill{Int64}: entries equal to 1
[ Info: wrap_chainrules_input(InplaceableThunk)
[ Info: !left thunk.val of pullback of [24] * [5] for 1×1 Fill{Int64}: entries equal to 1 = 1×1 Fill{Int64}: entries equal to 1 * [5]
[ Info: pullback of [6] * [4] for [5]
[ Info: wrap_chainrules_input(InplaceableThunk)
[ Info: !left thunk.val of pullback of [6] * [4] for [5] = [5] * [4]
[ Info: pullback of [2] * [3] for [20]
[ Info: !right thunk.val of pullback of [24] * [5] for 1×1 Fill{Int64}: entries equal to 1 = [24] * 1×1 Fill{Int64}: entries equal to 1
([24],)
```

but we still be some immediate benefit from this PR, since with the current release of Zygote, things look like this:

```
julia> Zygote.gradient(X -> sum(A * B * C * X), D)
[ Info: rrule for [2] * [3]
[ Info: rrule for [6] * [4]
[ Info: rrule for [24] * [5]
[ Info: pullback of [24] * [5] for 1×1 Fill{Int64}: entries equal to 1
[ Info: !left thunk.val of pullback of [24] * [5] for 1×1 Fill{Int64}: entries equal to 1 = 1×1 Fill{Int64}: entries equal to 1 * [5]
[ Info: !right thunk.val of pullback of [24] * [5] for 1×1 Fill{Int64}: entries equal to 1 = [24] * 1×1 Fill{Int64}: entries equal to 1
[ Info: pullback of [6] * [4] for [5]
[ Info: !left thunk.val of pullback of [6] * [4] for [5] = [5] * [4]
[ Info: !right thunk.val of pullback of [6] * [4] for [5] = [6] * [5]
[ Info: pullback of [2] * [3] for [20]
[ Info: !left thunk.val of pullback of [2] * [3] for [20] = [20] * [3]
[ Info: !right thunk.val of pullback of [2] * [3] for [20] = [2] * [20]
([24],)
```

oschulz:
Here are some benchmarks for the example above, with bigger arrays:

```julia
using Zygote, BenchmarkTools

A, B, C, D = [fill(i,100,100) for i in 2:5]

@btime Zygote.gradient(X -> sum($A * $B * $C * X), $D)

```

Using current Zygote v0.6.10:

```
julia> @btime Zygote.gradient(X -> sum($A * $B * $C * X), $D)
  3.066 ms (222 allocations: 792.28 KiB)

```

With this PR (requires FluxML/ZygoteRules.jl#17):

```
julia> @btime Zygote.gradient(X -> sum($A * $B * $C * X), $D)
  2.192 ms (245 allocations: 562.34 KiB)
```

And after disabling forced unthunking of rrules input (step 3 in the plan above):

```
julia> @inline Zygote.wrap_chainrules_input(x) = x

julia> @btime Zygote.gradient(X -> sum($A * $B * $C * X), $D)
  1.413 ms (229 allocations: 416.22 KiB)
```

oschulz:
Hey @DhairyaLGandhi , it seems all that's missing before we can move with this is your blessing. Would you have time to take a look?

DhairyaLGandhi:
I'll take a look this week, thanks for this!

oschulz:
@DhairyaLGandhi , gentle bump - I don't mean to nag, but I would love to start using the first stage of this for some projects.

DhairyaLGandhi:
On it :)

Not nagging at all, I was trying to explore what it would entail for writing rules in zygote and what it means for the ergonomics anyway

oschulz:
Thanks!

There should be no difference for users writing `ZygoteRules` rules. Starting with step 2 in the plan above, user writing `ChainRulesCore` rules (and I guess the idea is that this will be the preferred way in the future, @oxinabox ? ) will need to start using `unthunk` for the tangents their pullbacks receive, if the pullback uses functionality that isn't itself thunk-compatible. That should really be all, regarding ergonomics.

oxinabox:
@mzgubic is working as we speak for making sure all rules in ChainRules.jl support recieving thunks as inputs.
Mostly by adding missing overloads for common linear operators to ChainRulesCore, but probably in some cases via calling `unthunk` at the start.
With the intent that when we release ChainRulesCore 1.0 and ChainRulesTestUtils 1.0,
it will be a requirement of all rules.

DhairyaLGandhi:
Will the requirement be to ubthunk before anythibg in an rrule?

mzgubic:
Not necessarily, only in cases where the thunk would have been unthunked more than once. See https://github.com/JuliaDiff/ChainRules.jl/pull/449 (which is nearly there)

oschulz:
> > Will the requirement be to ubthunk before anythibg in an rrule?
> Not necessarily,

Also, quite a few rrules will not need to `unthunk` at all, as long as they're only calling on functions that have methods for thunks. From what I understand, there's an ongoing effort right now to significantly increase thunk-support in linear algebra, etc. So in some (maybe quite a few) cases we'll be able to pass thunks through. Correct, @oxinabox ?

mzgubic:
Yes, that's right, some thunks will pass through.

We will need to be careful that we do unthunk in cases where the thunk is used twice (to avoid the duplicate calculation). If the thunk is only used once, it will be passed through.

oschulz:
I wonder if Julia shouldn't have something like a thunk as a native concept. We have something along those lines in so many places/packages.

oschulz:
> unthunk in cases where the thunk is used twice (to avoid the duplicate calculation)

Could we use some result caching / lazy value mechanism?

oschulz:
> Could we use some result caching / lazy value mechanism?

Might lead to even more alloc than we already have, though ...

oxinabox:
Yes, thunks are an emulation of a compiler feature julia doesn't have.
Haskell (and TensorFlow) have lazy evaluation where only things that are used are computed.
Thunks are like 1 level of lazy evaluation.

And there is a more general concept of [program slicing](https://en.wikipedia.org/wiki/Program_slicing)
program slicing would be particularly useful for AD.

oschulz:
> Yes, thunks are an emulation of a compiler feature julia doesn't have.

Hm, this got the thinking - thunks are really monads, right? Especially now that we're adding methods for them to lots of basic function. Shouldn't `AbstractThunk` have a type parameter, then? Kinda like

```
abstract type AbstractThunk{T} end

struct Thunk{T,F<:Base.Callable} <: AbstractThunk{T}
    body::F
end

unthunk(t::Thunk{T}) where T = t.body()::T

Base.eltype(t::Thunk{T}) where T = T

macro thunk(body)
    func = Expr(:->, Expr(:tuple), Expr(:block, __source__, body))
    return quote
        f = $(esc(func))
        Thunk{Base._return_type(f,()), typeof(f)}(f)
    end
end


thnk = let A = rand(5,5), B = rand(5,5)
    @thunk A * B
end

eltype(thnk) == typeof(unthunk(thnk))
```

If seen type inference fail with thunks quite a few times - maybe having typed thunks would help with that, too?

oxinabox:
> > Could we use some result caching / lazy value mechanism?
> 
> Might lead to even more mem allocs than we already have, though ...

A very very ancient version of ChainRules had this.
It was removed long before I was involved.
Issue for adding it back is still open https://github.com/JuliaDiff/ChainRulesCore.jl/issues/7#issuecomment-491609475
I do worry about the same thing re mem allocs, here

oschulz:
> Issue for adding it back is still open JuliaDiff/ChainRulesCore.jl#7 (comment)
> I do worry about the same thing re mem allocs, here

Thanks, I added a suggestion there.

oschulz:
@DhairyaLGandhi do we get your blessing on the "plan"?

oxinabox:
Given having full support for thunks in all ChainRules functions is well and truly on its way, at this point the plan might be moot.
And we can soon basically just merge something like this PR.

oschulz:
> at this point the plan might be moot. And we can soon basically just merge something like this PR.

More than fine with me - we'll still need to merge FluxML/ZygoteRules.jl#17 first, though - or duplicate that functionality within Zygote (would duplicate most of `ZygoteRules.gradm`).

oschulz:
FluxML/ZygoteRules.jl#17 has been merged, I'll get this PR into shape as soon as there's new release of ZygoteRules, so CI can run.

oschulz:
Ok, new release of ZygoteRules is out (had overlooked it), I'll jump back on this.

oschulz:
Rebasing this now, but several things have changed in Zygote in the mean time ... hope it still works.

oschulz:
CI results don't look too horrible - at least some of the test failures look similar to the ones in #1104, so may be unrelated to this PR. Could an expert take a look?

oschulz:
I think this is almost there, if we can only figure the nested Params test out.

src/compiler/chainrules.jl
```diff
@@ -1,3 +1,8 @@
+@inline unthunk_tangent(x::AbstractThunk) = wrap_chainrules_output(unthunk(x))
+@inline unthunk_tangent(x::AbstractArray{<:AbstractThunk}) = map(unthunk_tangent, x)
+unthunk_tangent(d::IdDict) = IdDict([unthunk_tangent(k) => unthunk_tangent(v) for (k, v) in d])
```
ToucheSir, line 3:
```suggestion
unthunk_tangent(d::IdDict) = IdDict([unthunk_tangent(k) => unthunk_tangent(v) for (k, v) in d])
@non_differentiable unthunk_tangent(::IdDict)
```
This would probably fix the "map is not defined on dictionaries" error, but unsure if it's sufficient to pass the nested params test.

ToucheSir (commented):
@oschulz do you mind rebasing this? I just tried some local testing and there appear to be new test failures.

src/compiler/chainrules.jl
ToucheSir: After some more testing, I suspect that Zygote is relying on the identity of the inner `IdDict` for parameter tracking. Breaking that identity with `unthunk_tangent(d::IdDict)` seems to cause propagation of null gradients at unexpected places.

oschulz:
> @oschulz do you mind rebasing this? I just tried some local testing and there appear to be new test failures.

Will do! Really need to bring this one home ...

oschulz:
@ToucheSir rebase done.

src/compiler/chainrules.jl
ToucheSir:
```suggestion
function ChainRulesCore.rrule(::typeof(unthunk_tangent), d::IdDict)
  unthunk_iddict_pullback(_) = (NoTangent(), ChainRulesCore.@not_implemented "unthunking IdDict")
  return d, unthunk_iddict_pullback
end
```
This seems to resolve the follow-up test failure and not cause any new ones, but can not 100% vouch for its correctness.

ToucheSir:
Keeping track of test failure causes as I find them:

- https://github.com/FluxML/Zygote.jl/blob/4ed3a86db708a27bfe0afd5aeaa6408dd8d43a3e/src/lib/lib.jl#L324 is causing a test failure. I think the signature of `(back::Jnew)(...)` needs to be relaxed or another method added which handles thunks.

lassepe:
@ToucheSir that is probably not a useful test but I recall what happened there:

When you compute `y, back = pullback(f, Params[x])`, then evaluating `back` multiple times would  mutate the `Grad` object handed out earlier. Say I compute `g = back(1)` then `g1` is a `Grad` object which has `g1[w]` as the gradient of `f` w.r.t. `w`. If I now call `g2 = back(nothing)`, this mutates `g1` so that `g1[w] ==  nothing`.

I am not sure, what is the intended behavior actually. It does not seem to be documented at least and it was a really bad foot gun for me. I added this test to test the "default  case" to contrast it with the "fixed case" where I `copy` the grad object in between to avoid this unexpected coupling. So I guess I mostly added that as implicit documentation. I am sorry that this has caused confusion here. Maybe I should just open a PR to the docs instead.

ToucheSir:
Thanks for the quick response! I think in that case we can call it a footgun and consider that fixed (unintentionally) by this PR, no need for a docs update.

lassepe:
I guess this PR does not fix the problem in general, only for the `nothing` case, right? For example, if I call `g1 = back(1)` and then `g2 = back(2)` it seems like that would still mutate `g1` since both unthunk, right? I'm not raising this as an issue against this PR (I think the PR is great and I don't think it is its job to fix this behavior), I'm just trying to figure out whether this would need documentation perhaps.

oschulz:
Thanks for the changes! Let's see that CI say now ... :-)

ToucheSir:
> For example, if I call `g1 = back(1)` and then `g2 = back(2)` it seems like that would still mutate `g1` since both unthunk, right?

It appears that may be addressed as well:
```julia
using Zygote

x, y = ones(2), rand(2)
ps = Params([x, y])

l, back = pullback(() -> sum(x) + prod(y), ps)

julia> g1 = back(1.)
Grads(...)

julia> g1.grads
IdDict{Any, Any} with 4 entries:
  :(Main.x)            => Fill(1.0, 2)
  :(Main.y)            => [0.981219, 0.920662]
  [0.920662, 0.981219] => [0.981219, 0.920662]
  [1.0, 1.0]           => Fill(1.0, 2)

julia> g2 = back(2.)
Grads(...)

julia> g1.grads
IdDict{Any, Any} with 4 entries:
  :(Main.x)            => Fill(1.0, 2)
  :(Main.y)            => [0.981219, 0.920662]
  [0.920662, 0.981219] => [0.981219, 0.920662]
  [1.0, 1.0]           => Fill(1.0, 2)

julia> g2.grads
IdDict{Any, Any} with 4 entries:
  :(Main.x)            => Fill(3.0, 2)
  :(Main.y)            => [2.94366, 2.76199]
  [0.920662, 0.981219] => [1.96244, 1.84132]
  [1.0, 1.0]           => Fill(2.0, 2)
```

oschulz:
There's still a lot of new test failures (compared to pre-rebase). Of course Zygote has changed a bit - help tracking them down is welcome.

ToucheSir:
@oschulz that's what https://github.com/FluxML/Zygote.jl/pull/966#issuecomment-1065931188 is for :)

oschulz:
> @oschulz that's what https://github.com/FluxML/Zygote.jl/pull/966#issuecomment-1065931188 is for :)

Aw, sorry, overlooked that one - great, thanks!

mloubout:
Is there a plan/timeline on this PR? Not pushing it's just it's quite problematic for some of our (PDE based) codes where this is a huge blocker performance wise.

oxinabox:
I would still like this done.
AFAIK there are no remaining blockers -- and haven't been for about 12 months.

@DhairyaLGandhi @oschulz should we have a call and workout what is the way forward with this?

oschulz:
> I would still like this done.

Me too.

> AFAIK there are no remaining blockers

I don't think there are any blockers conceptually, just some test failures we hadn't managed to figure out.

> should we have a call and workout what is the way forward with this?

Gladly!

oxinabox:
@DhairyaLGandhi ?

DhairyaLGandhi:
Sure

oschulz:
> Sure, sounds good

Let's find a date via Slack?

DhairyaLGandhi:
Sure

ToucheSir:
As the one who raised all these pesky failure cases in https://github.com/FluxML/Zygote.jl/pull/966#issuecomment-1065931188, I'd be more than happy to help support this effort as well. Be that testing or trying to make sense of AD compiler output.

oschulz:
Hey, I haven't forgotten, I've just been swamped, I'll get to the rebase, promise.

oschulz:
Ok, I rebased this onto the current master branch ... let's see what CI says.

Sorry this took so horribly long, was a bit overloaded.

ToucheSir:
Thanks! Some notes on the test failures post-rebase:

**test**: test/features.jl:678
**line**: src/lib/array.jl#L264
- The thunk is not being detected in this type:
```julia
dy::Array{
	Tuple{
		Tuple{
			ChainRulesCore.Thunk{ChainRules.var"#1292#1326"{Int64, Int64, Int64, ChainRulesCore.ProjectTo{Float64, NamedTuple{(), Tuple{}}}, Int64}},
			Float64
		},
		Int64
	},
	4
}
```
- PRs only check for arrays of thunks. Arrays of tuples also need unthunking
- Maybe arrays of all non-numbers?

**test**: test/chainrules.jl#L307-L323
**line**: src/lib/lib.jl#L327
**msg** : Need an adjoint for constructor Foo. Gradient is of type Thunk{ChainRulesTestUtils.var"#61#65"{Tangent{Foo, NamedTuple{(:x, :y), Tuple{Float64, Float64}}}}}
- Should `Jnew` back unthunk, should it keep the thunk going, or should it not use @_adjoint_keepthunks! ?
- Likewise for `__splatnew__`

**test**: test/chainrules.jl:329
**line**: src/lib/lib.jl#L104
**msg** : Gradient Thunk(ChainRulesTestUtils.var"#...) should be a tuple
- `tuple(...)`'s pullback can't handle thunks
- Either ZygoteRules needs to handle the varargs somehow, or the pullback needs to unthunk

The failure at test/interface.jl:152 is actually a broken test working now, see https://github.com/FluxML/Zygote.jl/pull/966#issuecomment-1066170674.

oschulz:
Thanks for the analysis @ToucheSir!

> test: test/features.jl:678 ... The thunk is not being detected in this type

This one should hopefully be fixed with the push just now.

CarloLucibello:
What is the status of this? It would be a tremendous feature to have

ToucheSir:
Still failing tests, though only a couple I think. Unfortunately these PRs seem to take a good amount of work because of all the edge cases.

CarloLucibello:
Worth rebasing and see exactly where we are

pxl-th:
I've fixed some of the issues after https://github.com/FluxML/Zygote.jl/pull/966/commits/482ab1b4712978a09f5bab756664eb55ddca3087 in my local fork, but there's one I'm not sure where it'd be optimal to land a fix.

MWE for the problem (taken from one of the failing tests):
```julia
x = ones(Float32, 3, 2)
Zygote.gradient(x) do x
    sum(map(norm, eachcol(x)))
end
```
Returns:
```
(Float32[0.0 0.0; 0.0 0.0; 0.0 0.0],)
```

The reason this happens is because `ZBack` no longer `unthunk`s `Thunks` when being called (in `wrap_chainrules_output`), so [`rrule` for `eachcol`](https://github.com/JuliaDiff/ChainRules.jl/blob/e05500931b5424897d49ae184ed8e796e6e5357f/src/rulesets/Base/indexing.jl#L264) receives `dy` as:
```
Vector{ChainRulesCore.InplaceableThunk{ChainRulesCore.Thunk{ChainRules.var"#1071#1074"{Float32, SubArray{Float32, 1, Matrix{Float32}, Tuple{Base.Slice{Base.OneTo{Int64}}, Int64}, true}, Float32}}, ChainRules.var"#1070#1073"{Float32, SubArray{Float32, 1, Matrix{Float32}, Tuple{Base.Slice{Base.OneTo{Int64}}, Int64}, true}, Float32}}}
```

While previously due to `unthunk`ing in `ZBack` we'd get for `dy`:
```
Vector{Vector{Float32}}
```

And so the execution of `∇eachslice` woudn't terminate [here](https://github.com/JuliaDiff/ChainRules.jl/blob/e05500931b5424897d49ae184ed8e796e6e5357f/src/rulesets/Base/indexing.jl#L268).

I feel like this is due to **that** rrule not handling correctly unthinking and so the fix should go there.
We can simply unthunk each element of `dys` as we iterate it (if it is a thunk).
But maybe I'm missing something.

pxl-th:
I'm still looking at the remaining tests. But here're some initial results with lazy Zygote:

- [HiFiGAN.jl](https://github.com/pxl-th/HiFiGAN.jl):

Alternating rounds of generator vs 2 discriminators training (computing grads either w.r.t. generator or discriminators).

Batch size: `16`.
GPU memory utilization: from `22.36` GiB down to `18.72` GiB.
1 Epoch training time: from `16` minutes down to `14` minutes.

- [GaussianSplatting.jl](https://github.com/JuliaNeuralGraphics/GaussianSplatting.jl):

Using `Flux.Conv` layer only as a loss function with fixed weights (previously it'd compute grads w.r.t. conv weights).

1k training steps benchmark: from `93` seconds down to `46` seconds.

---

There are still places where it does unthinking when it shouldn't.
They can be fixed by turning:
`accum(x::AbstractThunk, y::AbstractThunk) = accum(unthunk_tangent(x), unthunk_tangent(y))`
into:
`accum(x::AbstractThunk, y::AbstractThunk) = @thunk(accum(unthunk_tangent(x), unthunk_tangent(y)))`

But then some tests start raising errors such as:
https://github.com/FluxML/Zygote.jl/blob/9b6dd0803441a874f6afdbdfa61e27c2b9b1b097/src/lib/lib.jl#L330

Where `T <: AbstractThunk`, but `Δ` is an `Array` for example, which I'm not sure how to deal with.

pxl-th:
With latest commit all of the newly introduced test failures are now fixed.

ToucheSir:
Do you mind rebasing one more time so we can see what CI looks like with https://github.com/FluxML/Zygote.jl/pull/1545?

pxl-th:
Hm... Still some failures. I'll take a look at them in a bit

mcabbott:
Current CI errors from `hessian_reverse` (Zygote over Zygote, 2nd order AD) look a bit like some I tried to avoid, once upon a time, by completely disabling the creation of thunks within AD (i.e. on the inner gradient call). 

My code for this was https://github.com/JuliaDiff/ChainRulesCore.jl/pull/568 , which could be revived, or perhaps re-written within Zygote alone.

pxl-th:
I did it via a hook at CRC level, but this creates some type instabilities.
Or is it fine to disable unconditionally at CRC level?

CarloLucibello:
The DiffEqFlux and DistributionsAD integration test failures are possibly related (they passed in last merged PR).

pxl-th:
All of the errors in DiffEqFlux seem to come from ComponentArrays.jl rrule not unthunking `Δ`:
https://github.com/SciML/ComponentArrays.jl/blob/882df4d7eadc8709cddeeb39deb6a740a230074d/src/compat/chainrulescore.jl#L2
Error:
https://github.com/FluxML/Zygote.jl/actions/runs/12596665169/job/35108214912?pr=966#step:6:2000

pxl-th:
MLUtils rrule for chunk needs to [unthunk each element](https://github.com/JuliaML/MLUtils.jl/blob/204b958b7c53af65cdfca5c56cdce2eb306cfc4a/src/utils.jl#L240) of `dys` similar to how we did it for `∇eachslice`.
Error:
https://github.com/FluxML/Zygote.jl/actions/runs/12596665169/job/35108214079?pr=966#step:6:912

pxl-th:
For DistributionsAD.jl, not sure yet, maybe rules needs to do unthunking?
https://github.com/TuringLang/DistributionsAD.jl/blob/242bd15fcada4f8d8f7f878fe9635240450d6fc1/src/zygote.jl#L18

pxl-th:
Fixed DistributionsAD (removed extra `unthunk_tangent`).

CarloLucibello (approved):

mcabbott:
The plan is to tag this as 0.7?

CarloLucibello:
Thanks to everyone, it has been a long journey!

CarloLucibello:
> The plan is to tag this as 0.7?

I'm not sure it is strictly necessary, but @pxl-th suggested this. I also think it is better to consider this breaking.

paulxshen:
Sadly this breaks my code :( I'm doing adjoint optimization of an electromagnetic time domain simulation (time stepped for loop). Worked in 0.6 but 0.7 throws stack overflow error of basically adjoint arg list repeated x # of for loop iters. Presumably tangents at each time step that were previously unthunked are now thunked together into a gigantic tuple. We don't have any custom adjoint rules.

ToucheSir:
Can you open an issue with a MWE? Otherwise it's not really possible to troubleshoot.

PR #972, opened 2021-05-14
fix gradient algebra on gpu

### LINKED ISSUE #971: Grads algebra on GPU causes `scalar getindex is disallowed` error
Hi! I was trying to do some gradient accumulation, but `Grads` algebra seems to cause error on GPU.
Am I missing something or is there indeen an issue?

I was using it in the context of UNet model training, but below is MWE.:

```julia
using CUDA
CUDA.allowscalar(false)
using Zygote

w = randn(Float32, 2) |> cu
x1 = randn(Float32, 2) |> cu
x2 = randn(Float32, 2) |> cu
p = Params([w])

g1 = gradient(() -> sum(w .* x1), p)
g2 = gradient(() -> sum(w .* x2), p)

g1 .+= g2
```

Error:

```bash
ERROR: scalar getindex is disallowed
Stacktrace:
  [1] error(s::String)
    @ Base .\error.jl:33
  [2] assertscalar(op::String)
    @ GPUArrays C:\Users\tonys\.julia\packages\GPUArrays\4n0iS\src\host\indexing.jl:62
  [3] getindex
    @ C:\Users\tonys\.julia\packages\GPUArrays\4n0iS\src\host\indexing.jl:104 [inlined]
  [4] hash(A::CuArray{Float32, 1}, h::UInt64)
    @ Base .\abstractarray.jl:2432
  [5] hash
    @ .\hashing.jl:18 [inlined]
  [6] hashindex
    @ .\dict.jl:169 [inlined]
  [7] ht_keyindex2!(h::Dict{Any, Nothing}, key::CuArray{Float32, 1})
    @ Base .\dict.jl:310
  [8] setindex!(h::Dict{Any, Nothing}, v0::Nothing, key::CuArray{Float32, 1})
    @ Base .\dict.jl:383
  [9] push!(s::Set{Any}, x::CuArray{Float32, 1})
    @ Base .\set.jl:57
 [10] union!(s::Set{Any}, itr::Params)
    @ Base .\abstractset.jl:91
 [11] Set
    @ .\set.jl:10 [inlined]
 [12] _Set
    @ .\set.jl:23 [inlined]
 [13] Set
    @ .\set.jl:21 [inlined]
 [14] issetequal(l::Params, r::Params)
    @ Base .\abstractset.jl:375
 [15] #65
    @ .\none:0 [inlined]
 [16] iterate
    @ .\generator.jl:47 [inlined]
 [17] _all
    @ .\reduce.jl:922 [inlined]
 [18] all
    @ .\reduce.jl:918 [inlined]
 [19] all
    @ .\reduce.jl:836 [inlined]
 [20] map!(::Function, ::Zygote.Grads, ::Zygote.Grads, ::Zygote.Grads)
    @ Zygote C:\Users\tonys\.julia\dev\Zygote\src\compiler\interface.jl:232
 [21] map
    @ C:\Users\tonys\.julia\dev\Zygote\src\compiler\interface.jl:228 [inlined]
 [22] broadcasted(f::Function, gs::Zygote.Grads, gss::Zygote.Grads)
    @ Zygote C:\Users\tonys\.julia\dev\Zygote\src\compiler\interface.jl:211
 [23] top-level scope
    @ REPL[10]:1
 [24] top-level scope
    @ C:\Users\tonys\.julia\packages\CUDA\k52QH\src\initialization.jl:81
```

CarloLucibello:
fix #971 

I could not fix the gradient + Dict case, if someone has any idea on how to that it would be appreciated, otherwise will just leave the broken test there

### DIFF
--- a/docs/src/utils.md
+++ b/docs/src/utils.md
@@ -42,8 +42,8 @@ gs = gs1 .+ gs2
 @test gs[w] ≈ gs1[w] + gs2[w] 
 @test gs[b] ≈ gs1[b] + gs2[b] 
 
-# gradients and dictionaries interact nicely
-gs .+= Dict(p => randn(size(p)) for p in keys(gs)) 
+# gradients and IdDict interact nicely
+gs .+= IdDict(p => randn(size(p)) for p in keys(gs)) 
 
 # clip gradients
 map(x -> clamp.(x, -0.1, 0.1), gs)
--- a/docs/src/utils.md
+++ b/docs/src/utils.md
@@ -43,6 +43,7 @@ gs = gs1 .+ gs2
 @test gs[b] ≈ gs1[b] + gs2[b] 
 
 # gradients and IdDict interact nicely
+# note that an IdDict must be used for gradient algebra on the GPU
 gs .+= IdDict(p => randn(size(p)) for p in keys(gs)) 
 
 # clip gradients
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -81,7 +81,6 @@ end
 Base.copy(ps::Params) = union!(Params(), ps)
 Base.union(ps::Params, itrs...) = union!(copy(ps), itrs...)
 Base.issetequal(ps1::Params, ps2::Params) = issetequal(ps1.params, ps2.params)
-# Base.issetequal(ps1::Params, x::AbstractSet) = issetequal(ps1.params, x)
 
 function Base.intersect!(ps::Params, itrs...)
   for itr in itrs
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -81,6 +81,7 @@ end
 Base.copy(ps::Params) = union!(Params(), ps)
 Base.union(ps::Params, itrs...) = union!(copy(ps), itrs...)
 Base.issetequal(ps1::Params, ps2::Params) = issetequal(ps1.params, ps2.params)
+Base.issetequal(ps1::Params, x::Base.AbstractSet) = issetequal(ps1.params, x)
 
 function Base.intersect!(ps::Params, itrs...)
   for itr in itrs
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -80,6 +80,8 @@ end
 
 Base.copy(ps::Params) = union!(Params(), ps)
 Base.union(ps::Params, itrs...) = union!(copy(ps), itrs...)
+Base.issetequal(ps1::Params, ps2::Params) = issetequal(ps1.params, ps2.params)
+# Base.issetequal(ps1::Params, x::AbstractSet) = issetequal(ps1.params, x)
 
 function Base.intersect!(ps::Params, itrs...)
   for itr in itrs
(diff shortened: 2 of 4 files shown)


test/cuda.jl
```diff
@@ -37,3 +39,46 @@ end
...
+  @test (gs3 .+ gs4)[b] ≈ gs4[b] 
+  
+  @test_broken begin
+    gs3 .+ Dict(w => similar(w), b => similar(b)) isa Grads
+    gs3 .+= Dict(p => randn!(similar(p)) for p in keys(gs3))
+    gs3 isa Grads 
+  end
+
+  @test_throws ArgumentError gs1 .+ gs4
+end
```
DhairyaLGandhi, line 84: Fine ending

darsnack:
Does
```julia
Base.issetequal(x::Params, y::Base.KeySet) = issetequal(x.params, y)
```
work? What was the error with `AbstractSet` (may not want to define it so broadly anyways)?

CarloLucibello:
> What was the error with AbstractSet (may not want to define it so broadly anyways)?

turns AbstractSet works indeed, the problem was in the dictionary creation, this gives an error
```julia
w = cu(rand(2))
Dict(w => w)
```
CuArrays are not hashable and IdDict should be used instead. 

I currently overload `issetequal(::Params, ::AbstractSet)` for generality, but I'm never sure when those kind of general method overloads would lead to ambiguities in the wild. Better go with the stricter `Base.KeySet`? Although that seems too limited, I think I prefer `AbstractSet`

CarloLucibello:
Nightly failed unrelated and fixed in https://github.com/FluxML/IRTools.jl/pull/86

src/compiler/interface.jl
```diff
@@ -80,6 +80,8 @@ end
 
 Base.copy(ps::Params) = union!(Params(), ps)
 Base.union(ps::Params, itrs...) = union!(copy(ps), itrs...)
+Base.issetequal(ps1::Params, ps2::Params) = issetequal(ps1.params, ps2.params)
+Base.issetequal(ps1::Params, x::Base.AbstractSet) = issetequal(ps1.params, x)
```
darsnack, line 84: Should we also define `issetequal(x::Base.AbstractSet, ps1::Params)` for generality (even though internally we don't use it so far)?

CarloLucibello:
```suggestion
Base.issetequal(x::Base.AbstractSet, ps1::Params) = issetequal(x, ps1.params)
```

docs/src/utils.md
```diff
@@ -42,8 +42,8 @@ gs = gs1 .+ gs2
 @test gs[w] ≈ gs1[w] + gs2[w] 
 @test gs[b] ≈ gs1[b] + gs2[b] 
 
-# gradients and dictionaries interact nicely
-gs .+= Dict(p => randn(size(p)) for p in keys(gs)) 
+# gradients and IdDict interact nicely
```
darsnack, line 45:
```suggestion
# gradients and IdDict interact nicely
# note that an IdDict must be used for gradient algebra on the GPU
```

darsnack (changes requested):

darsnack (approved):
LGTM!

PR #973, opened 2021-05-15
Don't `unbroadcast` some cases which don't need broadcasting

mcabbott:
This is a small optimisation of what I think are fairly common broadcasts:
```julia
julia> @btime gradient((x,y) -> sum(x .* y), $(rand(100,100)), pi);
  20.125 μs (7 allocations: 234.62 KiB)
  13.750 μs (3 allocations: 78.22 KiB)  # this PR

julia> @btime gradient((x,y) -> sum(x ./ y), $(rand(100,100)), pi);
  19.667 μs (7 allocations: 234.62 KiB)
  13.958 μs (3 allocations: 78.22 KiB)  # this PR
```
For best effect it will need https://github.com/JuliaLang/julia/pull/39053, which you can simulate via
```julia
@eval Base function mapreduce(f, op, A::AbstractArrayOrBroadcasted...; kw...)
   get(kw, :dims, :) === (:) && return mapreduce(A -> f(A...), op, zip(A...); kw...)
   return reduce(op, map(f, A...); kw...)
end
```

### DIFF
--- a/src/lib/grad.jl
+++ b/src/lib/grad.jl
@@ -124,7 +124,7 @@ julia> jacobian((a,d) -> prod(a, dims=d), [1 2; 3 4; 5 6], 2)
 !!! warning
     For arguments of any type except `Number` & `AbstractArray`, the result is `nothing`.
 
-```jldoctest; setup=:(using Zygote)
+```
 julia> jacobian((a,s) -> a.^length(s), [1,2,3], "str")
 ([3 0 0; 0 12 0; 0 0 27], nothing)
 
@@ -132,7 +132,7 @@ julia> jacobian((a,t) -> sum(a .* t[1]) + t[2], [1,2,3], (4,5))
 ([4 4 4], nothing)
 
 julia> gradient((a,t) -> sum(a .* t[1]) + t[2], [1,2,3], (4,5))  # gradient undersands the tuple
-([4, 4, 4], (6, 1))
+(Fill(4, 3), (6, 1))
 ```
 """
 function jacobian(f, args...)
--- a/src/lib/grad.jl
+++ b/src/lib/grad.jl
@@ -132,7 +132,7 @@ julia> jacobian((a,t) -> sum(a .* t[1]) + t[2], [1,2,3], (4,5))
 ([4 4 4], nothing)
 
 julia> gradient((a,t) -> sum(a .* t[1]) + t[2], [1,2,3], (4,5))  # gradient undersands the tuple
-(Fill(4, 3), (6, 1))
+([4 4 4], (6, 1))
 ```
 """
 function jacobian(f, args...)
--- a/test/cuda.jl
+++ b/test/cuda.jl
@@ -24,18 +24,6 @@ end
   
 end
 
-@testset "un-broadcasting .*, ./ with scalars" begin
-  cu12 = cu(Float32[1,2])
-  cu55 = cu(Float32[5,5])
-  @test gradient((x,y) -> sum(x .* y), cu12, 5) == (cu55, 3)
-  @test gradient((x,y) -> sum(x .* y), 5, cu12) == (3, cu55)
-  # @test gradient((x,y) -> sum(z -> z, x .* y), cu12, 5) == (cu55, 3)
-  # @test gradient((x,y) -> sum(z -> z, x .* y), 5, cu12) == (3, cu55)
-  cu345 = cu(Float32[3 4 5])
-  @test gradient((x,y) -> sum(x .* y), cu12, cu345) == (cu([12, 12]), cu([3 3 3]))
-  @test all(gradient((x,y) -> sum(x ./ y), cu12, 5) .≈ (cu([0.2, 0.2]), -0.12))
-end
-
 @testset "jacobian" begin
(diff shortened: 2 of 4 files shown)


DhairyaLGandhi:
Yeah, this seems too specific of an optimisation to be acceptable. It also seems to need some testing with the Julialang pr you mentioned.

mcabbott:
> It also seems to need some testing with the Julialang pr you mentioned.

Great, if you have a GPU handy, this would be a useful thing to do.

test/cuda.jl
```diff
@@ -24,6 +24,17 @@ end
   
 end
 
+@testset "un-broadcasting *, / with mapreduce" begin
+  cu12 = cu(Float32[1,2])
+  @test gradient((x,y) -> sum(x .* y), cu12, 5) == ([5, 5]), 3)
```
CarloLucibello, line 29:
```suggestion
  @test gradient((x,y) -> sum(x .* y), cu12, 5) == ([5, 5], 3)
```

CarloLucibello:
a 30% speedup for such common operations with a few lines of code seems great

test/cuda.jl
mcabbott: Oops, thanks. 

I've also now removed the tests like `gradient((x,y) -> sum(z -> z, x .* y), cu12, 5) == ([5, 5], 3)`, as these seem to give errors before this PR:
```
julia> gradient((x,y) -> sum(identity, x .* y), cu12, 5)  # same with z->z
ERROR: Compiling Tuple{typeof(task_local_storage), GPUArrays.var"#24#27"{CuArray{Float32, 1}}, Symbol, GPUArrays.ScalarIndexing}: try/catch is not supported.
```

mcabbott:
It looks like CUDA.jl supports `mapreducedim` fully, and efficiently, which is great -- it's ahead of Base. The interaction with FillArrays wasn't great, and only matters on CPU in the end... but after fixing that, this is a 3x improvement in both time and memory:
```julia
julia> @btime gradient((x,y) -> sum(x .* y), $(rand(100,100)), pi);
  20.125 μs (7 allocations: 234.62 KiB) # master
  6.108 μs (3 allocations: 78.22 KiB)   # this PR + FillArrays PR

julia> @btime gradient((x,y) -> sum(x ./ y), $(rand(100,100)), pi);
  19.667 μs (7 allocations: 234.62 KiB)
  6.781 μs (3 allocations: 78.22 KiB)
```
This is trimmed down from a bigger attempt to do this for more broadcasts. Things like `ones(10,10) ./ ones(10)` could also benefit (and are also common, e.g. normalising columns) but for things like `ones(1,10) .* ones(10)` you'd need something more general than `mapreducedim`, so it needs a bit of logic to avoid that. Maybe later.

src/lib/grad.jl
```diff
@@ -124,15 +124,15 @@ julia> jacobian((a,d) -> prod(a, dims=d), [1 2; 3 4; 5 6], 2)
 !!! warning
     For arguments of any type except `Number` & `AbstractArray`, the result is `nothing`.
 
-```jldoctest; setup=:(using Zygote)
+```
```
CarloLucibello, line 127: why you remove the doctest?

mcabbott: Because one of the gradient calls now returns a FillArray, which was making tests fail. And because I the printing of that may change soon, I realised I'd edited the text to the new printing not the old... still maybe I should put it back?

CarloLucibello: maybe the Fill is more confusing than helpful in this docstring, maybe we can keep the old version but remove the doctest. Or have the Fill but keep the doctest

mcabbott: With this PR + tagged FillArrays, it looks like this, which I think is a bit ugly, and not so helpful:
```
julia> jacobian((a,t) -> sum(a .* t[1]) + t[2], [1,2,3], (4,5))
([4 4 4], nothing)

julia> gradient((a,t) -> sum(a .* t[1]) + t[2], [1,2,3], (4,5))  # gradient undersands the tuple
(3-element Fill{Int64}: entries equal to 4, (6, 1))
```
Maybe best is to restore the old printing with no doctest here:
```
julia> jacobian((a,t) -> sum(a .* t[1]) + t[2], [1,2,3], (4,5))
([4 4 4], nothing)

julia> gradient((a,t) -> sum(a .* t[1]) + t[2], [1,2,3], (4,5))  # gradient undersands the tuple
([4, 4, 4], (6, 1))
```

CarloLucibello: agree

mcabbott:
In fact the `scalar .* array` cases are much easier, I was mislead by thinking about different arrays. In these cases we can just call the existing rules for un-broadcasted multiplication etc, which use `dot` instead of broadcasting & reduction. It'll be faster than the previous benchmarks (at least once some specialisations in FillArrays are added) and faster in other cases, and it's simpler.

src/lib/grad.jl
mcabbott: Done.

(The new method below still returns a Fill in this case, btw.)

CarloLucibello (approved):

PR #976, opened 2021-05-19
minor typo fix

marius311:
Minorest of typos, makes it so the user wont confuse the first clause as declarative.


DhairyaLGandhi (approved):
bors r+

Thank you, it had bugged me recently too!

PR #977, opened 2021-05-20
`unbroadcast` for tuples

### LINKED ISSUE #975: broadcast involving `NTuple` errors `no method matching unbroadcast(::Tuple{…` AND possible fix?
Hi, 
I have a mwe that looks something like 
```julia
import Zygote
f = x -> prod( x .^ (1, 2) )
Zygote.gradient(f, rand(2))
```
i.e, the exponents for `f` are provided as an `NTuple`.
This will result in an error
```
MethodError: no method matching unbroadcast(::Tuple{Int64, Int64}, ::Vector{Float64})
```

I have taken a look into src/lib/unbroadcast.jl.
Although I don't really know what is going on there, this seems to fix the issue (?)
```
Zygote.unbroadcast( x :: NTuple{N, <:Any} where N, x̄ ) = (Zygote.accum_sum(x̄),)
```

Edit: There still seems to be an issue with the return type?
When I do
```julia
Zygote.gradient(f, [1,2])
```
the gradient is a `Vector{Float64}`. Is this expected?

mcabbott:
Fixes #975

### DIFF
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -46,6 +46,7 @@ function Base.reducedim_init(::typeof(identity), ::typeof(accum), A::AbstractArr
 end
 
 trim(x, Δ) = reshape(Δ, ntuple(i -> size(Δ, i), Val(ndims(x))))
+trim(x::Tuple, Δ) = ntuple(k -> Δ[k], length(x))
 
 unbroadcast(x::AbstractArray, x̄) =
   size(x) == size(x̄) ? x̄ :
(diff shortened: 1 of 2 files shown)


src/lib/broadcast.jl
```diff
@@ -55,6 +56,7 @@ unbroadcast(x::AbstractArray, x̄) =
 unbroadcast(x::Number, x̄) = accum_sum(x̄)
 unbroadcast(x::Tuple{<:Any}, x̄) = (accum_sum(x̄),)
 unbroadcast(x::Base.RefValue, x̄) = (x=accum_sum(x̄),)
+unbroadcast(x::Tuple, x̄) = trim(x, length(x) == length(x̄) ? x̄ : accum_sum(x̄; dims=2:ndims(x̄))) # case length(x) > 1
```
CarloLucibello, line 59: I wonder if we can use dispatch on abstractarrays instead of this branching

CarloLucibello (approved):

src/lib/broadcast.jl
mcabbott: What this should avoid is summing a 9x1 matrix over the 2nd dimension, and hence making a copy. whether this can plausibly matter for arrays short enough to become tuples, I don't know. The function does the same thing for arrays (just above) where it can matter:
```
julia> @btime sum($(rand(100,1,100)); dims=2);
  3.713 μs (8 allocations: 78.31 KiB)

julia> @btime sum($(rand(100,2,100)); dims=2);
  5.229 μs (8 allocations: 78.31 KiB)

julia> @btime reshape($(rand(100,1,100)), (100,100));
  17.117 ns (1 allocation: 64 bytes)
```

PR #985, opened 2021-06-02
Update ChainRules and ChainRulesCore

oxinabox:
Replaces #983 and #984 doing both at once.
We should wait for https://github.com/JuliaMath/SpecialFunctions.jl/pull/319 to be merged first though as without that SpecialFunctions.jl holds us back to ChainRulesCore 0.9
which means CI will not actually tell us

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.6.11"
+version = "0.6.12"
 
 [deps]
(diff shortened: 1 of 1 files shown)


oxinabox:
@mzgubic can have approval?

mzgubic (approved):
Let's get CI to pass with ChainRulesCore 0.10 but LGTM

oxinabox:
> Let's get CI to pass with ChainRulesCore 0.10 but LGTM

It does, except for Julia nightly (which Zygote rarely works on)
https://github.com/FluxML/Zygote.jl/pull/985/checks?check_run_id=2730743543#step:5:20

bors r+

PR #988, opened 2021-06-08
delete rule for prod to use ChainRules'

oxinabox:
@mcabbott  added a rule for prod into ChainRules
https://github.com/JuliaDiff/ChainRules.jl/pull/335

It's better than the one in Zygote as it gets the right answer even if one of the elements is zero.

So we can delete the old one here.
But leaving the tests in place per our policy, as a double check against regressions in ChainRules


CarloLucibello (approved):

mcabbott:
Probably this should increase the bound on ChainRules to 0.7.66, from 0.7.55 right now.

PR #987, opened 2021-06-03
WIP: test Zygote-generated gradients using ChainRulesTestUtils `test_rrule`

mzgubic:
A real `rrule_f` example to go with https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/166.

To do:
- [ ] improve docstrings for `rrule_via_ad`
- [x] add an example where `rrule_via_ad` is used to test a gradient that is difficult to evaluate without finite differencing

Some benchmarking shows that `test_rrule` is about 40 times slower than `gradtest`. Do we understand why?
```julia
julia> @testset "time" begin @btime test_rrule((x, W, b) -> identity.(W*x .+ b), rand(5), rand(2,5), rand(2); rrule_f=rrule_via_ad) end
  2.186 ms (7482 allocations: 574.81 KiB)
Test Summary: |  Pass  Total
time          | 30536  30536
[...]

PR #990, opened 2021-06-09
Use ChainRules RuleConfig

oxinabox:
This PR's commit history is pretty gross.
built on top of the merge of https://github.com/FluxML/Zygote.jl/pull/987 and  https://github.com/FluxML/Zygote.jl/pull/967
and then hacks them up.

Right now its kinda just a sketch.
It might compile and work, but I assume it doesn't.

This needs: https://github.com/JuliaDiff/ChainRules.jl/pull/441 which it uses as a test-case.


test/runtests.jl
```diff
@@ -41,7 +42,7 @@ end
 end
 
 @testset "ChainRules" begin
-  include("chainrules.jl")
+  # include("chainrules.jl")
```
mzgubic, line 45: Just flagging this needs to be changed before merging

Project.toml
```diff
@@ -25,6 +25,7 @@ ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 AbstractFFTs = "0.5, 1.0"
 ChainRules = "0.7.55, 0.8"
 ChainRulesCore = "0.9.44, 0.10"
+ChainRulesTestUtils = "0.7.1"
```
mzgubic, line 28: this also needs to be removed

test/chainrules.jl
```diff
@@ -212,8 +209,61 @@ using Zygote, Test, ChainRules
         @test (nothing,) == Zygote.gradient(x->not_diff_kw_eg(x, 2), 10.4)
         @test (nothing,) == Zygote.gradient(x->not_diff_kw_eg(x, 2; kw=2.0), 10.4)
     end
+
+    @testset "rrule_via_ad" begin
+        @testset "basic" begin
+            test_rrule(round, 2.2; rrule_f=rrule_via_ad)
```
mzgubic, line 215: these tests need https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/166/files to be merged first

Project.toml
mzgubic, line 28:
```suggestion
ChainRulesTestUtils = "0.7.9"
```

oxinabox:
Still trying to debug why it won't pick up new rrules  til a `Zygote.refresh()` is done.
Generated functions in general don't see things defined after them,
but we should have the right extra back-edged in place to cause them to see them.
`Zygote.refresh()` fixes it, but we shouldn't need that, we didn't need it for this case before.

The following is what happens right now:
afterr defining the rrule for `cr_inner_demo`.

```
julia> Zygote._pullback(Zygote.Context(), cr_outer_demo, 11)[2](1)
T=Tuple{Zygote.ZygoteRuleConfig{Zygote.Context}, typeof(Main.cr_outer_demo), Int64}; return_type=Nothing
instance=rrule(Zygote.ZygoteRuleConfig{Zygote.Context}, typeof(Main.cr_outer_demo), Int64) from rrule(ChainRulesCore.RuleConfig{T} where T, Any, Any...)
T=Tuple{Zygote.ZygoteRuleConfig{Zygote.Context}, typeof(Main.cr_inner_demo), Int64}; return_type=Nothing
instance=rrule(Zygote.ZygoteRuleConfig{Zygote.Context}, typeof(Main.cr_inner_demo), Int64) from rrule(ChainRulesCore.RuleConfig{T} where T, Any, Any...)
T=Tuple{Zygote.ZygoteRuleConfig{Zygote.Context}, typeof(Base.:(*)), Int64, Int64}; return_type=Tuple{Int64, ChainRules.var"#times_pullback#1490"{Int64, Int64}}
T=Tuple{Zygote.ZygoteRuleConfig{Zygote.Context}, typeof(Base.:(*)), Int64, Int64}; return_type=Tuple{Int64, ChainRules.var"#times_pullback#1490"{Int64, Int64}}
(nothing, 50)

julia> Zygote._pullback(Zygote.Context(), cr_outer_demo, 11)[2](1)  # just uses what it found before by ADing fully wihtout rule
(nothing, 50)
```

The problem is this line
```
T=Tuple{Zygote.ZygoteRuleConfig{Zygote.Context}, typeof(Main.cr_inner_demo), Int64}; return_type=Nothing
```

but it should be (shown after doing a `Zygote.refresh()`
```
T=Tuple{Zygote.ZygoteRuleConfig{Zygote.Context}, typeof(Main.cr_inner_demo), Int64}; return_type=Tuple{Int64, Main.var"#cr_inner_demo_pullback#10"}
```

src/compiler/chainrules.jl
```diff
@@ -10,13 +13,17 @@ If it does not, then the second argument is a list of edges to attach to the Cod
 such that if a suitable rule is defined later, the generated function will recompile.
 """
 function has_chain_rrule(T)
-  m = meta(Tuple{typeof(rrule),T.parameters...})
-  if m.method !== chainrules_fallback
+  return_type = Core.Compiler.return_type(rrule, T)
```
vtjnash, line 16: Looks like you are trying to use `Core.Compiler.return_type` *inside* the generated function. That is not what you want: call this outside the the generated function, and possibly pass the answer into the generated function if necessary.

oxinabox: I am not sure that is possible.
There is no outside the generated functions.
Within the 
`_pullback` generated function is the first place we have the information that is needed to decide if we should use a ChainRule or not.

oxinabox: Is the problem that the content of the generated fuynction *always* runs in the world-age where it was declared.
Not in the world-age where it was run to generate the function?
So backedges can't fix it, as they don't change when it was declared they just cause it to recompile, but it still acts in that world age where it was declared regardless?

vtjnash: Calling `Core.Compiler.return_type` inside `@generated` should be entirely prohibited, as it also risks crashing the compiler, but I am not yet brave enough to address that problem

oxinabox: ok, I will find another solution then.

oxinabox:
Alternative solution for detecting if rules don't exist, since we can't ose `Core.Compiler.return_type`.

First check if we are hitting the `rrule(::RuleConfig, args...) = rrule(args...)` **redispatcher**.
then check if the `rrule(args...)` would hit the `rrule(args...) = nothing` **fallback**.
If that is the case then we have no rule and so need to attach the backedge, to the **redispatcher** (which will be invalidated in turn by backedged from the **fallback** anyway, so no need to attach backedges there).

Otherwise, we have hit a rule and we need no edges (because we will be returning an Expr that calls the rule, so natural edges apply).

src/compiler/chainrules.jl
```diff
@@ -1,7 +1,11 @@
 struct ZygoteRuleConfig{CTX<:AContext} <: RuleConfig{Union{HasReverseMode,NoForwardsMode}}
   context::CTX
 end
-ZygoteRuleConfig()=ZygoteRuleConfig(Context())
+ZygoteRuleConfig() = ZygoteRuleConfig(Context())
+
+
+const rrule_fallback_method = Base.which(rrule, Tuple{Any, Vararg{Any}})
+const rrule_redisatcher_method = Base.which(rrule, Tuple{RuleConfig, Any, Vararg{Any}})
```
mzgubic, line 8:
```suggestion
const rrule_redispatcher_method = Base.which(rrule, Tuple{RuleConfig, Any, Vararg{Any}})
```

src/compiler/chainrules.jl
```diff
@@ -13,17 +17,20 @@ If it does not, then the second argument is a list of edges to attach to the Cod
...
-    # no rule exists, or we hit a specialisation telling us to keep digging
-    m = meta(Tuple{typeof(rrule),T.parameters...})
-    Core.println("instance=", m.instance)
-    return false, m.instance
-  else
-    # found a rrule, no need to add any edges
-    return true, nothing
+  config_T, arg_Ts = Iterators.peel(T.parameters)
+  m_with_config = meta(Tuple{typeof(rrule), config_T, arg_Ts...})
+  if m_with_config.method === rrule_redisatcher_method
```
mzgubic, line 22:
```suggestion
  if m_with_config.method === rrule_redispatcher_method
```

oxinabox:
The core of this now works. All the tests in `test/chainrules.jl` pass.
Mostly what is left is chasing up edge cases in https://github.com/JuliaDiff/ChainRules.jl/pull/441/
of which there are surprisingly many (and getting that merged)

Project.toml
```diff
@@ -25,6 +25,7 @@ ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 AbstractFFTs = "0.5, 1.0"
 ChainRules = "0.7.66, 0.8"
```
oxinabox, line 26:
```suggestion
ChainRules = "0.8.12"
```

oxinabox:
This is all passing for me locally, so just need https://github.com/JuliaDiff/ChainRules.jl/pull/441 merged and it will pass on CI.

Project.toml
mzgubic: why was this resolved?

src/compiler/chainrules.jl
```diff
@@ -10,13 +17,20 @@ If it does not, then the second argument is a list of edges to attach to the Cod
...
-    # found a rrule, no need to add any edges
-    return true, nothing
+  config_T, arg_Ts = Iterators.peel(T.parameters)
+  m_with_config = meta(Tuple{typeof(rrule), config_T, arg_Ts...})
+  if m_with_config.method === rrule_redispatcher_method
+    # it is being redispatched without config, so check it that hits the fallback
+    m_without_config = meta(Tuple{typeof(rrule), arg_Ts...})
+    if m_without_config.method === rrule_fallback_method
+      # no rrule exists, return instance for m_with_config as that will be invalidated 
+      # directly if configured rule added, or indirectly if unconfigured ruleadded
```
mzgubic, line 27: I don't know what this comment means

test/chainrules.jl
```diff
@@ -1,8 +1,5 @@
-using Zygote, Test, ChainRules
-
-
-@testset "ChainRules Integration" begin
-    @testset "basic" begin
+@testset "ChainRules intergration" begin
```
mzgubic, line 1:
```suggestion
@testset "ChainRules integration" begin
```

src/compiler/interface2.jl
```diff
@@ -10,12 +10,18 @@ end
   T = Tuple{f,args...}
   ignore_sig(T) && return :(f(args...), Pullback{$T}(()))
 
-  iskw = is_kwfunc(f, args...)
-  # if it is_kw then `args[1]` are the keyword args, `args[2]` is actual function
-  base_T = iskw ? Tuple{args[2:end]...} : T
-  hascr, cr_edge = has_chain_rrule(base_T)
-  chain_rrule_f = iskw ? :chain_rrule_kw : :chain_rrule
-  hascr && return :($chain_rrule_f(f, args...))
+  if is_kwfunc(f, args...)
```
mzgubic, line 13: this is much nicer, thanks!

test/chainrules.jl
```diff
@@ -208,12 +208,69 @@ using Zygote, Test, ChainRules
...
+        makefoo(a, b) = Foo(a, b)
+        sumfoo(foo) = foo.x + foo.y
+
+
+        test_rrule(
+            ZygoteRuleConfig(), sumfoo, Foo(1.2, 2.3); rrule_f=rrule_via_ad, check_inferred=false
+        )
+        test_rrule(ZygoteRuleConfig(),
+            makefoo, 1.0, 2.0; rrule_f=rrule_via_ad, check_inferred=false
+        )
```
mzgubic, line 244:
```suggestion
        test_rrule(
            ZygoteRuleConfig(), makefoo, 1.0, 2.0;
            rrule_f=rrule_via_ad, check_inferred=false
        )
```

mzgubic (approved):
Thanks, looks great, excited about this!

One remaining concern is how do we opt out of using rules if we can't rely on the return value? I guess one way is to call `f/rrule_via_ad` inside the rule that we are opting out from. But that is far from ideal as it flops between the rule and AD and this results in redundant differential type changes. Is there a better way?

src/compiler/chainrules.jl
oxinabox:
```suggestion
      # directly if configured rule added, or indirectly if unconfigured rule added
```

> I don't know what this comment means

It is about how julia knows when to trigger recompilation because of new methods being added.
We can  have a call about it, it needs a fair bit of background info

oxinabox:
> One remaining concern is how do we opt out of using rules if we can't rely on the return value? 

Correct, that concern remains.
It is out of scope for this PR to solve, i had hoped we could solve it incidentally, but it seems the obvious way doens't work.

> I guess one way is to call f/rrule_via_ad inside the rule that we are opting out from. 

This is not trivial since, it will (if nothing special is done) just infinite-loop.
Maybe `invoke` can be used to prevent that?

PR #1003, opened 2021-06-19
use `Base.ismutabletype`

mcabbott:
May fix a test failure on master.

### DIFF
--- a/src/lib/lib.jl
+++ b/src/lib/lib.jl
@@ -1,4 +1,10 @@
-using Base: RefValue, ismutabletype
+using Base: RefValue
+
+if VERSION > v"1.7.0-DEV.204"
+  using Base: ismutabletype
+else
+  ismutabletype(::Type{T}) where T = T.mutable
+end
 
 # Interfaces
 
--- a/src/lib/lib.jl
+++ b/src/lib/lib.jl
@@ -1,4 +1,4 @@
-using Base: RefValue
+using Base: RefValue, ismutabletype
 
 # Interfaces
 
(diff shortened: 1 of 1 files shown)


src/lib/lib.jl
```diff
@@ -1,5 +1,11 @@
 using Base: RefValue
 
+if VERSION > v"1.7.0-DEV.204"
+  using Base: ismutabletype
+else
+  ismutabletype(::Type{T}) where T = T.mutable
```
johnnychen94, line 6:
```suggestion
  function ismutabletype(@nospecialize(t::Type))
    t = Base.unwrap_unionall(t)
    return isa(t, DataType) && t.mutable
  end
```

mcabbott: Maybe better? It used to literally call `T.mutable` so that's all I was trying to preserve.

johnnychen94: Just feel it better to keep consistent with the upstream implementation if we want to use the name `ismutabletype` 😆

mcabbott: Yes am using that on new versions; CI informed me that it's a recent addition. 

I guess if there any difference in specialisation performance, we don't care as it's only used inside `@generated`, in fact would prefer less specialisation always.

mcabbott: Let's call CI sufficiently happy

src/lib/lib.jl
```diff
@@ -1,5 +1,14 @@
 using Base: RefValue
 
+if VERSION > v"1.7.0-DEV.204"
+  using Base: ismutabletype
+else
+  function ismutabletype(@nospecialize(t::Type))
```
ChrisRackauckas, line 6: `@pure` this?

mcabbott: It's copied straight from Base, I see this was discussed here: https://github.com/JuliaLang/julia/pull/39037/files#r550261314

PR #1009, opened 2021-06-25
Use abstract GPU types for broadcast.

akashkgarg:
Replace GPU broadcast types to their more generic counterparts.

### DIFF
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -247,8 +247,8 @@ end
     end
   end
 
-  @adjoint CUDA.DenseCuArray{N,T}(xs::Array) where {N,T} =
-    CUDA.DenseCuArray{N,T}(xs), Δ -> (convert(Array, Δ), )
+  @adjoint CUDA.CuArray{N,T}(xs::Array) where {N,T} =
+    CUDA.CuArray{N,T}(xs), Δ -> (convert(Array, Δ), )
 
   @adjoint function sum(xs::CUDA.DenseCuArray; dims = :)
     placeholder = similar(xs)
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -250,19 +250,19 @@ end
   @adjoint CUDA.CuArray{N,T}(xs::Array) where {N,T} =
     CUDA.CuArray{N,T}(xs), Δ -> (convert(Array, Δ), )
 
-  @adjoint function sum(xs::CUDA.DenseCuArray; dims = :)
+  @adjoint function sum(xs::CUDA.AbstractGPUArray; dims = :)
     placeholder = similar(xs)
     sum(xs, dims = dims), Δ -> (placeholder .= Δ,)
   end
   
   # Make sure sum(f, ::CuArray) uses broadcase through forward-mode defined above
   # Not the ChainRules.rrule which will use the Zygote.Context and thus not be GPU compatible
-  @adjoint function sum(f, xs::CUDA.DenseCuArray; kws...)
+  @adjoint function sum(f, xs::CUDA.AbstractGPUArray; kws...)
     @assert !haskey(kws, :init) # TODO add init support (julia 1.6)
     return pullback(__context__, (f, xs) -> sum(f.(xs); kws...), f, xs)
   end
   
-  @adjoint function Base.convert(::Type{T}, xs::Array)  where {T<:CUDA.DenseCuArray}
+  @adjoint function Base.convert(::Type{T}, xs::Array)  where {T<:CUDA.AbstractGPUArray}
     Base.convert(T, xs), Δ -> (nothing, Base.convert(Array, Δ),)
   end
 
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -233,7 +233,7 @@ end
 end
 
 @init @require CUDA="052768ef-5323-5732-b1bb-66c8b64840ba" begin
-  const CuArrayStyle = CUDA.CuArrayStyle
+  const CuArrayStyle = CUDA.AbstractGPUArrayStyle
 
   if isdefined(CUDA, :cufunc)
     @eval @adjoint function broadcasted(::CuArrayStyle, f, args...)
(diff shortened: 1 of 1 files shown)


CarloLucibello:
Hi, thank you for this! Can you add some tests for some newly supported types?

mcabbott:
Agree test would be ideal, but they might be pretty involved to write. 

Motivation from @akashkgarg was the following:

> using here a custom CuUnifiedArray type that is similar to CuArray but lives in Unified GPU memory accessible across multiple GPUs

giving an error like so, which I think was caused by the new array not using something `<: CuArrayStyle` and hence not catching the dispatch to the simpler broadcasting. It seems likely that any GPU-array is going to have the same problem.

```
ERROR: LoadError: GPU compilation of kernel broadcast_kernel(CUDA.CuKernelContext, CuDeviceMatrix{Tuple{Float64, Zygote.ZBack{ChainRules.var"#cos_pullbac[103/3908]
oat64}}}, 1}, Base.Broadcast.Broadcasted{Nothing, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}, Zygote.var"#1096#1103"{Zygote.Context, typeof(cos)}, Tuple{Base.Broa
dcast.Extruded{CuDeviceMatrix{Float64, 1}, Tuple{Bool, Bool}, Tuple{Int64, Int64}}}}, Int64) failed                                
KernelError: passing and using non-bitstype argument                                                                                                         
Argument 4 to your kernel function is of type Base.Broadcast.Broadcasted{Nothing, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}, Zygote.var"#1096#1103"{Zygote.Contex
t, typeof(cos)}, Tuple{Base.Broadcast.Extruded{CuDeviceMatrix{Float64, 1}, Tuple{Bool, Bool}, Tuple{Int64, Int64}}}}, which is not isbits:                        
  .f is of type Zygote.var"#1096#1103"{Zygote.Context, typeof(cos)} which is not isbits.                                           
    .__context__ is of type Zygote.Context which is not isbits.                                                                    
      .cache is of type Union{Nothing, IdDict{Any, Any}} which is not isbits.                                                   
Stacktrace:                                                                                                              
  [1] check_invocation(job::GPUCompiler.CompilerJob)                                                                                         
    @ GPUCompiler ~/.julia/packages/GPUCompiler/OyOEp/src/validation.jl:66                                                                            
  [2] macro expansion                                                                                                                        
    @ ~/.julia/packages/GPUCompiler/OyOEp/src/driver.jl:318 [inlined]                                                                        
  [3] macro expansion                                                                                                                                 
    @ ~/.julia/packages/TimerOutputs/PZq45/src/TimerOutput.jl:226 [inlined]                                                                           
  [4] macro expansion                                                                                                                                 
    @ ~/.julia/packages/GPUCompiler/OyOEp/src/driver.jl:317 [inlined]                                                                                          
  [5] emit_asm(job::GPUCompiler.CompilerJob, ir::LLVM.Module; strip::Bool, validate::Bool, format::LLVM.API.LLVMCodeGenFileType)             
    @ GPUCompiler ~/.julia/packages/GPUCompiler/OyOEp/src/utils.jl:62                                                    
  [6] cufunction_compile(job::GPUCompiler.CompilerJob)                                                                                                       
    @ CUDA ~/.julia/environments/v1.6/dev/CUDA/src/compiler/execution.jl:313                                                                                    
  [7] cached_compilation(cache::Dict{UInt64, Any}, job::GPUCompiler.CompilerJob, compiler::typeof(CUDA.cufunction_compile), linker::typeof(CUDA.cufunction_link))
    @ GPUCompiler ~/.julia/packages/GPUCompiler/OyOEp/src/cache.jl:89                                                                                        
  [8] cufunction(f::GPUArrays.var"#broadcast_kernel#16", tt::Type{Tuple{CUDA.CuKernelContext, CuDeviceMatrix{Tuple{Float64, Zygote.ZBack{ChainRules.var"#cos_pullba
ck#1454"{Float64}}}, 1}, Base.Broadcast.Broadcasted{Nothing, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}, Zygote.var"#1096#1103"{Zygote.Context, typeof(cos)}, Tupl
e{Base.Broadcast.Extruded{CuDeviceMatrix{Float64, 1}, Tuple{Bool, Bool}, Tuple{Int64, Int64}}}}, Int64}}; name::Nothing, kwargs::Base.Iterators.Pairs{Union{}, Unio
n{}, Tuple{}, NamedTuple{(), Tuple{}}})                                                                                                                      
    @ CUDA ~/.julia/environments/v1.6/dev/CUDA/src/compiler/execution.jl:288                                                                 
  [9] cufunction                                                                                                                                             
    @ ~/.julia/environments/v1.6/dev/CUDA/src/compiler/execution.jl:282 [inlined]                                                                              
 [10] macro expansion                                                                                                                                        
    @ ~/.julia/environments/v1.6/dev/CUDA/src/compiler/execution.jl:102 [inlined]                                                                     
 [11] #launch_heuristic#234                                                                                                                                  
    @ ~/.julia/environments/v1.6/dev/CUDA/src/gpuarrays.jl:17 [inlined]                                                                                        
 [12] launch_heuristic                                                                                                                                       
    @ ~/.julia/environments/v1.6/dev/CUDA/src/gpuarrays.jl:17 [inlined]                                                                                        
 [13] copyto!                           
    @ ~/.julia/packages/GPUArrays/ymNij/src/host/broadcast.jl:63 [inlined]                                                                                         
 [14] copyto!                           
    @ ./broadcast.jl:936 [inlined]                                               
 [15] copy                              
    @ ~/.julia/packages/GPUArrays/ymNij/src/host/broadcast.jl:47 [inlined]                                                                                         
 [16] materialize                       
    @ ./broadcast.jl:883 [inlined]                                               
 [17] _broadcast(f::Zygote.var"#1096#1103"{Zygote.Context, typeof(cos)}, x::CuUnifiedMatrix{Float64})
 ```

akashkgarg:
@CarloLucibello I'd be happy to add some tests, being a little new to this can you point in the right direction on how I might test these abstract types? I think there are tests already test/cuda.jl for example that should test that the new abstract interfaces don't break existing CUDA support.

CarloLucibello:
let's just merge this, adding tests seems complicated

DhairyaLGandhi:
Is there some mwe that compiles with this PR and not without?

DhairyaLGandhi:
Also, please use bors.

DhairyaLGandhi:
I don't mean to say we shouldn't use abstract types - far from it. I am in favor of the PR.

I want an mwe to test so we can expand our coverage and have some checks on expected behavior available to us for future work.

DhairyaLGandhi:
In fact, for DDP I was going to make this PR myself since I had needed something similar for unified memory.

ChrisRackauckas:
I don't think the MWE will work until https://github.com/JuliaGPU/CUDA.jl/pull/946 is a bit further along. But the hope is that when that is done, then it'll just work with Flux, and then we'll have an MWE to add as a test.

ToucheSir:
On the note of broader GPU support, is there anything in https://github.com/FluxML/Zygote.jl/blob/2e6407e62a0ab3406bdf4a758cbb25f3dba46b5c/src/lib/broadcast.jl#L249-L285 that can't be moved to use exports from GPUArrays instead of CUDA? Both `AbstractGPUArrayStyle` and `AbstractGPUArray` are re-exported from there, after all. Would it make sense to split the require block in two to separate out the `CuArray` constructor and pre CUDA 3.0 `broadcasted` adjoint?

ChrisRackauckas:
Yes, I think it would make sense to make that require block not require CUDA.jl for most of it.

mcabbott:
The broadcast overload can probably be deleted, since (for sufficiently simple `f` over real numbers) the same `broadcast_forward` is now used on the CPU. 

(For stateful `f`, or complex numbers, you'll get an error instead of a wrong answer.)

PR #1010, opened 2021-06-26
Add convenience function to return primal & cotangents together

mcabbott:
This adds the long-requested feature of easily getting both the gradient and the loss at once. 

It's a new function to avoid breaking anything. Many worse names for it have been proposed. It returns a named tuple because why not; you can still destructure or index `res[2][1]`  if you prefer.

Closes #892. Also copies across a minimal docstring for `gradient` with implicit parameters.

### DIFF
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -53,6 +53,9 @@ Returns a tuple containing `∂f/∂x` for each argument `x`,
 the derivative (for scalar x) or the gradient.
 
 `f(args...)` must be a real number, see [`jacobian`](@ref) for array output.
+
+See also [`withgradient`](@ref) to keep the value `f(args...)`,
+and `pullback`](@ref) for value and back-propagator.
 """
 function gradient(f, args...)
   y, back = pullback(f, args...)
@@ -60,6 +63,25 @@ function gradient(f, args...)
 end
 
 Base.adjoint(f::Function) = x -> gradient(f, x)[1]
+
+"""
+    withgradient(f, args...)
+
+Returns both the value `f(args...)` and the [`gradient`](@ref), 
+`∂f/∂x` for each argument `x`, as a named tuple.
+
+```jldoctest
+julia> y, ∇ = withgradient(/, 1, 2)
+(val = 0.5, grad = (0.5, -0.25))
+
+julia> ∇ == gradient(/, 1, 2)
+true
+```
+"""
+function withgradient(f, args...)
+  y, back = pullback(f, args...)
+  (val=y, grad=back(sensitivity(y)))
+end
 
 # Param-style wrappers
 
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -66,9 +66,11 @@ Base.adjoint(f::Function) = x -> gradient(f, x)[1]
 
 """
     withgradient(f, args...)
+    withgradient(f, ::Params)
 
 Returns both the value `f(args...)` and the [`gradient`](@ref), 
 `∂f/∂x` for each argument `x`, as a named tuple.
+With imiplicit parameters, the value is `f()`.
 
 ```jldoctest
 julia> y, ∇ = withgradient(/, 1, 2)
@@ -85,10 +87,24 @@ end
 
 # Param-style wrappers
 
-# TODO store ids only
+"""
+    gradient(() -> loss(), ::Params) -> Grads
+
+Gradient with implicit parameters. Returns a container, from which
+`grads[W]` extracts the gradient with respect to some array `W`,
+if this is among those being tracked, for example via `Params([W, A, B])`.
+"""
+gradient
+
+"""
+    Params([A, B, C...])
+
+Container for implicit parameters, differentiating a zero-argument
+funtion `() -> loss()` with respect to `A, B, C`.
+"""
 struct Params
   order::Buffer # {Any, Vector{Any}}
-  params::IdSet{Any}
+  params::IdSet{Any} # TODO store ids only
 end
 
 Params() = Params(Buffer([], false), IdSet())
@@ -193,7 +209,13 @@ function copy!(x::AbstractVector, ps::Params)
   ps
 end
 
+"""
+    Grads(...)
 
+Dictionary-like container returned when taking gradients with
+respect to implicit parameters. For an array `W`, appearing 
+within `Params([W, A, B...])`, the gradient is `g[W]`.
+"""
 struct Grads
   grads::IdDict{Any,Any}
   params::Params
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -78,8 +78,8 @@ Base.adjoint(f::Function) = x -> gradient(f, x)[1]
     withgradient(f, args...)
     withgradient(f, ::Params)
 
-Returns both the value `f(args...)` and the [`gradient`](@ref)
-as a named tuple. With implicit parameters, the value is `f()`.
+Returns both the value of the function and the [`gradient`](@ref),
+as a named tuple. 
 
 ```jldoctest; setup=:(using Zygote)
 julia> y, ∇ = withgradient(/, 1, 2)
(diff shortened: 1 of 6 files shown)


src/compiler/interface.jl
```diff
@@ -61,12 +64,47 @@ end
...
+
 # Param-style wrappers
 
-# TODO store ids only
+"""
+    gradient(() -> loss(), ::Params) -> Grads
+
+Gradient with implicit parameters. Returns a container, from which
+`grads[W]` extracts the gradient with respect to some array `W`,
+if this is among those being tracked, for example via `Params([W, A, B])`.
```
darsnack, line 95:
```suggestion
Gradient with implicit parameters.
Returns a container indexable by the arrays in `Params` (see examples).

```jldoctest
julia> A, B = ones(3, 2), ones(3, 3);

julia> ps = Params([A, B]);

julia> grads = gradient(() -> sum(A) + sum(B), ps)
Grads(...)

julia> grads[A] == ones(3, 2)
true
```

```

Might need to tweak the output to get the test to pass. I wrote this on mobile.

File: src/compiler/interface.jl [hunk 2]
```diff
@@ -61,12 +64,47 @@ end
...
+"""
+gradient
+
+"""
+    Params([A, B, C...])
+
+Container for implicit parameters, differentiating a zero-argument
+funtion `() -> loss()` with respect to `A, B, C`.
```
darsnack, line 103:
```suggestion
Container for implicit parameters.
Useful for differentiating a zero-argument
funtion `() -> loss()` with respect to `A, B, C`.
```

src/lib/grad.jl
```diff
@@ -137,7 +137,19 @@ julia> gradient((a,t) -> sum(a .* t[1]) + t[2], [1,2,3], (4,5))  # gradient unde
...
+jacobian(f, args...) = withjacobian(f, args...).grad
+
+"""
+    withjacobian(f, args...)
+
+Returns both the value `f(args...)` and the [`jacobian`](@ref) as a named tuple.
+
+```jldoctest; setup=:(using Zygote)
+julia> withjacobian(cumsum, [1,2,3])
+(val = [1, 3, 6], grad = ([1 0 0; 1 1 0; 1 1 1],))
```
darsnack, line 149: Should it be `jacobian` instead of `grad`? Or we can standardize around `value` and `deriv`?

src/compiler/interface.jl
```diff
@@ -61,12 +64,47 @@ end
 
 Base.adjoint(f::Function) = x -> gradient(f, x)[1]
 
+"""
+    withgradient(f, args...)
+    withgradient(f, ::Params)
+
+Returns both the value `f(args...)` and the [`gradient`](@ref), 
+`∂f/∂x` for each argument `x`, as a named tuple.
```
darsnack, line 72:
```suggestion
`∂f/∂x` for each argument `x in args`/`x in Params(...)`, as a named tuple.
```

src/compiler/interface.jl
```diff
@@ -61,12 +64,47 @@ end
...
+With imiplicit parameters, the value is `f()`.
```
darsnack, line 73:
```suggestion
With implicit parameters, the value is `f()`.
```

darsnack (commented):
This looks great. I like the names `withgradient` and `withjacobian` a lot.

I left a few comments to hopefully make the docstrings clearer.

src/lib/grad.jl
mcabbott: I was hoping that `grad` was neutral enough, and abbreviated to a comparable extent to `val`. Zygote already has `Grads` for all purposes. 

Of course they are all derivatives, just with varying numbers of indices on... `deriv` somehow seems less pretty to me, and of course "derive" mostly means something unrelated to "derivative".

It could follow the function names exactly, but IMO it's a little weird to re-use a function name as a field name (when that function isn't the getter).

src/compiler/interface.jl
mcabbott: Wasn't sure how big to go here. The other `gradient` docsting does not give examples.

src/compiler/interface.jl
mcabbott: I've tweaked this, in the end removing `∂f/∂x` here, to focus on differences from `gradient`.

src/compiler/interface.jl
mcabbott: See if you like my example after update.

src/compiler/interface.jl
```diff
@@ -50,9 +50,22 @@ sensitivity(y) = error("Output should be scalar; gradients are not defined for o
     gradient(f, args...)
 
 Returns a tuple containing `∂f/∂x` for each argument `x`,
-the derivative (for scalar x) or the gradient.
+the derivative (for scalar `x`) or the gradient.
 
 `f(args...)` must be a real number, see [`jacobian`](@ref) for array output.
+
+See also [`withgradient`](@ref) to keep the value `f(args...)`,
+and `pullback`](@ref) for value and back-propagator.
```
CarloLucibello, line 58:
```suggestion
and [`pullback`](@ref) for value and back-propagator.
```

src/compiler/interface.jl
```diff
@@ -50,9 +50,22 @@ sensitivity(y) = error("Output should be scalar; gradients are not defined for o
...
+
+```jldoctest; setup=:(using Zygote)
+julia> gradient(*, 2, 3, 5)
+(15, 10, 6)
+
+julia> gradient([7,11,13]) do x
+         sum(abs2, x)
+       end
+([14, 22, 26],)
```
CarloLucibello, line 67:
```suggestion
julia> gradient(x -> sum(abs2, x), [7,11,13])
([14, 22, 26],)
```

CarloLucibello: just because the do syntax could be confusing for beginners

CarloLucibello:
yeah, nice names

src/compiler/interface.jl
mcabbott: We can see what CI thinks, I didn't think there was a doc entry for `pullback` so it might complain.

mcabbott: Turns out it has no complaints about this, but does have a few similar complaints... although they don't cause a failure so that's OK.

https://github.com/FluxML/Zygote.jl/pull/1010/checks#step:6:89

src/compiler/interface.jl
mcabbott: True. Although it would also be good to teach them, since this is a common & useful idiom for gradients. I've added this plus a `do` variant, see what you think.

src/compiler/interface.jl
darsnack: Yeah the new docstrings with examples all look good

src/lib/grad.jl
darsnack: How do you feel about `value` and `diff`?

mcabbott: I would say `val` & `diff` as it seems odd to abbreviate only one. One objection is that `diff` is actually a function, but I doubt that'll confuse anyone. 

I could live with this, but still prefer `grad` I think. To my ear at least, it seems generic enough to mean "the thing AD produces" not specifically what calculus class called the gradient.

darsnack: Yeah perhaps you are right. I am okay with either `val`/`diff` or `val`/`grad` for this PR.

mcabbott: Cool. Perhaps this is ready to go?

darsnack (approved):
LGTM!

src/lib/grad.jl
CarloLucibello: sorry to come late to the party, but can't we bypass the naming issue entirely by returning a tuple? Returning a named tuple doesn't seem to add much additional value

darsnack: 🤷🏾‍♂️ it's nice on the user end to make their code more readable. I don't think the current names are bad, I just wanted to make sure we discussed it.

mcabbott: Yes I guess the downside is that naming things is hard, as they say. But IMO there are already too many tuples of tuples of iterators of tuples etc. floating around, hence lots of slightly mysterious `[2]`s, seems nice to have some optional signposts.

Not sure I'm brave enough see what happens if `pullback` returned a named tuple, though!

PR #1016, opened 2021-07-02
Documentation edit on implicit parameters

JordiBolibar:
After struggling to handle implicit parameters with a Flux model, and following a discourse discussion (https://discourse.julialang.org/t/unrecognized-gradient-using-zygote-for-ad-with-universal-differential-equations/63791/2) , I have decided to add some extra details on how to access and when to use implicit parameters. I hope this helps new users like me to avoid wasting time looking for this.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -129,7 +129,7 @@ julia> gradient(colordiff, RGB(1, 0, 0), RGB(0, 1, 0))
 ((r = 0.4590887719632896, g = -9.598786801605689, b = 14.181383399012862), (r = -1.7697549557037275, g = 28.88472330558805, b = -0.044793892637761346))
 ```
 
-## Explicit and implicit parameters of ML models
+## Explicit and Implicit Parameters
 
 It's easy to work with even very large and complex models, and there are few ways to do this. Autograd-style models pass around a collection of weights. There are two ways of passing *explicit* parameters:
 
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -189,16 +189,9 @@ Grads(...)
 julia> grads[W], grads[b]
 ([0.652543 … 0.683588], [1.0, 1.0])
 ```
-Unlike with explicit gradients, in order to see implicit gradients one needs to do:
+To inspect the `Grads(...)` object returned for implicit parameters, you can index it using the parameters passed to `Params`:
 
 ```julia
-julia> grads.grads
-IdDict{Any, Any} with 5 entries:
-  [0.467471 0.597815 … 0.678126 … => [0.579671 0.215381 … 0.635058 0.623832; 0.579671 0.215381 … …
-  :(Main.x)                       => [1.3377, 0.930234, 0.499161, 1.33827, 1.37791]
-  :(Main.W)                       => [0.579671 0.215381 … 0.635058 0.623832; 0.579671 0.215381 … …
-  [0.106308, 0.705531]            => 2-element Fill{Float64}: entries equal to 1.0
-  :(Main.b)                       => 2-element Fill{Float64}: entries equal to 1.0
-```
+julia> [grads[p] for p in [W, b]]
 
 However, implicit parameters exist mainly for compatibility with Flux's current AD; it's recommended to use the other approaches unless you need this.
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -170,7 +170,7 @@ julia> dmodel = gradient(model -> sum(model(x)), model)[1]
 (W = [0.652543 … 0.683588], b = [1.0, 1.0])
 ```
 
-On the other hand, the *implicit* style is the one presently used by [Flux.jl](https://github.com/FluxML/Flux.jl), a closely related machine learning library. It uses structs like `Linear` above to define layers, and the function `Flux.params(model)` returns a `Params` object containing all the parameters of all layers. See [its documentation](https://fluxml.ai/Flux.jl/stable/models/basics/) for more details. When using Zygote for most other purposes, however, the explicit style is usually preferred.
+Zygote also support another way to take gradients, via *implicit parameters*. Here the loss function takes zero arguments, but the variables of interest are indicated by a special `Params` object. The function `linear` which depends on `W` and `b` is executed when the loss function `() -> sum(linear(x))` is called, and hence this dependence is visible to Zygote:
 
 ```julia
 julia> W = rand(2, 5); b = rand(2);
@@ -186,4 +186,3 @@ julia> grads[W], grads[b] # access gradients using arrays as keys
 ```
 
 Here `grads` is a dictionary-like object, whose keys are the same parameters we indicated in `Params`. (In fact it wraps a dictionary using `objectid(W)` as keys, which does not change if the values in `W` are mutated).
-
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -170,7 +170,7 @@ julia> dmodel = gradient(model -> sum(model(x)), model)[1]
 (W = [0.652543 … 0.683588], b = [1.0, 1.0])
 ```
 
-Zygote also support one more way to take gradients, via *implicit parameters* – this is a lot like autograd-style gradients, except we don't have to thread the parameter collection through all our code. When working with Flux models, this is the recommended way of passing the gradients, as it ensures compatibility with Flux's built-in optimizers. 
+On the other hand, the *implicit* style is the one presently used by [Flux.jl](https://github.com/FluxML/Flux.jl), a closely related machine learning library. It uses structs like `Linear` above to define layers, and the function `Flux.params(model)` returns a `Params` object containing all the parameters of all layers. See [its documentation](https://fluxml.ai/Flux.jl/stable/models/basics/) for more details. When using Zygote for most other purposes, however, the explicit style is usually preferred.
 
 ```julia
 julia> W = rand(2, 5); b = rand(2);
@@ -187,4 +187,3 @@ julia> grads[W], grads[b] # access gradients using arrays as keys
 
 Here `grads` is a dictionary-like object, whose keys are the same parameters we indicated in `Params`. (In fact it wraps a dictionary using `objectid(W)` as keys, which does not change if the values in `W` are mutated).
 
-This implicit style is the one presently used by [Flux.jl](https://github.com/FluxML/Flux.jl), a closely related machine learning library. It uses structs like `Linear` above to define layers, and the function `Flux.params(model)` returns a `Params` object containing all the parameters of all layers. See [its documentation](https://fluxml.ai/Flux.jl/stable/models/basics/) for more details. When using Zygote for most other purposes, however, the explicit style is usually preferred.
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -186,3 +186,5 @@ julia> grads[W], grads[b] # access gradients using arrays as keys
 ```
 
 Here `grads` is a dictionary-like object, whose keys are the same parameters we indicated in `Params`. (In fact it wraps a dictionary using `objectid(W)` as keys, which does not change if the values in `W` are mutated).
+
+This implicit style is the one presently used by [Flux.jl](https://github.com/FluxML/Flux.jl), a closely related machine learning library. It uses structs like `Linear` above to define layers, and the function `Flux.params(model)` returns a `Params` object containing all the parameters of all layers. See [its documentation](https://fluxml.ai/Flux.jl/stable/models/basics/) for more details. When using Zygote for most other purposes, however, the explicit style is usually preferred.
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -131,6 +131,8 @@ julia> gradient(colordiff, RGB(1, 0, 0), RGB(0, 1, 0))
 
 ## Gradients of ML models
 
+### Explicit parameters
+
 It's easy to work with even very large and complex models, and there are few ways to do this. Autograd-style models pass around a collection of weights.
 
 ```julia
@@ -170,7 +172,9 @@ julia> dmodel = gradient(model -> sum(model(x)), model)[1]
 (W = [0.652543 … 0.683588], b = [1.0, 1.0])
 ```
 
-Zygote also support one more way to take gradients, via *implicit parameters* – this is a lot like autograd-style gradients, except we don't have to thread the parameter collection through all our code.
+### Implicit parameters
+
+Zygote also support one more way to take gradients, via *implicit parameters* – this is a lot like autograd-style gradients, except we don't have to thread the parameter collection through all our code. When working with Flux models, this is the recommended way of passing the gradients, as it ensures compatibility with Flux's built-in optimizers. 
 
 ```julia
 julia> W = rand(2, 5); b = rand(2);
@@ -181,8 +185,20 @@ linear (generic function with 2 methods)
 julia> grads = gradient(() -> sum(linear(x)), Params([W, b]))
 Grads(...)
 
+# Apply gradients to model parameters
 julia> grads[W], grads[b]
 ([0.652543 … 0.683588], [1.0, 1.0])
+```
+Unlike with explicit gradients, in order to see implicit gradients one needs to do:
+
+```julia
+julia> grads.grads
+IdDict{Any, Any} with 5 entries:
+  [0.467471 0.597815 … 0.678126 … => [0.579671 0.215381 … 0.635058 0.623832; 0.579671 0.215381 … …
+  :(Main.x)                       => [1.3377, 0.930234, 0.499161, 1.33827, 1.37791]
+  :(Main.W)                       => [0.579671 0.215381 … 0.635058 0.623832; 0.579671 0.215381 … …
+  [0.106308, 0.705531]            => 2-element Fill{Float64}: entries equal to 1.0
+  :(Main.b)                       => 2-element Fill{Float64}: entries equal to 1.0
 ```
 
 However, implicit parameters exist mainly for compatibility with Flux's current AD; it's recommended to use the other approaches unless you need this.
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -153,7 +153,9 @@ Dict{Any,Any} with 2 entries:
   :W => [0.628998 … 0.433006]
 ```
 
-An extension of this is the Flux-style model in which we use call overloading to combine the weight object with the pullback pass (equivalent to a closure).
+We can combine the role of the dictionary and the function here by making a callable struct which
+contains the parameters, equivalent to a closure. Passed explicitly to `gradient`, we get a named tuple
+with the same field names:
 
 ```julia
 julia> struct Linear
@@ -170,7 +172,7 @@ julia> dmodel = gradient(model -> sum(model(x)), model)[1]
 (W = [0.652543 … 0.683588], b = [1.0, 1.0])
 ```
 
-Zygote also support another way to take gradients, via *implicit parameters*. Here the loss function takes zero arguments, but the variables of interest are indicated by a special `Params` object. The function `linear` which depends on `W` and `b` is executed when the loss function `() -> sum(linear(x))` is called, and hence this dependence is visible to Zygote:
+Zygote also supports another way to take gradients, via *implicit parameters*. Here the loss function takes zero arguments, but the variables of interest are indicated by a special `Params` object. The function `linear` which depends on `W` and `b` is executed when the loss function `() -> sum(linear(x))` is called, and hence this dependence is visible to Zygote:
 
 ```julia
 julia> W = rand(2, 5); b = rand(2);
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -180,18 +180,13 @@ linear (generic function with 2 methods)
 
 julia> grads = gradient(() -> sum(linear(x)), Params([W, b]))
 Grads(...)
-```
-To inspect the `Grads(...)` object returned for implicit parameters, you can access it using the parameters passed to `Params`:
 
-```julia
-# Apply gradients to model parameters
-julia> grads[W], grads[b]
+julia> grads[W], grads[b] # access gradients using arrays as keys
 ([0.652543 … 0.683588], [1.0, 1.0])
 ```
 
 Here `grads` is a dictionary-like object, whose keys are the same parameters we 
-indicated in `Params`. (In fact it contains a dictionary using `objectid(W)`, which 
-does not change if the values in `W` are mutated.) These parameters `W, b` are global 
-variables, but gradients with respect to other global variables are not stored.
+indicated in `Params`. (In fact it wraps a dictionary using `objectid(W)` as keys, which 
+does not change if the values in `W` are mutated).
 
 However, implicit parameters exist mainly for compatibility with Flux's current AD; it's recommended to use the other approaches unless you need this.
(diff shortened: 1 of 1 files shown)


docs/src/index.md
```diff
@@ -181,8 +185,20 @@ linear (generic function with 2 methods)
 julia> grads = gradient(() -> sum(linear(x)), Params([W, b]))
 Grads(...)
 
+# Apply gradients to model parameters
 julia> grads[W], grads[b]
 ([0.652543 … 0.683588], [1.0, 1.0])
 ```
+Unlike with explicit gradients, in order to see implicit gradients one needs to do:
```
DhairyaLGandhi, line 192: We can index `Grads` directly. Why would you say one needs to access an internal field? That may break in future versions.

DhairyaLGandhi:
Thanks for the PR! Always a treat to have clearer docstrings. Out of curiosity, why do we need to show an internal field?

JordiBolibar:
It's just for new users (like me) to be able to show implicit gradients as it can be done with explicit. I've been struggling with gradient structures and sizes. This is the result of [this Discourse discussion](https://discourse.julialang.org/t/unrecognized-gradient-using-zygote-for-ad-with-universal-differential-equations/63791/8). 

If you think there's a better wait to do it I would be happy to update my PR. The documentation is pretty minimalistic right now, it's kinda hard to get into it for new users. So I'm planning to update the docs as I "discover" new undocumented stuff that I've struggled with.

docs/src/index.md
```diff
@@ -181,8 +185,20 @@ linear (generic function with 2 methods)
...
+
+```julia
+julia> grads.grads
+IdDict{Any, Any} with 5 entries:
+  [0.467471 0.597815 … 0.678126 … => [0.579671 0.215381 … 0.635058 0.623832; 0.579671 0.215381 … …
+  :(Main.x)                       => [1.3377, 0.930234, 0.499161, 1.33827, 1.37791]
+  :(Main.W)                       => [0.579671 0.215381 … 0.635058 0.623832; 0.579671 0.215381 … …
+  [0.106308, 0.705531]            => 2-element Fill{Float64}: entries equal to 1.0
+  :(Main.b)                       => 2-element Fill{Float64}: entries equal to 1.0
+```
```
darsnack, line 202: We don't want to point people to `grads.grads` cause it contains extraneous info (it's quick for debugging, but not recommended). The suggestion below is the recommended way.

Can you add the printed out text at the prompt?

```suggestion
To inspect the `Grads(...)` object returned for implicit parameters, you can index it using the parameters passed to `Params`:

```julia
julia> [grads[p] for p in [W, b]]
```
```

File: docs/src/index.md [hunk 3]
```diff
@@ -181,8 +185,20 @@ linear (generic function with 2 methods)
 julia> grads = gradient(() -> sum(linear(x)), Params([W, b]))
 Grads(...)
 
+# Apply gradients to model parameters
 julia> grads[W], grads[b]
 ([0.652543 … 0.683588], [1.0, 1.0])
```
darsnack, line 190: Deleting in favor of the more descriptive example below (also note that `grads[W]` does not apply the gradient, it is just how we index `Grads`).

```suggestion
```

darsnack (changes requested):
Thanks for this! A couple quick changes and we should be good to merge.

docs/src/index.md
JordiBolibar: Cool, thanks for the clarification!

mcabbott: This seems much more direct than `[grads[p] for p in [W, b]]`, why remove it?

Also, I'm not sure we should describe this as indexing. It's looking something up in a dictionary, whose keys are the arrays themselves (or their IDs). Of course I know it's `Base.getindex` but that's a detail.

docs/src/index.md
```diff
@@ -181,8 +185,13 @@ linear (generic function with 2 methods)
...
 ```
+To inspect the `Grads(...)` object returned for implicit parameters, you can index it using the parameters passed to `Params`:
+
+```julia
+julia> [grads[p] for p in [W, b]]
```
darsnack, line 195:
```suggestion
```

To inspect the `Grads(...)` object returned for implicit parameters, you can index it using the parameters passed to `Params`:
```julia
julia> grads[W], grads[b]
([0.652543 … 0.683588], [1.0, 1.0])
```
```

File: docs/src/index.md (hunk 3 above)
Comment (darsnack) [reply docs/src/index.md:190]: You're right. I added a new suggestion that keeps the same clarifications as the other example but the less opaque code from this one.

File: docs/src/index.md [hunk 5]
```diff
@@ -131,6 +131,8 @@ julia> gradient(colordiff, RGB(1, 0, 0), RGB(0, 1, 0))
 
 ## Gradients of ML models
 
+### Explicit parameters
```
mcabbott, line 134: Do we really need 3 levels of headings? The Zygote docs are not very long, and elsewhere 2 is enough. 

If some section needs to be broken into two, that can be done. But this section seems fairly coherent to me, it describes two ways to deal with gradients of the same example.

Is the problem just that "implicit" isn't in a tile? Should the section title be something like "## Explicit and implicit parameters for ML models"?

docs/src/index.md
darsnack: Ooof Github does not like all the backticks here. Hopefully, you get the idea behind the suggestion.

docs/src/index.md
ToucheSir: Yeah, the subheadings are redundant if "implicit" and "explicit" are already mentioned in the text.

darsnack:
I am copy and pasting the full section here because Github is butchering the code suggestions in the review tool.

```md
Zygote also support one more way to take gradients, via *implicit parameters* – this is a lot like autograd-style gradients, except we don't have to thread the parameter collection through all our code. When working with Flux models, this is the recommended way of passing the gradients, as it ensures compatibility with Flux's built-in optimizers. 

```julia
julia> W = rand(2, 5); b = rand(2);

julia> linear(x) = W * x .+ b
linear (generic function with 2 methods)

julia> grads = gradient(() -> sum(linear(x)), Params([W, b]))
Grads(...)
```

To inspect the `Grads(...)` object returned for implicit parameters, you can access it using the parameters passed to `Params`:
```julia
julia> grads[W], grads[b]
([0.652543 … 0.683588], [1.0, 1.0])
```
`Grads(...)` is like a dictionary whose keys are the parameters.
```

**Comment by ToucheSir** [comment]
Just putting a note here for myself to add a pass for spelling and grammar once the content and structure have stabilized.

File: docs/src/index.md (hunk 5 above)
Comment (JordiBolibar) [reply docs/src/index.md:134]: Good point, to me it is a matter of clearly conveying the message that there are 2 main ways of passing parameters. I have nothing for or against 2 vs 3 levels of headings, so I can easily add both types of parameters in the main title as you suggested. When you're browsing through a ton of docs it is helpful to have the relevant information in the headings. 

All these comments are from a perspective of a new user coming without any previous knowledge, so my view on this will certainly differ from yours.

File: docs/src/index.md (hunk 4 above)
Comment (mcabbott) [reply docs/src/index.md:195]: I don't know how you tricked it to accept the backticks, but well done. 

I'm still not sure we should break this code block to insert the sentence. The others are kept together, and it means you can copy-paste them whole to the REPL. Can't whatever sentence of clarification come afterwards, like
```
julia> grads[W], grads[b]
([0.652543 … 0.683588], [1.0, 1.0])
` ` `
Here `grads` is a dictionary-like object, whose keys are the same parameters we 
indicated in `Params`. These parameters `W, b` are global variables, but gradients 
with respect to other global variables are not stored. 

However, implicit parameters exist mainly for compatibility with Flux...
```

File: docs/src/index.md (hunk 5 above)
Comment (ToucheSir) [reply docs/src/index.md:134]: Of course, I think @mcabbott's point was that this page is _so_ short that additional headings have a marginal benefit at best. Perhaps it would be better to have the first sentence of each paragraph use the keyword explicit/implicit and italicize it like how implicit is in the current version.

File: docs/src/index.md (hunk 4 above)
Comment (darsnack) [reply docs/src/index.md:195]: Yeah, let's keep the example together and move the text to after. I am updating my comment below.

Also, I discovered that you need to add extra backticks to the outer most block (so quandruple-backtick block instead of triple-backtick).

File: docs/src/index.md (hunk 5 above)
Comment (mcabbott) [reply docs/src/index.md]: Yes I'm sorry if the nit-picking comes off wrong. It's super-valuable to have feedback as to what was hard to understand, and this can be hard for doc-writers to figure out since they are the ones who do already understand. 

The alchemy is in finding ways to address what the problems were, without upsetting what was good before. There's some value to being concise, and to keeping the structure simple.

**Comment by JordiBolibar** [comment]
I have updated the PR following your suggestions. I hope I managed to capture most of the discussed stuff, as it is little bit chaotic to follow.

File: docs/src/index.md [hunk 6]
```diff
@@ -170,7 +172,9 @@ julia> dmodel = gradient(model -> sum(model(x)), model)[1]
 (W = [0.652543 … 0.683588], b = [1.0, 1.0])
 ```
 
-Zygote also support one more way to take gradients, via *implicit parameters* – this is a lot like autograd-style gradients, except we don't have to thread the parameter collection through all our code.
+### Implicit parameters
+
+Zygote also support one more way to take gradients, via *implicit parameters* – this is a lot like autograd-style gradients, except we don't have to thread the parameter collection through all our code. When working with Flux models, this is the recommended way of passing the gradients, as it ensures compatibility with Flux's built-in optimizers.
```
ToucheSir, line 173:
```suggestion
Zygote also supports taking gradients via *implicit parameters*. This is a lot like autograd-style gradients, except we don't have to thread the parameter collection through our code.
```
We're moving away from using implicit parameters in general, and the next iteration of Flux's optimizers are designed for explicit parameters/gradients first. I'm also leery of referencing a downstream package, especially when we already have similar documentation in https://fluxml.ai/Flux.jl/stable/models/basics/#Taking-Gradients and https://fluxml.ai/Flux.jl/stable/training/optimisers/. Perhaps an alternative would be to link

docs/src/index.md
darsnack: I am not sure the last part about globals is correct (if I understood the sentence properly). `grads.grads` will contain extra gradients including globals that are not part of `Params`.

docs/src/index.md
```diff
@@ -129,9 +129,9 @@ julia> gradient(colordiff, RGB(1, 0, 0), RGB(0, 1, 0))
 ((r = 0.4590887719632896, g = -9.598786801605689, b = 14.181383399012862), (r = -1.7697549557037275, g = 28.88472330558805, b = -0.044793892637761346))
 ```
 
-## Gradients of ML models
+## Explicit and implicit parameters of ML models
```
darsnack, line 132:
```suggestion
## Gradients with explicit and implicit parameters for ML models
```

docs/src/index.md
darsnack: In the long term, I agree, but the whole point of this PR is to prevent the frustration of reading Zygote's docs, taking explicit gradients, and then getting an error on `update!`. Flux is already mentioned in the docs.

docs/src/index.md
```diff
@@ -129,9 +129,9 @@ julia> gradient(colordiff, RGB(1, 0, 0), RGB(0, 1, 0))
...
 
-It's easy to work with even very large and complex models, and there are few ways to do this. Autograd-style models pass around a collection of weights.
+It's easy to work with even very large and complex models, and there are few ways to do this. Autograd-style models pass around a collection of weights. There are two ways of passing *explicit* parameters:
```
mcabbott, line 134: I think it should be "parameters for ML models" not "of".

But more importantly, what two ways are you seeing? In my thinking, `gradient(f, x)` which evaluates `f(x)` is one way, whether `x` is a number, a vector, a tuple, a dictionary, a struct -- the point is that it's an argument of the function.

docs/src/index.md
```diff
@@ -180,9 +180,18 @@ linear (generic function with 2 methods)
...
+# Apply gradients to model parameters
 julia> grads[W], grads[b]
 ([0.652543 … 0.683588], [1.0, 1.0])
 ```
 
+Here `grads` is a dictionary-like object, whose keys are the same parameters we 
+indicated in `Params`. (In fact it contains a dictionary using `objectid(W)`, which 
+does not change if the values in `W` are mutated.) These parameters `W, b` are global 
+variables, but gradients with respect to other global variables are not stored.
+
```
darsnack, line 196: Sorry this one got confusing. This is where I think we all landed.

````suggestion

julia> grads[W], grads[b] # access gradients using arrays as keys
([0.652543 … 0.683588], [1.0, 1.0])
```

Here `grads` is a dictionary-like object, whose keys are the same parameters we 
indicated in `Params`. (In fact it wraps a dictionary using `objectid(W)` as keys, which 
does not change if the values in `W` are mutated).

````

**Review by darsnack — COMMENTED** [review]
Thanks for bearing with us!

File: docs/src/index.md [hunk 10]
```diff
@@ -180,9 +180,18 @@ linear (generic function with 2 methods)
 
 julia> grads = gradient(() -> sum(linear(x)), Params([W, b]))
 Grads(...)
+```
+To inspect the `Grads(...)` object returned for implicit parameters, you can access it using the parameters passed to `Params`:
 
+```julia
+# Apply gradients to model parameters
 julia> grads[W], grads[b]
```
mcabbott, line 188: I can't make a suggestion block, but I think this "To inspect" should be removed, to restore one continuous code block which can be copy-pasted. The description of what grads is is covered below.

I also think "Apply gradients" sounds wrong, sounds like this is performing gradient descent or something. These are simply extracting elements of this container. I'm not sure it needs more description than what's said just below it, but if you think it does, can we find better words?

docs/src/index.md
mcabbott: I also confess that I don't really know what "Autograd-style" means, exactly. Does this mean Tracker.jl? Some python package everyone was familiar with back in the day?

docs/src/index.md
darsnack: Are you okay with my suggestion block above? (just to keep this less confusing for @JordiBolibar)

docs/src/index.md
darsnack: I assume PyTorch's AD called autograd.

docs/src/index.md
mcabbott: Yes I think so!

Sorry this is probably a maximally confusing review, we all seem to be typing in parallel...

I mark this resolved.

docs/src/index.md
mcabbott: Oh nice, good to know. 

And, perhaps clearer not to mention globals.

docs/src/index.md
ToucheSir: Why are people reading Zygote's docs instead of Flux's docs if they want to use functionality in Flux though? Seems like a rare edge case to only read the Zygote docs and the Flux optimizer docs without reading the Flux docs on taking gradients as well. I think updating the optimizer docs to add an example with `θ` would be a better fix than trying to shoehorn it in Zygote.

docs/src/index.md
mcabbott: I wonder if this is going to wrap to like 5 lines? 

Screenshot 2021-07-02 at 13 32 33

Maybe it should be something like `Explicit and implicit parameters`?

docs/src/index.md
darsnack: I think it is confusing that two packages that are closely tied in practice (even if in reality Zygote exists independently of Flux) have contradictory statements. Flux explicitly encourages implicit parameters, and it makes no mention of the complex structure that results from `gradient(m -> loss(m(x), y), m)`. Zygote explicitly discourages implicit parameters. And the wording makes it sound like implicit and explicit work in Flux, but implicit is the "backwards compatible" version. But only certain varieties of explicit which are cumbersome to deal with for real models actually do work in Flux. The more convenient syntax with `m` will not.

docs/src/index.md
darsnack: Yeah, I agree

docs/src/index.md
ToucheSir: > And the wording makes it sound like implicit and explicit work in Flux, but implicit is the "backwards compatible" version

I haven't seen this verbiage anywhere, is it elsewhere in the Zygote docs? I certainly don't think putting implicit params second implies that their use is discouraged, but rather that it flows naturally along the gradient of easier -> more complex ideas. Note that the Flux docs use the same order, but actually endorse using implicit params for models with a lot of parameters (so pretty much all of them).

What I am against is explicitly and specifically calling out implicit params as the recommended way of doing things in the _Zygote_ docs. It's fine if we do so in the Flux docs (and we do), or if the language is neutral here (which IMO it is just after that sentence is removed).

darsnack: > However, implicit parameters exist mainly for compatibility with Flux's current AD; it's recommended to use the other approaches unless you need this.

In the last sentence of https://fluxml.ai/Zygote.jl/dev/#Gradients-of-ML-models-1

ToucheSir: I missed that, lets nix it in this PR or a follow-up.

mcabbott: Maybe the solution is to beef up the last paragraph of this section, it's only a few sentences down. "implicit parameters exist mainly for compatibility with Flux's current AD;" sounds like it's a very temporary disclaimer, and "Flux-style model" makes it sounds like actual Flux works that way.

Maybe we delete "Flux-style model " and make the last paragraph something like this:

This implicit style is the one presently used by Flux, a closely related machine learning library. It uses structs like `Linear` above to define layers, and the function `Flux.params(model)` returns a `Params` object containing all the many parameters of its nested layers. For most other purposes, however, the explicit style is usually preferred.

darsnack: I like that!

mcabbott: What the words still don't spell out is that "implicit parameters" means that the parameters are not arguments of the loss function. It takes no arguments, and picks up global (or, fairly global...) variables. 

I think you're supposed to conclude this from `() -> sum(linear(x))` and from "thread the parameter collection through" but I also think I was confused by that for a while. Maybe this paragraph/sentence should lay that out a bit more, like "Differentiate a zero-argument function, which closes over the parameters, and indicate that you want this behaviour by providing a Params object which exists for this purpose". Something like that?

docs/src/index.md
```diff
@@ -129,9 +129,9 @@ julia> gradient(colordiff, RGB(1, 0, 0), RGB(0, 1, 0))
 ((r = 0.4590887719632896, g = -9.598786801605689, b = 14.181383399012862), (r = -1.7697549557037275, g = 28.88472330558805, b = -0.044793892637761346))
 ```
 
-## Gradients of ML models
+## Explicit and Implicit Parameters
 
-It's easy to work with even very large and complex models, and there are few ways to do this. Autograd-style models pass around a collection of weights.
+It's easy to work with even very large and complex models, and there are few ways to do this. Autograd-style models pass around a collection of weights. There are two ways of passing *explicit* parameters:
```
darsnack, line 134:
```suggestion
It's easy to work with even very large and complex models, and there are few ways to do this. Autograd-style models pass around a collection of weights. Depending on how you write your model, there are multiple ways to *explicity* take gradients with respect to parameters. For example, the function `linear` accepts the parameters as an argument to the model. So, we directly pass in the parameters, `θ`, as an argument to the function being differentiated.
```

darsnack: For some reason, I cannot comment on lines that aren't changed by the PR (which is silly). But here is my suggestion for L156 that pairs with this one.

```
An extension of this is the Flux-style model in which we use call overloading to combine the weight object with the pullback pass (equivalent to a closure). Here, we take the gradient with respect to `model` which is an instance of the `Linear` type. Zygote returns a `NamedTuple` of gradients which shares the same structure as `Linear` (the keys of the tuple correspond to the fields of the struct).

File: docs/src/index.md (hunk 6 above)
Comment (darsnack) [reply docs/src/index.md:173]: @JordiBolibar I believe we still want to remove this added sentence (in favor of @mcabbott's paragraph).

**Review by darsnack — CHANGES_REQUESTED** [review]
Just a few more changes to address the concern about "two explicit ways."

File: docs/src/index.md (hunk 6 above)
Comment (JordiBolibar) [reply docs/src/index.md]: I have made a small commit replacing it with @mcabbott 's paragraph, adding a short sentence before to make it flow better after the explicit section.

File: docs/src/index.md [hunk 12]
```diff
@@ -170,7 +170,7 @@ julia> dmodel = gradient(model -> sum(model(x)), model)[1]
 (W = [0.652543 … 0.683588], b = [1.0, 1.0])
 ```
 
-Zygote also support one more way to take gradients, via *implicit parameters* – this is a lot like autograd-style gradients, except we don't have to thread the parameter collection through all our code.
+On the other hand, the *implicit* style is the one presently used by [Flux.jl](https://github.com/FluxML/Flux.jl), a closely related machine learning library. It uses structs like `Linear` above to define layers, and the function `Flux.params(model)` returns a `Params` object containing all the parameters of all layers. See [its documentation](https://fluxml.ai/Flux.jl/stable/models/basics/) for more details. When using Zygote for most other purposes, however, the explicit style is usually preferred.
```
mcabbott, line 173: I still think the paragraph describing what Flux does should go at the end. The flow is first simple explicit things, then Zygote's implicit things, then comments on other packages. 

What wants to be described here is how this differs from explicit & why it's called that. My attempt:

```suggestion
Zygote also support another way to take gradients, via *implicit parameters*. Here the loss function takes zero arguments, but the variables of interest are indicated by a special `Params` object. The function `linear` which depends on `W` and `b` is executed when the loss function `() -> sum(linear(x))` is called, and hence this dependence is visible to Zygote:
```
What also needs fixing is the comment on "the Flux-style model" here: 
https://github.com/FluxML/Zygote.jl/pull/1016/files#diff-36c852523b6a7513a10b8a6bae21f1037eaa27484efc9042f75bbda46d9f529bR156
which is confusing -- it has structs like Flux does, but isn't using implicit stuff.

docs/src/index.md
```diff
@@ -181,8 +181,8 @@ linear (generic function with 2 methods)
 julia> grads = gradient(() -> sum(linear(x)), Params([W, b]))
 Grads(...)
 
-julia> grads[W], grads[b]
+julia> grads[W], grads[b] # access gradients using arrays as keys
 ([0.652543 … 0.683588], [1.0, 1.0])
 ```
 
-However, implicit parameters exist mainly for compatibility with Flux's current AD; it's recommended to use the other approaches unless you need this.
+Here `grads` is a dictionary-like object, whose keys are the same parameters we indicated in `Params`. (In fact it wraps a dictionary using `objectid(W)` as keys, which does not change if the values in `W` are mutated).
```
mcabbott, line 188: This is the paragraph which is migrating downwards again:
```suggestion
Here `grads` is a dictionary-like object, whose keys are the same parameters we indicated in `Params`. (In fact it wraps a dictionary using `objectid(W)` as keys, which does not change if the values in `W` are mutated).

This implicit style is the one presently used by [Flux.jl](https://github.com/FluxML/Flux.jl), a closely related machine learning library. It uses structs like `Linear` above to define layers, and the function `Flux.params(model)` returns a `Params` object containing all the parameters of all layers. See [its documentation](https://fluxml.ai/Flux.jl/stable/models/basics/) for more details. When using Zygote for most other purposes, however, the explicit style is usually preferred.
```

Also todo, I think it should not say "Flux-style model" above. 

And I wonder whether having (and referring to) two different functions called `linear` is such a bright idea. Maybe the 2nd one should be `linearWb(x)`?

mcabbott (approved):
This needs one more thing, and then should be merged. I can't make a suggestion box, but I think this paragraph:
```
An extension of this is the Flux-style model in which we use call overloading to combine the weight object with the pullback pass (equivalent to a closure).
```
should avoid saying Flux, and become something like 
```
We can also combine the role of the dictionary and the function by making a callable struct which
contains the parameters, equivalent to a closure. Passed explicitly to `gradient`, we get a named tuple
with the same field names:
```

mcabbott:
Thanks @JordiBolibar !

JordiBolibar:
My pleasure! Thanks for taking my little PR as a chance to improve the docs!

Issue #961, opened 2021-05-03
`abs2` of complex CUDA array fails with `Zygote.gradient` 


roflmaostc:
Hey,

when applying `abs2` to a complex CUDA array I get an `ERROR: MethodError: no method matching iterate(::Nothing)`. 
I'm using CUDA 3.1.0, Julia 1.6.1 and Zygote 0.6.10.
But I also tried it on Julia 1.5.4, CUDA v2.4.0, Zygote v0.5.0 so it must be not a recent introduced issue.

See the MWE below:

```julia
julia> using Zygote, CUDA

julia> x = rand(ComplexF32, (2,2))
2×2 Matrix{ComplexF32}:
 0.0598111+0.678913im  0.767138+0.77825im
  0.548067+0.98656im   0.306103+0.166084im

julia> x_c = CuArray(x);

julia> f(x) = sum(abs2.(x))
f (generic function with 1 method)

julia> g(x) = sum(real(x .* conj.(x)))
g (generic function with 1 method)

julia> f(x) ≈ f(x_c) ≈ g(x) ≈ g(x_c)
true

julia> Zygote.gradient(f, x)
(ComplexF32[0.11962223f0 + 1.3578255f0im 1.534275f0 + 1.5565007f0im; 1.0961342f0 + 1.9731205f0im 0.61220574f0 + 0.33216715f0im],)

julia> Zygote.gradient(g, x)
(ComplexF32[0.11962223f0 + 1.3578255f0im 1.534275f0 + 1.5565007f0im; 1.0961342f0 + 1.9731205f0im 0.61220574f0 + 0.33216715f0im],)

julia> Zygote.gradient(f, x_c)
ERROR: MethodError: no method matching iterate(::Nothing)
Closest candidates are:
  iterate(::Union{LinRange, StepRangeLen}) at range.jl:664
  iterate(::Union{LinRange, StepRangeLen}, ::Int64) at range.jl:664
  iterate(::T) where T<:Union{Base.KeySet{var"#s79", var"#s78"} where {var"#s79", var"#s78"<:Dict}, Base.ValueIterator{var"#s77"} where var"#s77"<:Dict} at dict.jl:693
  ...
Stacktrace:
  [1] (::Zygote.var"#1209#1210"{Zygote.var"#1104#1108"})(ȳ::CuArray{Float32, 2})
    @ Zygote ~/.julia/packages/Zygote/6HN9x/src/lib/broadcast.jl:231
[...]

PR #1018, opened 2021-07-04
Tidy up a few things in broadcasting

### LINKED ISSUE #1027: Error (no method matching iterate(::Nothing)) when counting non-zero elements in loss function
The following code fails in Flux v0.12.4 with the error message:
```
caused by: MethodError: no method matching iterate(::Nothing)
Closest candidates are:
```

```julia
using Flux, Zygote
m = Chain(Dense(10,1)) |> gpu
b = (rand(10,10), rand(1,10)) |> gpu
# slightly contrived example :-)
l(x,y) = sum( (m(x) - y).^2 ) / sum(y .!= 0)
grad = Flux.gradient(Flux.params(m)) do
   l(b...)
end
```
The full error message is:
```
ERROR: ERROR: a exception was thrown during kernel execution.
       Run Julia on debug level 2 for device stack traces.
ERROR: a exception was thrown during kernel execution.
       Run Julia on debug level 2 for device stack traces.
ERROR: a exception was thrown during kernel execution.
       Run Julia on debug level 2 for device stack traces.
ERROR: a exception was thrown during kernel execution.
       Run Julia on debug level 2 for device stack traces.
ERROR: a exception was thrown during kernel execution.
       Run Julia on debug level 2 for device stack traces.
ERROR: a exception was thrown during kernel execution.
       Run Julia on debug level 2 for device stack traces.
ERROR: a exception was thrown during kernel execution.
       Run Julia on debug level 2 for device stack traces.
ERROR: a exception was thrown during kernel execution.
       Run Julia on debug level 2 for device stack traces.
ERROR: a exception was thrown during kernel execution.
       Run Julia on debug level 2 for device stack traces.
ERROR: a exception was thrown during kernel execution.
       Run Julia on debug level 2 for device stack traces.
KernelException: exception thrown during kernel execution on device NVIDIA GeForce RTX 3080
Stacktrace:
 [1] check_exceptions()
   @ CUDA ~/.julia/packages/CUDA/02Kjq/src/compiler/exceptions.jl:37
 [2] synchronize(stream::CUDA.CuStream; blocking::Bool)
   @ CUDA ~/.julia/packages/CUDA/02Kjq/lib/cudadrv/stream.jl:143
 [3] synchronize (repeats 2 times)
   @ ~/.julia/packages/CUDA/02Kjq/lib/cudadrv/stream.jl:117 [inlined]
 [4] top-level scope
   @ ~/.julia/packages/CUDA/02Kjq/src/initialization.jl:54

caused by: MethodError: no method matching iterate(::Nothing)
Closest candidates are:
  iterate(::Union{LinRange, StepRangeLen}) at range.jl:664
  iterate(::Union{LinRange, StepRangeLen}, ::Int64) at range.jl:664
  iterate(::T) where T<:Union{Base.KeySet{var"#s79", var"#s78"} where {var"#s79", var"#s78"<:Dict}, Base.ValueIterator{var"#s77"} where var"#s77"<:Dict} at dict.jl:693
  ...
Stacktrace:
  [1] (::Zygote.var"#1253#1254"{Zygote.var"#1121#1125"})(ȳ::CUDA.CuArray{Bool, 2})
    @ Zygote ~/.julia/packages/Zygote/0da6K/src/lib/broadcast.jl:246
  [2] (::Zygote.var"#558#back#1255"{Zygote.var"#1253#1254"{Zygote.var"#1121#1125"}})(Δ::CUDA.CuArray{Bool, 2})
    @ Zygote ~/.julia/packages/ZygoteRules/OjfTt/src/adjoint.jl:59
  [3] (::Zygote.var"#193#194"{Tuple{NTuple{4, Nothing}, Tuple{}}, Zygote.var"#558#back#1255"{Zygote.var"#1253#1254"{Zygote.var"#1121#1125"}}})(Δ::CUDA.CuArray{Bool, 2})
    @ Zygote ~/.julia/packages/Zygote/0da6K/src/lib/lib.jl:203
  [4] (::Zygote.var"#1714#back#195"{Zygote.var"#193#194"{Tuple{NTuple{4, Nothing}, Tuple{}}, Zygote.var"#558#back#1255"{Zygote.var"#1253#1254"{Zygote.var"#1121#1125"}}}})(Δ::CUDA.CuArray{Bool, 2})
    @ Zygote ~/.julia/packages/ZygoteRules/OjfTt/src/adjoint.jl:59
  [5] Pullback
    @ ./broadcast.jl:1315 [inlined]
  [6] Pullback
    @ ./REPL[51]:1 [inlined]
  [7] (::typeof(∂(l)))(Δ::Float32)
    @ Zygote ~/.julia/packages/Zygote/0da6K/src/compiler/interface2.jl:0
  [8] (::Zygote.var"#193#194"{Tuple{Tuple{Nothing, Nothing}}, typeof(∂(l))})(Δ::Float32)
    @ Zygote ~/.julia/packages/Zygote/0da6K/src/lib/lib.jl:203
  [9] (::Zygote.var"#1714#back#195"{Zygote.var"#193#194"{Tuple{Tuple{Nothing, Nothing}}, typeof(∂(l))}})(Δ::Float32)
    @ Zygote ~/.julia/packages/ZygoteRules/OjfTt/src/adjoint.jl:59
 [10] Pullback
    @ ./REPL[52]:2 [inlined]
 [11] (::typeof(∂(#23)))(Δ::Float32)
    @ Zygote ~/.julia/packages/Zygote/0da6K/src/compiler/interface2.jl:0
 [12] (::Zygote.var"#74#75"{Params, typeof(∂(#23)), Zygote.Context})(Δ::Float32)
    @ Zygote ~/.julia/packages/Zygote/0da6K/src/compiler/interface.jl:255
 [13] gradient(f::Function, args::Params)
    @ Zygote ~/.julia/packages/Zygote/0da6K/src/compiler/interface.jl:59
 [14] top-level scope
    @ REPL[52]:1
 [15] top-level scope
    @ ~/.julia/packages/CUDA/02Kjq/src/initialization.jl:52
```

However, if the `Zygote.dropgrad` function is used for `sum(y .!= 0)`, the gradient calculation does no longer give an error.
Apparently, Flux also computes the gradient relative to `y` which is surprising as `y` does not depend on the weights to be optimized.

```julia
using Flux, Zygote
m = Chain(Dense(10,1)) |> gpu
b = (rand(10,10), rand(1,10)) |> gpu
l(x,y) = sum( (m(x) - y).^2 ) / Zygote.dropgrad(sum(y .!= 0))
grad = Flux.gradient(Flux.params(m)) do
   l(b...)
end
```
Maybe I am not using the `Flux.gradient` function properly.

I am using Julia 1.6.1 (official binaries), Zygote v0.6.14, Flux v0.12.4 and CUDA v3.3.3 on Linux.

mcabbott:
This removes some functions like `z, Δ -> (nothing, back(Δ)...)` by calling `_pullback` immediately, or  by altering `broadcast_forward` (which is also tidied up a bit). Closes #1027 which was a bug in that function.

~~It also deletes the special case of broadcasting for CuArrays.~~ [now removed] Broadcasting over real numbers, with a pure enough `f`, now always uses `broadcast_forward`, thus should work on CuArrays too. Broadcasting over complex numbers will give an error instead of a wrong answer, seen in #961. Broadcasting a closure, or a stateful function, ditto. However, I'm less than 100% confident that there aren't some functions which currently do give correct answers, but will now give an error. Can anyone see how to construct one?

Note BTW buildkite Julia 1.5 tests CUDA 2 so misses this change, and Julia 1.6 seems broken before this change.

### DIFF
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -253,10 +253,10 @@ end
     @eval @adjoint broadcasted(::CuArrayStyle, f, args...) =
       broadcast_forward(CUDA.cufunc(f), args...)
 
-  # else CUDA >= 3.0 -- don't need cufunc(f), and ordinary broadcasting calls broadcast_forward when safe
+  else CUDA >= 3.0 -- don't need cufunc(f), and ordinary broadcasting calls broadcast_forward when safe
 
-    # @eval @adjoint function broadcasted(::CuArrayStyle, f, args...) =
-    #   broadcast_forward(f, args...)
+    @eval @adjoint function broadcasted(::CuArrayStyle, f, args...) =
+      broadcast_forward(f, args...)
 
   end
 
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -255,7 +255,7 @@ end
 
   else # CUDA >= 3.0 -- don't need cufunc(f), and ordinary broadcasting calls broadcast_forward when safe
 
-    @eval @adjoint function broadcasted(::CuArrayStyle, f, args...) =
+    @eval @adjoint broadcasted(::CuArrayStyle, f, args...) =
       broadcast_forward(f, args...)
 
   end
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -75,23 +75,17 @@ unbroadcast(x::AbstractArray, x̄::Nothing) = nothing
 
 @adjoint broadcasted(::typeof(*), x::Numeric, y::Numeric) = x.*y,
    Δ -> (nothing, unbroadcast(x, Δ .* conj.(y)), unbroadcast(y, Δ .* conj.(x)))
-@adjoint function broadcasted(::typeof(*), x::Number, y::AbstractArray{<:Number})
-  z, back = pullback(*, x, y)  # this uses dot(y,Δ) instead of Δ .* conj.(y)
-  z, Δ -> (nothing, back(Δ)...)
-end
-@adjoint function broadcasted(::typeof(*), x::AbstractArray{<:Number}, y::Number)
-  z, back = pullback(*, x, y)
-  z, Δ -> (nothing, back(Δ)...)
-end
+@adjoint broadcasted(::typeof(*), x::Number, y::AbstractArray{<:Number}) =
+  _pullback(*, x, y)  # this uses dot(y,Δ) instead of sum(Δ .* conj.(y))
+@adjoint broadcasted(::typeof(*), x::AbstractArray{<:Number}, y::Number) = 
+  _pullback(*, x, y)
 
 @adjoint function broadcasted(::typeof(/), x::Numeric, y::Numeric)
   res = x ./ y
   res, Δ -> (nothing, unbroadcast(x, Δ ./ conj.(y)), unbroadcast(y, .-Δ .* conj.(res ./ y)))
 end
-@adjoint function broadcasted(::typeof(/), x::AbstractArray{<:Number}, y::Number)
-  z, back = pullback(/, x, y)
-  z, Δ -> (nothing, back(Δ)...)
-end
+@adjoint broadcasted(::typeof(/), x::AbstractArray{<:Number}, y::Number) =
+  _pullback(/, x, y)
 
 @adjoint function broadcasted(::typeof(Base.literal_pow), ::typeof(^), x::Numeric, exp::Val{p}) where p
   y = Base.literal_pow.(^, x, exp)
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -174,10 +174,9 @@ _dual_safearg(x) = false
   T = Broadcast.combine_eltypes(f, args)
   # Avoid generic broadcasting in two easy cases:
   if T == Bool
-    return f.(args...), _->nothing 
+    return (f.(args...), _ -> nothing) 
   elseif T <: Real && isconcretetype(T) && _dual_purefun(F) && all(_dual_safearg, args)
-    y, back = broadcast_forward(f, args...)
-    return y, ȳ -> (nothing, nothing, back(ȳ)...)
+    return broadcast_forward(f, args...)
   end
   len = inclen(args)
   y∂b = _broadcast((x...) -> _pullback(__context__, f, x...), args...)
@@ -189,7 +188,7 @@ _dual_safearg(x) = false
     end
     (nothing, accum_sum(dxs[1]), map(unbroadcast, args, Base.tail(dxs))...)
   end
-  y, ∇broadcasted
+  return y, ∇broadcasted
 end
 
 @adjoint function broadcasted(::AbstractArrayStyle{0}, f, args...)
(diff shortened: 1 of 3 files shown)


mcabbott:
Here's one example of something it gets silently wrong at present (which would become an error), and something which doesn't work (but might become a different error):
```
julia> using Zygote, CUDA
 
julia> gradient((x,y) -> sum((z->z^2+y[1]).(x)), [1,2,3], [4,5])  # correct
([2, 4, 6], [3, 0])

julia> gradient((x,y) -> sum((z->z^2+y[1]).(x)), cu([1,2,3]), cu([4,5]))
([2, 4, 6], nothing)

julia> mutable struct Mu{T}; x::T; end  # contains mutable state

julia> (m::Mu)(y) = (m.x = m.x * y)^2;

julia> gradient((x,ys) -> begin m = Mu(x); sum(map(m, ys)) end, 0.1, [1, 1/2, 1/3])  # correct if we trust map's order
(0.2555555555555556, [0.02555555555555556, 0.011111111111111113, 0.0016666666666666668])

julia> gradient((x,ys) -> begin m = Mu(x); sum(broadcast(m, ys)) end, 0.1, [1, 1/2, 1/3])  # wrong
(0.061111111111111116, [0.020000000000000004, 0.030000000000000006, 0.009166666666666668])

julia> gradient((x,ys) -> begin m = Mu(x); sum(broadcast(m, ys)) end, 0.1f0, cu([1, 1/2, 1/3]))
ERROR: GPU broadcast resulted in non-concrete element type Union{}.
This probably means that the function you are broadcasting contains an error or type instability.
Stacktrace:
...
  [4] broadcast_forward
    @ ~/.julia/packages/Zygote/0da6K/src/lib/broadcast.jl:227 [inlined]
```

test/cuda.jl
```diff
@@ -24,6 +24,11 @@ end
   g_gpu = gradient(x -> w(x), a_gpu)[1]
   @test g_gpu isa CuArray
   @test g_gpu |> collect ≈ g
+
+  # https://github.com/FluxML/Zygote.jl/issues/1027
+  @test gradient(x -> sum(x .!= 0), a_gpu) == (nothing,)
+  g3 = gradient(x -> sum(x .^ 3) ./ count(x .> 3), a)[1]
+  @test cu(g3) ≈ gradient(x -> sum(x .^ 3) ./ sum(x .> 3), a_gpu)[1]
```
mcabbott, line 31: This new test causes "ERROR: a exception was thrown during kernel execution." https://buildkite.com/julialang/zygote-dot-jl/builds/515

src/lib/broadcast.jl
```diff
@@ -237,28 +231,33 @@ function dual_function(f::F) where F
 end
 
 @inline function broadcast_forward(f, args::Vararg{Any,N}) where N
-  T = Broadcast.combine_eltypes(f, args)
+  valN = Val(N)
   out = dual_function(f).(args...)
   eltype(out) <: Dual || return (out, _ -> nothing)
```
mcabbott, line 236: `broadcast_forward` is mostly edited for clarity. The exception is that this shortcut `return (out, _ -> nothing)` didn't actually work before, as the caller would do what the line `(nothing, nothing, dargs...)` now does here, which isn't desired in this case.

src/lib/broadcast.jl
```diff
@@ -106,10 +100,10 @@ end
 end
 
 @adjoint broadcasted(::typeof(conj), x::Numeric) =
-  conj.(x), z̄ -> (nothing, conj.(z̄))
+  conj(x), z̄ -> (nothing, conj(z̄))
```
mcabbott, line 103: This ought to save a copy, when not needed, since
```
julia> @btime conj.($(rand(10^6)));
  285.667 μs (2 allocations: 7.63 MiB)

julia> @btime conj($(rand(10^6)));
  1.500 ns (0 allocations: 0 bytes)
```
And since arrays are being regarded as immutable, it's safe to remove trivial copies.

src/lib/broadcast.jl
```diff
@@ -75,23 +75,17 @@ unbroadcast(x::AbstractArray, x̄::Nothing) = nothing
...
 @adjoint function broadcasted(::typeof(/), x::Numeric, y::Numeric)
   res = x ./ y
   res, Δ -> (nothing, unbroadcast(x, Δ ./ conj.(y)), unbroadcast(y, .-Δ .* conj.(res ./ y)))
 end
-@adjoint function broadcasted(::typeof(/), x::AbstractArray{<:Number}, y::Number)
-  z, back = pullback(/, x, y)
-  z, Δ -> (nothing, back(Δ)...)
-end
+@adjoint broadcasted(::typeof(/), x::AbstractArray{<:Number}, y::Number) =
+  _pullback(/, x, y)
```
mcabbott, line 88: This is just tidying up, `_pullback` has the extra `nothing` so this saves one layer of lambdas.

test/cuda.jl
```diff
@@ -24,6 +24,11 @@ end
   g_gpu = gradient(x -> w(x), a_gpu)[1]
   @test g_gpu isa CuArray
   @test g_gpu |> collect ≈ g
+
+  # https://github.com/FluxML/Zygote.jl/issues/1027
+  @test gradient(x -> sum(x .!= 0), a_gpu) == (nothing,)
+  g3 = gradient(x -> sum(x .^ 3) / count(x .> 3), a)[1]
+  @test cu(g3) ≈ gradient(x -> sum(x .^ 3) / sum(x .> 3), a_gpu)[1]
```
mcabbott, line 31: Besides marking count as non-differentiable, the change to `broadcast_forward` removes an error from boolean-valued broadcasts like `x .> 3`.

test/cuda.jl
mcabbott: The last line `gradient(x -> sum(x .^ 3) ./ sum(x .> 3), a_gpu)` still gives a kernel error, I have no idea why. The pieces are fine separately, `sum(x .^ 3)` and `sum(x .> 3)` (and now tested). 

It's not a regression, fails with a similar error on Zygote v0.6.14, CUDA v3.3.0.

mcabbott:
Bump. Would be good to get this in, to build on top of. Only does minor things now -- the change mentioned above to remove the special path for CUDA broadcasting, has been removed, as I'm not 100% confident it won't break valid cases which now work.

test/cuda.jl
mcabbott: GPU failure on Julia 1.5 is this, surely unrelated:
```
Removing coverage files
--
  | Running Pkg.gc()
  | ERROR: type Context has no field verbose
  | Stacktrace:
  | [1] #Context!#44 at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Pkg/src/Types.jl:384 [inlined]
  | [2] gc(::Pkg.Types.Context; collect_delay::Day, kwargs::Base.Iterators.Pairs{Symbol,Bool,Tuple{Symbol},NamedTuple{(:verbose,),Tuple{Bool}}}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Pkg/src/API.jl:341
  | [3] top-level scope at none:1
  | 🚨 Error: Error tearing down bootstrap: The plugin julia pre-exit hook exited with status 1
```

CarloLucibello (approved):

PR #1019, opened 2021-07-04
Whitespace fixes

DhairyaLGandhi:
to be consistent


ToucheSir:
Can we set up a JuliaFormatter config for this repo? It would be nice to have this enforced so that CI can help with formatting (a pre-commit hook would be good too).

PR #1024, opened 2021-07-12
Make broadcasting over `Params` in the gradient an error

mcabbott:
According to https://github.com/FluxML/Flux.jl/pull/1614 this never worked. But at least it could have a less mysterious error message. 

And, there are actually 3 different "Mutating arrays is not supported" errors, with the same message, which aren't trivial to distinguish from their stack traces. So I made them more informative.

### DIFF
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -200,7 +200,6 @@ function Base.delete!(ps::Params, x)
 end
 
 Base.Broadcast.broadcasted(f, ps::Params) = broadcasted(f, ps.order)
-Base.Broadcast.broadcastable(ps::Params) = ps.order
 
 @adjoint function Broadcast.broadcasted(f::Function, ps::Params)
   f.(ps), _ -> throw(ArgumentError("Zygote.Params does not support broadcasting within gradients, try iteration `for p in ps`"))
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -200,6 +200,11 @@ function Base.delete!(ps::Params, x)
 end
 
 Base.Broadcast.broadcasted(f, ps::Params) = broadcasted(f, ps.order)
+# Broadcast.broadcastable(ps::Params) = ps.order
+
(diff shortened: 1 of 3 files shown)


ToucheSir (approved):
Looks good to me modulo any bikeshedding folks may have over the message syntax.

 Surprising to see the same message used in 3 different places, I'd always assumed there was a shared codepath!

darsnack (approved):
No bike shedding from me, the messages are perfect!

test/interface.jl
```diff
@@ -38,7 +38,10 @@ using Zygote: Grads
     x, y = [1,2], [1]
     ps = Params([x, y])
     @test length.(ps) == length.([x, y]) # 617
+    @test size.(ps, 1) == [2, 1]
     @test all(Params([[1,1]]) .== Params([[1,1]]))
+
+    @test_throws ArgumentError gradient(() -> sum(sum.(ps)), ps)
```
DhairyaLGandhi, line 44: Can drop this since we may want to enable this functionality with some changes in zygote.

src/compiler/interface.jl
```diff
@@ -200,6 +200,11 @@ function Base.delete!(ps::Params, x)
 end
 
 Base.Broadcast.broadcasted(f, ps::Params) = broadcasted(f, ps.order)
+# Broadcast.broadcastable(ps::Params) = ps.order
```
DhairyaLGandhi, line 203: This can be removed.

test/interface.jl
CarloLucibello: Maybe we want a test_broken here. In the end, this is a bug that we will want to fix at some point

mcabbott: Someone make a suggestion. Write the test which would pass if this were fixed. But if someone just deletes the line making the error, then that should produce an error here.

darsnack: I think this is fine. It correctly tests the current behavior (most important), and when this is fixed, this will throw an error and be modified for the new behavior.

PR #1025, opened 2021-07-12
Remove incorrect `push!` and `pop!` gradients

mcabbott:
This is a replacement for #876, which doesn't seem to get the right gradients at all.

It also looks into fixing #992, where similar gradients appear never to be called with nontrivial input.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -79,41 +79,17 @@ for f in [push!, pop!, pushfirst!, popfirst!]
 end
 
 # Exceptions:
-@adjoint! function push!(dst::AbstractVector{<:AbstractArray}, xs::AbstractArray...)
+@adjoint! function push!(dst::AbstractVector, xs::AbstractArray...)
   num_xs = length(xs)
   push!(dst, xs...), Δ -> begin
     (Δ[1:end-num_xs], Δ[end-num_xs+1:end]...)
   end
 end
 
-@adjoint! function pop!(src::AbstractVector{<:AbstractArray{T}}) where T
+@adjoint! function pop!(src::AbstractVector{<:AbstractArray})
   zs = fill(nothing, length(src)-1)
   pop!(src), Δ -> (vcat(zs, [Δ]),)
 end
-
-#=
-
-
-julia> gradient((xs, y) -> sum(abs2, push!(xs, y)[1]), [[1,2], [3,4]], [5,6])
-([[2, 4], nothing, nothing], 2-element Ones{Int64})
-([[2, 4], nothing], nothing)  # correct
-
-julia> gradient((xs, y) -> sum(abs2, push!(xs, y)[2]), [[1,2], [3,4]], [5,6])
-([nothing, [6, 8], nothing], 2-element Ones{Int64})
-([nothing, [6, 8]], nothing)  # correct
-
-julia> gradient((xs, y) -> sum(abs2, push!(xs, y)[3]), [[1,2], [3,4]], [5,6])
-([nothing, nothing, [10, 12]], 2-element Ones{Int64})
-([nothing, nothing], [10, 12])  # correct
-
-(jl_aA0cjy) pkg> st Zygote
-      Status `/private/var/folders/yq/4p2zwd614y59gszh7y9ypyhh0000gn/T/jl_aA0cjy/Project.toml`
-  [e88e6eb3] Zygote v0.6.14
-
-julia> gradient(xs -> sum(abs2, pop!(xs)), [[1,2], [3,4]])
-(Union{Nothing, Vector{Int64}}[nothing, nothing, [6, 8]],)
-
-=#
 
 # General
 
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -74,8 +74,8 @@ _droplike(dy::Union{LinearAlgebra.Adjoint, LinearAlgebra.Transpose}, dxv::Abstra
   _ -> error("Mutating arrays is not supported -- called copyto!(::$(typeof(xs)), _...)")
 
 for f in [push!, pop!, pushfirst!, popfirst!]
-  @eval @adjoint! $f(xs, x...) = $f(xs, x...), 
-    _ -> error("Mutating arrays is not supported -- called $($f)(::$(typeof(xs)), _...)")
+  @eval @adjoint! $f(x::AbstractVector, ys...) = $f(x, ys...), 
+    _ -> error("Mutating arrays is not supported -- called $($f)(::$(typeof(x)), _...)")
 end
 
 # General
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -79,17 +79,25 @@ for f in [push!, pop!, pushfirst!, popfirst!]
 end
 
 # Exceptions:
-@adjoint! function push!(dst::AbstractVector, xs::AbstractArray...)
-  num_xs = length(xs)
+@adjoint! function push!(dst::AbstractVector, xs...)
+  # num_xs = length(xs)
+  n = length(xs)
+  valn = Val(n)
   push!(dst, xs...), Δ -> begin
-    (Δ[1:end-num_xs], Δ[end-num_xs+1:end]...)
+    # (Δ[1:end-num_xs], Δ[end-num_xs+1:end]...)
+    (Δ[1:end-n], ntuple(i -> Δ[end-n+i], valn)...)
   end
 end
 
-@adjoint! function pop!(src::AbstractVector{<:AbstractArray})
+@adjoint! function pop!(src::AbstractVector)
   zs = fill(nothing, length(src)-1)
   pop!(src), Δ -> (vcat(zs, [Δ]),)
 end
+@adjoint! function pop!(src::AbstractVector{<:Number})
+  zs = falses(length(src)-1)
+  pop!(src), Δ -> (vcat(zs, Δ),)
+end
+
 
 # General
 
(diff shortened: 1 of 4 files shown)


src/lib/array.jl
```diff
@@ -80,26 +80,20 @@ _droplike(dy::Union{LinearAlgebra.Adjoint, LinearAlgebra.Transpose}, dxv::Abstra
...
-# This is kind of bad, but at least we don't materialize the whole
-# array. Prefer to use `Buffer`
-# function _pullback(cx::Context, ::typeof(push!), xs::AbstractVector{<:AbstractArray}, x::AbstractArray{T}...) where T
-@adjoint! function push!(xs::AbstractVector{<:AbstractArray}, x::AbstractArray{T}...) where T
-  sz_xs = size.(xs)
-  sz_x = size.(x)
-  push!(xs, x...), Δ -> begin
-    (Δ, map(x -> Ones{T}(x...), sz_x)...)
+# Exceptions:
+@adjoint! function push!(dst::AbstractVector, xs::AbstractArray...)
```
DhairyaLGandhi, line 87:
```suggestion
@adjoint! function push!(dst::AbstractVector{<: AbstractArray}, xs::AbstractArray...)
```

DhairyaLGandhi: Also retain the comments.

src/lib/array.jl
```diff
@@ -80,26 +80,20 @@ _droplike(dy::Union{LinearAlgebra.Adjoint, LinearAlgebra.Transpose}, dxv::Abstra
...
 
-@adjoint! function pop!(xs::AbstractVector{<:AbstractArray{T}}) where T
-  sz_xs = size.(xs)
-  op = pop!(xs)
-  op, Δ -> begin
-    ([Ones{T}(sz...) for sz in sz_xs], )
-  end
+@adjoint! function pop!(src::AbstractVector{<:AbstractArray})
+  zs = fill(nothing, length(src)-1)
+  pop!(src), Δ -> (vcat(zs, [Δ]),)
```
DhairyaLGandhi, line 96: Use `push` for clarity sake. 
```suggestion
  pop!(src), Δ -> (push!(zs, Δ),)
```

DhairyaLGandhi (changes requested):

src/lib/array.jl
mcabbott: Tests will fail if you do this.

src/lib/array.jl
mcabbott: Tests will fail if you do this. If you think it too wide, then can you invent a test case which it gets wrong?

mcabbott:
After fiddling a bit, I'm not really sure this can be fixed. Here's the core problem:
```
julia> using Zygote: @showgrad

julia> gradient([1,2,3], 4) do xs, y
             a = sum(@showgrad(xs))
             b = sum(push!(@showgrad(xs), y))
             c = sum(@showgrad xs)
             a+b+c
           end
∂(xs) = Fill(1, 4)  # c, final size
∂(xs) = Fill(1, 3)  # b, size before push!
∂(xs) = Fill(1, 4)  # a, surprising, a bug?
ERROR: DimensionMismatch("arrays could not be broadcast to a common size; got a dimension with lengths 3 and 4")
Stacktrace:
...
  [8] accum(::FillArrays.Fill{Int64, 1, Tuple{Base.OneTo{Int64}}}, ::FillArrays.Fill{Int64, 1, Tuple{Base.OneTo{Int64}}}, ::FillArrays.Fill{Int64, 1, Tuple{Base.OneTo{Int64}}})
```
pushing to / popping from `xs` means the same variable corresponds to a different size of array at different times, and somehow the gradients for it which are being accumulated will need to keep track of this. One could hack `accum` not to error, but not to do the right thing. Whether the deep internals can keep track of this I'm not quite sure.

src/lib/array.jl
DhairyaLGandhi: It'll be a memory hog otherwise and be slow.

src/lib/array.jl
DhairyaLGandhi: Why?

mcabbott: You can try it out pretty easily:
```
julia> let
        z = fill(nothing, 3)
        push!(z, [4, 5])
       end
ERROR: MethodError: Cannot `convert` an object of type Vector{Int64} to an object of type Nothing
```

DhairyaLGandhi: I'm sorry, `z` is not `Vector{Array}}` anyway. `zs` needs to have a union eltype at construction. Apologies if that wasn't clear.

darsnack:
Yeah seems like you'd need element-level tracking to make sure things lined up when accumulating even if we did have some padding hack for the size mismatch.

src/lib/array.jl
mcabbott: I can't read your mind. If you write something in the suggestion box, I take that as your suggestion.

More broadly I think a bit of focus on the big picture would be great. This is a super-minor style point, on a WIP PR, while there are major "can this work at all?" questions in play.

mcabbott:
It's possible that if you always wrote `xs = push!(xs, y)` then Zygote would understand that the label `xs` is attached to different-length arrays at different points of the code, and accumulate their gradients correctly. Maybe some attempt was made to automate that, when trying to allow mutation via `@adjoint!` etc? Have never looked that deep into the internals. Maybe @simeonschaub knows things.

But without something like that, we should probably just make these back into errors.

ToucheSir:
Rolling back the clock a bit, is there a way we could tackle the original issue in https://github.com/FluxML/Zygote.jl/issues/37#issuecomment-771532104 without something like #876? As noted in https://github.com/FluxML/Flux.jl/pull/1614#issuecomment-877873705 that would get us back to a known point that is less buggy than what exists now.

mcabbott:
That does sound like a good path. #37 doesn't have a stacktrace, I can block out time on my calendar to compile the DiffEqverse and try it sometime soon...

ToucheSir:
Partially answering my own question now that this is up for review, @ChrisRackauckas's comment at https://github.com/FluxML/Zygote.jl/pull/876#issuecomment-771534007 suggests that permitting `push!`/`pop!` for `AbstractVector` wouldn't have done much for `pop!(::BinaryHeap)` anyhow. This seems to be corroborated by the [DataStructures.jl method](https://github.com/JuliaCollections/DataStructures.jl/blob/c400f30052719a92e5a285f58931b1f1772ac1b8/src/heaps/arrays_as_heaps.jl#L71) too, which not only calls `push!`, but further performs scalar array mutation in a loop [here](https://github.com/JuliaCollections/DataStructures.jl/blob/c400f30052719a92e5a285f58931b1f1772ac1b8/src/heaps/arrays_as_heaps.jl#L39-L40).

test/features.jl
```diff
@@ -350,10 +350,10 @@ end
...
 @test Zygote.@code_adjoint(f(1)) isa Zygote.Adjoint
 
-@test_throws ErrorException Zygote.gradient(1) do x
-  push!([], x)
-  return x
-end
+# @test_throws ErrorException Zygote.gradient(1) do x
+#   push!([], x)
+#   return x
+# end
```
ToucheSir, line 356:
```suggestion
@test_throws ErrorException Zygote.gradient(1) do x
  push!([], x)
  return x
end
```
This should work again, if I'm not mistaken?

mcabbott: Yes I think so, I missed that.

mcabbott:
Ok, then you've got more out of these scattered threads than I did. If the goal is to make certain objects in DataStructures like `BinaryHeap` AD-able, then I think someone should directly state that on an issue there, and we can discuss. 

The `percolate_up!` and `heappush!` functions you link to don't look like good candidates, for the same reason as `push!(::Vector, ...)` above, and for the generic reason that allowing `f!(x)` assumes that no other operation has closed over `x` to re-use this in the backward pass. But maybe something can be attached to a higher-level function. Or maybe the bad things can have errors attached to them to leave a safe happy path. 

The code in https://github.com/FluxML/Zygote.jl/issues/37#issuecomment-771532104 leads me (eventually) to this error in all 4 cases, from which I learn nothing about what's desired:
```
ERROR: MethodError: no method matching fast_materialize(::Vector{Float64})
Closest candidates are:
  fast_materialize(::Base.Broadcast.Broadcasted{S}) where S at /Users/me/.julia/packages/FastBroadcast/WP7Ws/src/FastBroadcast.jl:18
Stacktrace:
  [1] macro expansion
    @ ~/.julia/packages/Zygote/nsu1Y/src/compiler/interface2.jl:0 [inlined]
  [2] _pullback(ctx::Zygote.Context, f::typeof(FastBroadcast.fast_materialize), args::Vector{Float64})
    @ Zygote ~/.julia/packages/Zygote/nsu1Y/src/compiler/interface2.jl:9
```

mcabbott:
New CI error on 1.3 is not caused by this PR. It's a bad interaction between complex numbers in the gradient of `^` and Dual numbers from `hessian`. It's fixed (well, skipped) here: https://github.com/FluxML/Zygote.jl/pull/1044/files#diff-3483c521f73b08fdb6b00f014614cc0d69f87ea7b098a24aff838cbdc812704dR25

ToucheSir:
Apparently my attempt to re-run CI was a bust, I guess this needs a rebase?

Rolling things back sucks, but it'll buy us some time to ponder how best to address `push!`, `pop!` and co. without more issues like https://github.com/FluxML/Zygote.jl/pull/992#issuecomment-929517421 popping up. So if there the tests pass and there aren't any unrelated regressions, this is probably good to go.

ToucheSir (approved):

PR #1031, opened 2021-07-19
construct Params with empty tuple

CarloLucibello:
Fixes a bug encountered in Flux's tests https://github.com/FluxML/Flux.jl/pull/1531#discussion_r672035454.
The test has been there since forever, and it creates a `Params` from an empty tuple. I don't know what changed on the Zygote side now leading to the error, but the change in this PR is pretty safe

Fix. #1032

### DIFF
--- a/src/tools/idset.jl
+++ b/src/tools/idset.jl
@@ -3,18 +3,22 @@ struct IdSet{T} <: AbstractSet{T}
   IdSet{T}() where T = new(IdDict{T,Nothing}())
 end
 
-Base.eltype(::IdSet{T}) where T = T
+IdSet(xs) = IdSet{eltype(xs)}(xs)
 
 IdSet() = IdSet{Any}()
+
+function IdSet{T}(xs) where T
+  s = IdSet{T}()
+  for x in xs
+    push!(s, x)
+  end
+  return s
+end
 
 Base.push!(s::IdSet{T}, x::T) where T = (s.dict[x] = nothing; s)
 Base.delete!(s::IdSet{T}, x::T) where T = (delete!(s.dict, x); s)
 Base.in(x, s::IdSet) = haskey(s.dict, x)
-
-IdSet{T}(xs) where T = push!(IdSet{T}(), xs...)
-
-IdSet(xs) = IdSet{eltype(xs)}(xs)
-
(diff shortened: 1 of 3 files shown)


test/interface.jl
```diff
@@ -82,6 +82,11 @@ using Zygote: Grads
     @test ps isa Params
     @test issetequal(ps, Set([y]))
   end
+
+  @testset "constructor with empty args" begin
+    @test length(Params()) == 0
+    @test length(Params(())) == 0
```
DhairyaLGandhi, line 88:
```suggestion
    @test length(Params([])) == 0
```
Params typically expects a mutable container.

DhairyaLGandhi (approved):
lgtm

src/tools/idset.jl
```diff
@@ -3,18 +3,16 @@ struct IdSet{T} <: AbstractSet{T}
   IdSet{T}() where T = new(IdDict{T,Nothing}())
 end
 
-Base.eltype(::IdSet{T}) where T = T
+IdSet(xs) = IdSet{eltype(xs)}(xs)
```
DhairyaLGandhi, line 6: Is this needed anymore?

test/interface.jl
CarloLucibello: we do suuport tuples though https://github.com/FluxML/Zygote.jl/blob/4d22edf4a27db0302badb12e1faec6d5ee31fd9d/src/compiler/interface.jl#L142

src/tools/idset.jl
DhairyaLGandhi: Nevermind, its deleted and redefined in this PR.

CarloLucibello: I think so, unless we remove the T from `IdSet{T}(xs)` and deduce it internally

test/interface.jl
DhairyaLGandhi: Nope, notice that `collect` will convert it to an array.

CarloLucibello: so? We support constructing Params out of tuples e.g. `Param((w, b))

DhairyaLGandhi:
Please use bors to merge.

PR #1033, opened 2021-07-19
Add some downstream testing

ChrisRackauckas:
Adds downstream testing to CI, with a few application packages that should flex the API a bit.

### DIFF
--- a/.github/workflows/Downstream.yml
+++ b/.github/workflows/Downstream.yml
@@ -1,0 +1,55 @@
+  
+name: IntegrationTest
+on:
+  push:
+    branches: [master]
+    tags: [v*]
+  pull_request:
+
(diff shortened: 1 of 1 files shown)


DhairyaLGandhi:
Related https://github.com/FluxML/Flux.jl/pull/1555

CarloLucibello:
Can this be linked to bors so that we run downstream tests only before merging a PR instead of on every commit?

CarloLucibello:
An `@inferred` is failing in NNlib, we get `Any` instead of `Float16` for `elu`, slightly annoying but not too serious. I would remove FastAI and GeometricFlux to save some CI, they don't do anything particularly fancy and most of their functionality relies on Flux and NNlib.

Issue #705, opened 2020-06-21
Incorrect complex gradients, Zygote 0.4.21


mcabbott:
The last two here are incorrect:
```julia
julia> Zygote.gradient(imag∘exp, 1)[1] # fine
0.0 + 2.718281828459045im

julia> Zygote.gradient(imag∘exp, 2)[1] # fine 
0.0 + 7.38905609893065im

julia> Zygote.gradient(imag∘exp, 2+0.00001im)[1] # small perturbation
7.3890560988075e-5 + 7.3890560985611975im

julia> Zygote.gradient(x -> imag(sum(exp, x)), [1,2,3])[1] # fine
3-element Array{Complex{Float64},1}:
 0.0 + 2.718281828459045im
 0.0 + 7.38905609893065im
 0.0 + 20.085536923187668im

julia> Zygote.gradient(x -> imag(sum(exp, x)), [1+0im,2,3])[1] # wtf?
3-element Array{Complex{Float64},1}:
 0.0 + 1.0im
 0.0 + 1.0im
 0.0 + 1.0im

julia> Zygote.gradient(x -> imag(sum(exp, x)), [1+0.00001im,2,3])[1]
3-element Array{Complex{Float64},1}:
 2.7182818284137405e-5 + 2.718281828323131im
                   0.0 + 1.0im
                   0.0 + 1.0im

(@v1.5) pkg> st Zygote 
Status `~/.julia/environments/v1.5/Project.toml`
  [e88e6eb3] Zygote v0.4.21
```
This also occurs with Zygote 0.5.0, and on Julia 1.3. 

It does not occur with Zygote 0.4.20, where instead you get results like the scalar ones above:
```
julia> Zygote.gradient(x -> imag(sum(exp, x)), [1+0im,2,3])[1] # now OK
3-element Array{Complex{Float64},1}:
 0.0 + 2.718281828459045im
 0.0 + 7.38905609893065im
 0.0 + 20.085536923187668im

julia> Zygote.gradient(x -> imag(sum(exp, x)), [1+0.01im,2,3])[1]
3-element Array{Complex{Float64},1}:
 0.02718236523988427 + 2.718145915500236im
                 0.0 + 7.38905609893065im
                 0.0 + 20.085536923187668im

(sjkfhsd) pkg> st Zygote
Status `~/trash/sjkfhsd/Project.toml`
  [e88e6eb3] Zygote v0.4.20
```

sethaxen:
[...]

PR #1035, opened 2021-07-20
Move to ChainRulesCore v1.0 (in OptingOut)

oxinabox:
This PR will make the changes needed to support ChainRulesCore v1.0
A number of commits will be added to it to fix each breaking change.

Right now I am aware of 2 such changes

- [x]  The new officially supported way of `@opt_out` to say not to do AD. For https://github.com/JuliaDiff/ChainRulesCore.jl/pull/398 (this is the biggest change)
- [x]  Removing of convert. For https://github.com/JuliaDiff/ChainRulesCore.jl/issues/359
- [x] Change Project.toml


src/compiler/chainrules.jl
```diff
@@ -18,19 +17,45 @@ such that if a suitable rule is defined later, the generated function will recom
...
+  configured_rrule_m = meta(Tuple{typeof(rrule), config_T, arg_Ts...})
+  if _is_rrule_redispatcher(configured_rrule_m.method)
+    # it is being redispatched without config, so get the method it redispatches to
+    rrule_m = meta(Tuple{typeof(rrule), arg_Ts...})
+    no_rrule_m = meta(Tuple{typeof(ChainRulesCore.no_rrule), arg_Ts...})
+  else
+    # Not being redispatched
+    rrule_m = configured_rrule_m
+    no_rrule_m = meta(Tuple{typeof(ChainRulesCore.no_rrule), config_T, arg_Ts...})
+  end
```
oxinabox, line 29: I keep having to convince myself that this does the right thing.
Maybe it needs more comments and tests.

Especially ones with Configs involved.

src/compiler/chainrules.jl
```diff
@@ -18,19 +17,45 @@ such that if a suitable rule is defined later, the generated function will recom
...
-  return true, nothing
+end
+
+matching_cr_sig(t, s) = matching_cr_sig(t.method.sig, s.method.sig)
+matching_cr_sig(::DataType, ::UnionAll) = false
+matching_cr_sig(::UnionAll, ::DataType) = false
+matching_cr_sig(t::Type, s::Type) = type_tuple_tail(t) == type_tuple_tail(s)
+ 
+type_tuple_tail(d::DataType) = Tuple{d.parameters[2:end]...}
+function type_tuple_tail(d::UnionAll)
```
mzgubic, line 55: What is this function doing?

mzgubic (commented):
Generally the mechanism seems fine. I guess that `args_T` is always just concrete types. I do still have some questions on how the functions work - but seems fine overall

src/compiler/chainrules.jl
```diff
@@ -18,19 +17,45 @@ such that if a suitable rule is defined later, the generated function will recom
...
+    rrule_m = meta(Tuple{typeof(rrule), arg_Ts...})
+    no_rrule_m = meta(Tuple{typeof(ChainRulesCore.no_rrule), arg_Ts...})
+  else
+    # Not being redispatched
+    rrule_m = configured_rrule_m
+    no_rrule_m = meta(Tuple{typeof(ChainRulesCore.no_rrule), config_T, arg_Ts...})
+  end
+
+  do_not_use_rrule = matching_cr_sig(no_rrule_m, rrule_m)
+  if do_not_use_rrule
```
DhairyaLGandhi, line 32: Is this meant to do decomposition?

test/chainrules.jl
```diff
@@ -232,6 +232,36 @@ using ChainRulesCore, ChainRulesTestUtils, Zygote
...
+
+        # make sure don't hit the one we have opted out
+        oa_id_rrule_hitcount[] = 0
+        @test ([1.0],) == Zygote.gradient(oa_id_outer, [π])
+        @test oa_id_rrule_hitcount[] == 0
+
+        # Now try opting out After we have already used it 
+        @opt_out ChainRulesCore.rrule(::typeof(oa_id), x::Real)
+        oa_id_rrule_hitcount[] = 0
+        oa_id_outer(x) = sum(oa_id(x))
```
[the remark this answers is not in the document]
oxinabox: Probably a mistake

src/compiler/chainrules.jl
```diff
@@ -18,19 +17,45 @@ such that if a suitable rule is defined later, the generated function will recom
...
+    # be revisiting this decision when a new opt-out is added
+    return true, nothing
   end
-  # otherwise found a rrule, no need to add any edges, as it will generate code with
-  # natural edges.
-  return true, nothing
+end
+
+matching_cr_sig(t, s) = matching_cr_sig(t.method.sig, s.method.sig)
+matching_cr_sig(::DataType, ::UnionAll) = false
```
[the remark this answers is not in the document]
oxinabox:
```julia
julia> typeof(Tuple{typeof(rrule), Array{T}} where T)
UnionAll
```

src/compiler/chainrules.jl
```diff
@@ -18,19 +17,45 @@ such that if a suitable rule is defined later, the generated function will recom
...
+matching_cr_sig(::UnionAll, ::DataType) = false
+matching_cr_sig(t::Type, s::Type) = type_tuple_tail(t) == type_tuple_tail(s)
```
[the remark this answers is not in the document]
oxinabox: Correct. `method.sig` is a `UnionAll` or a `DataType`

src/compiler/chainrules.jl
oxinabox: `Tuple{typeof(rrule), typeof(+), Array{T}} where T` --> `Tuple{typeof(+), Array{T}} where T`

src/compiler/chainrules.jl
oxinabox: This function is `has_chain_rrule` and if it returns `false,...` then `pullback` will indeed end up calling `generate_pullback_via_decomposition`.

mzgubic:
minimal failing example for the remaining test failure is:
```julia
    using Zygote 

    W = ones(Float32, 3)
    ps = Zygote.Params([W, ])

    gs = gradient(ps) do
        p, pb = pullback(ps) do
            sum(W)
        end
        g = pb(p)
        sum(g[W])
    end
```

<details>
<summary>(stacktrace)</summary>
<p>

```julia
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:33
  [2] Pullback
    @ ./iddict.jl:87 [inlined]
  [3] (::typeof(∂(get)))(Δ::Nothing)
    @ Zygote ~/JuliaEnvs/Zygote.jl/src/compiler/interface2.jl:0
  [4] Pullback
    @ ~/JuliaEnvs/Zygote.jl/src/lib/lib.jl:68 [inlined]
  [5] (::typeof(∂(accum_global)))(Δ::Nothing)
    @ Zygote ~/JuliaEnvs/Zygote.jl/src/compiler/interface2.jl:0
  [6] Pullback
    @ ~/JuliaEnvs/Zygote.jl/src/lib/lib.jl:79 [inlined]
  [7] (::typeof(∂(λ)))(Δ::Nothing)
    @ Zygote ~/JuliaEnvs/Zygote.jl/src/compiler/interface2.jl:0
  [8] Pullback
  [9] (::typeof(∂(λ)))(Δ::Nothing)
    @ Zygote ~/JuliaEnvs/Zygote.jl/src/compiler/interface2.jl:0
 [10] getindex
    @ ./tuple.jl:29 [inlined]
 [11] (::typeof(∂(λ)))(Δ::Nothing)
    @ Zygote ~/JuliaEnvs/Zygote.jl/src/compiler/interface2.jl:0
 [12] Pullback
    @ ~/JuliaEnvs/Zygote.jl/src/compiler/interface.jl:348 [inlined]
 [13] (::typeof(∂(λ)))(Δ::Nothing)
    @ Zygote ~/JuliaEnvs/Zygote.jl/src/compiler/interface2.jl:0
 [14] Pullback
    @ ./REPL[5]:5 [inlined]
 [15] (::typeof(∂(#5)))(Δ::Float32)
    @ Zygote ~/JuliaEnvs/Zygote.jl/src/compiler/interface2.jl:0
 [16] (::Zygote.var"#90#91"{Params, typeof(∂(#5)), Zygote.Context})(Δ::Float32)
    @ Zygote ~/JuliaEnvs/Zygote.jl/src/compiler/interface.jl:348
 [17] gradient(f::Function, args::Params)
    @ Zygote ~/JuliaEnvs/Zygote.jl/src/compiler/interface.jl:76
 [18] top-level scope
    @ REPL[5]:1
```

where the failing expression is
`ex = :($(Expr(:foreigncall, :(:jl_eqtable_get), Any, svec(Any, Any, Any), 0, :(:ccall), %5, %3, %4)))`
</p>
</details>

oxinabox:
> minimal failing example for the remaining test failure is:

That fails to me with the current released version of Zygote  for me also.
```
      Status `/tmp/jl_QkhRV2/Project.toml`
  [082447d4] ChainRules v0.8.23
  [d360d2e6] ChainRulesCore v0.10.13
  [e88e6eb3] Zygote v0.6.17
```

DhairyaLGandhi:
Check with https://github.com/FluxML/Zygote.jl/pull/823?

oxinabox:
> Check with #823?

Yes it works with that branch.
I will see if it works in this branch with that one merged into it

oxinabox:
Minor varient of @mzgubic 's MWE that actually passes on current release of Zygote, but doesn't on this PR:
wrap everything in a `@testset`

```
using Zygote, Test
@testset "mwe" begin
    W = ones(Float32, 3)
    ps = Zygote.Params([W, ])

    gs = gradient(ps) do
        p, pb = pullback(ps) do
            sum(W)
        end
        g = pb(p)
        sum(g[W])
    end
end
```

oxinabox:
That particular issue is fixed by https://github.com/JuliaDiff/ChainRulesCore.jl/pull/414

but other issues still remain.

DhairyaLGandhi:
Probably worth it to merge #823 with a cleanup then.

oxinabox:
> Probably worth it to merge #823 with a cleanup then.

No, i fixed it to not need that.
I think #823  fixed it kinda by coincidence.
Or maybe it didn't even fix the real thing at all, it was a bad MWE.

----

This is now passing all tests locally.
It is just blocked by https://github.com/JuliaMath/SpecialFunctions.jl/pull/335
once that is in this should be ready to go.

test/chainrules.jl
```diff
@@ -275,7 +304,7 @@ end
             ZygoteRuleConfig(), my_namedtuple, 1., 2., 3.; rrule_f=rrule_via_ad
         )
         test_rrule(
-            ZygoteRuleConfig(), my_namedtuple, 1., (2.0, "str"), 3.; rrule_f=rrule_via_ad
+            ZygoteRuleConfig(), my_namedtuple, 1., (2.0, 2.4), 3.; rrule_f=rrule_via_ad
```
oxinabox, line 307: Had to change this due to an issue in ChainRulesTestUtils
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/194

The Zygote and ChainRules code handles the string fine,
but ChainRulesTestUtils kinda freaks out about it.

test/complex.jl
```diff
@@ -18,7 +18,7 @@ using Zygote, Test, LinearAlgebra
 @test gradient(x -> real(logabsdet(x)[1]), [1 2im; 3im 4])[1] ≈ [4 3im; 2im 1]/10
 
 # https://github.com/FluxML/Zygote.jl/issues/705
-@test gradient(x -> imag(sum(exp, x)), [1,2,3])[1] ≈ im .* exp.(1:3)
+@test gradient(x -> imag(sum(exp, x)), [1,2,3])[1] ≈ real(im .* exp.(1:3))
```
oxinabox, line 21: This is because ChainRules takes embedded subspaces seriously.
Derivative of a real array can not be imaginary.

The thing that #705 was worried about is still fixed

test/features.jl
```diff
@@ -449,12 +449,12 @@ end
 @test pullback(type_test)[1] == Complex{<:Real}
 
 @testset "Pairs" begin
-    @test (x->10*pairs((a=x, b=2))[1])'(100) === 10
+    @test (x->10*pairs((a=x, b=2))[1])'(100) === 10.0
```
oxinabox, line 452: this is because we now take embedded sub-spaces seriously.
Integers are always considered to be a subspace of Floats.
Not just went it is convenient.

Project.toml
```diff
@@ -6,6 +6,7 @@ version = "0.6.17"
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
 ChainRules = "082447d4-558c-5d27-93f4-14fc19e9eca2"
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
+ChainRulesTestUtils = "cdddcdb0-9152-4a09-a978-84456f9df70a"
```
oxinabox, line 9:
```suggestion
```

test/chainrules.jl
DhairyaLGandhi: Maybe we can keep it as a `gradtest` then? I'm not super sure if CRTU is expected to be a dependency for every AD test suite.

test/features.jl
DhairyaLGandhi: Shouldn't this be compiling to integer code anyway?

test/utils.jl
```diff
@@ -24,10 +24,25 @@ end
 
   xs, y = randn(2,3), rand()
   f34(xs, y) = xs[1] * (sum(xs .^ (1:3)') + y^4)  # non-diagonal Hessian, two arguments
+
+  # Follow is should work ones we workout what ForwardDiff should do when `Float64` is called on a `Dual`
```
DhairyaLGandhi, line 28:
```suggestion
  # Following should work once we workout what ForwardDiff should do when `Float64` is called on a `Dual`
```

test/chainrules.jl
oxinabox: I am pretty sure `gradtest` can't handle it either -- gradtest can't handle namedtuples containing tuples.

> I'm not super sure if CRTU is expected to be a dependency for every AD test suite.

It is,  if you have `rrule_via_ad` overloaded. (or `rrule`)
Then you can use it to test all things.
It's a more robust version of `gradtest` that can handle more types, and doesn't e.g. get tricked by antisymmetric errors.

test/features.jl
oxinabox: Integers are not a good type to use to represent tangents.
Because if you are going to do gradeient decent you are going to apply a learning rate like `0.1*dx`.
So we call `float` on them to get the corresponding floating point type.
(Unless it is a index or something, then it would be `NoTangent()`)

Arguably we could keep integers here, but encouraging people to use integers to repressent continous values that just so happen do be integers feels like not taking subspace types seriously
:shrug:

DhairyaLGandhi: I think it's a matter of letting the language take charge of these things. It's possible to work with just ints.

src/compiler/chainrules.jl
```diff
@@ -18,19 +17,68 @@ such that if a suitable rule is defined later, the generated function will recom
...
+    # it is being redispatched without config, so we need the method it redispatches to
+    rrule_m = meta(Tuple{typeof(rrule), arg_Ts...})
+    # Thus any no_rrule that might apply must also not have a config because if there was a
+    # no_rrule with a config that applied then there would also be a rrule with config that applied
+    no_rrule_m = meta(Tuple{typeof(ChainRulesCore.no_rrule), arg_Ts...})
+  else
+    # Not being redispatched: it does have a config
+    rrule_m = configured_rrule_m
+     # Thus any no_rrule that might apply must also have a config because if it applied
+     # it will be identical, and if it doesn't we don't care what it is.
```
DhairyaLGandhi, line 32:
```suggestion
    # Thus any no_rrule that might apply must also have a config because if it applied
    # it will be identical, and if it doesn't we don't care what it is.
```

src/compiler/interface.jl
```diff
@@ -58,11 +58,11 @@ See also [`withgradient`](@ref) to keep the value `f(args...)`,
...
 
 ```jldoctest; setup=:(using Zygote)
-julia> gradient(*, 2, 3, 5)
-(15, 10, 6)
+julia> gradient(*, 2.0, 3.0, 5.0)
+(15.0, 10.0, 6.0)
 
-julia> gradient(x -> sum(abs2,x), [7, 11, 13])
-([14, 22, 26],)
+julia> gradient(x -> sum(abs2,x), [7.0, 11.0, 13.0])
```
DhairyaLGandhi, line 64: We still expect the integer inputs (and also primal) to return integer gradient in this case, yes?

src/compiler/chainrules.jl
```diff
@@ -18,19 +17,68 @@ such that if a suitable rule is defined later, the generated function will recom
...
+  # To understand why we only need to check if the sigs match between no_rrule_m and rrule_m
+  # in order to decide if to use, one must consider the following facts:
+  # - for every method in `no_rrule` there is a identical one in `rrule` that returns nothing
+  # - this includes the general fallback `rrule(::Any...)=nothing`.
+  # - a configured rrule/no_rrule is always more specific than a otherwise equivalent unconfigured rrule/no_rrule
+  #  
+  # Consider the following truth table, for what can occur:
+  # rrule: fallback, no_rrule: fallback =>  matches => do not use rrule.
+  # rrule: specific, no_rrule: fallback => !matches => do use rrule, as haven't opted out.
+  # rrule: fallback, no_rrule: specific =>  IMPOSSIBLE, every no_rule us identical to some rrule
```
DhairyaLGandhi, line 45:
```suggestion
  # rrule: fallback, no_rrule: specific =>  IMPOSSIBLE, every no_rule is identical to some rrule
```

src/compiler/chainrules.jl
```diff
@@ -18,19 +17,68 @@ such that if a suitable rule is defined later, the generated function will recom
...
+  # rrule: specific, no_rrule: specific =>  matches => do not use rrule as opted out
+  # rrule: specific, no_rrule: general  => !matches => do use rrule as a more specific rrule takes preciedent over more general opted out
```
DhairyaLGandhi, line 47: So its the kind of complement of `@nograd f(x::SomeType, y::SomeOtherType)` where we can still get grads for some methods, and not for others? Kind of defining more specific rrules and dispatching to them, but doing so manually here, rather than letting Julia do it?

src/compiler/chainrules.jl
```diff
@@ -18,19 +17,68 @@ such that if a suitable rule is defined later, the generated function will recom
...
+  # rrule: general , no_rrule: specific =>  IMPOSSIBLE, every no_rule us identical to some rrule so can't have a more general rrule being hit, as the specific one would hit first
+  #
+  # Note that the fallback  cases are the same outcome as the general cases as fallback is just most general.
+  # It can be seen that checking if it matches is the correct way to decide if we should ue the rrule or not.
```
DhairyaLGandhi, line 51:
```suggestion
  # Note that the fallback cases are the same outcome as the general cases as fallback is just most general.
  # It can be seen that checking if it matches is the correct way to decide if we should use the rrule or not.
```

test/chainrules.jl
DhairyaLGandhi: Its good to have the coverage over different data types, but I guess its going to need a future PR. I can already sense some SMILES like application making use of strings outside of embeddings.

src/compiler/interface.jl
DhairyaLGandhi: Guess not, its the same as a previous review comment.

src/compiler/chainrules.jl
oxinabox: It is like defining a more specific `rrule` where that more specific rrule is "Let AD work it out".
But we can't use `rrule_via_ad` for this because you hit a stackoverflow.

It is not really like `@nograd` (`@non_differentiable` is like `@nograd`) except that both participate in `rrule` dispatch via specificity

test/features.jl
oxinabox: Can this be moved to a issue on ChainRulesCore.jl?
Its not going to be resolved in this PR.

DhairyaLGandhi (approved):

PR #1044, opened 2021-07-27
Use `ProjectTo` in broadcasting & `gradient`

### LINKED ISSUE #342: Complex gradient on real function with complex intermediates

### LINKED ISSUE #402: Adjoints for functions of specialized matrices

### LINKED ISSUE #431: Gradient of inv and logdet UpperTriangular matrix is not UpperTriangular

### LINKED ISSUE #917: Trying to differentiate a function with complex and real inputs results in error

mcabbott:
This starts building ChainRulesCore's type projection story into how Zygote handles broadcasting, and into its user-facing functions. This will already be called in some rules handled by ChainRules, but this applies it a bit more broadly. 

After:
```julia
julia> gradient(x->imag(x + 2.0*im), 3.0)  # https://github.com/FluxML/Zygote.jl/issues/342
(0.0,)

julia> gradient(x -> getindex(x,2,1), Diagonal(rand(3,3)))[1]  # https://github.com/FluxML/Zygote.jl/issues/402
3×3 Diagonal{Float64, Vector{Float64}}:
 0.0   ⋅    ⋅ 
  ⋅   0.0   ⋅ 
  ⋅    ⋅   0.0

julia> gradient(x -> sum(sqrt.(x .+ 1)), [1,2,3]')[1]  # previously became a matrix
1×3 adjoint(::Vector{Float64}) with eltype Float64:
 0.353553  0.288675  0.25
```
Before:
```julia
julia> gradient(x->imag(x + 2.0*im), 3.0) 
(0.0 + 1.0im,)

julia> gradient(x -> getindex(x,2,1), Diagonal(rand(3,3)))[1]
3×3 Zygote.OneElement{Float64, 2, Tuple{Int64, Int64}, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}}:
 0.0  0.0  0.0
 1.0  0.0  0.0
 0.0  0.0  0.0

julia> gradient(x -> sum(sqrt.(x .+ 1)), [1,2,3]')[1]
1×3 Matrix{Float64}:
 0.353553  0.288675  0.25
```

Replaces #965, or most of it.

~~Many tests will fail, including most of the FFT tests I think, since those tend to return a complex gradient for a real input.~~ FFT tests are unchanged. 

Closes #342, closes #402. Fixes #917, fixes #431.

Closes https://github.com/FluxML/Flux.jl/issues/886


src/compiler/interface.jl
```diff
@@ -123,26 +144,15 @@ julia> haskey(g, z)  # only x and y are parameters
...
-
-Container for implicit parameters, used when differentiating
-a zero-argument funtion `() -> loss(A, B)` with respect to `A, B`.
-"""
-struct Params
-  order::Buffer # {Any, Vector{Any}}
-  params::IdSet{Any} # TODO store ids only
+function gradient(f, ps::Params)
+  y, back = pullback(f, ps)
+  back(sensitivity(y))
```
mcabbott, line 149: `struct Params` hasn't changed. I just had to make a separate method `gradient(f, ps::Params)` and move that below it, the diff is confused. 

The reason that's separate is that projection doesn't happen for Params at all right now. But perhaps it should.

src/compiler/interface.jl
```diff
@@ -73,7 +73,8 @@ julia> gradient([7, 11], 0, 1) do x, y, d
 """
 function gradient(f, args...)
   y, back = pullback(f, args...)
-  return back(sensitivity(y))
+  grad = back(sensitivity(y))
+  map(_project, args, grad)
```
mzgubic, line 77: do we want this at the `gradient` or the `pullback` level?

mcabbott:
One side-effect of this is that it makes this wrong answer into an error:
```julia
julia> gradient((x,y) -> sum(map(+,x,y)), [1,2], [3,4,5,6])  # before
([1, 1], [1, 1])

julia> gradient((x,y) -> sum(map(+,x,y)), [1,2], [3,4,5,6])  # after
ERROR: DimensionMismatch("variable with size(x) == (4,) cannot have a gradient with size(dx) == (2,)")
Stacktrace:
 [1] (::ChainRulesCore.ProjectTo{AbstractArray, NamedTuple{(:element, :axes), Tuple{ChainRulesCore.ProjectTo{Float64, NamedTuple{(), Tuple{}}}, Tuple{Base.OneTo{Int64}}}}})(dx::Vector{Int64})
   @ ChainRulesCore ~/.julia/packages/ChainRulesCore/ySyqy/src/projection.jl:197
```

src/compiler/interface.jl
mcabbott: My thinking was to start small! Applying it to `gradient` applies it to the user-facing calls, once. Applying it to `pullback` or `_pullback` inserts it into many more places internally... maybe it'll make `sin'''(1.0)` unhappy.

DhairyaLGandhi:
That error seems awkward to me. Previously, the Julia behaviour of the function was the reason behind this operation. Presumably, the resultant gradient should be sized appropriately, not error.

src/compiler/interface.jl
```diff
@@ -95,11 +97,32 @@ true
 """
 function withgradient(f, args...)
   y, back = pullback(f, args...)
-  (val = y, grad = back(sensitivity(y)))
+  grad = back(sensitivity(y))
+  isnothing(grad) && return (val=y, grad=nothing)
```
DhairyaLGandhi, line 101: Why is this check necessary?

src/compiler/chainrules.jl
```diff
@@ -128,6 +128,39 @@ Convert `x` from the format Zygote uses internally to differentials types ChainR
   ChainRules.Tangent{Any, typeof(xp)}(xp)
 end
 
+"""
+  _project(x)(dx)
```
DhairyaLGandhi, line 132: This sound like it should be in CRC, not zygote.

cc @oxinabox

src/lib/broadcast.jl
```diff
@@ -45,18 +45,20 @@ function Base.reducedim_init(::typeof(identity), ::typeof(accum), A::AbstractArr
   Base.reducedim_initarray(A, region, nothing, Union{Nothing,eltype(A)})
 end
 
-trim(x, Δ) = reshape(Δ, ntuple(i -> size(Δ, i), Val(ndims(x))))
```
DhairyaLGandhi, line 48: I think doing this makes unbroadcast less generic, we don't need to define projections here afaict. Let's retain the current definition.

src/compiler/interface.jl
mcabbott: It might not be. WIP, as it says.

src/compiler/chainrules.jl
mcabbott: If you check the CRC issues, you will see much discussion of this exact point. WIP, as it says.

src/lib/broadcast.jl
mcabbott: What case exactly is not handled, if this is less generic?

DhairyaLGandhi: It restricts it to what can be handled by `_project` as opposed to simple sizes and lengths of arrays.

src/lib/broadcast.jl
```diff
@@ -45,18 +45,20 @@ function Base.reducedim_init(::typeof(identity), ::typeof(accum), A::AbstractArr
...
-  length(x) == length(x̄) ? trim(x, x̄) :
-    trim(x, accum_sum(x̄, dims = ntuple(i -> size(x, i) == 1 ? i : ndims(x̄)+1, Val(ndims(x̄)))))
-
+function unbroadcast(x::AbstractArray, x̄)
+  N = ndims(x̄)
+  if length(x) == length(x̄)
+    _project(x, x̄)  # ProjectTo handles reshape, offsets, structured matrices, row vectors
+  else
+    tup = filter(d -> size(x, d) == 1, ntuple(identity, N))
+    dims = length(tup) == 1 ? only(tup) : tup  # avoid sum(xbar, dims=(1,)) as e.g. sum(SA[1 2; 3 4], dims=(1,)) fails
```
DhairyaLGandhi, line 54: Is there a downside to passing the tuple?

mcabbott: Did you try the example provided on precisely the line you quoted?

DhairyaLGandhi: Oh the comment goes beyond the dims. Still, that seems like a SA issue. Do they have a corresponding issue to support the syntax? Do you have a link?

DhairyaLGandhi: Found https://github.com/JuliaArrays/StaticArrays.jl/pull/909

DhairyaLGandhi: Yeah, so we can avoid this line in that case and SA can make a release that fixes the bug.

src/compiler/chainrules.jl
```diff
@@ -123,11 +123,35 @@ Convert `x` from the format Zygote uses internally to differentials types ChainR
...
+  _project(x, dx)
+
+Uses `ChainRulesCore.ProjectTo` to standardise the gradient `dx` for type & shape.
+Also handles some Zygote-specific corrections, such as `x::Array, dx::Tuple`.
+Safe to apply to arbitrary input.
+"""
+@inline function _project(x::Union{Numeric, Ref{<:Numeric}}, dx)
+  wrap_chainrules_output(ProjectTo(x)(wrap_chainrules_input(dx)))
+end
+_project(x::AbstractArray, dx) = dx isa AbstractArray ? reshape(dx, axes(x)) : dx
```
DhairyaLGandhi, line 142: This can be broken down into a different method

src/compiler/interface.jl
```diff
@@ -68,15 +68,20 @@ julia> gradient([7, 11], 0, 1) do x, y, d
...
-([14.0, 22.0], 2, nothing)
+([14.0, 22.0], 2.0, nothing)
 ```
 """
 function gradient(f, args...)
   y, back = pullback(f, args...)
-  return back(sensitivity(y))
+  grad = back(sensitivity(y))
+  isnothing(grad) ? nothing : map(_project, args, grad)
 end
```
DhairyaLGandhi, line 78: You can add a method to `_project` and avoid this change

DhairyaLGandhi (changes requested):

oxinabox:
Is there a reason not to pull all of broadcasting down into ChainRules.jl?
Probably combined with setting up Zygote to claim that ForwardDiff is it's ForwardMode AD?

DhairyaLGandhi:
Why don't we give Zygote.Forward more love? It's better for neural networks.

mcabbott:
> Is there a reason not to pull all of broadcasting down into ChainRules.jl?

One reason not to is that Zygote's un-fused broadcast might not be the last word here. Maybe you can write a fused forward broadcast in Diffractor which would be hopelessly slow here. I think there's a lot of exploring left to be done. Unlike the basic rules in ChainRules, where we can write a pretty close to optimal rule once & let everything use it.

Anyway this PR has much more modest goals. In the linked Flux issues it comes pretty close to entirely removing the penalty for mixing up your eltypes. And it fixes a lot of Zygote issues about real/complex.

DhairyaLGandhi:
Mixing eltypes is going to get really important with low precision work picking up the pace. We shouldn't have to write custom passes for every operation related to 16 bit floats

mcabbott:
Yes there have been rumours of mixed-precision training for ages. I don't see any obvious problem though. It does not involve randomly mixing types and hoping that Julia's promotion will figure it out.

Complex/real has been discussed at great length. This PR really isn't the place to argue it; if you think it's wrong you should open an issue on ChainRulesCore and make your case.

> Wouldn't we expect complex numbers to have gradients with complex types?

Err, they do? There would be a lot of broken tests if that were altered. I think you may have misunderstood what problem this projection solves. The first message has examples, and links to issues closed.

src/compiler/chainrules.jl
mcabbott: Can you write exactly what method you prefer? There are obviously always other ways to write things.

src/compiler/interface.jl
mcabbott: > You can add a method to _project and avoid this change

Can you write exactly what method that would be?

src/compiler/chainrules.jl
```diff
@@ -123,11 +123,33 @@ Convert `x` from the format Zygote uses internally to differentials types ChainR
 """
 @inline wrap_chainrules_input(x) = x
 @inline wrap_chainrules_input(::Nothing) = ChainRules.ZeroTangent()
+@inline wrap_chainrules_input(::AbstractArray{Nothing}) = ChainRules.ZeroTangent()
```
oxinabox, line 126: This takes us another step closer to #1057 
(but avoids the worst cases of it) still

src/compiler/chainrules.jl
```diff
@@ -123,11 +123,33 @@ Convert `x` from the format Zygote uses internally to differentials types ChainR
...
+
+# Restore splatted arrays
+_project(x::AbstractArray, dx::Tuple) = _project(x, reshape(collect(dx), axes(x)))
+
+# Piracy:
+# wrap_chainrules_input doesn't handle array of Union{Int,Nothing}
+(::ChainRulesCore.ProjectTo)(::Nothing) = ChainRulesCore.NoTangent()
+
+# CRC likes Tangent{<:Complex}, but Zygote makes Tangent{Any}
+(project::ProjectTo{<:Complex})(dx::Tangent) = project(Complex(dx.re, dx.im))
```
oxinabox, line 151: we could probably relax that one in CRC really.

src/compiler/chainrules.jl
```diff
@@ -123,11 +123,33 @@ Convert `x` from the format Zygote uses internally to differentials types ChainR
...
+@inline function _project(x, dx)
+  wrap_chainrules_output(ProjectTo(x)(wrap_chainrules_input(dx)))
+end
+
+# Restore splatted arrays
+_project(x::AbstractArray, dx::Tuple) = _project(x, reshape(collect(dx), axes(x)))
+
+# Piracy:
+# wrap_chainrules_input doesn't handle array of Union{Int,Nothing}
+(::ChainRulesCore.ProjectTo)(::Nothing) = ChainRulesCore.NoTangent()
```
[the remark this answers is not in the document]
mcabbott: I think that didn't work. CRC was trying to do `map((f,x) -> f(x), project.elements, [1,2,nothing])` and this doesn't call `_project`.

src/compiler/chainrules.jl
mcabbott: Ok. But this should have no bad side effects right? No extra memory, no conflict with doing #1057 later?

src/compiler/chainrules.jl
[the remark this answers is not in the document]
mcabbott: This could be in CRC I guess, harmless to accept `nothing` since what else could it mean? But maybe you'd rather that be an error in general?

test/gradcheck.jl
```diff
@@ -1403,6 +1403,17 @@ end
...
+
+  # Many of these tests check a complex gradient to a function with real input. This is now
+  # clamped to real by ProjectTo, but to run the old tests, use here the old gradient function:
+  function oldgradient(f, args...)
+    y, back = Zygote.pullback(f, args...)
+    back(Zygote.sensitivity(y))
+  end
+  # Eventually these rules and tests will be moved to ChainRules.jl, at which point the tests
+  # can be updated to use real / complex consistently.
+  # https://github.com/JuliaMath/AbstractFFTs.jl/pull/58
```
oxinabox, line 1415: Normally I would not be OK with this defining `oldgradient` as it is a super-gross hack.
But if we can outright just delete all these tests entirely once that PR is merge than I am down with it.

src/compiler/chainrules.jl
oxinabox: Correct.

test/gradcheck.jl
mcabbott: Yea I lost quite a few hours trying to fix these. They are pretty convoluted tests.

src/compiler/chainrules.jl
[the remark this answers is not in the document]
mcabbott: Yea so then better here as piracy. Tests of CR etc. will run without this, and catch stray `nothing`s. It's a bit ugly but not the worst thing in Zygote for sure.

src/compiler/chainrules.jl
mcabbott: I can if you like. But do we want it allowed in general?

src/lib/broadcast.jl
```diff
@@ -45,18 +45,20 @@ function Base.reducedim_init(::typeof(identity), ::typeof(accum), A::AbstractArr
...
-  size(x) == size(x̄) ? x̄ :
-  length(x) == length(x̄) ? trim(x, x̄) :
-    trim(x, accum_sum(x̄, dims = ntuple(i -> size(x, i) == 1 ? i : ndims(x̄)+1, Val(ndims(x̄)))))
-
+function unbroadcast(x::AbstractArray, x̄)
+  N = ndims(x̄)
+  if length(x) == length(x̄)
+    _project(x, x̄)  # ProjectTo handles reshape, offsets, structured matrices, row vectors
+  else
+    tup = filter(d -> size(x, d) == 1, ntuple(identity, N))
```
oxinabox, line 53:
```suggestion
    tup = filter(d -> size(x, d) == 1, ntuple(identity, N))  # same as `Tuple(findall(==(1), size(x)))` but carefully optimized
```

mcabbott: If it needs a comment, can it be even less optimised? 
```suggestion
    tup = filter(d -> size(x, d) == 1, ntuple(identity, N))  # like findall(size(x) .== 1) but optimised 
```

oxinabox: If we didn't need a `Tuple` here then it is much faster if we can do `findall==(1), size(x))`

```julia
julia> const x = rand(3,1,1,2,1);
WARNING: redefinition of constant x. This may fail, cause incorrect answers, or produce other errors.

julia> @btime filter(d -> size(x, d) == 1, ntuple(identity, $(ndims(x))))
  118.460 ns (3 allocations: 80 bytes)
(2, 3, 5)

julia> @btime Tuple(findall(==(1), size($x)))
  154.346 ns (4 allocations: 240 bytes)
(2, 3, 5)

julia> @btime findall(==(1), size($x))
  54.909 ns (3 allocations: 208 bytes)
3-element Vector{Int64}:
 2
 3
 5
```

oxinabox (approved):
Other than a few last comments this seem good to me.

Might want to wait for @DhairyaLGandhi  to reply to the comments he made initially when this was openned, but it has been a while waiting.
I will ping him on slack.
Maybe give him a day or so, then if he is too busy we can merge and make follow ups later.

src/lib/broadcast.jl
mcabbott: The weird hack of `ndims() + 1` is still the fastest. We could revert to that. My hope was to let more things work with StaticArrays but maybe we settle for not making things worse. 
```
julia> @btime unbroadcast($(rand(1,3)), $(rand(3,3)));  # original
  154.998 ns (5 allocations: 224 bytes)

julia> @btime unbroadcast2($(rand(1,3)), $(rand(3,3)));  # as in PR
  889.191 ns (3 allocations: 112 bytes)

julia> @btime unbroadcast3($(rand(1,3)), $(rand(3,3)));  # with size(x) .== 1
  198.840 ns (8 allocations: 432 bytes)
```

src/lib/broadcast.jl
```diff
@@ -45,18 +45,20 @@ function Base.reducedim_init(::typeof(identity), ::typeof(accum), A::AbstractArr
...
+    dims = length(tup) == 1 ? first(tup) : tup  # avoid sum(xbar, dims=(1,)) as e.g. sum(SA[1 2; 3 4], dims=(1,)) fails
```
mcabbott, line 54:
```suggestion
    dims = ntuple(d -> size(x, d) == 1 ? d : ndims(x̄)+1, ndims(x̄))
```

src/lib/broadcast.jl
oxinabox: lets wrap whatever we come up with in a function like `singleton_dims(x)` or something, so we can encapsulate that complexity away form the implementation of `unbroadcast`

mcabbott: For now I've put it back how it was; we can tackle some tidy up on another day.

src/compiler/interface.jl
DhairyaLGandhi: Something like `_project(x, ::Nothing) = nothing` maybe

test/cuda.jl
```diff
@@ -6,7 +6,11 @@ CUDA.allowscalar(false)
 # Test GPU movement inside the call to `gradient`
 @testset "GPU movement" begin
   r = rand(Float32, 3,3)
-  @test gradient(x -> sum(cu(x)), r)[1] isa Array{Float32, 2}
+  if VERSION < v"1.6"
+    @test gradient(x -> sum(cu(x)), r)[1] isa Matrix{Float32}
+  else
+    @test_broken gradient(x -> sum(cu(x)), r)[1] isa Matrix{Float32}
```
DhairyaLGandhi, line 12: Is this still the case? Should be fixed on master

mcabbott: Indeed, looks like it is fixed.

src/compiler/interface.jl
mcabbott: This is easy to try:
```
julia> _project(x, ::Nothing) = nothing
_project (generic function with 1 method)

julia> map(_project, (1,2,3), nothing)
ERROR: MethodError: no method matching length(::Nothing)
```

DhairyaLGandhi:
I did respond on the slack where I'd mentioned wanting to take a look at it today.

PR #1047, opened 2021-07-29
remove artifact warning

mzgubic:
I’ve copied most of the `zygote2differential` code from an old PR where I tried to make Zygote use ChainRules types internally. It shouldn’t hit that ever anyway, because `z2d` transforms zygote differential types (like `nothing` or `namedtuple`s) into the ChainRules differential types (e.g. `ZeroTangent()` and `Tangent{Primal}`.

### DIFF
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -194,7 +194,7 @@ z2d(x, ::Any) = x
 z2d(::Nothing, ::Any) = NoTangent()
 z2d(a::AbstractArray{<:Number}, primal::AbstractArray{T}) where T = a
 z2d(a::AbstractArray, primal::AbstractArray{T}) where T = z2d.(a, primal)
-z2d(x::Union{AbstractZero, Tangent}, ::Any) = (difftype_warn(x); return x)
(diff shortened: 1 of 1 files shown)


mcabbott (approved):
Looks fine

CarloLucibello:
ahah, double borsed

src/compiler/chainrules.jl
```diff
@@ -194,7 +194,7 @@ z2d(x, ::Any) = x
 z2d(::Nothing, ::Any) = NoTangent()
 z2d(a::AbstractArray{<:Number}, primal::AbstractArray{T}) where T = a
 z2d(a::AbstractArray, primal::AbstractArray{T}) where T = z2d.(a, primal)
-z2d(x::Union{AbstractZero, Tangent}, ::Any) = (difftype_warn(x); return x)
+z2d(x::Union{AbstractZero, Tangent}, ::Any) = return x
```
oxinabox, line 197:
```suggestion
# Note: this should never be hit if we are converting things right, but it seems to be
# happening in the wild for sufficiently weird functions/types.
# This fixes most (all?) cases, but it would be good to find what we miss.
z2d(x::Union{AbstractZero, Tangent}, ::Any) = return x
```

PR #1049, opened 2021-07-29
nograd XOR RNGs

DhairyaLGandhi:
Maybe we should do this in RandomNumbers itself, if it can take a ChainRulesCore dep.

cc @ChrisRackauckas

### DIFF
--- a/src/Zygote.jl
+++ b/src/Zygote.jl
@@ -60,6 +60,10 @@ end
   @nograd Colors.ColorTypes._parameter_upper_bound
 end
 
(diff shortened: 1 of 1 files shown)


darsnack:
Also https://github.com/JuliaDiff/ChainRules.jl/pull/484

darsnack:
I believe this is no longer needed with https://github.com/JuliaDiff/ChainRules.jl/pull/484?

ToucheSir:
I believe the backport in https://github.com/JuliaDiff/ChainRules.jl/pull/488 should handle stable Zygote too.

PR #1051, opened 2021-08-02
Remove rules for Array, vect, circshift, fill, reverse -- ChainRules 1.5

willtebbutt:
New rules added in CR-1.5 should make the rules removed here redundant. I've left the corresponding tests to ensure nothing breaks.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -24,7 +24,7 @@ ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 [compat]
 AbstractFFTs = "0.5, 1.0"
 ChainRules = "1.5"
-ChainRulesCore = "1.0.1"
+ChainRulesCore = "1.1"
 ChainRulesTestUtils = "1"
 DiffRules = "1.0"
 FillArrays = "0.8, 0.9, 0.10, 0.11, 0.12"
--- a/Project.toml
+++ b/Project.toml
@@ -23,7 +23,7 @@ ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 
 [compat]
 AbstractFFTs = "0.5, 1.0"
-ChainRules = "1"
+ChainRules = "1.5"
 ChainRulesCore = "1.0.1"
 ChainRulesTestUtils = "1"
 DiffRules = "1.0"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.6.18"
+version = "0.6.19"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -3,14 +3,10 @@ using FillArrays: AbstractFill, getindex_value
 using Base.Broadcast: broadcasted, broadcast_shape
 using Distributed: pmap, AbstractWorkerPool
 
-@adjoint (::Type{T})(::UndefInitializer, args...) where T<:Array = T(undef, args...), Δ -> nothing
-
 @adjoint Array(xs::AbstractArray) = Array(xs), ȳ -> (ȳ,)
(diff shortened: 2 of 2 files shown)


willtebbutt:
GPU failure doesn't seem to be related to this PR.

DhairyaLGandhi:
Pretty weird that 1.5 errors. 

Could you link to the rules in CR? Mostly lgtm.

willtebbutt:
> Could you link to the rules in CR?

Where / in what format would you like me to do this?

mcabbott:
Could remove most of the `cat` rules too, although not sure this one test will pass on CR's rules, so `vcat` may have to stay:

https://github.com/FluxML/Zygote.jl/blob/master/test/cuda.jl#L105

Could also remove `repeat`, the rules here appear to be in CR:

https://github.com/JuliaDiff/ChainRules.jl/blob/master/src/rulesets/Base/array.jl#L63

willtebbutt:
@mcabbott I would prefer to keep the scope of this PR to the changes introduced in CR 1.5 -- would you mind if we made those other changes in a different PR?

mcabbott:
Sure, no obligation to do more.

Project.toml
```diff
@@ -23,7 +23,7 @@ ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 
 [compat]
 AbstractFFTs = "0.5, 1.0"
-ChainRules = "1"
+ChainRules = "1.5"
 ChainRulesCore = "1.0.1"
```
mcabbott, line 27: Probably doesn't matter, but CR 1.5 needs CRC 1.1
```suggestion
ChainRulesCore = "1.1"
```

mcabbott (approved):

src/lib/array.jl
```diff
@@ -3,22 +3,17 @@ using FillArrays: AbstractFill, getindex_value
...
 
-@adjoint Base.vect(xs...) = Base.vect(xs...), Δ -> (Δ...,)
-
 @adjoint copy(x::AbstractArray) = copy(x), ȳ -> (ȳ,)
 
 @adjoint collect(x::Tuple) = collect(x), dy -> (Tuple(dy),)
 @adjoint collect(x::AbstractArray) = collect(x), dy -> (dy,)
 
 # Array Constructors
-@adjoint (::Type{T})(x::T) where T<:Array = T(x), ȳ -> (ȳ,)
```
DhairyaLGandhi, line 21: Do we have an equivalent in CR?

src/lib/array.jl
```diff
@@ -106,17 +101,6 @@ end
 
 @adjoint collect(x::Array) = collect(x), Δ -> (Δ,)
 
-@adjoint fill(x::Real, dims...) = fill(x, dims...), Δ->(sum(Δ), map(_->nothing, dims)...)
-
-@adjoint function circshift(A, shifts)
-  circshift(A, shifts), Δ -> (circshift(Δ, map(-, shifts)), nothing)
-end
-
-@adjoint function reverse(x::AbstractArray, args...; kwargs...)
```
DhairyaLGandhi, line 115: Ditto here - this is needed for bidirectional recurrent cells

DhairyaLGandhi (approved):
Lgtm, couple of quick qs before merging

src/lib/array.jl
willtebbutt: Added in 1.5: https://github.com/JuliaDiff/ChainRules.jl/blob/7593339839832ffb7ac83401918390ffe9d4eb42/src/rulesets/Base/array.jl#L314

The tests appear to pass -- are there not tests for `reverse` in Zygote?

src/lib/array.jl
willtebbutt: Added in 1.5: https://github.com/JuliaDiff/ChainRules.jl/commit/7593339839832ffb7ac83401918390ffe9d4eb42#diff-7a31ac47520a0ad235f15cead3f4c48fcb595dc7f2421ebbee8ffce23e2510f5R7

src/lib/array.jl
DhairyaLGandhi: Ah we are still testing, excellent?

PR #1052, opened 2021-08-02
Bump patch

willtebbutt:
As per discussion with @oxinabox on slack.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.6.18-DEV"
+version = "0.6.18"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"


DhairyaLGandhi:
Hmm ci seems to be out to lunch.
bors try

DhairyaLGandhi (approved):

willtebbutt:
@DhairyaLGandhi would you like me to wait until a successful CI run has occured before merging and tagging?

DhairyaLGandhi:
I've retriggered BK, hopefully its alright on 1.5 and we should be good to merge

willtebbutt:
The buildkite failures are weird -- the tests pass, it's seems to fall over after they've completed...

DhairyaLGandhi:
Yeah I wonder if we should remove the coverage to see if that would help. I wouldn't want to remove testing on 1.5 over this esp when the package runs fine. It seems to be when `gc` is triggered. We can also just merge for now and check back later. 

@maleadt did something change in BK?

maleadt:
Yes, cc @DilumAluthge, I'll revert the change.

PR #1055, opened 2021-08-19
Support kwargs in `rrule_via_ad`

mzgubic:
Adds support for kwargs in `rrule_via_ad`. cc @niklasschmitz @oxinabox

### DIFF
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -175,8 +175,9 @@ As per [`chain_rrule`](@ref) but with support for kwargs.
 end
 
 
-function ChainRulesCore.rrule_via_ad(config::ZygoteRuleConfig, f, args...)
-    y, pb = _pullback(config.context, f, args...)
+function ChainRulesCore.rrule_via_ad(config::ZygoteRuleConfig, f, args...; kwargs...)
+    kwf(args...) = f(args...; kwargs...)
+    y, pb = _pullback(config.context, kwf, args...)
     ad_pullback(Δ) = zygote2differential(pb(wrap_chainrules_output(Δ)), (f, args...))
     return y, ad_pullback
 end
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -180,7 +180,7 @@ function ChainRulesCore.rrule_via_ad(config::ZygoteRuleConfig, f_args...; kwargs
     y, pb = if !isempty(kwargs)
         kwf() = first(f_args)(Base.tail(f_args)...; kwargs...)
         _y, _pb = _pullback(config.context, kwf)
-        _y, Δ -> only(_pb(Δ)).f_args
+        _y, Δ -> first(_pb(Δ)).f_args  # `first` should be `only`
     else
         _pullback(config.context, f_args...)
     end
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -177,9 +177,19 @@ end
 
 function ChainRulesCore.rrule_via_ad(config::ZygoteRuleConfig, f, args...; kwargs...)
     kwf(args...) = f(args...; kwargs...)
-    y, pb = _pullback(config.context, kwf, args...)
+    y, _pb = _pullback(config.context, kwf, args...)
+    pb = _isclosure(f) ? _unwrap_closure ∘ _pb : _pb
     ad_pullback(Δ) = zygote2differential(pb(wrap_chainrules_output(Δ)), (f, args...))
     return y, ad_pullback
+end
+
+function _isclosure(f)
+    f isa Type && return false # constructor
+    return fieldcount(typeof(f)) > 0 ? true : false
+end
+function _unwrap_closure(pbres)
+    # take :f of first element to extract the tangent of the inner closure
+    return ntuple(i -> i==1 ? pbres[i][:f] : pbres[i], length(pbres))
 end
 
 """
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -185,9 +185,7 @@ function ChainRulesCore.rrule_via_ad(config::ZygoteRuleConfig, f_args...; kwargs
         _pullback(config.context, f_args...)
     end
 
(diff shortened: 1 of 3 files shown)


niklasschmitz:
Thanks @mzgubic !

mzgubic:
The issue seems to be that hacking around `_pullback` not accepting kwargs using `kwf` function falls over for functors:
```julia
julia> struct MyStruct
           x::Vector{Float64}
       end

julia> (ms::MyStruct)(a; dims=:) = ms.x[1] + a

julia> function reproduce(f, args...; kwargs...)
           @show f
           @show args
           kwf(args...) = f(args...; kwargs...)
           y, pb = Zygote._pullback(kwf, args...)
           @show y
           @show pb(1.0)
           return nothing
       end
reproduce (generic function with 1 method)

julia> ms = MyStruct([1.0, 2.0, 3.0])
MyStruct([1.0, 2.0, 3.0])

julia> reproduce(ms, 1.0; dims=1)
f = MyStruct([1.0, 2.0, 3.0])
args = (1.0,)
y = 2.0
pb(1.0) = ((kwargs = nothing, f = (x = [1.0, 0.0, 0.0],)), 1.0)
```
where we wanted the last line to in fact be
```julia
pb(1.0) = ((x = [1.0, 0.0, 0.0],), 1.0)
```
I don't think this is a bug because `kwf` is also a closure over `kwargs` and `f`, it's just annoying in this case because we don't care about `kwargs`.

I can think of two ways of getting around this:
a) an annoying hack which checks if something is a closure and extracts the right part
b) support passing kwargs to `_pullback` (but not differentiating w.r.t. them) which removes the need for `kwf` in the first place.

I don't now how hard it would be to have b), we may have reasons against providing that? a) is probably pretty easy

oxinabox:
It is pretty funky that if `f` is a closure/functor,  we get that named tuple first arg with both, 
but if it's not we get `nothing` as first return.

```julia
julia> ff(x; kwargs...) = first(x)
ff (generic function with 1 method)

julia> reproduce(ff, 1.0; dims=1)
f = ff
args = (1.0,)
y = 1.0
pb(1.0) = (nothing, 1.0)
```

I am guessing the reason for this is that zygote collapes named tuples where all elements are `nothing` to just a `nothing`.
So actually before the collapsing step would have contained `(kwargs=nothing, f=nothing)`

Anyway,
related to this, rather than detecting he closure in advance, we can just deal with the consiquence.
Either we get back `nothing` and should return `nothing`
or we get back a named tuple with a `f` field and should return that field.

Another way to write the unwrap_closure code to to that is
```julia
_unwrap_kwf_pullback(Δ::Tuple{Nothing,Vararg}) = (nothing, Base.tail(Δ)...)
_unwrap_kwf_pullback(Δ::Tuple{NamedTuple,Vararg}) = (first(Δ).f, Base.tail(Δ))
```
which I think should actually be pretty fast because `Base.tail` is optimized for tuples to return a tuple, and splatting tuples into tuples is I think also optimized.

mzgubic:
Nice, its much cleaner than the one I had initially.

Now there are some inference issues
```julia

julia> function reproduce2(f_args...; kwargs...)
           kwf() = first(f_args)(Base.tail(f_args)...; kwargs...)
           y, pb = Zygote._pullback(kwf)
           @inferred pb(1.0)
       end
reproduce2 (generic function with 1 method)

julia> reproduce2(sum, rand(4); dims=1)
ERROR: return type Tuple{NamedTuple{(:kwargs, :f_args), Tuple{Nothing, Tuple{Nothing, Vector{Float64}}}}} does not match inferred return type Tuple{Any}
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] reproduce2(::Function, ::Vararg{Any, N} where N; kwargs::Base.Iterators.Pairs{Symbol, Int64, Tuple{Symbol}, NamedTuple{(:dims,), Tuple{Int64}}})
   @ Main ./REPL[219]:4
 [3] top-level scope
   @ REPL[220]:1
```

I tried doing the functor dance like we did in projections, but it didn't work (it's not the same issue anyway). Just for reference, it was
```julia
struct Kwf
    f_args::Tuple
    kwargs::NamedTuple
end
(kwf::Kwf)() = first(kwf.f_args)(Base.tail(kwf.f_args)...; kwf.kwargs...)
```

mzgubic:
This seems to work, hopefully CI agrees.

mzgubic:
nightly and GPU errors appear unrelated? I don't really know much about either

src/compiler/chainrules.jl
```diff
@@ -175,23 +175,16 @@ As per [`chain_rrule`](@ref) but with support for kwargs.
...
-function ChainRulesCore.rrule_via_ad(config::ZygoteRuleConfig, f, args...; kwargs...)
-    kwf(args...) = f(args...; kwargs...)
-    y, _pb = _pullback(config.context, kwf, args...)
-    pb = _isclosure(f) ? _unwrap_closure ∘ _pb : _pb
-    ad_pullback(Δ) = zygote2differential(pb(wrap_chainrules_output(Δ)), (f, args...))
+function ChainRulesCore.rrule_via_ad(config::ZygoteRuleConfig, f_args...; kwargs...)
+    kwf() = first(f_args)(Base.tail(f_args)...; kwargs...)
+    y, pb = _pullback(config.context, kwf)
+    function ad_pullback(Δ)
+        pbres = only(pb(wrap_chainrules_output(Δ))).f_args
```
oxinabox:
```suggestion
        # Extract tuple of tangents for f_args. (leave the ones for other closed field kwargs behind)
        pbres = only(pb(wrap_chainrules_output(Δ))).f_args
```

`only` was introduced in julia 1.4, but I guess it works here because of having Compat.jl loaded.
(ChainRulesCore loads that)

oxinabox (approved):

oxinabox:
Yeah, GPU is also failing for this totally different PR
https://buildkite.com/julialang/zygote-dot-jl/builds/568#0cb347e5-aca5-477a-8168-d9c5d6fd7fb8
so it does indeed seem unrelated.

And Nightly is some kinda of issues with the IRTransform

PR #1057, opened 2021-08-25
Use zygote2differential to wrap chainrules inputs

oxinabox:
@mzgubic  implemented zygote2differential as a better version of wrap_chainrules_inputs and added it to use in the code for `rrule_via_ad`.
But it was not added to the normal path for when Zygote uses ChainRules.
I guess because it requires keeping the primal values in memory.
Which is probably a lot?

Anyway this would give us more consistent chainrules types.
No more `Tangent{Any}` or `nothings` that are hidden with-in arrays.

We probably do not want to merge this as is because of the extra memory use.
or maybe it is not too bad. Do we have a benchmark for it?

But hopefully this will fix the problems in https://github.com/TuringLang/DistributionsAD.jl/pull/197
cc @devmotion .
If it does we can look at reworking `zygote2differential` to not have to store so much.
We learnt a lot about doing that for `ProjectTo`
same techniques can be applied here.

NB: I am putting this PR up at 9:30 at night, and I have not even run it locally.
Might have typos etc and just not work.
It also has no tests, yet.

### DIFF
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -123,8 +123,8 @@ Wrapper for a ChainRules pullback `back`, that causes it to follow Zygote conven
 (A functor here is used rather than a closure to avoid boxing issues);
 """
 struct ZBack{F,P} <: Function
-  primals::P
   back::F
+  primals::P
 end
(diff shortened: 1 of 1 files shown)


CarloLucibello:
closing as outdated

PR #1059, opened 2021-09-02
Support functions that splat namedtuples as keyword arguments

oxinabox:
@lsindoni ran into this.
This problem occurs for Dict and NamedTuples with a error about the methods of `Base.setindex`.

The `Dict` in the `pairs` pullback in this case ends up containing `Symbol`s not `Integer`s.
So we branch to support that.
(unfortunately it is a `Dict{Any,Any}` so we can't branch earlier, and so it is dynamic dispatch).

The chain to the rule for `pairs` is enough to make NamedTuple's work,
If we also want to support Dicts, we need the rule for `merge` as well

I am surprised noone has run into this before.

### DIFF
--- a/src/lib/base.jl
+++ b/src/lib/base.jl
@@ -133,6 +133,15 @@ end
   return pairs(t), pairs_namedtuple_pullback
 end
 
+# For merge between NamedTuple and Dict, we will just convert the Dict to a NamedTuple.
+# and then call `pullback`, which should overall be pretty efficient code generated,
+# and it avoids trying to AD the problematic generic `merge(::NamedTuple, ::iter)` method which uses `push!`.
+if VERSION >= v"1.6"
+  @adjoint merge(nt::NamedTuple, dict::Dict) = pullback(merge, nt, NamedTuple(dict))
+else
+  @adjoint merge(nt::NamedTuple, dict::Dict) = pullback(merge, nt, (;dict...))
+end
+
 @adjoint function Base.getfield(p::Pair, i::Int)
     function pair_getfield_pullback(Δ)
         f, s = i == 1 ? (Δ, nothing) : (nothing, Δ)
--- a/src/lib/base.jl
+++ b/src/lib/base.jl
@@ -118,17 +118,18 @@ end
 
 # named tuple
 @adjoint function pairs(t::NamedTuple{N}) where N
-
-  pairs_namedtuple_pullback(dx::NamedTuple) = (dx.data,)
   
+  pairs_namedtuple_pullback(dx::NamedTuple) = (dx.data,)
+
   function pairs_namedtuple_pullback(Δ::Dict)
     t0 = map(zero, t)
     for (idx, v) in Δ
-      t0 = NamedTuple{N}(Base.setindex((t0...,), v, idx))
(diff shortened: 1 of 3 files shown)


test/features.jl
```diff
@@ -458,6 +458,13 @@ end
     @test (x->10*(x => 2)[2])'(100) === 0
 end
 
+@testset "kwarg splatting, pass in object" begin
+  g(; kwargs...) = kwargs[:x] * kwargs[:z]
+  h(somedata) = g(; somedata...)
+  @test gradient(h, (; x=3.0, y=4.0, z=2.3)) == ((x = 2.3, y = 0.0, z = 3.0),)
+  @test gradient(h, Dict(:x=>3.0, :y=>4.0, :z=>2.3)) == ((y = 0.0, z = 3.0, x = 2.3),)
```
oxinabox, line 465: I am not sure what type Zygote wants to use to represent `Dict`.
`accum` isn't defined for `Dict` AFAICT.
So `NamedTuple` seemed reasonable.

src/lib/base.jl
```diff
@@ -122,13 +122,23 @@ end
...
     return pairs(t), pairs_namedtuple
 end
 
+# For merge between NamedTuple and Dict, we will just convert the Dict to a NamedTuple.
+# and then call `pullback`, which should overall be pretty efficient code generated,
+# and it avoids trying to AD the problematic generic `merge(::NamedTuple, ::iter)` method which uses `push!`.
+if VERSION >= v"1.6"
+  @adjoint merge(nt::NamedTuple, dict::Dict) = pullback(merge, nt, NamedTuple(dict))
+else
+  @adjoint merge(nt::NamedTuple, dict::Dict) = pullback(merge, nt, (;dict...))
```
DhairyaLGandhi, line 139: Just keep this?

DhairyaLGandhi (commented):
Thanks for the follow up!

src/lib/base.jl
oxinabox: What do you mean?
You mean use if for both 1.6 and pre-1.6?

DhairyaLGandhi:
The CUDA failures seem related to CUDA.jl v3.4 -- We should get some kind of fix in Flux/ Zygote or CUDA since `cu` seems to have changed behaviour in a couple places. @vchuravy would you know what needs to be done here?

src/lib/base.jl
DhairyaLGandhi: Yeah, exactly.

oxinabox: ok

oxinabox:
Bump

test/features.jl
CarloLucibello: the `Dict` story is not consistent at the moment and there are many missing features
https://github.com/FluxML/Zygote.jl/issues?q=is%3Aissue+is%3Aopen+label%3Adictionary
I guess generally the returned gradient should be a dictionary, but for dicts with symbol keys maybe a namedtuple is good enough for the time being

CarloLucibello (approved):

oxinabox:
CI failures are unrelated.

PR #1060, opened 2021-09-02
Return grads for structs with closed over params

DhairyaLGandhi:
This can cause issues with double counting, so I want to see what CI says in terms of breakage

### DIFF
--- a/src/lib/lib.jl
+++ b/src/lib/lib.jl
@@ -55,8 +55,7 @@ end
   quote
     if haskey(cache(cx), x)
       cache(cx)[x] = accum(cache(cx)[x],Δ)
-      @show cache(cx)[x]
-      return cache(cx)[x]
+      return
     else
       return Δ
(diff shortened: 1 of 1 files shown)


CarloLucibello:
Related to #767 #717

Why do we need `accum_param` at all, can't we get rid of it?

src/lib/lib.jl
```diff
@@ -109,6 +109,16 @@ using Base: tail
   val, back
 end
 
+@adjoint function literal_getindex(xs::Dict, ::Val{k}) where k
+  val = xs[k]
+  ks = keys(xs)
+  function back(Δ)
+    accum_param(__context__, val, Δ) === nothing && return
```
CarloLucibello, line 116: Why do we need `accum_param` at all, can't we get rid of it?

CarloLucibello:
closing as stale

PR #1069, opened 2021-09-09
Add short-circuit to `rrule_via_ad`

mcabbott:
This makes some cases I tested much quicker, and none much slower. 

What would be a great test of whether this adds overhead sometimes?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.6.20"
+version = "0.6.21"
 
 [deps]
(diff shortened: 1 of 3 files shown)


oxinabox:
Yes, I think this should be fast.
The place it might cause slow down still is nested AD.

oxinabox:
Yes, I think this should be fast.
It should always optimise away.
Can you check the code typed?

The place it might cause slow down still is nested AD.

mcabbott:
Will try some cases. 

Does it need ` return y, ZBack(back)`, btw?

oxinabox (approved):
This is the correct code.

Maybe it could do with a test?

Issue #1071, opened 2021-09-19
error with NamedTuple constructor for iterator


CarloLucibello:
```julia
julia> x = [:a=>1, :b=>2]
2-element Vector{Pair{Symbol, Int64}}:
 :a => 1
 :b => 2

# This is working as expected
julia> gradient(x ->  x[1].second, x)
(Union{Nothing, NamedTuple{(:first, :second), Tuple{Nothing, Int64}}}[(first = nothing, second = 1), nothing],)

julia> gradient(x ->  NamedTuple(x).a, x)
ERROR: Mutating arrays is not supported -- called push!(::Vector{Any}, _...)
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:33
  [2] (::Zygote.var"#444#445"{Vector{Any}})(#unused#::Nothing)
    @ Zygote ~/.julia/packages/Zygote/ajuwN/src/lib/array.jl:78
  [3] (::Zygote.var"#2360#back#446"{Zygote.var"#444#445"{Vector{Any}}})(Δ::Nothing)
    @ Zygote ~/.julia/packages/ZygoteRules/OjfTt/src/adjoint.jl:59
  [4] Pullback
    @ ./namedtuple.jl:287 [inlined]
  [5] (::typeof(∂(merge)))(Δ::NamedTuple{(:a, :b), Tuple{Int64, Nothing}})
    @ Zygote ~/.julia/packages/Zygote/ajuwN/src/compiler/interface2.jl:0
  [6] Pullback
    @ ./namedtuple.jl:107 [inlined]
  [7] (::typeof(∂(NamedTuple)))(Δ::NamedTuple{(:a, :b), Tuple{Int64, Nothing}})
    @ Zygote ~/.julia/packages/Zygote/ajuwN/src/compiler/interface2.jl:0
  [8] Pullback
    @ ./REPL[44]:1 [inlined]
  [9] (::Zygote.var"#50#51"{typeof(∂(#37))})(Δ::Int64)
    @ Zygote ~/.julia/packages/Zygote/ajuwN/src/compiler/interface.jl:41
 [10] gradient(f::Function, args::Vector{Pair{Symbol, Int64}})
    @ Zygote ~/.julia/packages/Zygote/ajuwN/src/compiler/interface.jl:76
 [11] top-level scope
    @ REPL[44]:1
 [12] top-level scope
    @ ~/.julia/packages/CUDA/9T5Sq/src/initialization.jl:66
```

[...]

PR #1073, opened 2021-09-19
Add adjoint for NamedTuple + Generator

DhairyaLGandhi:
Check for #1071 

Currently this does not handle stateful `f` in the generator. Maybe we can lower to map there? It would be slightly slower


CarloLucibello:
closing as stale / unifinished / missing tests

PR #1075, opened 2021-09-23
Project `nothing` to `nothing`

DhairyaLGandhi:
Ref https://github.com/SciML/DiffEqSensitivity.jl/pull/493

### DIFF
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -145,7 +145,7 @@ _project(x::AbstractArray, dx::Tuple) = _project(x, reshape(collect(dx), axes(x)
 
 # Piracy:
 # wrap_chainrules_input doesn't handle array of Union{Int,Nothing}
-(::ChainRulesCore.ProjectTo)(::Nothing) = ChainRulesCore.NoTangent()
+(::ChainRulesCore.ProjectTo)(::Nothing) = nothing # ChainRulesCore.NoTangent()
 
 # CRC likes Tangent{<:Complex}, but Zygote makes Tangent{Any}
 (project::ProjectTo{<:Complex})(dx::Tangent) = project(Complex(dx.re, dx.im))


DhairyaLGandhi:
This is likely to break some tests, but seems like the cogent thing to do for some tests.

oxinabox:
I am not sure if this is the right place for this fix.
Or if we need a more general insertion of the code to convert CRC types in to Zygote types.
I think I saw it happening with other types than just Nothing.

@mcabbott  should review.

DhairyaLGandhi:
Since `_project` is the last transform Zygote makes now, it should generically output Zygote types.

src/compiler/chainrules.jl
```diff
@@ -106,6 +106,7 @@ Convert `x` from the differentials types ChainRules uses to the format Zygote us
 # Zygote convention: even if many AbstractZero partials (i.e. multi-input function), make just 1 nothing.
 @inline wrap_chainrules_output(x::Tuple{Vararg{ChainRules.AbstractZero}}) = nothing
 @inline wrap_chainrules_output(x::ChainRules.AbstractZero) = nothing
+@inlint wrap_chainrules_output(x::ChainRules.NoTangent) = nothing
```
mcabbott: This seems completely safe to me.

Re PR's original change  https://github.com/FluxML/Zygote.jl/pull/1075/commits/c53df992fa37faee417db245ca1d7c569d59b201 to propagate nothings, I was surprised this broke no tests. I don't recall now what test motivated that method in the first place, and it's possible that some later change removed the need.

DhairyaLGandhi:
ref https://github.com/SciML/DiffEqSensitivity.jl/pull/493#issuecomment-926969231

I believe we have this setup to handle `OneElement` which doesn't have information about the type of the array. Why do we have `OneElement`? Our accumulation speedup came from generalising `accum`, and incrementally through `OneElement`. Judging by the few number of times it coming up in issues, it might be that we can have a more stable and simpler implementation by removing it without losing much performance at all.

CarloLucibello:
closing as stale, possibly also not needed anymore

PR #1076, opened 2021-09-24
Update README

willtebbutt:
The current README is a bit misleading in terms of performance, because Zygote really doesn't have good performance when control flow is involved. I'm open to other suggestions for re-wording, but it seems reasonable to me that we temper what is currently there a bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ top:
 
 "Source-to-source" means that Zygote hooks into Julia's compiler, and generates the backwards pass for you – as if you had written it by hand.
 
-Without compromising on performance, Zygote supports the full flexibility and dynamism of the Julia language, including control flow, recursion, closures, structs, dictionaries, and more.
+Zygote supports the full flexibility and dynamism of the Julia language, including control flow, recursion, closures, structs, dictionaries, and more.
 
 ```julia
 julia> fs = Dict("sin" => sin, "cos" => cos, "tan" => tan);


DhairyaLGandhi:
Zygote uses the IR Julia produces, and the compiler can strip off the branching when it's not needed, so it doesn't need to carry it around unnecessarily. We need a compiler pass post optimisation too, I think that would massively change what all we can do it terms of handling code.

willtebbutt:
> Zygote uses the IR Julia produces, and the compiler can strip off the branching when it's not needed, so it doesn't need to carry it around unnecessarily. We need a compiler pass post optimisation too, I think that would massively change what all we can do it terms of handling code.

Certainly, you'll find no disagreement from me there, but that's future work, and doesn't reflect the current state of the package. Are you happy to update the readme as I propose, or is there something else that you would prefer?

DhairyaLGandhi:
I feel like we can add it. Zygote is pretty close to optimal on enough of Julia that I wouldn't call it slow, by any means. I think that message should get spoken about too.

willtebbutt:
> I feel like we can add it.

Sorry, add which bit? Please feel free to provide a suggestion :)

DhairyaLGandhi:
Add (or rather remove the statement) the suggestion in the PR.

willtebbutt:
Cool. If you approve, I'll merge :)

DhairyaLGandhi (approved):

PR #1079, opened 2021-09-24
ProjectTo{AbstractArray}(::Tangent{Any}) fix

### LINKED ISSUE #1078: _project error
The problem occurs when one tries to take the `Zygote.gradient` of a function an `AbstractArray` with a structural tangent.

`ProjectTo(my_array)` produces a `ProjectTo{AbstractArray}`, and Zygote winds up producing a `Tangent{Any}` for it. This means that [this](https://github.com/JuliaDiff/ChainRulesCore.jl/blob/2ec2549b73b22bc08f554dae864fb650cfb9c3d7/src/projection.jl#L139) method doesn't apply.

MWE:
```julia
julia> using ChainRules: Tangent, ProjectTo, NoTangent

julia> using KernelFunctions: MOInputIsotopicByFeatures

julia> x = MOInputIsotopicByFeatures(randn(3), 2);

julia> dx = Tangent{Any}(x=randn(3), out_dim=NoTangent());

julia> ProjectTo(x)(dx)
ERROR: MethodError: no method matching (::ProjectTo{AbstractArray, NamedTuple{(:elements, :axes), Tuple{Vector{typeof(identity)}, Tuple{Base.OneTo{Int64}}}}})(::Tangent{Any, NamedTuple{(:x, :out_dim), Tuple{Vector{Float64}, NoTangent}}})
Closest candidates are:
  (::ProjectTo{T, D} where D<:NamedTuple)(::Tangent{var"#s15", T} where {var"#s15"<:T, T}) where T at /Users/willtebbutt/.julia/packages/ChainRulesCore/ChM7X/src/projection.jl:139
  (::ProjectTo{AbstractArray, D} where D<:NamedTuple)(::Union{LinearAlgebra.Adjoint{T, var"#s832"}, LinearAlgebra.Transpose{T, var"#s832"}} where {T, var"#s832"<:(AbstractVector{T} where T)}) at /Users/willtebbutt/.julia/packages/ChainRulesCore/ChM7X/src/projection.jl:239
  (::ProjectTo{AbstractArray, D} where D<:NamedTuple)(::Number) at /Users/willtebbutt/.julia/packages/ChainRulesCore/ChM7X/src/projection.jl:245
  ...
Stacktrace:
 [1] top-level scope
   @ REPL[6]:1
```

What is the appropriate fix @mcabbott @oxinabox @mzgubic ? I can work around this by using `pullback` or `_pullback` directly, but it's obviously sub-optimal not to be able to use `gradient`.

Where's the discussion about adding the `_project` call [here](https://github.com/FluxML/Zygote.jl/blob/528e0be677d1feb9ccf6fc4ab298f4d8a106de10/src/compiler/interface.jl#L77)? I feel like I've seen it somewhere, but I'm not sure where...

willtebbutt:
Resolves #1078

### DIFF
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -151,6 +151,9 @@ _project(x::AbstractArray, dx::Tuple) = _project(x, reshape(collect(dx), axes(x)
 (project::ProjectTo{<:Complex})(dx::Tangent) = project(Complex(dx.re, dx.im))
 
 # CRC likes Tangent{AbstractArray}, but Zygote makes Tangent{Any}
+# in particular this would hit https://github.com/JuliaDiff/ChainRulesCore.jl/blob/2ec2549b73b22bc08f554dae864fb650cfb9c3d7/src/projection.jl#L139
+# if we were not losing track of the Primal in the Tangent
+# This type piracy is just giving up that safety check.
 (project::ProjectTo{AbstractArray})(dx::Tangent) = dx
 
 """
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -150,6 +150,9 @@ _project(x::AbstractArray, dx::Tuple) = _project(x, reshape(collect(dx), axes(x)
 # CRC likes Tangent{<:Complex}, but Zygote makes Tangent{Any}
 (project::ProjectTo{<:Complex})(dx::Tangent) = project(Complex(dx.re, dx.im))
 
+# CRC likes Tangent{AbstractArray}, but Zygote makes Tangent{Any}
+(project::ProjectTo{AbstractArray})(dx::Tangent) = dx
+
 """
   ZBack{F}(back) <: Function
 
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.6.22"
+version = "0.6.23"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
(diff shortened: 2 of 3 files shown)


src/compiler/chainrules.jl
```diff
@@ -150,6 +150,9 @@ _project(x::AbstractArray, dx::Tuple) = _project(x, reshape(collect(dx), axes(x)
 # CRC likes Tangent{<:Complex}, but Zygote makes Tangent{Any}
 (project::ProjectTo{<:Complex})(dx::Tangent) = project(Complex(dx.re, dx.im))
 
+# CRC likes Tangent{AbstractArray}, but Zygote makes Tangent{Any}
```
oxinabox, line 153:
```suggestion
# CRC likes Tangent{AbstractArray}, but Zygote makes Tangent{Any}
# in particular this would hit https://github.com/JuliaDiff/ChainRulesCore.jl/blob/2ec2549b73b22bc08f554dae864fb650cfb9c3d7/src/projection.jl#L139
# if we were not losing track of the Primal in the Tangent
```

oxinabox (approved):
A test might be good?

DhairyaLGandhi:
Should we merge #1075 as well?

willtebbutt:
>  But I am sure you can add one without me needed to rereview it.

But... there's already one there, is there not?

oxinabox:
> But... there's already one there, is there not?

Oops I missed it

oxinabox:
> Should we merge #1075 as well?

It is a different thing.
I understand exactly what this PR is fixing and am pretty sure it covers the whole case.
(The thing it fixes is not being able to hit https://github.com/JuliaDiff/ChainRulesCore.jl/blob/2ec2549b73b22bc08f554dae864fb650cfb9c3d7/src/projection.jl#L139)

#1075 seems like it is working around some Zygote types being in a place where ChainRules types are expected.
And that needs more understanding about why it is happening in the first place.
Which I don't have,  but hopefully @mcabbott  does (so I have pinged them there)

PR #1090, opened 2021-10-01
Improve Map Type Stability

willtebbutt:
As the title suggests, improves the type-stability of map. Does so by getting rid of a branch. Tests pass locally, and I've added a few more.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.6.24"
+version = "0.6.25"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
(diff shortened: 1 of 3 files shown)


DhairyaLGandhi (approved):
Lgtm, can we add another test case for non empty and empty collections too?

willtebbutt:
> Lgtm, can we add another test case for non empty and empty collections too?

Sorry, I'm not sure what kind of thing you have in mind here -- could you elaborate?

willtebbutt:
Hmmm confusingly, I can't replicate the error I'm seeing locally...

willtebbutt:
> Sorry, I'm not sure what kind of thing you have in mind here -- could you elaborate?

Added additional tests for type-stability.

> Hmmm confusingly, I can't replicate the error I'm seeing locally...

Have fixed this and added unit tests designed to catch the specific edge case on which map was failing.

Will bors when I know that CI passes.

PR #1096, opened 2021-10-07
Support 0-dimensional array indexing

ChrisRackauckas:
MWE:

```julia
x = Array{Float64,0}(undef)
x[1] = 0.7
axes(x) # ()
dx = Zygote.OneElement(1.0, x, axes(x)) # fails
Zygote.gradient(x->x[1],x) # fails
```

It's somewhat of an edge case found in https://github.com/SciML/DiffEqSensitivity.jl/pull/498

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -28,7 +28,9 @@ end
 @adjoint view(x::AbstractArray, inds...) = view(x, inds...), ∇getindex(x, inds)
 
 ∇getindex(x::AbstractArray{T,N}, inds) where {T,N} = dy -> begin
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
6.25? I don't see a later tag.

CarloLucibello:
I erroneously bumped the version and the registration was unsuccessful. Registering again right now

ChrisRackauckas:
Okay yeah, I see it working on master. I'll wait on that tag and rerun https://github.com/SciML/DiffEqSensitivity.jl/pull/498 . 

Would there still be a performance benefit to avoid the OneElement in this case though?

PR #1102, opened 2021-10-15
Make `wrap_chainrules_input` understand `mutable struct`s

mcabbott:
For the gradient of a mutable struct, Zygote will make `Base.RefValue{Any}((value = 7.0,))` instead of just a NamedTuple. This was not handled at all by the code for converting to ChainRules types, causing Zygote's types to appear in unexpected places.

This PR makes `wrap_chainrules_input` send them to the same `Tangent` used for immutable `struct`s. This means that `wrap_chainrules_output` back to Zygote's types will create an ordinary NamedTuple, without the Ref. This doesn't seem to cause problems in the examples I could think of.

The simplest mutable struct is `Ref`, for which there appear to be no tests at all, apart from its use in broadcasting. Except one mutation test, which still works.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.6.27"
+version = "0.6.28"
 
 [deps]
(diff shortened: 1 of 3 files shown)


oxinabox:
(on your head be it :grin: )

PR #1104, opened 2021-10-16
Insert `_project` into `getproperty`'s gradient, and then improve `z2d` etc. to restore stability

mcabbott:
The goal here is to avoid things like this:
```julia
julia> pullback(x -> abs2(x.im * ((1-im) * x).re), 4+5im)[2](1.0)
# Zygote v0.6.0:
((re = 450.0 + 450.0im, im = 810.0),)
# Zygote v0.6.28:
ERROR: MethodError: no method matching Complex(::ComplexF64, ::ZeroTangent)
# this PR:
(450.0 + 1260.0im,)
```
While projection was intended as a mechanism to encode & enforce mathematical properties like real vs complex, I think it's also a good way to encode (when desired) a preference for the "natural gradient" of a complex number, instead of a NamedTuple, because this can participate in other stages of the calculation. One example of this found in the wild was fixed in https://github.com/JuliaDiff/ChainRules.jl/pull/509, a made-up example in the same spirit is:
```julia
julia> gradient(x -> sqrt(sum(parent(x .^ 2))), Diagonal([1,2,3]))[1]
# latest tagged:
ERROR: ArgumentError: broadcasting over dictionaries and `NamedTuple`s is reserved
# this PR + CRC#446:
3×3 Diagonal{Float64, Vector{Float64}}:
 0.267261   ⋅         ⋅ 
  ⋅        0.534522   ⋅ 
  ⋅         ⋅        0.801784
```
I'm a little concerned this may affect what #909 just fixed. Maybe CI will tell us, but cc @simeonschaub ?

### DIFF
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -269,6 +269,7 @@ zygote2differential(t::Tuple, primal::Tuple) = map(z2d, t, primal)
 zygote2differential(t::Tuple, primal) = (@warn "primal should be a tuple, not $primal"; return t)
 
 z2d(::Nothing, ::Any) = NoTangent()
+z2d(::Tuple{Vararg{Nothing}}, ::Tuple) = NoTangent()  # collapse all-zero case
 z2d(dx, ::Any) = dx
 z2d(dx::AbstractArray{<:Number}, primal::AbstractArray) = dx
 z2d(dx::AbstractArray{<:AbstractArray{<:Number}}, primal::AbstractArray) = dx
@@ -292,7 +293,11 @@ z2d(x::Union{AbstractZero, Tangent}, ::Any) = return x
 
 function z2d(delta::Tuple, primal::Tuple)
   backing = map(z2d, delta, primal)
-  return canonicalize(Tangent{typeof(primal), typeof(backing)}(backing))
+  if backing isa Tuple{Vararg{AbstractZero}}
+    return NoTangent()  # collapse all-zero case
+  else
+    return canonicalize(Tangent{typeof(primal), typeof(backing)}(backing))
+  end
 end
 
 z2d(dx::NamedTuple, primal::AbstractDict) = dx  # uses a NamedTuple but not for fields!
@@ -300,8 +305,13 @@ function z2d(delta::NamedTuple, primal::T) where T  # arbitrart struct
   fnames = fieldnames(T)
   deltas = map(n -> get(delta, n, nothing), fnames)
   primals = map(n -> getfield(primal, n), fnames)
-  backing = NamedTuple{fnames}(map(z2d, deltas, primals))  # recurse into fields
-  return canonicalize(Tangent{T, typeof(backing)}(backing))
+  inner = map(z2d, deltas, primals)  # recurse into fields
+    if inner isa Tuple{Vararg{AbstractZero}}
+    return NoTangent()  # collapse all-zero case
+  else
+    backing = NamedTuple{fnames}(inner)
+    return canonicalize(Tangent{T, typeof(backing)}(backing))
+  end
 end
 # On Julia <= 1.6, this fixes easy cases which do not require recursion into fields, e.g.
 # @inferred Zygote.z2d((re=1, im=nothing), 3.0+im)
@@ -316,9 +326,13 @@ z2d(dx::NamedTuple{L,S}, primal::AbstractDict) where {L,S<:Tuple{Vararg{Union{Nu
       :(delta.$n)
     end
   end
-  return quote
-    backing = NamedTuple{$fnames}(($(deltas...),))
-    Tangent{$T, typeof(backing)}(backing)
+  if all(d -> d == :(NoTangent()), deltas)
+    return :(NoTangent())  # collapse all-zero case
+  else
+    return quote
+      backing = NamedTuple{$fnames}(($(deltas...),))
+      Tangent{$T, typeof(backing)}(backing)
+    end
   end
 end
 
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -159,16 +159,13 @@ _project(x::AbstractArray, dx::Tuple) = _project(x, reshape(collect(dx), axes(x)
 
 # Piracy:
 # wrap_chainrules_input doesn't handle array of Union{Int,Nothing}
-(::ChainRulesCore.ProjectTo)(::Nothing) = ChainRulesCore.NoTangent()
-
-# CRC likes Tangent{<:Complex}, but Zygote makes Tangent{Any}
-# (project::ProjectTo{<:Complex})(dx::Tangent) = project(Complex(dx.re, dx.im))
+# (::ChainRulesCore.ProjectTo)(::Nothing) = ChainRulesCore.NoTangent()
 
 # CRC likes Tangent{AbstractArray}, but Zygote makes Tangent{Any}
 # in particular this would hit https://github.com/JuliaDiff/ChainRulesCore.jl/blob/2ec2549b73b22bc08f554dae864fb650cfb9c3d7/src/projection.jl#L139
 # if we were not losing track of the Primal in the Tangent
 # This type piracy is just giving up that safety check.
-(project::ProjectTo{AbstractArray})(dx::Tangent) = dx
+# (project::ProjectTo{AbstractArray})(dx::Tangent) = dx
 
 """
   ZBack{F}(back) <: Function
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -244,7 +244,8 @@ zygote2differential(t::Tuple, primal::Tuple) = map(z2d, t, primal)
 zygote2differential(t::Tuple, primal) = (@warn "primal should be a tuple, not $primal"; return t)
 
 z2d(::Nothing, ::Any) = NoTangent()
-z2d(dx, ::Any) = dx  # numbers, Dict, etc.
+z2d(dx, ::Any) = dx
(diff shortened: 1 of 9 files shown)


mcabbott:
Ok, good. Yes I need to stare a bit more at the mutable case, and think up some tests.

willtebbutt:
Should we need this here? There's nothing in the default `getproperty` implementation that would warrant it -- if something needs projecting here, presumably it's because a rule somewhere else was incorrect, no?

mcabbott:
> because a rule somewhere else was incorrect, no?

No. These examples aren't doing "mathematical" projection at all. They just change the representation, from a NamedTuple to something more like the original. This is a second meaning to overload onto `_project`, and there is some chance that it ought to be called something else... although combining these seems fairly natural and tidy to me.

> default getproperty implementation

There's nothing wrong with the default, except that it returns a NamedTuple. The examples above have steps before `getproperty`, which the gradient wants to flow backwards into, and these steps don't work with a NamedTuple -- you get the errors shown or the nonsensical answers shown.

devmotion:
FYI this PR fixes the remaining Zygote test errors in https://github.com/TuringLang/DistributionsAD.jl/pull/203 (https://github.com/TuringLang/DistributionsAD.jl/pull/203#issuecomment-955584059).

DhairyaLGandhi:
Cool, let me take a pass at this in my morning. @devmotion does this address the `accum(::NamedTuple, ::Tangent)` case for you?

devmotion:
>  @devmotion does this address the `accum(::NamedTuple, ::Tangent)` case for you?

I don't know which error you refer to, it's none of the errors in https://github.com/TuringLang/DistributionsAD.jl/pull/203#issuecomment-949570421. Both Zygote problems in this comment (https://github.com/TuringLang/DistributionsAD.jl/pull/203#issuecomment-949570421) are fixed with this PR though.

src/compiler/chainrules.jl
```diff
@@ -236,32 +260,85 @@ end
...
+
+# Dict case matches signature for ambiguity reasons:
+z2d(dx::NamedTuple{L,S}, primal::AbstractDict) where {L,S<:Tuple{Vararg{Union{Number,Nothing}}}} = dx
+# On Julia <= 1.6, this fixes easy cases which do not require recursion into fields, e.g.
+# @inferred Zygote.z2d((re=1, im=nothing), 3.0+im)
+@generated function z2d(delta::NamedTuple{L,S}, primal::T) where {L,S<:Tuple{Vararg{Union{Number,Nothing}}}, T}
+  fnames = fieldnames(T)
+  deltas = map(fnames) do n
+    i = findfirst(isequal(n), L)
+    if i == nothing || S.parameters[i] == Nothing
```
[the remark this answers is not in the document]
DhairyaLGandhi: I don't think I understand the need of the generated function still. A regular function should work here.

src/compiler/chainrules.jl
```diff
@@ -236,32 +260,85 @@ end
...
-function z2d(t::NamedTuple, primal)
-  primal_type = typeof(primal)
-  fnames = fieldnames(primal_type)
-  complete_t = NamedTuple{fnames}(fn in keys(t) ? t[fn] : nothing for fn in fnames)
-  primals = NamedTuple{fnames}(getfield(primal, fn) for fn in fnames)
-  tp::NamedTuple = map(z2d, complete_t, primals)
-  return canonicalize(Tangent{primal_type, typeof(tp)}(tp))
+# Dict handling in Zygote is a mess... should this become a  `Tangent{Dict,Dict}` ?
+# Right now it uses a NamedTuple but not for fields of the AbstractDict struct
+z2d(dx::NamedTuple, primal::AbstractDict) = dx
```
DhairyaLGandhi, line 305: Why should it be the fields of the `AbstractDict` struct?

src/compiler/chainrules.jl
```diff
@@ -236,32 +260,85 @@ end
...
 z2d(x::Union{AbstractZero, Tangent}, ::Any) = return x
-function z2d(t::Tuple, primal::Tuple)
-  tp::Tuple = map(z2d, t, primal)
-  primal_type = typeof(primal)
-  return canonicalize(Tangent{primal_type, typeof(tp)}(tp))
+
+function z2d(delta::Tuple, primal::Tuple)
+  backing = map(z2d, delta, primal)
+  if backing isa Tuple{Vararg{AbstractZero}}
+    return NoTangent()  # collapse all-zero case
```
DhairyaLGandhi, line 297: There seems to be a lot of "collapsing zero" going on. The accumulation step already takes care of the collapsing for the most part. Why can't we let the zeros cases get handled in the accumulation step? Is this related to zygote sometimes producing `Array{Nothing}`?

DhairyaLGandhi:
Given the extensive breakage in the ecosystem, we would definitely need a hot patch so we can get PPL and SciML up and running again. We definitely need a more robust solution than we have currently to handle the zygote <-> CR interaction.

src/compiler/chainrules.jl
devmotion: Without it the DistributionsAD tests fail. The main reason is that ChainRules expects non-differentiable objects to have a `NoTangent` derivative but without this fix we end up with stuff like `Tangent{T}(; a = NoTangent(), b = [NoTangent(), NoTangent()], c = Tangent{S}(; d = NoTangent()))`.

DhairyaLGandhi: A stacktrace would be helpful.

devmotion:
IMO the integration tests added in this PR are a big improvement and hopefully ensure that Zygote support in Distributions and Turing is not broken accidentally again.

I don't think it's the main problem right now but I assume the CR-Zygote integration could be simplified if Zygote would use `NoTangent` and `ZeroTangent` instead of `nothing`.

src/compiler/chainrules.jl
devmotion: There are more explanations in https://github.com/TuringLang/DistributionsAD.jl/pull/203#issuecomment-950240079. Unfortunately, it seems there are no GH actions logs of the original issue since I fixed it (partially) before adding and pushing a new commit.

src/compiler/chainrules.jl
oxinabox: No statement is being made here about should or shouldn't.
Per the comment Dict handling is currently a mess, and is inconsistent.
But most of the time it is  a named tuple that is not the fields, but rather is the values.
And that is fine.

src/compiler/chainrules.jl
```diff
@@ -115,28 +115,60 @@ for T_outer in (:Tuple, :NamedTuple)
...
-@inline wrap_chainrules_input(xs::AbstractArray{<:NamedTuple}) = wrap_chainrules_input.(xs)
-@inline wrap_chainrules_input(xs::AbstractArray{<:Union{Nothing, <:NamedTuple}}) = wrap_chainrules_input.(xs)
+@inline wrap_chainrules_input(dx::Ref) = wrap_chainrules_input(dx[])
+# For arrays, whitelist the safe ones, but always look inside Any[]:
+@inline wrap_chainrules_input(dxs::AbstractArray{<:Number}) = dxs
+@inline wrap_chainrules_input(dxs::AbstractArray{<:AbstractArray{<:Number}}) = dxs
+@inline wrap_chainrules_input(dxs::AbstractArray) = map(wrap_chainrules_input, dxs)
+# Could `reinterpret` instead here...
+# One easy case, but can this go wrong?
+# @inline wrap_chainrules_input(xs::Base.ReinterpretArray{<:NamedTuple, <:Tangent}) = parent(@show xs)
```
DhairyaLGandhi, line 161: This seems like debugging code

src/compiler/chainrules.jl
```diff
@@ -115,28 +115,60 @@ for T_outer in (:Tuple, :NamedTuple)
     ChainRulesCore.backing(xp)  # this is accessing ChainRulesCore internals, but it is prob safe enough, and it is fastest
   end
 end
-# Could `reinterpret` instead of broadcasting here -- TODO
-@inline wrap_chainrules_output(xs::AbstractArray{<:ChainRules.Tangent}) = wrap_chainrules_output.(xs)
+wrap_chainrules_output(dxs::AbstractArray{<:Number}) = dxs
+wrap_chainrules_output(dxs::AbstractArray{<:AbstractArray{<:Number}}) = dxs
```
DhairyaLGandhi, line 119: Is there an mwe for this being needed? Is it for RecursiveArrayTools or does it come up elsewhere?

mcabbott: For all of these, the minimal correct thing is `wrap_chainrules_output(dxs::AbstractArray) = map(wrap_chainrules_output, dxs)`. 

But for some common types, we know that the representation used by Zygote and by ChainRules agrees, so we can save a copy by doing nothing. That's the case for `AbstractArray{<:Number}` and `AbstractArray{<:AbstractArray{<:Number}}`.

src/compiler/chainrules.jl
mcabbott: It's a comment, about further optimisations? Again 1112 is the issue.

DhairyaLGandhi: We can get rid of the show either way. In general, it's best to leave reinterpreted arrays be, they can have unexpected behavior if we try to manage them manually. Even in a comment, I still doubt the show was intentionally expected to be part of committed code?

DhairyaLGandhi (approved):
bors r+

ToucheSir:
Cryptic error from Bors: https://app.bors.tech/repositories/20723/log

mcabbott:
Indeed. 

This did pass buildkite tests, BTW, after being rebased on master (a8306bc), and master hasn't changed since.

PR #1114, opened 2021-11-04
remove `@adjoint function cholesky`

st--:
The adjoint of `cholesky` is already [defined in ChainRules.jl](https://github.com/JuliaDiff/ChainRules.jl/blob/edf3a1f48fb5c9af01820aeca6ced94d4f97fa1a/src/rulesets/LinearAlgebra/factorization.jl#L441), but it is overridden by Zygote's definition, which is missing the case for Hermitian:
```julia
C = rand(5,5)
f(x) = sum(cholesky(Hermitian(x'x)).L)
Zygote.gradient(f, C)  # errors
```

This PR bumps the julia compat to 1.6 and ChainRules to 1.35.3, at which point we can safely remove the adjoint. It also adds a test for the Hermitian case.


DhairyaLGandhi:
I'm generally on board. The integration test failures are related to DistributionsAD but the buildkite errors are curious seeing as they shouldn't be involved here at all.

st--:
Yeah, this failure in the CI / Julia 1 test looks like something is actually not going quite right in the cholesky gradient: https://github.com/FluxML/Zygote.jl/runs/4106058861?check_suite_focus=true#step:6:189

devmotion:
Maybe one should check if https://github.com/FluxML/Zygote.jl/pull/1104 fixed (some of) the issues and rerun the tests based on the master branch?

st--:
> Maybe one should check if #1104 fixed (some of) the issues and rerun the tests based on the master branch?

There's 28 more tests in total and 3 more broken tests, comparing master-merged-in vs previous run... so doesn't seem to have helped unfortunately...

devmotion:
I just remembered https://github.com/JuliaDiff/ChainRules.jl/issues/611, I think it causes most of the test errors here (https://github.com/JuliaDiff/ChainRules.jl/issues/611#issuecomment-1149650865).

devmotion:
Looks great 👍

The remaining test error is caused by a newly added feature of the rule in ChainRules - it also supports failed factorizations for some input types now. So updating the test should be sufficient I assume.

ChainRules 1.35.3 also only supports Julia >= 1.6 (the Julia compat in CR was updated quite a while ago), so the Julia compat should be updated or the Zygote rule only be removed on Julia >= 1.6 if you want to keep support for older Julia versions.

The downstream test error of DynamicPPL was fixed in the latest release of DistributionsAD (just released around 1-2 hours ago).

ToucheSir:
@FluxML/ad @CarloLucibello should we take the opportunity to bump compat to 1.6? Both upstream (CR) and downstream (Flux) have already done so now.

CarloLucibello:
Yes let's bump

st--:
@devmotion I've tried to think about how to _fix_ the test but I couldn't think of anything, so I've just removed it for now; if you've got any better idea what a test should look like please just suggest

test/gradcheck.jl
```diff
@@ -820,6 +819,18 @@ end
     @test back′(C̄)[1] isa Diagonal
     @test diag(back′(C̄)[1]) ≈ diag(back(C̄)[1])
   end
+  @testset "cholesky - Hermitian" begin
+    rng, N = MersenneTwister(123456), 3
+    A = randn(rng, N, N) + im * randn(rng, N, N)
```
devmotion, line 824: I think you can just use
```suggestion
    A = randn(rng, Complex{Float64}, N, N)
```

devmotion (commented):
Can you also update the Github action to 1.6?

test/gradcheck.jl
st--: thank you, much better :+1:

st--: I thought there should have been something like this:)

ToucheSir:
If this looks good, we can roll it and https://github.com/FluxML/Zygote.jl/pull/1226 into v0.6.41.

ToucheSir:
I'll take the +1 as approval from @devmotion, so if you wouldn't mind rebasing @st-- we can get this merged.

devmotion (approved):
A more formal approval 🙂 Looks good to me, tests seem to pass now (apart from nightlies and NeuralPDE which fail on master as well).

PR #1123, opened 2021-11-23
support specialfunctions 2.x

stevengj:
The only breaking change in SpecialFunctions 2.0 [was deleting](https://github.com/JuliaMath/SpecialFunctions.jl/pull/297) the `factorial(x::Real) = gamma(x+1)` function, and since you don't use `factorial` this shouldn't affect you.


DhairyaLGandhi:
Interesting, I wonder why compat helper didn't open the pr yet. Thanks for the bump!

stevengj:
Looks like you got a compathelper PR.   Closing in favor of #1122

PR #1125, opened 2021-11-25
Make tests pass on 1.8

### LINKED ISSUE #1066: Zygote broken on nightly
I guess I'm stating the obvious (tests are also failing), but I didn't find any previous issue. 
 
Just curious about the reason. Feel free to close.

```               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.8.0-DEV.480 (2021-09-08)
 _/ |\__'_|_|_|\__'_|  |  Commit 62f0ccda46* (0 days old master)
|__/                   |

julia> using Zygote

julia> Zygote.gradient(x -> 2x, 2)
Internal error: encountered unexpected error in runtime:
BoundsError(a=Array{Any, (3,)}[
  Core.Compiler.VarState(typ=Core.Const(val=ZygoteRules._pullback), undef=false),
  Core.Compiler.VarState(typ=Zygote.Context, undef=false),
  Core.Compiler.VarState(typ=Core.Const(val=Main.var"#5"), undef=false)], i=(4,))
```

mcabbott:
Previously, `sum(::Tuple)` had `check_inferred=false` but NamedTuple did not. Now both do. That was the one failing test.

But some other tests marked `check_inferred=false` no longer need that. So perhaps it's a wash.

Closes #1066

### DIFF
--- a/test/chainrules.jl
+++ b/test/chainrules.jl
@@ -267,14 +267,8 @@ end
 
 @testset "ChainRulesCore.rrule_via_ad" begin
     @testset "basic" begin
-        # broken because Zygoye compresses `(NoTangent(), NoTangent())` into just NoTangent()
-        # which ChainRulesTestUtils does not think is valid:
-        @test_broken(rrule_via_ad(ZygoteRuleConfig(), round, 2.2) isa Tuple{NoTangent,NoTangent})
-        # Later: That makes no sense, would always be broken. If you call the pullback,
-        # then right now Zygote agrees with the rrule, but gives 0.0 not NoTangent:
+        # Not marked as tests since perhaps ZeroTangent would be better.
         rrule_via_ad(ZygoteRuleConfig(), round, 2.2)[2](1) == (NoTangent(), 0.0)
-        rrule(round, 2.2)[2](1) == (NoTangent(), 0.0)
-        # Not marked as tests since perhaps NoTangent would be better.
         # But test_rrule is happy:
         test_rrule(ZygoteRuleConfig(), round, 2.2; rrule_f=rrule_via_ad)
 
--- a/test/chainrules.jl
+++ b/test/chainrules.jl
@@ -1,4 +1,6 @@
 using ChainRulesCore, ChainRulesTestUtils, Zygote
+using Zygote: ZygoteRuleConfig
+
 @testset "ChainRules integration" begin
     @testset "ChainRules basics" begin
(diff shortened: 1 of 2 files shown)


test/chainrules.jl
```diff
@@ -268,10 +270,15 @@ end
         # broken because Zygoye compresses `(NoTangent(), NoTangent())` into just NoTangent()
         # which ChainRulesTestUtils does not think is valid:
         @test_broken(rrule_via_ad(ZygoteRuleConfig(), round, 2.2) isa Tuple{NoTangent,NoTangent})
```
mzgubic, line 272: The comment would be correct if
```suggestion
        # broken because Zygoye compresses `(NoTangent(), NoTangent())` into just NoTangent()
        # which ChainRulesTestUtils does not think is valid:
        @test_broken(rrule_via_ad(ZygoteRuleConfig(), round, 2.2)[2](1) isa Tuple{NoTangent,NoTangent})
```
but the rrule should return a `y, pb`.

In fact it looks fine below, though the rule should probably return `ZeroTangent()` instead. (Not `NoTangent()` since we definitely can perturb the input)

test/chainrules.jl
```diff
@@ -268,10 +270,15 @@ end
...
-        # uncomment below when/if above is fixed
-        # test_rrule(ZygoteRuleConfig(), round, 2.2; rrule_f=rrule_via_ad)
-
-        test_rrule(ZygoteRuleConfig(), vcat, rand(3), rand(4); rrule_f=rrule_via_ad, check_inferred=false)
+        # Later: That makes no sense, would always be broken. If you call the pullback,
+        # then right now Zygote agrees with the rrule, but gives 0.0 not NoTangent:
+        rrule_via_ad(ZygoteRuleConfig(), round, 2.2)[2](1) == (NoTangent(), 0.0)
```
mzgubic, line 275: doesn't this just use the `rrule` under the hood anyway?

test/chainrules.jl
mcabbott: OK, I've deleted it. I was just a bit puzzled by whether this was broken at some point.

mzgubic (approved):

PR #1134, opened 2021-12-13
Fix Buffer when storing arrays

avik-pal:
MWE:

```julia
using Zygote, Flux


struct MultiParallelNet{L}
    layers::L

    function MultiParallelNet(args...)
        layers = tuple(args...)
        return new{typeof(layers)}(layers)
    end

    MultiParallelNet(layers::Tuple) = new{typeof(layers)}(layers)

    MultiParallelNet(layers::Vector) = MultiParallelNet(layers...)
end

Flux.@functor MultiParallelNet

function (mpn::MultiParallelNet)(x::Union{Tuple,Vector})
    buf = Zygote.Buffer([])
    for l in mpn.layers
        push!(buf, l(x...))
    end
    return copy(buf)
end

function (mpn::MultiParallelNet)(args...)
    buf = Zygote.Buffer([])
    for l in mpn.layers
        push!(buf, l(args...))
    end
    return copy(buf)
end

mpn = MultiParallelNet(Dense(1, 1), Dense(1, 1))
x = rand(Float32, 1, 1)
gradient(() -> sum(vcat(mpn((x,))...)), Flux.params(mpn))
```

```julia
ERROR: ArgumentError: indexed assignment with a single value to possibly many locations is not supported; perhaps use broadcasting `.=` instead?
```


DhairyaLGandhi (approved):
Seems fine but needs a test

ToucheSir:
Existing buffer test failure needs to be resolved as well...

ToucheSir (changes requested):
https://github.com/FluxML/Zygote.jl/blob/master/test/gradcheck.jl#L1428 this one

avik-pal:
Added tests and fixed the mentioned test case.

The other failures seem to be unrelated:
```
ERROR: LoadError: InitError: could not load symbol "LLVMExtraInitializeAllTargets":
/opt/hostedtoolcache/julia/nightly/x64/bin/julia: undefined symbol: LLVMExtraInitializeAllTargets
```

ToucheSir (approved):
Yeah, something is screwy with the CUDA dep resolving and I'm not sure why, perhaps package server lag causing issues? Buffer tests look good though, thanks!

avik-pal:
bump

src/lib/buffer.jl
```diff
@@ -45,8 +45,17 @@ _pullback(cx::AContext, ::typeof(Broadcast.materialize!), b::Buffer, x::Abstract
...
-  copy(b), function (b̄)
-    grad_mut(__context__, b)[:] = b̄
+  res = copy(b)
+
+  function copy_sensitivity(b̄)
+    grad_mut(__context__, b)[:] .= vec(b̄)
     return
   end
+
+  function copy_sensitivity(b̄::Union{Tuple,Vector{T}}) where {T<:AbstractArray}
```
DhairyaLGandhi, line 55:
```suggestion
  function copy_sensitivity(b̄::Union{Tuple,AbstractVector{T}}) where {T<:AbstractArray}
```

Fwiw,  `Buffer` mimics `similar` more than it does a literal array; and `similar(::Tuple)` would be an error.

DhairyaLGandhi (approved):

Issue #1135, opened 2021-12-13
Possible typo in `copy!(x::AbstractVector, ps::Params)`


pakk-minidose:
Hello,
the following function
https://github.com/FluxML/Zygote.jl/blob/601f888c701d21971c9857adad90318cd8c3ea10/src/compiler/interface.jl#L231

should return `x` instead of `ps`, in order to be consistent with the Julia documentation, which states that the first argument should be returned

```
help?> copy!
search: copy! copyto! circcopy! unsafe_copyto! copy copysign deepcopy

  copy!(dst, src) -> dst

  In-place copy of src into dst, discarding any pre-existing elements in dst. If dst and src are of the same type, dst == src should hold after the call. If dst and src are 
  multidimensional arrays, they must have equal axes. See also copyto!.
```

Is this a typo, or is it intentional? I wanted to ask before submitting a PR.

Thank you for your reply.

DhairyaLGandhi:
There isn't a strict requirement and returns on mutating functions are mostly for convenience but in this case it does seem better to return `x` instead.

ToucheSir:
Judging by the original commit, it may well have been a copy paste typo. Please do file a PR!

pakk-minidose:
Thank you for your replies!

@ToucheSir done!
@DhairyaLGandhi I have expanded on your point in the PR

mcabbott:
Note, for future, that "closes #1135" has to be in the first post for github to notice it. But this is fixed, thanks.

PR #1136, opened 2021-12-13
Make value returned by `copy!` consistent

pakk-minidose:
Hello,

following the discussion from #1135, I have spotted two instances, where the `copy!` function returns the second argument instead of the first.

I personally think that the function should always return the first argument in order for the behavior to be consistent.

Further, even though this behavior is not enforced in the Julia documentation, the line
```copy!(dst, src) -> dst```
in the [documentation](https://docs.julialang.org/en/v1/base/arrays/#Base.copy!) can lead users to a false conclusion that the first argument is always returned.

Best regards,
Dominik

### DIFF
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -288,7 +288,7 @@ function copy!(gs::Grads, x::AbstractVector)
     gs[p] .= reshape(x[i+1:i+length(p)], size(p))
     i += length(p)
   end
-  x
+  gs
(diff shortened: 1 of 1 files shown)


pakk-minidose:
closes #1135

DhairyaLGandhi:
Thank you for this!

ToucheSir (approved):
Thanks!

Issue #1137, opened 2021-12-16
Tag new release?


devmotion:
Is it possible to tag a new release? Since the last release two PRs were merged, and in particular the support of SpecialFunctions 2 would be very helpful for the ecosystem (e.g., https://github.com/JuliaManifolds/Manifolds.jl/pull/444). ForwardDiff already supports it and hence `] add Zygote#master` actually pulls in SpecialFunctions 2 now. Also the upcoming CUDA release will support SpecialFunctions 2.

devmotion:
Probably Zygote and CUDA are also the missing pieces for https://github.com/FluxML/Flux.jl/issues/1802.

devmotion:
Closing as a new release will be available soon: https://github.com/JuliaRegistries/General/pull/51051

PR #1138, opened 2021-12-16
Tag for SpecialFunctions@2

DhairyaLGandhi:
Ref #1137

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.6.32"
+version = "0.6.33"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"


devmotion:
The test error (https://github.com/FluxML/Zygote.jl/runs/4554123422?check_suite_focus=true#step:6:178) is caused by CUDA, it appears in multiple other packages as well (e.g. DiffRules and ChainRules). The problem is that an ancient version of CUDA is pulled in since the newer releases are not compatible with SpecialFunctions 2 (this does not affect the installation of Zygote since it is only installed and loaded in the tests). The upcoming release of CUDA will support SpecialFunctions 2 which should resolve this problem.

DhairyaLGandhi:
Right, I'm assuming CUDA is ready to be tagged? I saw that the version was bumped on master but hasn't been registered yet. We can register zygote with passing CI since it seems only registration of CUDA is the remaining piece for SpecialFunctions@2 which shouldn't hold us up too much.

devmotion:
CUDA 3.6.0 is available now, can you rerun the tests? Hopefully they should pass 🙂

devmotion:
Tests pass (apart from buildkite which wasn't resrarted) 🎉

devmotion:
Good to go? Buildkite passed as well now :slightly_smiling_face:

DhairyaLGandhi:
Yes, I'd wanted to get ci passing, thanks for bumping!

devmotion:
Thanks, can you (or may I) register the new release? Or do you want to wait until tests pass on the master branch as well?

PR #1145, opened 2022-01-07
Downstream test for Molly.jl

jgreener64:
A while ago I talked with @DhairyaLGandhi about including downstream tests for differentiable simulation in Molly.jl. This PR adds those downstream tests, using the `group` flag to only run the relevant tests. The CPU-only tests take about 10 mins.

I am still figuring out the best tolerance levels for the tests so there may be intermittent failures. Also, Molly makes use of Zygote's broadcast internals so failures may be acceptable.


mcabbott (approved):
Sounds good, and is far from being the slowest such test.

Issue #1126, opened 2021-11-27
Significant compile time latency in Flux with default optimization


Alexander-Barth:
I am trying to optimize a super-resolution GAN, and I see very significant compile time.
This issue is discussed here:
 https://discourse.julialang.org/t/significant-compile-time-latency-in-flux-with-a-gan/68518

@ToucheSir was able to reduce this issue to the code below (thanks a lot!).

I am using Julia 1.6.3 on Linux and these packages:

```
(@v1.6) pkg> st Zygote ZygoteRules Flux
      Status `~/.julia/environments/v1.6/Project.toml`
  [587475ba] Flux v0.12.8 `https://github.com/FluxML/Flux.jl.git#master`
  [e88e6eb3] Zygote v0.6.30
  [700de1a5] ZygoteRules v0.2.2
```

It appears that this issue is sensitive to the optimization level and to the julia version used.

with julia 1.6.3 (default optimization):

```
 2.129568 seconds (6.12 M allocations: 352.760 MiB, 5.21% gc time, 99.96% compilation time)
  0.000323 seconds (1.80 k allocations: 138.938 KiB)
2740.221683 seconds (62.97 M allocations: 3.596 GiB, 0.05% gc time, 0.44% compilation time)
  0.004320 seconds (12.04 k allocations: 2.761 MiB)
```
Note the timing of the first call the `loss_grad(lr_images, ps)`.

In julia 1.5.3 I get the following timing:

```
  1.969913 seconds (8.92 M allocations: 458.885 MiB, 6.31% gc time)
  0.000265 seconds (1.30 k allocations: 102.344 KiB)
 25.426727 seconds (57.88 M allocations: 2.927 GiB, 3.93% gc time)
  0.005641 seconds (14.71 k allocations: 1.047 MiB)
```

with "julia 1.6.3 -O1":

```
 1.389465 seconds (6.12 M allocations: 352.764 MiB, 25.61% gc time, 99.94% compilation time)
  0.000403 seconds (2.10 k allocations: 143.625 KiB)
 28.301099 seconds (62.99 M allocations: 3.596 GiB, 3.96% gc time, 26.40% compilation time)
  0.004889 seconds (12.94 k allocations: 2.775 MiB)
```

```julia
using Flux

channels = 4

function resblock(channels)
    return SkipConnection(Chain(
        Conv((3, 3), channels => channels, pad=1),
        Conv((3, 3), channels => channels, pad=1),
    ), +)
end

model = Chain(
[...]

PR #1147, opened 2022-01-10
use Base.Experimental macros to reduce load time 

mcabbott:
Aims to reduce startup time, xref https://github.com/FluxML/Zygote.jl/issues/1126 . Needs more benchmarking, to see if this slows down other things.

Here's one way to get awful results:
```
julia> @btime gradient((x,y) -> sum(sqrt.(x./2 .+ y) .- y' .+ 10 .* x.^2), $(rand(100)), $(rand(100)));
  min 6.625 μs, mean 19.931 μs (22 allocations, 167.00 KiB)   # 0.6.33
  min 43.125 μs, mean 56.539 μs (24 allocations, 167.02 KiB)  # PR
  min 131.416 μs, mean 142.481 μs (27 allocations, 167.06 KiB)  # julia -O1, and Zygote 0.6.33
```
I suspect for this to be useful, you would have to partition the package into sub-modules, only some of which are `-O1`. It seems quite messy to do that.

### DIFF
--- a/src/Zygote.jl
+++ b/src/Zygote.jl
@@ -1,5 +1,13 @@
 module Zygote
 
+# This reduces time to first gradient, by 4x in the Flux example in Zygote#1126
+if isdefined(Base, :Experimental) && isdefined(Base.Experimental, Symbol("@optlevel"))
+  @eval Base.Experimental.@optlevel 1  # Julia 1.5
+end
+if isdefined(Base, :Experimental) && isdefined(Base.Experimental, Symbol("@max_methods"))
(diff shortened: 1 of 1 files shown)


ToucheSir:
Would isolating just the compiler internals work? i.e. still optimize the code that Zygote generates, but don't optimize the code that does that generation.

mcabbott:
Yes, that's my hope with the submodule idea -- either put all the internals in `Zygote.Slow`, or else leave them in `Zygote` and put the run-time rules in `Zygote.Fast`. But the code isn't organised to make this division very easy, there are dozens of weird functions which seem to cross the border.

PR #1157, opened 2022-02-07
Companion PR to new ChainRules pmap rule

gaurav-arya:
Companion PR to https://github.com/JuliaDiff/ChainRules.jl/pull/566

Removes Zygote's `pmap` rule and adds extra downstream tests for the rule in `ChainRules`.

In `test/gradcheck.jl`, there were a number of functions inside of the loop on `mapfunc` that just used `map`; I took those out of the for loop.


gaurav-arya:
See https://github.com/JuliaDiff/ChainRules.jl/pull/566#issuecomment-1074295189

Issue #1162, opened 2022-02-15
Type error with `zip` and heterogeneous vectors in pullback


SBuercklin:
The following code errors on `Zygote@0.6.34` and Julia `1.7.2`. The code below does work on Julia `1.6.2`. 

The issue seems to be the `NTuple{N}(dx)` call [here](https://github.com/FluxML/Zygote.jl/blob/master/src/lib/array.jl#L187), which led [this issue](https://github.com/JuliaLang/julia/issues/44179), which reveals that apparently this is an intentional change on `1.7`.

```julia
struct Bar{X}
    x::X
end
struct BarCache{XC}
    xc::XC
end

function toplevel(bars, bar_caches)
    results = [one_bar(b, bc) for (b, bc) in zip(bars, bar_caches)]
    return results[2][1]
end

function one_bar(b, bc)
    return [b.x^2, bc.xc^2]
end

bars = ntuple(_ -> Bar(randn()), 3)
caches = ntuple(_ -> BarCache(randn()), 3)

toplevel(bars, caches) 
# Works

Zygote.gradient(toplevel, bars, caches) 
```
which errors with stacktrace
```julia
ERROR: TypeError: in typeassert, expected Tuple{T, T} where T, got a value of type Tuple{NamedTuple{(:x,), Tuple{Float64}}, Nothing}
Stacktrace:
  [1] _totuple
    @ ./tuple.jl:333 [inlined]
  [2] Tuple
    @ ./tuple.jl:317 [inlined]
  [3] _restore(dx::Vector{Union{Nothing, NamedTuple{(:x,), Tuple{Float64}}}}, #unused#::Val{3})
    @ Zygote ~/.julia/packages/Zygote/FPUm3/src/lib/array.jl:187
  [4] (::Zygote.var"#622#624"{Vector{Union{Nothing, Tuple{NamedTuple{(:x,), Tuple{Float64}}, NamedTuple{(:xc,), Tuple{Float64}}}}}, Tuple{Tuple{Bar{Float64}, Bar{Float64}, Bar{Float64}}, Tuple{BarCache{Float64}, BarCache{Float64}, BarCache{Float64}}}, Tuple{Val{3}, Val{3}}})(d::Int64)
    @ Zygote ~/.julia/packages/Zygote/FPUm3/src/lib/array.jl:306
  [5] ntuple
    @ ./ntuple.jl:19 [inlined]
  [6] back
[...]

PR #1163, opened 2022-02-15
`ntuple` for `_restore` regardless of length

SBuercklin:
This is a fix for #1162 by removing the ternary check in `_restore` to see if the length of `dx` is less than the length of the tuple and always using `ntuple` instead of `NTuple{N}`. 

As far as I can tell the behavior of `_restore` should be unchanged, since if `length(dx) < N` the code path is the same, whereas if `length(dx) >= N` we still retrieve the first `N` elements of `dx`.

### DIFF
--- a/test/lib/array.jl
+++ b/test/lib/array.jl
@@ -6,7 +6,3 @@ using Zygote: ZygoteRuleConfig
 
 test_rrule(ZygoteRuleConfig(), x->sum(sin, Diagonal(x)), ones(2); rrule_f=rrule_via_ad, check_inferred=false)
 test_rrule(ZygoteRuleConfig(), x->sum(sin, Diagonal(x)), rand(3); rrule_f=rrule_via_ad, check_inferred=false)
-
-# issue 44179
-
-@test Zygote._restore([nothing, 1, nothing], Val(3)) == (nothing, 1, nothing)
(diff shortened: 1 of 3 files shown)


test/lib/array.jl
```diff
@@ -6,3 +6,7 @@ using Zygote: ZygoteRuleConfig
 
 test_rrule(ZygoteRuleConfig(), x->sum(sin, Diagonal(x)), ones(2); rrule_f=rrule_via_ad, check_inferred=false)
 test_rrule(ZygoteRuleConfig(), x->sum(sin, Diagonal(x)), rand(3); rrule_f=rrule_via_ad, check_inferred=false)
+
+# issue 44179
+
+@test Zygote._restore([nothing, 1, nothing], Val(3)) == (nothing, 1, nothing)
```
mcabbott, line 12: The change looks safe to me. 

But can you replace this test with a `gradient` example (or a few) which failed before? I think we should mostly test the end result, not internal functions. If someone re-writes the rule for `map`, they are free to delete `_restore`, but shouldn't break actual uses.

SBuercklin:
Added a `gradient` test which didn't use types, just `Tuple`s. The only tests I was able to nail down were all variations on this: two `Tuple`s or `Vector`s that are `zip`ped together and then have some `NoTangent()/nothing`s in the pullback. 

Surely there are likely more cases that could be constructed, but this captures the essence of the problem as I was able to reproduce it. I can add more tests if you have ideas for what those pathological cases may be.

mcabbott (approved):
Looks good, can confirm that the test fails for me.

Issue #1098, opened 2021-10-10
Decide what to do about `pushforward` and co.


ToucheSir:
We get at least a couple of confused questions about it every month, but AFAICT there isn't much appetite to bring it up to a point of general usefulness. How does a big flashing banner saying "don't use this" sound until if/when someone feels the urge to pick it back up?

DhairyaLGandhi:
We added the forward mode to be used, and it's actually the fastest forward mode on neural networks right now, so I would say if we get questions we should help resolve them and flesh it out.

ToucheSir:
Are there benchmarks floating out there that demonstrate this? I was under the impression that rule support was not fleshed out enough to support all but the most basic of Flux models.

DhairyaLGandhi:
We can do mlps which are of most importance to small model cases such as SciML. Small - medium sized models are also the ones that are poised to gain most from a fast forward mode, beyond which a reverse mode would start making sense, esp where a lot of arguments are required.

Also see https://github.com/FluxML/Zygote.jl/pull/503#issuecomment-635408758

gdalle:
What is the status of this in 2024? @oxinabox any insights? I just had a question asking whether we should support Zygote's forward mode in DI

oxinabox:
I have never heard of anyone using it.
I have never used it.
It was one of the last things Mike did before leaving

PR #1172, opened 2022-02-23
Fix deprecations in DiffRules 1.4, add forward-mode rules for LogExpFunctions

### LINKED ISSUE #1155: DiffRules warning
I am getting this warning:

> ┌ Warning: the implicit keyword argument `filter_modules=(:Base, :SpecialFunctions, :NaNMath)` in `diffrules()` is deprecated and will be changed to `filter_modules=nothing` in an upcoming breaking release of DiffRules (i.e., `diffrules()` will return all rules defined in DiffRules)
> │   caller = top-level scope at number.jl:6

and a similar one from line 14.

https://github.com/FluxML/Zygote.jl/blob/3415724c00ed1e15056365c4d3edbce6f32ccae4/src/forward/number.jl#L6

https://github.com/FluxML/Zygote.jl/blob/3415724c00ed1e15056365c4d3edbce6f32ccae4/src/forward/number.jl#L14

devmotion:
This PR fixes the deprecations in DiffRules 1.4, as done similarly already in ReverseDiff (https://github.com/JuliaDiff/ReverseDiff.jl/pull/191) and ForwardDiff.

Closes #1155.


mcabbott:
This is all for forward mode, which is a bit of a prototype, xref https://github.com/FluxML/Zygote.jl/issues/1098

devmotion:
I don't anything about this part of Zygote - I just don't want to see the deprecation warnings anymore 😅

devmotion:
>  Is there any value to expanding the list of functions for which derivatives are defined, if nobody is using them?

Since LogExpFunctions is already an indirect dependency (through DiffRules, SpecialFunctions, and ForwardDiff) I thought there's no harm in adding these rules. Using the upcoming default keyword argument `filter_modules=nothing` has the additional advantage that one is notified if additional packages are added to DiffRules. However, this could be achieved also by silently filtering out `M === :LogExpFunctions` in the for loop.

devmotion:
Bump 🙂

mcabbott:
I'd support doing the minimum to silence the warning. But I think that adding more rules to forward mode needs some argument that this is useful.

devmotion:
I think the advantage is that it makes use of all existing rules in DiffRules and does not, perhaps rather surprisingly, only work with a subset of them. Generally, I would assume that one rather has to argue why specific rules should be omitted.

Additionally, if LogExpFunctions (or even DiffRules, SpecialFunctions, or NaNMath) are removed as direct dependencies, they are all still indirect dependencies via ForwardDiff, and LogExpFunctions and SpecialFunctions are also dependencies of StatsFuns.

But if the general consensus is that not all rules should be included (@DhairyaLGandhi? @CarloLucibello?), I'll update the PR accordingly. I really would like to get rid of the deprecation warnings :slightly_smiling_face:

mcabbott:
If there's a one-line fix to shut the warning up, let's do that. Any complexity beyond that needs an argument *for* it, rather than an argument that adding new code and new dependencies and new compat bounds probably won't hurt anything.

devmotion:
> and new dependencies and new compat bounds which we don't need

LogExpFunctions is already an indirect dependency, which Zygote won't get rid of even if DiffRules is removed. Moreover, Zygote even defines custom adjoints for LogExpFunctions in https://github.com/FluxML/Zygote.jl/blob/master/src/lib/logexpfunctions.jl that are loaded conditionally. I think they should be loaded unconditionally since LogExpFunctions is loaded anyway if recent versions of DiffRules, ForwardDiff, StatsFuns, and/or SpecialFunctions are used. Hence a proper compat bound should be added anyway (and maybe even if it's only loaded conditionally).

> If there's a one-line fix to shut the warning up, can we just do that?
> Why not do the simple thing?

There's no one-line fix. This PR is the simple fix that is used eg in ForwardDiff and ReverseDiff. Basically all downstream packages of DiffRules, including Zygote, were missing checks for whether the functions used in the rules are available at all. The diff if rules for LogExpFunctions and possibly other packages (it's not possible to just exclude LogExpFunctions) are excluded is:
```diff
diff --git a/Project.toml b/Project.toml
index 97af0610..43ca7373 100644
--- a/Project.toml
+++ b/Project.toml
@@ -13,7 +13,6 @@ ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 IRTools = "7869d1d1-7146-5819-86e3-90919afe41df"
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
-LogExpFunctions = "2ab3a3ac-af41-5b50-aa03-7779005ae688"
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
 NaNMath = "77ba4419-2d1f-58cd-9bb1-8ffee604a2e3"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
@@ -32,7 +31,6 @@ DiffRules = "1.4"
 FillArrays = "0.8, 0.9, 0.10, 0.11, 0.12, 0.13"
 ForwardDiff = "0.10"
 IRTools = "0.4.4"
-LogExpFunctions = "0.3"
 MacroTools = "0.5"
 NaNMath = "0.3, 1"
 Requires = "1.1"
diff --git a/src/forward/number.jl b/src/forward/number.jl
index 322f76f8..517a8e4e 100644
--- a/src/forward/number.jl
+++ b/src/forward/number.jl
@@ -1,9 +1,9 @@
-using DiffRules, SpecialFunctions, NaNMath, LogExpFunctions
+using DiffRules, SpecialFunctions, NaNMath
 using Base.FastMath: fast_op, make_fastmath
 
 # TODO use CSE here
 
-for (M, f, arity) in DiffRules.diffrules(; filter_modules=nothing)
+for (M, f, arity) in DiffRules.diffrules(; filter_modules=(:Base, :SpecialFunctions, :NaNMath))
   if !(isdefined(@__MODULE__, M) && isdefined(getfield(@__MODULE__, M), f))
     @warn "$M.$f is not available and hence rule for it can not be defined"
     continue  # Skip rules for methods not defined in the current scope
```

mcabbott:
Too many words! The minimal fix is `diffrules(; filter_modules=(:Base, :SpecialFunctions, :NaNMath))` no?

All the rest is extra functionality.

mcabbott:
That's a start. After that, it still seems to add a bunch more method definitions, why? 

And maybe it shouldn't print warnings, since the whole point is not to print a warning?

devmotion:
> That's a start. After that, it still seems to add a bunch more method definitions, why?

No, it doesn't. It's exactly the same definitions as before, just in a single for loop instead of two since this reduces the amount of code and avoids code duplication, in particular with the additional checks.

> And maybe it shouldn't print warnings, since the whole point is not to print a warning? 

The point of the PR is to fix the deprecation warning caused by `DiffRules.diffrules`. The warning in the PR was suggested by a reviewer in ReverseDiff (IIRC) to make it clear when not all rules in DiffRules are used, eg because a package (such as LogExpFunctions) was added to DiffRules but is not available in the current scope. The check can't be removed but the warning could, of course, in particular if we don't want to use any rules besides Base, SpecialFunctions, and NaNMath.

devmotion:
> I think they should be loaded unconditionally since LogExpFunctions is loaded anyway if recent versions of DiffRules, ForwardDiff, StatsFuns, and/or SpecialFunctions are used. Hence a proper compat bound should be added anyway (and maybe even if it's only loaded conditionally)

I opened https://github.com/FluxML/Zygote.jl/pull/1221 which fixes this issue and makes LogExpFunctions a proper dependency. When that PR is merged, the discussion about dependencies in this PR would be resolved automatically.

devmotion:
Any update? The PR doesn't add any dependencies anymore, so the only remaining question seems to be whether to exclude rules for LogExpFunctions or not.

Issue #1158, opened 2022-02-08
Different jacobian than FiniteDiff for rfft


judober:
Maybe I'm doing something terribly wrong here. I tried getting the jacobian from an rfft.
I get different results with Zygote compared to a numerical evaluation with FiniteDiff:

```[julia]
using FFTW, FiniteDiff, Zygote

function Compare1()
    # x
    xlen = 100
    xre = zeros(xlen)
    xre[1] = 1

    # Numerical Jacobian
    JacNum = FiniteDiff.finite_difference_jacobian(test1real, xre)

    # Zygote
    JacZyg = jacobian(test1real, xre)[1]

    return (JacNum[1:8,1:8], JacZyg[1:8,1:8]) # compare the first values
end

function test1(xinC)  # irfft
    TAnz = 2 * size(xinC, 1) - 2   # Even
    xoutR = irfft(xinC, TAnz)
end

function test1real(xinR) # Real wrapper
    iseven(length(xinR)) || error("Expect even vector length")
    xinC = xinR[1:2:end] .+ im .* xinR[2:2:end]
    xoutR = test1(xinC)
end

function Compare2()
    # x
    xlen = 100
    xre = zeros(xlen)
    xre[1] = 1

    # Numerical Jacobian
    JacNum = FiniteDiff.finite_difference_jacobian(test2real, xre)

    # Zygote
    JacZyg = jacobian(test2real, xre)[1]

    return (JacNum[1:8,1:8], JacZyg[1:8,1:8]) # compare the first values
end

function test2(xinR) # rfft
    xoutC = rfft(xinR)
end

function test2real(xinR) # Real wrapper
    xoutC = test2(xinR)
    entry(i) = begin
        if iseven(i)
            imag(xoutC[div(i, 2)])
        else
            real(xoutC[div(i, 2, RoundUp)])
        end
    end
    xoutR = [entry(i) for i in 1:2*length(xoutC)]
end

(JacNumIrfft, JacZygIrfft) = Compare1();
(JacNumRfft, JacZygRfft) = Compare2();

JacNumIrfft - JacZygIrfft
JacNumRfft - JacZygRfft
```

My results:
```
julia> JacNumIrfft - JacZygIrfft
8×8 Matrix{Float64}:
 2.37582e-11  0.0  0.0102041    0.0          0.0102041    0.0         0.0102041    0.0
[...]

PR #1173, opened 2022-02-23
ScaledPlan backwards rule

gaurav-arya:
The current adjoint for FFT plans in Zygote doesn't handle `ScaledPlan`s. See for example:

https://github.com/JuliaMath/FFTW.jl/issues/182
https://github.com/SciML/OperatorLearning.jl/issues/11

This issue mostly arises because the inverse plan of e.g. an FFT makes use of `ScaledPlan`. This rule adds a `ScaledPlan` rule, and adds more comprehensive tests which explicitly test every possible forward and inverse plan (and therefore test for the `ScaledPlan` issue that was not being handled previously).

Also, it seems like backwards rules for `fft`s are being moved into `AbstractFFTs.jl` using ChainRules. However, there are no backwards rules there for FFT plans yet, see https://github.com/JuliaMath/AbstractFFTs.jl/issues/63. (The same issue holding that back -- that the type of an FFT plan does not convey enough information for the backwards rule for real FFTs -- leads to a few errored tests in Zygote, which are manually skipped in `gradcheck.jl`). But until that happens, it seems like the rules in Zygote are the only ones that handle FFT plans to at least some extent, so this patch should be helpful until plans are properly handled in `AbstractFFTs.jl`.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -814,7 +814,21 @@ end
 @adjoint function \(P::AbstractFFTs.Plan, xs)
   return P \ xs, function(Δ)
     N = prod(size(Δ)[[P.region...]])
-    return (nothing, (P * Δ)/N)
+    return (nothing, (P * Δ) / N)
+  end
+end
+
+@adjoint function *(P::AbstractFFTs.ScaledPlan, xs)
+  return P * xs, function(Δ)
+    N = prod(size(xs)[[P.p.region...]])
+    return (nothing, N * P.scale^2 * (P \ Δ))
+  end
+end
+
+@adjoint function \(P::AbstractFFTs.ScaledPlan, xs)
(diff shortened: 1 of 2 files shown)


devmotion:
Just a heads up: The rules for `rfft` etc. are incorrect in Zygote but implemented correctly in AbstractFFTs (https://github.com/FluxML/Zygote.jl/issues/1158).

gaurav-arya:
I see. I think that this PR, which partially handles FFT plans, still make sense, do you agree? But perhaps this PR / a separate PR should get rid of all the non-plan FFT rules in Zygote, what do you think?

gaurav-arya:
Closing because we should hopefully have ChainRules for plans somewhat soon, after which this rule can just go.

PR #1182, opened 2022-03-12
Require GPUArrays instead of CUDA

mcabbott:
This replaces the special code for CUDA with that for GPUArrays, which is loaded by CUDA. One motivation is things like  https://github.com/FluxML/OneHotArrays.jl/pull/9, which runs GPU tests with JLArray.

Could be loaded unconditionally, but would need to check load times? Not the only use of Requires in Zygote.

### DIFF
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -251,43 +251,44 @@ end
   return y, bc_fwd_back
 end
 
-@init @require CUDA="052768ef-5323-5732-b1bb-66c8b64840ba" begin
+@init @require GPUArrays = "0c68f7d7-f131-5f86-a1c3-88cf8149b2d7" begin
 
-  const CuArrayStyle = CUDA.AbstractGPUArrayStyle
+  # const CuArrayStyle = AbstractGPUArrayStyle
 
-  if isdefined(CUDA, :cufunc)  # CUDA < 3.0
+  # if isdefined(CUDA, :cufunc)  # CUDA < 3.0
 
-    @eval @adjoint broadcasted(::CuArrayStyle, f, args...) =
-      broadcast_forward(CUDA.cufunc(f), args...)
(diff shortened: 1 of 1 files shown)


src/lib/broadcast.jl
```diff
@@ -251,43 +251,44 @@ end
...
-       # Ordinary broadcasting calls broadcast_forward anyway when certain its' safe,
-       # so perhaps this can be deleted? Possible edge case here:
-       # https://github.com/FluxML/Zygote.jl/pull/1018#issuecomment-873629415
+  # else # CUDA >= 3.0 -- don't need cufunc(f). 
+  #      # Ordinary broadcasting calls broadcast_forward anyway when certain its' safe,
+  #      # so perhaps this can be deleted? Possible edge case here:
+  #      # https://github.com/FluxML/Zygote.jl/pull/1018#issuecomment-873629415
 
-    @eval @adjoint broadcasted(::CuArrayStyle, f, args...) =
-      broadcast_forward(f, args...)
```
mcabbott, line 269: Not sure it's safe to delete these.

ToucheSir: If not, changing `CuArrayStyle` -> `AbstractGPUArrayStyle` seems pretty safe.

src/lib/broadcast.jl
```diff
@@ -251,43 +251,44 @@ end
...
+  #   @eval @adjoint broadcasted(::CuArrayStyle, f, args...) =
+  #     broadcast_forward(f, args...)
 
-  end
+  # end
 
-  @adjoint (::Type{T})(xs::Array) where {T <: CUDA.CuArray} =
```
ToucheSir, line 273: Good opportunity to change these over to `rrule(::ZygoteRuleConfig, ...)` instead? IME they're much easier to debug than `@adjoint` definitions.

PR #1183, opened 2022-03-12
Use Base.IdSet

ToucheSir:
GitHub tells me this has been available since at least 0.7.


ToucheSir:
The `IdSet` interface in Base is not a complete 1-1 match with the one in Zygote, but given we seem to consider this type as an internal-only thing and downstream tests pass, it should be fine?

mcabbott (approved):

Issue #1184, opened 2022-03-14
Ambiguous type signature of Zygote's `unbroadcast`


SebastianAment:
I ran into the following method error while differentiating a model:
```Julia
LoadError: MethodError: unbroadcast(::Tuple{GaussianProcessPhaseMapping.var"#f#64"}, ::Nothing) is ambiguous. Candidates:
  unbroadcast(x::Tuple{Any}, x̄) in Zygote at /Users/sebastianament/.julia/packages/Zygote/3I4nT/src/lib/broadcast.jl:58
  unbroadcast(x::Tuple, x̄::Nothing) in Zygote at /Users/sebastianament/.julia/packages/Zygote/3I4nT/src/lib/broadcast.jl:61
Possible fix, define
  unbroadcast(::Tuple{Any}, ::Nothing)
```

It appears to be fixed by adding
```Julia
Zygote.unbroadcast(x::Tuple{<:Any}, x̄::Nothing) = nothing
```

PR #1185, opened 2022-03-14
Fixing type ambiguity of `unbroadcast`

SebastianAment:
Fixing #1184.

### DIFF
--- a/test/gradcheck.jl
+++ b/test/gradcheck.jl
@@ -360,7 +360,7 @@ end
   @test gradient(x -> map(+, x, (1,2,3))[1], [4,5,6]) == ([1,0,0],)
 
   # mismatched lengths, should zip
-  @test gradient(x -> map(+, x, [1,2,3,99])[1], (4,5,6)) == ((1.0, 0.0, 0.0),) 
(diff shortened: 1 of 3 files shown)


mcabbott:
Looks good.

Is it easy to trigger this problem? Would be nice to have a test, ideally calling `gradient` of something.

SebastianAment:
Great, I just came up with this:
```Julia
using Zygote
n, d = 128, 3
x = [randn(d) for _ in 1:n]

f = sin
g(x) = sum.((f,), x)
h(x) = sum(abs2, g(x))
gradient(h, x)
```
Which triggers the error. Will add it to the testing suite.

test/gradcheck.jl
```diff
@@ -1487,17 +1487,6 @@ using Zygote: Buffer
   @test ∇x == 6 .* x
 end
 
-@testset "FillArrays" begin
```
SebastianAment, line 1490: There were two separate "FillArrays" test sets, I moved the code here without modifications to the second test set starting line 1614.

SebastianAment:
I think this is ready to be merged.

ToucheSir (approved):
Thanks!

PR #1205, opened 2022-04-13
Handle `ChainRulesCore.NotImplemented`

### LINKED ISSUE #1204: AD error due to interaction of missing gradient of `besselk` and struct with vector
Minimal reproducible example - note that `Foo` errors, whereas `Bar` works.
```julia
using Zygote
using SpecialFunctions: besselk
using Test

struct Foo{Tv<:Real}
    v::Vector{Tv}

    function Foo(v::Real)
        return new{typeof(v)}([v])
    end
end

struct Bar{Tv<:Real}
    v::Tv
end

function check_ad(k)
    return only(Zygote.gradient(x -> besselk(only(k.v), x), 2.1))
end

@test_throws MethodError check_ad(Foo(1.5))
@test check_ad(Bar(1.5)) isa Real
```

devmotion:
Return `nothing` instead of `ChainRulesCore.NotImplemented`, similar to e.g. `ChainRules.AbstractZero`.

While this means - once again - dropping information present in CR types and hence makes the type less useful when used together with Zygote, it fixes problems such as #1204 that occur because Zygote pullbacks do not handle `NotImplemented`. `nothing`, however, is already supported throughout Zygote.

Fixes #1204.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.6.37"
+version = "0.6.38"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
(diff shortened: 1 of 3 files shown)


devmotion:
The test error with Julia nightly seems unrelated.

CarloLucibello (approved):

Issue #1201, opened 2022-04-11
hvcat not compatible with non-number types


ntausend:
Hey, consider the following example
```
struct Point
    x
    y
end

function tst(p)
   pp = [p[1] p[2]; p[3] p[4]]
   return pp
end

pointar = [Point(1,2), Point(3,4), Point(5,6), Point(7,8)]

back_x = x -> Zygote.pullback(tst, pointar)[2](x)[1] 

display(back_x(pointar))
```
which throws an error stating that `Mutating arrays is not supported -- called setindex!(::Matrix{point}, _...)`. The problem can be pulled back to the call of `hvcat` to build the righthand site in the `tst` function. After some research I found out that this is due to the definition of the adjoint of the `hvcat` function in `array.jl` Line number `98 - 100`:
```
@adjoint function hvcat(rows::Tuple{Vararg{Int}}, xs::Number...)
  hvcat(rows, xs...), ȳ -> (nothing, permutedims(ȳ)...)
end
```
It states that the `xs` can only be of type number. However, this seems to be restrictive, and comparing to the defnition of the function in Base/abstractarray.jl:

```
hvcat(rows::Tuple{Vararg{Int}}, xs::AbstractVecOrMat...) = ...
```
which seems to except abstract Vectors or Matrices the solution could be simply allowing for the same types.

mcabbott:
There are slightly more general rules for these in CR, but still restricted to numbers or arrays:
https://github.com/JuliaDiff/ChainRules.jl/blob/main/src/rulesets/Base/array.jl#L387

They could be widened to accept anything. Maybe you'd just have to replace `size` (which isn't sure to be defined) with some `size_if_array` which gives `()` not just on numbers, but on unknown types. 

And then Zygote's rules should be deleted.

ToucheSir:
I've folded  https://github.com/FluxML/Zygote.jl/issues/1223 into this issue and updated the title accordingly. There's no reason for us not to track all these *cat methods in one place.

CarloLucibello:
`[hv]cat` over non-number types (e.g. `Point` structs) differentiates now (the old hvcat mutation adjoint was removed). Closing as fixed.

_(Triaged against Zygote v0.7.10 on Julia 1.12.6.)_

PR #1212, opened 2022-04-26
Remove `cat` adjoint in favour of ChainRules

ToucheSir:
Now that this has been effectively superseded by https://github.com/JuliaDiff/ChainRules.jl/blob/v1.28.2/src/rulesets/Base/array.jl#L345, we can ease some of our maintenance burden. I've kept the tests around for now.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.6.38"
+version = "0.6.39"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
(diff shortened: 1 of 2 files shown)


mcabbott (approved):
Sounds fine. Why not `hcat`, `vcat`, `hvcat` too?

mcabbott:
I see `rrule(::typeof(hcat), Xs::Union{AbstractArray, Number}...)` there, vs. here:
```
@adjoint hcat(xs::Number...)
@adjoint hcat(A::Union{AbstractVector, AbstractMatrix}...)
```
In #1201 it was suggested to allow non-Number types too, which should be fine.

DhairyaLGandhi:
Is there a plan on adressong the inferrability/ stability in the general case. With say multiple dimensions?

PR #1219, opened 2022-05-07
Use `setglobal!` on nightly

ToucheSir:
Ref. https://github.com/JuliaLang/julia/commit/0deb28a4c114704c940f8eb53d3e93aca43185d3. This should address a CI failure.

### DIFF
--- a/src/lib/lib.jl
+++ b/src/lib/lib.jl
@@ -81,8 +81,12 @@ unwrap(ref, x) = x
 end
 
 function global_set(ref, val)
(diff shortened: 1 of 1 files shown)


ToucheSir:
Hard to tell if this worked because nothing does now :upside_down_face:. Looks like this will require IRTools changes to accommodate https://github.com/JuliaLang/julia/pull/45204.

CarloLucibello (approved):

ToucheSir:
Seems like only one failure left with threads, that will need a follow-up PR.

PR #1221, opened 2022-05-10
Make LogExpFunctions a proper dependency and clean adjoints

devmotion:
This PR makes LogExpFunctions a proper dependency and cleans its adjoints (ie, removes everything that is implemented identically or better in LogExpFunctions with ChainRules).

There's no need to hide LogExpFunctions behind a Requires block since it's already an indirect dependency of Zygote via ForwardDiff, DiffRules, and SpecialFunctions.

The remaining adjoints are for broadcasting arrays and/or reuse the primal and hence are potentially more efficient than the CR definitions in LogExpFunctions. I don't think the first ones will be added to LogExpFunctions in the near future (for most scalar functions, e.g. also in ChainRules, no special broadcast rules are implemented and one focus of LogExpFunctions is to keep it lightweight with reasonably low code complexity) but the other optimizations could be added to LogExpFunctions at some point.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.6.40"
+version = "0.6.41"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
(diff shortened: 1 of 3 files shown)


devmotion:
I opened a PR that optimizes the rules for `xlogx`, `xlogy`, and `xlog1py`: https://github.com/JuliaStats/LogExpFunctions.jl/pull/52 However, in contrast to the definitions in Zygote they only operate with scalars and hence don't affect this PR.

ToucheSir (approved):
I had no idea we were `@require`ing LogExpFunctions. Given that the responsible alternative would involve adding a compat entry anyhow, and that this removes a heavier dependency, it all makes sense to me.

Issue #1218, opened 2022-05-06
Reverse over forward differentiation


nmheim:
I would like to compute the gradient of a function that contains a `ForwardDiff.gradient`/`ForwardDiff.derivative`.
Computing that higher order gradient with ForwardDiff works, but will be slow for my case be cause its a many input function.
MWE below.
```julia
using ForwardDiff
using ReverseDiff
using Zygote

f(x::Real,y::AbstractVector) = x*sum(y)

df(x,y) = ForwardDiff.derivative(x->f(x,y), x)

# this returns (nothing,)
Zygote.gradient(y->df(0.1,y), rand(5))

# this works; returning vector of ones
ForwardDiff.gradient(y->df(0.1,y), rand(5))

# this works as well
ReverseDiff.gradient(y->df(0.1,y), rand(5))
```
It would be really great to get this to work, because for my use case I have to deal with complex numbers and it seems like ReverseDiff does not like them that much.

Tagging @mcabbott as we have discussed this on slack. I would be happy to help in any way I can.

nmheim:
As this was part of the discussion on slack: I am getting the same behaviour when using `ForwardDiff.gradient` inside
```julia
f(x::AbstractVector,y::AbstractVector) = sum(x)*sum(y)

df(x,y) = ForwardDiff.gradient(x->f(x,y), x)[1]

x = [0.1]
y = rand(5)

# this returns (nothing,)
Zygote.gradient(y->df(x,y), y) |> display

# this works; returning vector of ones
ForwardDiff.gradient(y->df(x,y), y) |> display

# this works as well
ReverseDiff.gradient(y->df(x,y), y)
```

mcabbott:
Thanks for making an issue.

The `nothing` comes from #968, which sends these to `Zygote.forwarddiff(f, x)`, forward over forward, which does not, cannot, track variables closed over in its `f`. Only IMO the minimal solution here is to remove that, and return some of these to errors. 

In #769 there was also a proposal to turn Zygote over ForwardDiff around automatically to ForwardDiff over Zygote. Not sure whether this can handle all cases, or be made to work really.

Alternatively, maybe we can figure out what ReverseDiff is doing. The last example is the surprising one, since `ForwardDiff.gradient` inside makes an array for its output and then writes into it.

mcabbott:
Ok I see. ReverseDiff is dropping down to scalar TrackedReal, which it can trace individually through mutation:
```julia
[...]

PR #1224, opened 2022-05-11
Add warnings to ForwardDiff functions

mcabbott:
This tries to print a warning on any use of `ForwardDiff.gradient` etc. likely to give a wrong answer. Xref discussion in #1218

I don't know of any functions which give wrong results without triggering this. It's easy to invent examples which trigger it without giving wrong results, but whether they happen in the wild I'm not sure. It we were more sure about this, they could be made errors.

### DIFF
--- a/src/lib/forward.jl
+++ b/src/lib/forward.jl
@@ -137,8 +137,28 @@ end
 
 # Use this to allow second derivatives -- this is forward-over-forward,
 # see  https://github.com/FluxML/Zygote.jl/issues/769  for a forward-over-reverse proposal
-@adjoint ForwardDiff.gradient(f, x) = pullback(forwarddiff, x -> ForwardDiff.gradient(f, x), x)
-@adjoint ForwardDiff.jacobian(f, x) = pullback(forwarddiff, x -> ForwardDiff.jacobian(f, x), x)
+@adjoint function ForwardDiff.gradient(f, x)
+  F = typeof(f)
+  Base.issingletontype(F) || @warn """`ForwardDiff.gradient(f, x)` within Zygote cannot track gradients with respect to `f`
+  typeof(f) = $F is not a singleton type""" # maxlog=1 _id=hash(F)
+  pullback(forwarddiff, x -> ForwardDiff.gradient(f, x), x)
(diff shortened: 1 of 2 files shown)


src/lib/forward.jl
```diff
@@ -137,8 +137,28 @@ end
...
+@adjoint function ForwardDiff.gradient(f, x)
+  F = typeof(f)
+  Base.issingletontype(F) || @warn """`ForwardDiff.gradient(f, x)` within Zygote cannot track gradients with respect to `f`
+  typeof(f) = $F is not a singleton type""" # maxlog=1 _id=hash(F)
+  pullback(forwarddiff, x -> ForwardDiff.gradient(f, x), x)
+end
+
+@adjoint function ForwardDiff.jacobian(f::F, x) where F
+  Base.issingletontype(F) || @warn """`ForwardDiff.jacobian(f, x)` within Zygote cannot track gradients with respect to `f`
+  typeof(f) = $F is not a singleton type""" # maxlog=1 _id=hash(F)
```
ToucheSir, line 149: Is there a wording that cuts down on the jargon here? e.g.
```suggestion
  Base.issingletontype(F) || @warn """`ForwardDiff.jacobian(f, x)` within Zygote cannot track gradients with respect to `f`
  typeof(f) = $F may be a closure or a callable type with fields""" # maxlog=1 _id=hash(F)
```

mcabbott: Didn't think hard at all. Might be nice to mention `issingletontype` so people know precisely what the test is.

ToucheSir (approved):

CarloLucibello:
merge this?

ToucheSir:
Let's do it.

mcabbott:
Note BTW that this test is not sensitive enough to notice the problem of https://github.com/FluxML/Flux.jl/issues/1966, I think because the model is a global variable:
```julia
julia> let f = y->sum(model(y))
         Base.issingletontype(typeof(f))
       end
true

julia> let m = model, f = y->sum(m(y))
         Base.issingletontype(typeof(f))
       end
false
```

Issue #1215, opened 2022-04-28
Cannot compute gradients when using FFTW


pawbz:
The following snippet works well on CPU i.e., it gives the correct gradient but fails on GPU.
```
y=complex.([4,1]) |> gpu
x=complex.([3,2]) |> gpu
p=plan_fft(x) |> gpu
function f(x, y) 
    x=p*x
    x=p*x
    return Flux.mse(x,y);
end
gs=gradient(()->f(x,y), Flux.params(x))
gs[x] # results nothing
```

mcabbott:
The gradient of broadcasting used for CuArrays doesn't handle complex numbers. It's a bit of a nasty surprise but nobody has got around to making it at least an error, or better to making it work. 

Xref https://github.com/FluxML/Zygote.jl/issues/961 among others.

CarloLucibello:
closed in #1324

PR #1225, opened 2022-05-12
Add an error for broadcasting with CUDA + complex numbers, etc

mcabbott:
Xref https://github.com/FluxML/Zygote.jl/issues/1215

### DIFF
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -226,9 +226,17 @@ end
 import ForwardDiff
 using ForwardDiff: Dual
 
-dual(x, p) = x
 dual(x::Real, p) = Dual(x, p)
-dual(x::Bool, p) = x
+dual(x::Bool, p) = x  # safe to ignore
+function dual(x, p)
+    if Base.singletontype(typeof(x)) || ChainRulesCore.ProjectTo(x) isa ProjectTo{ChainRulesCore.NoTangent}()
+        x  # safe to ignore
+    else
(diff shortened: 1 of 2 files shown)


src/lib/broadcast.jl
```diff
@@ -239,7 +238,13 @@ function dual_function(f::F) where F
   end
 end
 
-@inline function broadcast_forward(f, args::Vararg{Any,N}) where N
+@inline function broadcast_forward(f::F, args::Vararg{Any,N}) where {F,N}
+  Base.issingletontype(F) || error("""Zygote's dual number broadcasting (as used on GPU arrays) cannot handle this function.
+      typeof(f) = $(F)""")
```
mcabbott, line 243: Maybe this should be a warning like #1224. We don't know that `f` ought to have a gradient, but it might.

jgreener64:
I think making this an error is a good idea. Ideally there would be a way to not error for custom types when you know it is okay to not track the gradients. I guess you can define `_dual_safearg` for your custom type but this might be worth describing in the error message, or exposing in a different way.

mcabbott:
Yes. I guess the thing you overload should eventually be something like  https://github.com/JuliaDiff/ChainRulesCore.jl/pull/528 

However, at the moment I believe you get errors from `unbroadcast` not having appropriate methods, if you try to use some weird type (even just a Symbol, IIRC). So it ought to be safe to make these deliberate errors now, and adjustable later.

PR #1226, opened 2022-05-12
Use CR adjoint for `logdet(::Cholesky)`

devmotion:
Once https://github.com/JuliaDiff/ChainRules.jl/pull/613 is merged, the Zygote adjoint for `logdet(::Cholesky)` can be removed. (FYI the CR PR also adds a rule for `det(::Cholesky)` and hence fixes some existing issues since no such rule exists in Zygote.)

Unfortunately, using this CR rule requires updating the Julia compat entry from 1.3 to 1.6. According to the [RegistryCI docs](https://juliaregistries.github.io/RegistryCI.jl/stable/guidelines/#New-versions-of-existing-packages) a patch release may not narrow the compat range for Julia. Hence you might want to postpone this PR until the next breaking release of Zygote.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.3' # Replace this with the minimum Julia version that your package supports.
+          - '1.6' # Replace this with the minimum Julia version that your package supports.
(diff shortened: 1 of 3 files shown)


devmotion:
The CR PR was merged, so this PR is ready.

CarloLucibello (approved):

PR #1231, opened 2022-05-20
Fix non-holomorphic tests

### LINKED ISSUE #1229: Possible mistake in complex tests
Hi, thanks all for your work on this package!

I think there may be a mistake in the non-holomorphic tests. Primarily, they appear to only reference the set of holomorphic functions used for the holomorphic tests:
https://github.com/FluxML/Zygote.jl/blob/dad65a8066669d4c8d7019c28a3dda06ad14c963/test/complex.jl#L72-L89
Further, I feel this should test the gradients of both the real and imaginary parts.

Finally, a question on my original concern which lead me to this: if we have a function which is fairly complicated, and somewhere in the middle it uses a non-holomorphic function, how does Zygote account for this? I imagine around this function the full jacobian of the real and imaginary outputs with respect to the real and imaginary inputs would be needed to ensure the correct final derivatives, is this used?

DomCRose:
Fixes #1229.

- Uses non-holomorphic function list in test.
- Adds test for derivative of imaginary component of function.
- Adds one more unusual function to test.

### DIFF
--- a/test/complex.jl
+++ b/test/complex.jl
@@ -63,11 +63,11 @@ fs_C_to_C_holomorphic = (cos,
             ε = 1e-8
             grad_fd_r = (f(z+ε)-f(z))/ε
             grad_fd_i = (f(z + ε * im) - f(z)) / (ε * im)
-            # check the function is indeed holomorphic
+            # Check the function is indeed holomorphic
             @assert abs(grad_fd_r - grad_fd_i) < sqrt(ε)
-            # check Zygote derivatives agree with holomorphic definition
-            @test abs(grad_zygote_r + im*grad_zygote_i) < sqrt(ε)
-            # check derivative agrees with finite differences
+            # Check Zygote derivatives agree with holomorphic definition
+            @test grad_zygote_r ≈ -im*grad_zygote_i
+            # Check derivative agrees with finite differences
             @test abs(grad_zygote_r - conj(grad_fd_r)) < sqrt(ε)
         end
     end
--- a/test/complex.jl
+++ b/test/complex.jl
@@ -58,12 +58,17 @@ fs_C_to_C_holomorphic = (cos,
 @testset "C->C holomorphic" begin
     for f in fs_C_to_C_holomorphic
         for z in (1.0+2.0im, -2.0+pi*im)
-            grad_zygote = gradient(real∘f, z)[1]
+            grad_zygote_r = gradient(real∘f, z)[1]
+            grad_zygote_i = gradient(imag∘f, z)[1]
             ε = 1e-8
             grad_fd_r = (f(z+ε)-f(z))/ε
-            grad_fd_i = (f(z+ε*im)-f(z))/(ε*im)
-            @assert abs(grad_fd_r - grad_fd_i) < sqrt(ε) # check the function is indeed holomorphic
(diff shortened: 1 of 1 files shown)


test/complex.jl
```diff
@@ -76,14 +76,20 @@ fs_C_to_C_non_holomorphic = (conj,
                              z->im*abs2(z),
                              z->z'z,
                              z->conj(z)*z^2,
+                             z->imag(z)^2+real(sin(z))^3*1im,
                              )
 @testset "C->C non-holomorphic" begin
-    for f in (fs_C_to_C_holomorphic...,fs_C_to_C_holomorphic...)
+    for f in (fs_C_to_C_holomorphic...,fs_C_to_C_non_holomorphic...)
```
mcabbott, line 82: 20 lines up there's another loop over only the holomorphic list. By joining the two lists here, it looks like the intention may have been to delete the other loop?

DomCRose: Indeed, I suppose that includes an assertion to check the function is holomorphic but thats the only difference. We could add that to this loop conditional on it being a function in the holomorphic list?

DomCRose: Having thought a little more, perhaps its best to keep it as two distinct loops and remove it here? They are somewhat distinct, and it could be good to add a check that Zygote produces a holomorphic gradient (i.e. compare the gradients of the real and imag outputs) for holomorphic functions, similar to how this is done with finite-diff to check the function actually is holomorphic.

DomCRose: Done!

test/complex.jl
```diff
@@ -58,12 +58,17 @@ fs_C_to_C_holomorphic = (cos,
...
+            grad_zygote_r = gradient(real∘f, z)[1]
+            grad_zygote_i = gradient(imag∘f, z)[1]
             ε = 1e-8
             grad_fd_r = (f(z+ε)-f(z))/ε
-            grad_fd_i = (f(z+ε*im)-f(z))/(ε*im)
-            @assert abs(grad_fd_r - grad_fd_i) < sqrt(ε) # check the function is indeed holomorphic
-            @test abs(grad_zygote - conj(grad_fd_r)) < sqrt(ε)
+            grad_fd_i = (f(z + ε * im) - f(z)) / (ε * im)
+            # check the function is indeed holomorphic
+            @assert abs(grad_fd_r - grad_fd_i) < sqrt(ε)
```
mcabbott, line 67: Why assert, not test?

test/complex.jl
```diff
@@ -58,12 +58,17 @@ fs_C_to_C_holomorphic = (cos,
...
+            # check Zygote derivatives agree with holomorphic definition
+            @test abs(grad_zygote_r + im*grad_zygote_i) < sqrt(ε)
```
mcabbott, line 69: I think this can just test `grad_zygote_r ≈ -im * grad_zygote_i`, it should not have error from finite differencing, just numerical rounding.

test/complex.jl
DomCRose: This was in the original version, and I'm guessing the logic is that a real error here would indicate an error in base, rather than Zygote, since its comparing the two finite differences. Instead, I think the aim of this is to provide a check that anyone editing the tests hasn't accidentally put a non-holomorphic function in `fs_C_to_C_holomorphic`.

Issue #1233, opened 2022-06-01
Deepcopy of Dictionaries in combination with mutating leads to errors.


ntausend:
Hey, I discovered a strange behavior which I am not sure it is intended in that way.

Consider the following example code:
```julia
function mutating1(d, a)
	upd = d[1] * a
	d[1] = upd
	return d
end

function mutating2(d,a)
	nd = deepcopy(d)
		
	upd = d[1] * a
	nd[1] = upd
	return nd
end

function mutating3(d,a)
	nd = deepcopy(d)
		
	upd = nd[1] * a
	d[1] = upd
	return d
end

function nrmsq(d)
	res = norm(d[1]*d[2])
	return res
end

d = 3
arry = map(_ -> randn(d,d), 1:2)
dic = Dict{Any, Any}()
for (jj, tm) in enumerate(arry)
    dic[jj] = tm
end

dicco = deepcopy(dic)
f1 = x -> nrmsq(mutating1(x, 5))
f2 = x -> nrmsq(mutating2(x, 5))
f3 = x -> nrmsq(mutating3(x, 5))

# all three are equal, make sure f1 which modifies the input is called last
# same holds for f3, so make sure to call it on an copy.
f2(dic) |> display
f3(dicco) |> display
f1(dic) |> display

# works
g1 = gradient(f1, dic)[1]

# do not work
g2 = gradient(f2, dic)[1]
g3 =gradient(f3, dic)[1]
```

The versions utilizing `f2` and `f3` are leading to strange `AssertionError: a === b` from `accum` while the version with `f1` works flawless.

Actually I have no idea how to deal with this and how to get an workaround.

Thanks for help in advance!

ToucheSir:
The assertion in question: https://github.com/FluxML/Zygote.jl/blob/v0.6.40/src/lib/base.jl#L27. I don't see any reason to not support this form of mutation as well, but perhaps there are some edge cases I've not thought of.

ntausend:
Ok thanks for the reply. 

Yes i guess thats the raised assertion, but I do not understand why its raised since I also not understand how mutation is implemented.

I tried understanding the `grad_mut(__contex__)` appearing in all functions. 

So if its possible for you to including this case it would be great, otherwise I will find an alternative to that problem :)

CarloLucibello:
Current status of this is that the gradient doesn't throw an error anymore but g2 and g3 are different from g1:
```julia

julia> g1
Dict{Any, Any} with 2 entries:
[...]

PR #1237, opened 2022-06-05
Allow accumulating distinct Dicts

ToucheSir:
Fixes the MWE in https://github.com/FluxML/Zygote.jl/issues/1233. Removing the `@assert` does not appear to be a problem, but there may be unknown consequences that haven't been captured by the existing test suite.

NB: failure on nightly are unrelated.


CarloLucibello:
@ToucheSir what happened to this PR? Can be merged if tests pass?

ToucheSir:
I believe so, thanks for the bump.

CarloLucibello (approved):

PR #1242, opened 2022-06-10
Fix #1241

### LINKED ISSUE #1241: Gradient missing when using Params within Dicts
Consider the following:
```
d = Dict{String, Vector{Float64}}("key"=>ones(4))
fn() = d["key"][2]
gs = gradient(fn, Params([d["key"]]))
```
Confusingly, `gs[d["key"]]` is `nothing`. 

If, on the other hand, you have the following:
```
x = d["key"]
fn2() = x[2]
gs = gradient(fn2, Params([x]))
```
...you get a `Zygote.OneElement` as expected. 

I think this is another manifestation of #1232.

bogiebro:
This applies the same pattern used for tracking implicit gradients in tuples to track implicit gradients in Dicts.

### DIFF
--- a/src/lib/base.jl
+++ b/src/lib/base.jl
@@ -29,11 +29,14 @@ function accum(a::AbstractDict, b::AbstractDict)
 end
 
 @adjoint function getindex(d::AbstractDict, k)
-  d[k], function (Δ)
+  val = d[k]
+  function dict_getindex_pullback(Δ)
+    accum_param(__context__, val, Δ) === nothing && return
     grad = grad_mut(__context__, d)
     grad[k] = accum(get(grad, k, nothing), Δ)
     return (grad, nothing)
   end
+  val, dict_getindex_pullback
 end
 
 @adjoint! function setindex!(d::AbstractDict, v, k)
--- a/test/lib/base.jl
+++ b/test/lib/base.jl
@@ -1,0 +1,13 @@
+@testset "base.jl" begin
+    @testset "dict_param" begin
+        d = Dict{String, Vector{Float64}}("key"=>ones(4))
+        fn() = d["key"][2]
+        result1 = gradient(fn, Params([d["key"]]))[d["key"]]
+
+        x = d["key"]
+        fn2() = x[2]
+        result2 = gradient(fn2, Params([x]))[x]
+
+        @test result1 == result2
+    end
+end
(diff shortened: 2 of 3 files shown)


src/lib/base.jl
```diff
@@ -29,7 +29,9 @@ function accum(a::AbstractDict, b::AbstractDict)
 end
 
 @adjoint function getindex(d::AbstractDict, k)
-  d[k], function (Δ)
+  val = d[k]
+  val, function (Δ)
```
ToucheSir, line 33:
```suggestion
  val, function Dict_getindex_pullback(Δ)
```

ToucheSir (commented):
Thanks! Could you add a test for this? Something like the MWE form #1241 would work.

src/lib/base.jl
bogiebro: Sure! Should I add it to the `lib.jl` test set in the file `test/lib/lib.jl`? Or, mirroring, the source, should I make a new `test/lib/base.jl` file and test set?

bogiebro: Regarding naming the pullback: none of the other pullbacks created in `base.jl` seem to have names. Returning an anonymous function seems like the preferred style.

ToucheSir: The latter, ideally. Currently a bunch of tests are all jammed into https://github.com/FluxML/Zygote.jl/blob/master/test/features.jl (including the current `getindex(::Dict)` ones, which you need not worry about for this PR), and we might as well take this opportunity to start breaking that file up into a saner file structure :)

CarloLucibello: Ideally we would like to turn all the anonymous functions into named ones and also move them to ChainRules' rrules

ToucheSir: What Carlo said, but that will be a longer term project and `getindex` is a little special. Were this most other functions, I would've recommended rewriting it as a `rrule` off the bat.

> none of the other pullbacks created in base.jl seem to have names. Returning an anonymous function seems like the preferred style.

Yes, this is a legacy code style and has caused no shortage of headaches while debugging. We plan on either removing them or giving them a name sooner than later, but in the meantime we can at least advocate for new pullbacks to have descriptive names.

bogiebro:
I added a test for #1241 and named the pullback.

test/lib/base.jl
```diff
@@ -0,0 +1,13 @@
+@testset "base.jl" begin
+    @testset "dict_param" begin
```
ToucheSir, line 2:
```suggestion
    @testset "Dict getindex with implicit params" begin
```

ToucheSir (approved):
Thanks @bogiebro !

Issue #1235, opened 2022-06-03
Deprecate `@nograd` and `dropgrad` in favour of `ChainRulesCore.@non_differentiable` and `ChainRulesCore.ignore_derivatives`


ToucheSir:
The ChainRules equivalents work just as well, are more broadly applicable and switching to them will let us remove more Zygote-isms.

PR #1245, opened 2022-06-17
deprecate `dropgrad` and `ignore`

mzgubic:
Contributes to #1235 

I'll do `nograd` separately since it is used for a few adjoints that may need to be ported first.

### DIFF
--- a/docs/src/utils.md
+++ b/docs/src/utils.md
@@ -20,7 +20,6 @@ Zygote.@showgrad
 Zygote.hook
 Zygote.Buffer
 Zygote.forwarddiff
-Zygote.ignore
 Zygote.checkpointed
 ```
 
--- a/docs/src/utils.md
+++ b/docs/src/utils.md
@@ -18,7 +18,6 @@ Zygote.withgradient
 Zygote.withjacobian
 Zygote.@showgrad
 Zygote.hook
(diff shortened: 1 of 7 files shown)


ToucheSir:
Perhaps we could add a note in the docs about these deprecations pointing to CRC, or keep the autodocs and add pointers in the docstrings?

CarloLucibello:
The integration test failures could be real

mzgubic:
>Perhaps we could add a note in the docs about these deprecations pointing to CRC, or keep the autodocs and add pointers in the docstrings?

Good idea, I've added a note to the docs.

>The integration test failures could be real

NeuralPDE.jl seems to be failing with the same error as master.

DynamicPPL fails locally for me (but CI runs). Maybe @devmotion could advise whether there are any special requirements on running tests locally?

devmotion:
> Maybe @devmotion could advise whether there are any special requirements on running tests locally?

No, there shouldn't. I assume locally you don't run them with the latest release of DistributionsAD, most likely because you pinned or checked out some packages or (possibly unintentially) created a Manifest.toml file.

mzgubic:
DynamicPPL seems to have been fixed. Good to go? I don't have merge rights so please just merge when ready.

PR #1246, opened 2022-06-19
Make errors for array mutation, try/catch, and foreign call nicer (+ docs)

### LINKED ISSUE #1077: Limitations Docs
It's well understood that Zygote doesn't support mutation (particularly of arrays). My recollection was that this features prominently in the docs, but I now can't see where it's mentioned -- certainly it's not obvious. To my mind it's important that we mention this prominently so that new users expectations are set appropriately.

@DhairyaLGandhi would you be open to a PR adding a `Limitations` section to the readme?

darsnack:
This just updates the mutation error to be more beginner friendly. It seems pretty common for non-ML parts of the ecosystem to use in-place for performance reasons, and when these first-time users combine their code with Zygote...well they get really confused. This should help.

Closes #1077.

### DIFF
--- a/docs/src/limitations.md
+++ b/docs/src/limitations.md
@@ -9,9 +9,11 @@ In this section, we will introduce examples where each of these errors occurs as
 
 ## Array mutation
 
-Array mutation is by far the most commonly encountered Zygote limitation. Unfortunately, supporting it natively in Zygote is tricky, though it may happen eventually. For now, let's focus on what counts as mutation, and how to fix it.
+Array mutation is by far the most commonly encountered Zygote limitation.
 
-Here we define a simple mutating function, `f!`, which modifies the elements of its input argument, `x`, in place.
+Automatic differentiation (AD) systems like Zygote are built on basic principles of calculus where we encounter _pure_ functions. This means that the function, ``y = f(x)``, does not modify ``x`` and only produces the output ``y`` based on ``x``. If we have a chain of functions, such as ``y = h(g(f(x)))``, we can apply the chain rule to differentiate it. AD systems are built to programmatically apply the chain rule to a series of function calls. Unfortunately, typical programs do not behave this way. We might allocate some memory, `x`, then call a function `y = f!(x)` that modifies `x` to produce the output `y`. This mutating behavior is a _side-effect_ of `f!`. Side-effects are difficult for AD systems to handle, because the must track changes to mutated variables and store older versions of the variable. For these reasons, Zygote does not handle array mutation for now.
+
+Let's explore this with a more concrete example. Here we define a simple mutating function, `f!`, which modifies the elements of its input argument, `x`, in place.
 ```julia
 function f!(x)
   x .= 2 .* x
--- a/docs/src/limitations.md
+++ b/docs/src/limitations.md
@@ -1,0 +1,148 @@
+# Limitations
+
+Zygote aims to support differentiating any code you might write in Julia, but it still has a few limitations. Notably, you might encounter errors when trying to differentiate:
+- array mutation
+- `try`/`catch` statements
+- "foreign call" expressions
+
+In this section, we will introduce examples where each of these errors occurs as well as possible work-arounds.
+
+## Array mutation
+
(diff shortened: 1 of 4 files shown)


ToucheSir (approved):
We should explore pulling this out into an `ArrayMutationError` type at some point with a custom `showerror`, but until then this LGTM.

ToucheSir:
Also yikes, were we hardcoding function names into error messages before? :exploding_head:

darsnack:
Well I don't know if what I wrote is any better in that sense...I'll make it slightly more re-usable, but I don't know how many places this error message will realistically be used.

ToucheSir:
Oh I wouldn't worry about that for this PR, but we'll definitely want to get rid of it when refactoring mutation errors.

ToucheSir:
Shall we merge this and tag?

darsnack:
I figured there ought to be a more comprehensive guide too that can be linked from the error. So I made a start towards #1077 w.r.t. this error (un-pushed).

You can tag, or merge this then tag and I can file a separate PR. I'll probably finish it up while bored on my flight.

ToucheSir:
Oh, very nice. Take your time then and we can incorporate them together.

ToucheSir (approved):
Looks great!

docs/src/limitations.md
```diff
@@ -0,0 +1,148 @@
...
+Zygote aims to support differentiating any code you might write in Julia, but it still has a few limitations. Notably, you might encounter errors when trying to differentiate:
+- array mutation
+- `try`/`catch` statements
+- "foreign call" expressions
+
+In this section, we will introduce examples where each of these errors occurs as well as possible work-arounds.
+
+## Array mutation
+
+Array mutation is by far the most commonly encountered Zygote limitation. Unfortunately, supporting it natively in Zygote is tricky, though it may happen eventually. For now, let's focus on what counts as mutation, and how to fix it.
```
mcabbott, line 12: I wonder if this should start by saying something more broad, to explain what the problem is. 

* Functions you consider in calculus class are pure: `y = f(x)` always does the same thing, depends only on `x`
* The simple programs to AD are things like `y = f10(f9(f8(...(f1(x))...)`, a chain of function applications. For these, Zygote works by looking up the rule for differentiating `f10`, and feeding its result into the rule for `f9`, etc.
* Real programs don't look like that. And in particular, they often work by allocating a piece of memory, and then sequentially modifying its contents. This is tricky for Zygote, for two reasons. First, side effects. And second, where to put the old values. For now it doesn't try to handle these, and gives an error.

Haven't tried to sketch words. Perhaps working through the fact that `*` calls `mul!`, which calls BLAS, is a good example?

ToucheSir: Perfect being the enemy of the good, perhaps this can be a follow-up PR? IMO getting the better error message out is a priority, while improving docs can be done at a more leisurely pace.

darsnack: Yeah, I agree walking through `*` -> `mul!` -> BLAS would be a good example here. Since I'm currently out of the country and my time to work on PRs is sporadic, I'm going to merge this for now. A follow-up PR would be good.

darsnack: Okay I added a little more without going into the `*` -> `mul!` -> BLAS example.

src/compiler/reverse.jl
```diff
@@ -118,7 +118,10 @@ function instrument(ir::IR)
     if isexpr(ex, :foreigncall, :isdefined)
       continue
     elseif isexpr(ex, :enter, :leave)
-      error("try/catch is not supported.")
+      error("""try/catch is not supported.
+            Refer to the Zygote documentation for fixes.
+            https://fluxml.ai/Zygote.jl/dev/limitations.html#Try-catch-statements-1
```
mcabbott, line 123: I think this link causes https://github.com/FluxML/Flux.jl/issues/2015 

The correct link seems to be https://fluxml.ai/Zygote.jl/dev/limitations/#Try-catch-statements-1 , but I wonder if we should avoid automatic `#heading` things in links, as they will break if anyone touches the heading text.

PR #1248, opened 2022-06-23
Improved type stability with explicit params

### LINKED ISSUE #1243: Zygote hangs when taking explicit gradients of NaiveGAFlux model
Continuing from https://github.com/FluxML/Flux.jl/issues/1986#issuecomment-1154524802. @DrChainsaw are you able to capture a profile or at least a stacktrace mid-hang? I think that would be the easiest way to get started troubleshooting this and trying to put together a MWE.

ToucheSir:
We can disable accumulating (implicit) parameters to the gradient cache in explicit mode. This can dramatically improve type stability because `accum_param` will return a `Union{Nothing, [grad type]}` otherwise.

One impact of this PR is that taking gradients of functions with both implicit and explicit parameters (i.e. calling `pullback` twice) may involve some additional compilation. However, given that we're trying to move users off of using implicit params anyhow, I see it as a small price to pay for being friendlier to the compiler.

Benchmarking TTFG on the MWE in https://github.com/FluxML/Zygote.jl/issues/1126, modified to use explicit params:
```julia-repl
julia> @time loss_grad(model, lr_images)
 85.027100 seconds (61.36 M allocations: 3.216 GiB, 1.17% gc time, 99.98% compilation time) # 0.6.40
 59.024238 seconds (60.72 M allocations: 3.174 GiB, 1.69% gc time, 99.98% compilation time) # this PR
```

Closes #1243.

### DIFF
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -4,6 +4,9 @@ using Core: Typeof
 import Base: copy!, IdSet
 import Base.Broadcast: broadcasted, materialize!
 
+# Internal container used to track accumulated gradients of mutable types (including params).
+# Type param I ∈ (true, false) indicates whether implicit params are in use.
+# By default, this should be false unless pullback(f, ::Params) is called.
 mutable struct Context{I} <: AContext
   cache::Union{IdDict{Any,Any},Nothing}
 end
@@ -47,11 +50,11 @@ function pullback(cx::Context, f, args...)
     Incorrect argument order for pullback, please use:
 
       pullback(f, __context__::Context, args)
-    
+
     instead of:
 
       pullback(__context__::Context, f, args)
-    
+
     This is usually caused by a call to pullback in a higher-order @adjoint.
     The above warning will become an error in Zygote 0.7.
     """
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -4,11 +4,11 @@ using Core: Typeof
 import Base: copy!, IdSet
 import Base.Broadcast: broadcasted, materialize!
 
-mutable struct Context <: AContext
+mutable struct Context{I} <: AContext
   cache::Union{IdDict{Any,Any},Nothing}
 end
 
-Context() = Context(nothing)
+Context() = Context{false}(nothing)
 
 cache(cx::Context) = cx.cache === nothing ? (cx.cache = IdDict()) : cx.cache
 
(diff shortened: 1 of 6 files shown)


ToucheSir:
Ok, now we're back to the known failures on Nightly and downstream NeuralPDE. Molly one appears to be an intermittent one which only showed up recently, but that should be investigated separately.

One unexpected find while working on this PR is that Zygote and downstream packages were calling `pullback` with the wrong argument order. This worked by sheer coincidence, but I've added a warning so that it can be rectified before we disallow it in the next breaking release.

src/compiler/interface.jl
```diff
@@ -4,11 +4,11 @@ using Core: Typeof
 import Base: copy!, IdSet
 import Base.Broadcast: broadcasted, materialize!
 
-mutable struct Context <: AContext
+mutable struct Context{I} <: AContext
```
CarloLucibello, line 7: What is this I for? Can you add some documentation comments?

CarloLucibello: I guess it stands for "implicit"

ToucheSir: Added a small explainer, see if that looks good to you :)

mcabbott:
This doesn't sound crazy, but FWIW I do not see the same speedup:
```
julia>@time gradient(loss, lr_images);
 51.949997 seconds (59.92 M allocations: 3.781 GiB, 4.14% gc time, 99.92% compilation time)  # tagged
 54.520161 seconds (60.06 M allocations: 3.801 GiB, 2.68% gc time, 99.94% compilation time)  # this PR
```
(Julia master, M1 mac.)

ToucheSir:
That may be because `loss` is closing over a global model. My version passes both as args.

mcabbott:
Avoiding globals makes a surprisingly large difference here:
```
julia> @time gradient((m, x) -> sum(m(x)), model, lr_images);
 35.009302 seconds (58.93 M allocations: 3.719 GiB, 8.44% gc time, 99.92% compilation time)  # tagged
 34.581913 seconds (59.08 M allocations: 3.738 GiB, 2.97% gc time, 99.93% compilation time)  # this PR
```

ToucheSir:
Perhaps a newer Julia version helps close the gap, but I'm consistently seeing this ~15s difference on 1.7.3.

This is the full script I'm using:
```julia
using Flux

channels = 4

function resblock(channels)
    return SkipConnection(Chain(
        Conv((3, 3), channels => channels, pad=1),
        Conv((3, 3), channels => channels, pad=1),
    ), +)
end

model = Chain(
    SkipConnection(
        Chain(
            resblock(channels),
            resblock(channels),
            resblock(channels),
            resblock(channels),
            resblock(channels),

            resblock(channels),
            resblock(channels),
            resblock(channels),
            resblock(channels),
            resblock(channels),

            resblock(channels),
            resblock(channels),
            resblock(channels),
            resblock(channels),
            resblock(channels),
        ),
    +),
    AdaptiveMeanPool((1, 1))
)

@show typeof(model)

loss(m, x) = sum(m(x))

lr_images = randn(Float32, 2, 2, channels, 1)
@time loss(model, lr_images)
@time loss(model, lr_images)

loss_grad(m, x) = gradient(m -> loss(m, x), m)
# This gives the same numbers:
# loss_grad(m, x) = gradient((m, x) -> loss(m, x), m, x)

@time loss_grad(model, lr_images)
@time loss_grad(model, lr_images)
```

src/compiler/interface.jl
```diff
@@ -36,10 +39,28 @@ _pullback(f, args...) = _pullback(Context(), f, args...)
...
 tailmemaybe(x::Tuple) = Base.tail(x)
 
-function pullback(f, args...)
-  y, back = _pullback(f, args...)
+@inline pullback(f, args...) = pullback(f, Context(), args...)
+function pullback(f, cx::AContext, args...)
+  y, back = _pullback(cx, f, args...)
   y, Δ -> tailmemaybe(back(Δ))
 end
+function pullback(cx::Context, f, args...)
```
DhairyaLGandhi, line 47: What is the motivation to switch the argument order?

ToucheSir: It's not switched. Per the line below, this signature was simply incorrect and only works due to sheer coincidence.

mcabbott (approved):
Let's do it.

PR #1249, opened 2022-06-23
deprecate `Zygote.@nograd`

### LINKED ISSUE #1235: Deprecate `@nograd` and `dropgrad` in favour of `ChainRulesCore.@non_differentiable` and `ChainRulesCore.ignore_derivatives`

mzgubic:
Closes #1235 

Needs https://github.com/JuliaDiff/ChainRules.jl/pull/637 to work.

The controversial thing is changing the tests in `test/lib/number.jl`. This is what ChainRules does, though I am not 100% certain why. In any case, if we decide against following the ChainRules convention here, it's possible to do it.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -25,7 +25,7 @@ ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 
 [compat]
 AbstractFFTs = "0.5, 1.0"
-ChainRules = "1.35.3"
+ChainRules = "1.37"
 ChainRulesCore = "1.9"
 ChainRulesTestUtils = "1"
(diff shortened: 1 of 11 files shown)


mzgubic:
Is CI passing good enough or do we want some manual inference checks here?

mzgubic:
bump

ToucheSir (approved):
Thanks! There was one rule where I found `@nograd` inferred better than `@non_differentiable`, but I can't think of it now and we can always write a manual `_pullback` for it if need be.

Issue #1244, opened 2022-06-16
nested gradient fails with "Mutating arrays is not supported"


tmbdev:
Source code:

```Julia
using Zygote

println("start")

function f(w, x)
    return sum(maximum((w * x).^2, dims=1))
end

function g(w, x)
    d = gradient(f, w, x)[2]
    return sum(d.^2)
end

function h(w, x)
    t = gradient(g, w, x)[1]
    return t
end

w = rand(10, 100)
x = rand(100, 7)

x = f(w, x)
println("x=", x)
y = g(w, x)
println("y=", y)
z = h(w, x)
println("z=", z)

print("done")
```

Output:
```
$ julia ztest.jl
start
x=4589.3400747042815
y=5.4911019133806146e11
ERROR: LoadError: Mutating arrays is not supported -- called setindex!(::Matrix{Float64}, _...)
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:33
  [2] (::Zygote.var"#437#438"{Matrix{Float64}})(#unused#::Nothing)
    @ Zygote ~/.julia/packages/Zygote/DkIUK/src/lib/array.jl:71
  [3] (::Zygote.var"#2337#back#439"{Zygote.var"#437#438"{Matrix{Float64}}})(Δ::Nothing)
    @ Zygote ~/.julia/packages/ZygoteRules/AIbCs/src/adjoint.jl:67
  [4] Pullback
    @ ~/.julia/packages/Zygote/DkIUK/src/lib/array.jl:327 [inlined]
  [5] (::typeof(∂(λ)))(Δ::Tuple{Matrix{Float64}})
    @ Zygote ~/.julia/packages/Zygote/DkIUK/src/compiler/interface2.jl:0
  [6] Pullback
    @ ~/.julia/packages/ZygoteRules/AIbCs/src/adjoint.jl:73 [inlined]
  [7] (::typeof(∂(λ)))(Δ::Tuple{Nothing, Nothing, Nothing, Matrix{Float64}})
    @ Zygote ~/.julia/packages/Zygote/DkIUK/src/compiler/interface2.jl:0
  [8] Pullback
    @ ~/exp/jadv/ztest.jl:6 [inlined]
  [9] (::typeof(∂(λ)))(Δ::Tuple{Nothing, Nothing, Float64})
    @ Zygote ~/.julia/packages/Zygote/DkIUK/src/compiler/interface2.jl:0
 [10] Pullback
[...]

PR #1250, opened 2022-06-24
rm rules for `maximum`, `minimum`, `dropdims`

mcabbott:
Closes https://github.com/FluxML/Zygote.jl/issues/1244

Could perhaps use a test

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -313,33 +313,9 @@ end
   sum(xs, dims = dims), Δ -> (nothing,)
 end
 
-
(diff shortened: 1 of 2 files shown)


ToucheSir:
Is there any way to add nested AD tests on the CR side? Otherwise one here would be great.

mcabbott:
CR has some tests that components used to make rules are themselves differentiable, in this case here:
https://github.com/JuliaDiff/ChainRules.jl/blob/main/test/rulesets/Base/array.jl#L369

But it's fairly easy to imagine missing something, so complete 2nd order tests using AD are probably better.

ToucheSir (approved):

PR #1252, opened 2022-06-25
rm rules for `Statistics.var`, `std`, `mean`

mcabbott:
ChainRules.jl should handle these.


ToucheSir:
Molly is using some of the internals removed here: https://github.com/JuliaMolSim/Molly.jl/blob/master/src/zygote.jl#L114. @jgreener64 how much of that file do you still actively require? Is there anything that must use the Zygote API instead of ChainRules?

jgreener64:
Thanks for the heads up. I can switch Molly to use ChainRules for this.

jgreener64:
On the off chance that this is waiting on changes in Molly, just go ahead here and I will change it later. The Zygote compat is upper-bounded in Molly since it relies on Zygote internals.

ToucheSir (approved):

PR #1254, opened 2022-06-26
Avoid materializing `I`, simplify `_eyelike` with CUDA and make it differentiable

DhairyaLGandhi:


src/lib/broadcast.jl
```diff
@@ -272,7 +272,8 @@ end
 
   end
 
-  @adjoint (::Type{T})(xs::Array) where {T <: CUDA.CuArray} =
+  _eyelike(y::CUDA.CuVector{T}) where T = CUDA.CuArray(I(length(y)))
```
ToucheSir, line 275:
```suggestion
  _eyelike(y::CUDA.CuVector) = CUDA.CuArray(I(length(y)))
```
To avoid the unbound type param.

src/lib/broadcast.jl
```diff
@@ -272,7 +272,8 @@ end
...
+  @adjoint (::Type{T})(xs::AbstractArray) where {T <: CUDA.CuArray} =
```
ToucheSir, line 276: IIUC the intent of the PR, would we not want a dedicated path `UniformScaling`? Taking `<:AbstractArray` presumes we can convert back to that subtype, but the implementation can only convert back to `Array` at present. Adapt.jl may be able to help us here.

ToucheSir (commented):
Do you have a link or other reference to what inspired this PR? The changes make sense to me, but given the [only place](https://github.com/FluxML/Zygote.jl/blob/7f2b16987dc4eab236aabd889444cabdcc6f9caf/src/lib/grad.jl#L160) `_eyelike` is called is not AD-safe in the first place, I'm confused as to the need for making only it differentiable.

CarloLucibello:
closing as outdated

Issue #1257, opened 2022-06-30
Remove `@adjoint` for `adjoint`/`Adjoint` in favor of `ChainRules`


mtfishman:
Can we remove the `@adjoint` definitions for `adjoint`/`Adjoint`:
https://github.com/FluxML/Zygote.jl/blob/4777767737b4c95d2cea842933c5b2edae2771b2/src/lib/array.jl#L371-L381
in favor of the `rrule`s in `ChainRules` (found [here](https://github.com/JuliaDiff/ChainRules.jl/blob/e9d7d0acb7a034bedeba84fefee2029eb821022a/src/rulesets/LinearAlgebra/structured.jl#L114-L158))?

I want to overload the `rrule` for `adjoint` for a custom type, but I found that it ignores my `rrule` in favor of Zygote's rules, for example:
```julia
julia> using ChainRulesCore

julia> using ZygoteRules

julia> using Zygote

julia> struct MyType end

julia> Base.adjoint(x::MyType) = x

julia> f(x) = (MyType()'; x)
f (generic function with 1 method)

julia> f'(1.0)
1.0

julia> ChainRulesCore.rrule(::typeof(adjoint), x::MyType) = error("ChainRulesCore.rrule")

julia> f'(1.0)
1.0

julia> ZygoteRules.@adjoint Base.adjoint(x::MyType) = error("ZygoteRules.@adjoint")

julia> f'(1.0)
ERROR: ZygoteRules.@adjoint
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] adjoint
   @ ./REPL[10]:1 [inlined]
 [3] _pullback
   @ ~/.julia/packages/ZygoteRules/AIbCs/src/adjoint.jl:65 [inlined]
 [4] _pullback
   @ ./REPL[6]:1 [inlined]
 [5] _pullback(ctx::Zygote.Context, f::typeof(f), args::Float64)
   @ Zygote ~/.julia/packages/Zygote/IoW2g/src/compiler/interface2.jl:0
 [6] _pullback(f::Function, args::Float64)
   @ Zygote ~/.julia/packages/Zygote/IoW2g/src/compiler/interface.jl:34
 [7] pullback(f::Function, args::Float64)
   @ Zygote ~/.julia/packages/Zygote/IoW2g/src/compiler/interface.jl:40
 [8] (::Zygote.var"#62#63"{typeof(f)})(x::Float64)
[...]

PR #1259, opened 2022-07-03
rm `adjoint` & `transpose` adjoints

mcabbott:
closes https://github.com/FluxML/Zygote.jl/issues/1257

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -337,34 +337,6 @@ end
 # LinearAlgebra
 # =============
 
-@adjoint function transpose(x)
-  back(Δ) = (transpose(Δ),)
-  back(Δ::NamedTuple{(:parent,)}) = (Δ.parent,)
-  return transpose(x), back
-end
-
-@adjoint function LinearAlgebra.Transpose(x)
-  back(Δ) = (LinearAlgebra.Transpose(Δ),)
-  back(Δ::NamedTuple{(:parent,)}) = (Δ.parent,)
-  return LinearAlgebra.Transpose(x), back
-end
-
-
-@adjoint function Base.adjoint(x)
-  back(Δ) = (Δ',)
-  back(Δ::NamedTuple{(:parent,)}) = (Δ.parent,)
-  return x', back
-end
-
-@adjoint function LinearAlgebra.Adjoint(x)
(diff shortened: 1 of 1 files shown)


src/lib/array.jl
```diff
@@ -337,34 +337,6 @@ end
...
-end
-
-@adjoint function LinearAlgebra.Adjoint(x)
-  back(Δ) = (LinearAlgebra.Adjoint(Δ),)
-  back(Δ::NamedTuple{(:parent,)}) = (Δ.parent,)
-  return LinearAlgebra.Adjoint(x), back
-end
-
-@adjoint parent(x::LinearAlgebra.Adjoint) = parent(x), ȳ -> (LinearAlgebra.Adjoint(ȳ),)
-@adjoint parent(x::LinearAlgebra.Transpose) = parent(x), ȳ -> (LinearAlgebra.Transpose(ȳ),)
```
mcabbott, line 366: I forgot that  https://github.com/JuliaDiff/ChainRulesCore.jl/pull/446  isn't merged, so removing these lines probably isn't a good idea. 

Before:
```
julia> gradient(x -> first(parent(x)), [1,2,3]')[1]
1×3 adjoint(::Vector{Float64}) with eltype Float64:
 1.0  0.0  0.0
```
This PR, with tagged CRC:
```
julia> gradient(x -> first(parent(x)), [1,2,3]')[1]
(parent = [1.0, 0.0, 0.0],)
```

ToucheSir: One alternative idea is to desguar the `@adjoint` and overload `_pullback` directly. That'll have the benefit of removing some unnecessary layers, while making it easier to do stuff like https://github.com/FluxML/ZygoteRules.jl/issues/22.

mcabbott: Glad to see there were tests for this, at least: fails on `isapprox(::NamedTuple{(:parent,), ...}, ::Matrix{Float64})`.

mcabbott:
Ok to merge?

ToucheSir (approved):
Yes, I hadn't noticed https://github.com/FluxML/Zygote.jl/pull/1259/commits/26ae24b9cb104428d7e69e4717810a6917dc15e6

PR #1265, opened 2022-07-18
Fix link in error message

### LINKED ISSUE #1260: Error message link 404 error

ziyiyin97:
The link in the error message doesn't work now https://fluxml.ai/Zygote.jl/dev/limitations.html#Array-mutation-1
should be https://fluxml.ai/Zygote.jl/latest/limitations#Array-mutation-1 instead

[Edit: Closes #1260]

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -75,7 +75,7 @@ the elements of arrays in place (e.g. setting values with x .= ...)
 Possible fixes:
 - avoid mutating operations (preferred)
 - or read the documentation and solutions for this error
-  https://fluxml.ai/Zygote.jl/dev/limitations.html#Array-mutation-1
+  https://fluxml.ai/Zygote.jl/latest/limitations
 """)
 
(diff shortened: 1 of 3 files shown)


src/lib/array.jl
```diff
@@ -75,7 +75,7 @@ the elements of arrays in place (e.g. setting values with x .= ...)
 Possible fixes:
 - avoid mutating operations (preferred)
 - or read the documentation and solutions for this error
-  https://fluxml.ai/Zygote.jl/dev/limitations.html#Array-mutation-1
+  https://fluxml.ai/Zygote.jl/latest/limitations#Array-mutation-1
```
darsnack, line 78:
```suggestion
  https://fluxml.ai/Zygote.jl/latest/limitations
```

darsnack (changes requested):
Thanks! I think we decided that we prefer a link like this (no header) because it will be more stable. Do you mind fixing it here and elsewhere?

PR #1272, opened 2022-07-25
Replace `@require CUDA` with `using GPUArraysCore`

mcabbott:
Should have no functional changes for CUDA, but make the same paths used for any GPUArray.

(Many of these paths should probably be removed, but not in this PR.)

Tests will need GPUArrays 8.4.2... (strange that buildkite manages to load GPUArrays v8.3.2 which should be forbidden?)


CarloLucibello:
I cannot understand the buildkite error but it looks real

mcabbott:
This PR requires GPUArraysCore@0.1.1, which was added in https://github.com/JuliaGPU/GPUArrays.jl/pull/417 and is registered. The corresponding version of GPUArrays is not yet registered.

The registry should not allow GPUArrays@8.4.1 with anything except GPUArraysCore@0.1.0:  https://github.com/JuliaRegistries/General/pull/64447 . So instead it has loaded an even older GPUArrays@8.3.2, before the Core package existed. 

And then tests fail, as the overloads needed for CuArrays aren't present.

So perhaps this package needs also to bound GPUArrays@8.4.2 to prevent this.

mcabbott:
This should be ready to go.

Test failures on nightly are Cholesky etc. changes, unrelated.

CarloLucibello:
Why we have to depend on GpuArrays instead of GpuArrausCore only?

mcabbott:
Just to bound the version.

Using a version of CUDA + GPUArrays earlier than the latest will mean that you get a `GPUArraysCore.AbstractGPUArray` type which is completely unrelated to CuArray.

I don't love it but I think it shouldn't add to loading time, as the big package is never loaded.

CarloLucibello:
ok, it's a bit tricky but still better than `@require CUDA` I guess

mcabbott:
Besides just avoiding `@require`, this should also mean that other GPUArrays take the same path as CuArrays. One nice effect of that is that testing with JLArrays will see the same behaviour in more cases.

CarloLucibello (approved):

Issue #1234, opened 2022-06-01
Error from `has_chain_rrule` on an array of `Bool`


mcabbott:
The bug reported here https://discourse.julialang.org/t/help-me-pin-this-bug-in-flux/81901 can be simplified a little further. Note that the error occurs on the forward pass, and requires an array of booleans:
```julia
julia> z, back = pullback([true, false]) do x
           y2 = reshape(x, 1, 2, 1)
           y3 = repeat(y2, 3)
           y3[1]
       end
ERROR: type Nothing has no field method
Stacktrace:
  [1] getproperty(x::Nothing, f::Symbol)
    @ Base ./Base.jl:37
  [2] matching_cr_sig(t::IRTools.Inner.Meta, s::Nothing)
    @ Zygote ~/.julia/packages/Zygote/DkIUK/src/compiler/chainrules.jl:72
  [3] has_chain_rrule(T::Type)
    @ Zygote ~/.julia/packages/Zygote/DkIUK/src/compiler/chainrules.jl:54
  [4] #s53#1205
    @ ~/.julia/packages/Zygote/DkIUK/src/compiler/interface2.jl:20 [inlined]
  [5] var"#s53#1205"(::Any, ctx::Any, f::Any, args::Any)
    @ Zygote ./none:0
  [6] (::Core.GeneratedFunctionStub)(::Any, ::Vararg{Any})
...

julia> gradient([true, false]) do x
           y2 = reshape(x, 1, 2, 1) .+ 0.0
           y3 = repeat(y2, 3)
           y3[1]
       end
(nothing,)

julia> gradient([1, 0]) do x
           y2 = reshape(x, 1, 2, 1)
           y3 = repeat(y2, 3)
           y3[1]
       end
([1.0, 0.0],)
```

In testing variants, I also got this error, which is unrelated I think. But similar to  https://github.com/JuliaDiff/ChainRules.jl/issues/627 
```julia
julia> gradient([true, false]) do x
           y2 = reshape(x, 1, 2, 1)
           @show summary(y2)
           y3 = repeat(y2, outer=3)
           @show summary(y3)
           y3[1]
       end
summary(y2) = "1×2×1 Array{Bool, 3}"
ERROR: ArgumentError: number of outer repetitions (1) cannot be less than number of dimensions of input array (3)
Stacktrace:
  [1] check(arr::Array{Bool, 3}, inner::Tuple{Int64, Int64, Int64}, outer::Int64)
[...]

PR #1278, opened 2022-08-02
Propagate ambiguities from rrule lookup instead of failing inexplicably

ToucheSir:
Fixes https://github.com/FluxML/Zygote.jl/issues/1234.


mzgubic (approved):
Nice, this is great!

ToucheSir:
All credit goes to @oxinabox for the snippet in https://github.com/FluxML/Zygote.jl/issues/1234#issuecomment-1201149208 :)

PR #1280, opened 2022-08-03
Passthrough safe ccalls in threading code

ToucheSir:
Should fix the 2 compiler-related failures on nightly, e.g. https://github.com/FluxML/Zygote.jl/runs/7613462813?check_suite_focus=true. The rest look to be factorization-related&mdash;@sethaxen could I bug you about those?


sethaxen:
I'm only seeing Cholesky-related errors. I tested locally that these are resolved by https://github.com/JuliaDiff/ChainRules.jl/pull/633.

mcabbott (approved):

Issue #1065, opened 2021-09-08
broken iteration over dictionary


CarloLucibello:
Issue related to #725 :
```julia
julia> d = Dict(:a => 5, :b => 6);

julia> function f(d)
         s = 0
         for (k,v) in d
           s += v
         end
         s
       end
f (generic function with 1 method)

julia> gradient(f, d)
ERROR: MethodError: no method matching getindex(::Dict{Any, Any})
Closest candidates are:
  getindex(::Dict{K, V}, ::Any) where {K, V} at dict.jl:480
  getindex(::AbstractDict, ::Any) at abstractdict.jl:494
  getindex(::AbstractDict, ::Any, ::Any, ::Any...) at abstractdict.jl:504
Stacktrace:
  [1] (::Zygote.var"#back#222"{:vals, Zygote.Context, Dict{Symbol, Int64}, Vector{Int64}})(Δ::Zygote.OneElement{Int64, 1, Tuple{Int64}, Tuple{Base.OneTo{Int64}}})
    @ Zygote ~/.julia/dev/Zygote/src/lib/lib.jl:233
  [2] (::Zygote.var"#1789#back#223"{Zygote.var"#back#222"{:vals, Zygote.Context, Dict{Symbol, Int64}, Vector{Int64}}})(Δ::Zygote.OneElement{Int64, 1, Tuple{Int64}, Tuple{Base.OneTo{Int64}}})
    @ Zygote ~/.julia/packages/ZygoteRules/OjfTt/src/adjoint.jl:59
  [3] Pullback
    @ ./Base.jl:33 [inlined]
  [4] (::typeof(∂(getproperty)))(Δ::Zygote.OneElement{Int64, 1, Tuple{Int64}, Tuple{Base.OneTo{Int64}}})
    @ Zygote ~/.julia/dev/Zygote/src/compiler/interface2.jl:0
  [5] Pullback
    @ ~/.julia/packages/ZygoteRules/OjfTt/src/ZygoteRules.jl:11 [inlined]
  [6] (::typeof(∂(literal_getproperty)))(Δ::Zygote.OneElement{Int64, 1, Tuple{Int64}, Tuple{Base.OneTo{Int64}}})
    @ Zygote ~/.julia/dev/Zygote/src/compiler/interface2.jl:0
  [7] Pullback
    @ ./dict.jl:684 [inlined]
  [8] Pullback
    @ ./dict.jl:688 [inlined]
[...]

PR #1285, opened 2022-08-10
Add rule for Dict iteration

ToucheSir:
Fixes the first example in https://github.com/FluxML/Zygote.jl/issues/1065. N.B. that `_pullback` is used over `@adjoint` because we're trying to get rid of it, and over `rrule` because support for Dict tangents in CR is still spotty (+ we don't have to be nearly as general).

### DIFF
--- a/src/lib/base.jl
+++ b/src/lib/base.jl
@@ -47,6 +47,45 @@ end
   end
 end
 
+# This rule behaves much like the getindex adjoint,
+# just with an (internal) ordinal index instead of a key.
+function _pullback(cx::AContext, ::typeof(iterate), d::Dict, i)
+  iter = iterate(d, i)
+  function dict_iterate_pullback(Δ)
+    (iter === nothing || Δ === nothing) && return
+    k, v = iter[1]
+    _, dv = Δ[1]
+    accum_param(cx, v, dv) === nothing && return
+    grad = grad_mut(cx, d)
+    grad[k] = accum(get(grad, k, nothing), dv)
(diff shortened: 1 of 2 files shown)


CarloLucibello:
NNlib failure real but unrelated?

ToucheSir:
Nightly failure has been around for ages and should be fixed by https://github.com/FluxML/Zygote.jl/pull/1280. NNlib failure I'm unsure of, need to investigate.

ToucheSir:
CI looks green now that nightly and downstream failures have been addressed.

mohamed82008:
What's blocking this PR?

mohamed82008:
Any update here?

oxinabox (approved):
This looks right to me.
Except we should either mutate the gradient and return `nothing`
Or out of place compute the change in gradient and return it to be `accum`ed later.
Not both.

But this is a problem with all the `Dict` code. Which is why `accum(::Dict, ::Dict)` is defined as a no-op

So approving

ToucheSir:
> Except we should either mutate the gradient and return `nothing`
> Or out of place compute the change in gradient and return it to be `accum`ed later.
> Not both.

I actually tested this after #1288 and found it broke some tests, but I don't recall which ones. Shall we continue the discussion there?

mohamed82008:
Thanks for the PR and the review everyone! Can we get a release?

ToucheSir:
https://github.com/JuliaRegistries/General/pull/67749

PR #1287, opened 2022-08-15
`@adjoint broadcasted` for unary minus

mcabbott:
The absence of this rule means that the one from ChainRules gets called, which is lazier. 

Closes  https://github.com/FluxML/NNlib.jl/issues/432


ToucheSir (approved):

ToucheSir:
Not sure what you mean? The code snippet in that comment is copied almost verbatim from an existing test.

PR #1295, opened 2022-08-26
Handle nothing grads for `Pairs.data`

ToucheSir:
MWE:
```julia
no_kwarg_grad(x; kwargs...) = x[kwargs[:i]]
gradient(x -> no_kwarg_grad(x; i=1), [1]) # MethodError: no method matching getindex(::Nothing, ::Int64) on master
```
@ChrisRackauckas it also appears that whatever [SciMLSensitivity downstream tests](https://github.com/FluxML/Zygote.jl/blob/master/.github/workflows/Downstream.yml#L25) are running are inadequate to catch stuff like this. In fact, I'm not sure if any _are_ running given that file still reads DiffEqFlux. What should we replace that line with?

### DIFF
--- a/src/lib/base.jl
+++ b/src/lib/base.jl
@@ -162,7 +162,7 @@ function _pullback(cx::AContext, ::typeof(literal_getindex),
                    ps::Iterators.Pairs{<:Any,<:Any,<:Any,<:NamedTuple}, ::Val{K}) where K
   val, gf_back = _pullback(cx, literal_getfield, NamedTuple(ps), Val(K))
   function kwargs_literal_getindex_pullback(Δ)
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
It's just in a different test group. https://github.com/FluxML/Zygote.jl/pull/1296

ToucheSir:
Is there a way to run both groups on the same worker? The Layers group is pretty fast and amortizing the startup cost would be nice.

ChrisRackauckas:
Bump?

darsnack (approved):

ToucheSir:
https://github.com/JuliaRegistries/General/pull/67317

PR #1296, opened 2022-08-26
Add DiffEqFlux BasicNeuralDE Test

ChrisRackauckas:


ToucheSir:
Is there a way to run both groups on the same worker? The Layers group is pretty fast and amortizing the startup cost would be nice.

ChrisRackauckas:
You could use `All`, but that might be long 😓

ToucheSir (approved):

PR #1298, opened 2022-08-30
Expand limitations page of docs

mcabbott:
This proposes to add a new section to the "limitations" page, indicating some frequent bugs. Thoughts?

### DIFF
--- a/docs/src/limitations.md
+++ b/docs/src/limitations.md
@@ -163,7 +163,9 @@ Zygote has limited support for mutation, and in particular will allow you to cha
 
 However, this is buggy. Searching for ["mutable struct"](https://github.com/FluxML/Zygote.jl/issues?q=is%3Aissue+is%3Aopen+mutable+struct) on the issue tracker will return current problems.
 
-The simple solution is to use only immutable `struct`s.
+The simple solution is to use only immutable `struct`s. 
+
+If you need to modify them, using something like `@set` from [Accessors.jl](https://github.com/JuliaObjects/Accessors.jl) should work well. This returns a new object, but does not have side-effects on other copies of it. 
 
 ## Re-using variable names
 
--- a/docs/src/limitations.md
+++ b/docs/src/limitations.md
@@ -1,11 +1,14 @@
-# Limitations
+# Design Limitations
 
-Zygote aims to support differentiating any code you might write in Julia, but it still has a few limitations. Notably, you might encounter errors when trying to differentiate:
-- array mutation
-- `try`/`catch` statements
-- "foreign call" expressions
+Zygote aims to support differentiating any Julia code, but it still has a few limitations.
+Notably, you might encounter errors when trying to differentiate:
+- array mutation,
+- `try`/`catch` statements,
+- "foreign call" expressions.
 
-In this section, we will introduce examples where each of these errors occurs as well as possible work-arounds.
+This section gives examples where each of these errors occurs, as well as possible work-arounds.
+
+Below, it also describes some known bugs in expressions Zygote ought to be able to handle.
 
 ## Array mutation
 
@@ -123,7 +126,7 @@ Stacktrace:
 ```
 `jclock` will multiply the result of our C function by an argument. When we try to differentiate with respect to this argument, we get an `foreigncall` error.
 
-## Solutions
+# Solutions
 
 For all of the errors above, the suggested solutions are similar. You have the following possible work arounds available (in order of preference):
 1. avoid the error-inducing operation (e.g. do not use mutating functions)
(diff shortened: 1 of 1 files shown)


docs/src/limitations.md
```diff
@@ -148,3 +151,42 @@ julia> gradient(jclock, rand())
...
+
+Zygote's issue tracker has the current list of open [bugs](https://github.com/FluxML/Zygote.jl/issues?q=is%3Aissue+is%3Aopen+label%3Abug). There are some general principles about things you may wish to avoid if you can:
+
+## `mutable struct`s
+
+Zygote has limited support for mutation, and in particular will allow you to change a field in some `mutable struct X; a; b; end` by setting `x.a = val`.
+
+However, this is buggy. Searching for ["mutable struct"](https://github.com/FluxML/Zygote.jl/issues?q=is%3Aissue+is%3Aopen+mutable+struct) on the issue tracker will return current problems.
+
+The simple solution is to use only immutable `struct`s.
```
ToucheSir, line 166: A mention of libraries like Accessors.jl for those who still want to "mutate"/update immutable structs in their code could be a nice addition.

docs/src/limitations.md
```diff
@@ -148,3 +151,42 @@ julia> gradient(jclock, rand())
...
+
+# Known Bugs
+
+Zygote's issue tracker has the current list of open [bugs](https://github.com/FluxML/Zygote.jl/issues?q=is%3Aissue+is%3Aopen+label%3Abug). There are some general principles about things you may wish to avoid if you can:
+
+## `mutable struct`s
+
+Zygote has limited support for mutation, and in particular will allow you to change a field in some `mutable struct X; a; b; end` by setting `x.a = val`.
+
+However, this is buggy. Searching for ["mutable struct"](https://github.com/FluxML/Zygote.jl/issues?q=is%3Aissue+is%3Aopen+mutable+struct) on the issue tracker will return current problems.
```
ToucheSir, line 164: Should we dissuade users from using Dicts too? If anything they seem to be _more_ buggy than mutable structs.

docs/src/limitations.md
```diff
@@ -148,3 +151,42 @@ julia> gradient(jclock, rand())
...
+The simple solution is to use only immutable `struct`s.
+
+## Re-using variable names
+
+It is common to accumulate values in a loop by re-binding the same variable name to a new value
+many times, for example:
+```
+function mysum(x::Real, n::Int)
+  tot = 0.0
+  for i in 1:n
```
ToucheSir, line 175: "Use loops and control flow sparingly (favouring `ifelse` and manual unrolling via `ntuple`/`@nexprs`/etc.)" is another potential heading.

docs/src/limitations.md
```diff
@@ -148,3 +151,42 @@ julia> gradient(jclock, rand())
...
+## Second derivatives
+
+In principle Zygote supports taking derivatives of derivatives. There are, however, a few problems:
+* Quite a few of its rules are not written in a way that is itself differentiable. For instance they may work by making an array then writing into it, which is mutation of the sort forbidden above. 
+* The complexity of the code grows rapidly, as Zygote differentiates its own un-optimised output.
+* Reverse mode over reverse mode is seldom the best algorithm.
+
+The issue tracker has a label for [second order](https://github.com/FluxML/Zygote.jl/issues?q=is%3Aissue+is%3Aopen+label%3A%22second+order%22), which will outline where the bodies are buried.
+
+Often using a different AD system over Zygote is a better solution. This is what [`hessian`](@ref) does, using ForwardDiff over Zygote.
```
ToucheSir, line 192: If we were to revive Tracker, that would be a good future addition to this section. Or maybe even now, given Zygote defines an adjoint for it.

docs/src/limitations.md
mcabbott:
```suggestion
The simple solution is to use only immutable `struct`s. 

If you need to modify them, using something like `@set` from [Accessors.jl](https://github.com/JuliaObjects/Accessors.jl) should work well. This returns a new object, but does not have side-effects on other copies of it. 
```

docs/src/limitations.md
mcabbott: Sure. I have completely ignored any Dict issues because the idea horrifies me.

docs/src/limitations.md
mcabbott: I know people have favourite ways to use ReverseDiff and Zygote together, in some order... I do not want to write a section making recommendations about which will work, but if someone else does that would be fine.

mcabbott: Maybe this:
```suggestion
Often using a different AD system over Zygote is a better solution.
This is what [`hessian`](@ref) does, using ForwardDiff over Zygote, but other combinations are possible.
(Note that rules defined here mean that Zygote over ForwardDiff is translated to ForwardDiff over ForwardDiff.)
```
And then we should either merge or close this.

Issue #1301, opened 2022-09-01
zygote broadcast type stability


maartenvd:
I have a package that defines a simple utility Array type : https://github.com/maartenvd/MPSKit.jl/blob/diskarray/src/utility/periodicarray.jl
Zygote changes types when taking the derivative, which later on makes my backward rules fail. Here is a minimal example:

```
julia> f_add(x) = x + 3;
julia> function myfun(x)
       y = f_add.(x);
       @show typeof(y)
       norm(y)
       end
myfun (generic function with 1 method)
julia> myfun'(PeriodicArray(rand(5,5)));
typeof(y) = Matrix{Float64}

julia> myfun(PeriodicArray(rand(5,5)));
typeof(y) = PeriodicArray{Float64, 2}
```

This type change causes failures, as it then calls rrule with a tangent type of PeriodicArray, but a (wrong) primal type of Matrix

mcabbott:
What it's doing is this: https://github.com/FluxML/Zygote.jl/blob/17a5673bdfd7d8fcdf27454b75b15efdb6477c9e/src/lib/broadcast.jl#L250-L252
```julia
julia> out = Zygote.dual_function(f_add).(PeriodicArray(rand(2,3)))
2×3 PeriodicArray{ForwardDiff.Dual{Nothing, Float64, 1}, 2}:
 Dual{Nothing}(3.18449,1.0)  Dual{Nothing}(3.32056,1.0)  Dual{Nothing}(3.9397,1.0)
 Dual{Nothing}(3.20766,1.0)  Dual{Nothing}(3.20073,1.0)  Dual{Nothing}(3.19711,1.0)

julia> y = map(x -> x.value, out)
2×3 Matrix{Float64}:
 3.18449  3.32056  3.9397
 3.20766  3.20073  3.19711
```
I am not sure why the second is `map` , since calling `broadcast(x -> x.value, out)` would preserve this type. Want to make a PR and see what breaks?

FWIW, ChainRules uses https://github.com/JuliaDiff/ChainRules.jl/blob/main/src/rulesets/Base/broadcast.jl#L108-L114 which does not preserve such types, would be nice if it did too:
```
julia> myfun(PeriodicArray(rand(5,5)));
typeof(y) = PeriodicArray{Float64, 2}

julia> gradient(myfun, PeriodicArray(rand(5,5)));
┌ Debug: split broadcasting forwards
│   frule_fun = frule_via_ad (generic function with 1 method)
[...]

PR #1302, opened 2022-09-01
preserve type over broadcast

maartenvd:
fixes https://github.com/FluxML/Zygote.jl/issues/1301

### DIFF
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -195,7 +195,7 @@ _dual_safearg(x) = false
   end
   len = inclen(args)
   y∂b = _broadcast((x...) -> _pullback(__context__, f, x...), args...)
-  y = map(first, y∂b)
+  y = broadcast(first, y∂b)
   function ∇broadcasted(ȳ)
     dxs_zip = map(((_, pb), ȳ₁) -> pb(ȳ₁), y∂b, ȳ)
     dxs = ntuple(len) do i
@@ -249,7 +249,7 @@ end
   valN = Val(N)
   out = dual_function(f).(args...)
   eltype(out) <: Dual || return (out, _ -> nothing)
-  y = map(x -> x.value, out)
+  y = broadcast(x -> x.value, out)
   function bc_fwd_back(ȳ)
     dargs = ntuple(valN) do i
       unbroadcast(args[i], broadcast((y1, o1) -> y1 * o1.partials[i], ȳ, out))
--- a/test/lib/base.jl
+++ b/test/lib/base.jl
@@ -1,3 +1,5 @@
+using LinearAlgebra;
+
 @testset "base.jl" begin
     @testset "Dict getindex with implicit params" begin
         d = Dict{String, Vector{Float64}}("key"=>ones(4))
@@ -9,5 +11,12 @@
         result2 = gradient(fn2, Params([x]))[x]
 
         @test result1 == result2
+    end
+
+    @testset "Type stability under broadcast" begin
+        g_simple = gradient(a->sum(broadcast(x->x+3,a)),Diagonal([1,2,3]));
+        g_hard = gradient((a,b)->sum(broadcast(x->x*b,a)),Diagonal([1,2,3]),4);
+        @test first(g_simple) isa Diagonal
+        @test first(g_hard) isa Diagonal
     end
 end


mcabbott:
Glad existing tests pass with this.

Is there an easy way to test the new behaviour? Ideally hitting both paths (i.e. broadcasting a fairly simple function and a nasty one). Is there a simple package with an array type for which the old code didn't work & the new does?

maartenvd:
I guess LinearAlgebra.Diagonal could serve as a test (it does fail on master, and works here)? I'm not sure what the cleanest way is of hitting the slow fallback though

mcabbott:
One way to be sure of hitting the slow path is to broadcast a closure, like
```
gradient((x,y) -> sum(broadcast((z -> (z*x)^2), y)), 1.0, [2,3,4.0])
```

mcabbott:
Am not sure Diagonal will work as a test, and the one in [ea21486](https://github.com/FluxML/Zygote.jl/pull/1302/commits/ea21486bcd50b2e161d3aff21652f78f743711e6) seems to pass on tagged Zygote. 

In general its map & broadcast seem to agree:
```
julia> map(sqrt, Diagonal([2,3]))
2×2 Diagonal{Float64, Vector{Float64}}:
 1.41421   ⋅ 
  ⋅       1.73205

julia> broadcast(sqrt, Diagonal([2,3]))
2×2 Diagonal{Float64, Vector{Float64}}:
 1.41421   ⋅ 
  ⋅       1.73205

julia> map(cos, Diagonal([2,3]))
2×2 Matrix{Float64}:
 -0.416147   1.0
  1.0       -0.989992

julia> broadcast(cos, Diagonal([2,3]))
2×2 Matrix{Float64}:
 -0.416147   1.0
  1.0       -0.989992
```
I thought perhaps there were some disagreements for Adjoint, but now I don't see them. Something like `[1,2]' .+ 3` makes a Matrix, but that's un-`map`-like. For types from packages, many overload `map` to preserve themselves:
```
julia> using NamedDims

julia> sqrt.(NamedDimsArray([1,2], :a))
2-element NamedDimsArray(::Vector{Float64}, :a):
↓ a  1.0
     1.4142135623730951

julia> map(sqrt, NamedDimsArray([1,2], :a))
2-element NamedDimsArray(::Vector{Float64}, :a):
↓ a  1.0
     1.4142135623730951
```
Is it possible that the issue here is just that the original array type lacks such an overload? Or maybe there's a reason it cannot?

maartenvd:
I am also not entirely sure anymore - the documentation about implementing your own array types at the very least does not mention overloading map, but does mention specifying the broadcaststyle (which determines the output type of broadcasts) (see https://docs.julialang.org/en/v1/manual/interfaces/ )

This map vs broadcast thing could cause other problems, but I feel that in general julia really should make both broadcast and map use the same output type automatically. Using only base types I could come up with this example:

```
julia> a = Set((1,2,3))
Set{Int64} with 3 elements:
  2
  3
  1

julia> broadcast(x->x+3,a)
3-element Vector{Int64}:
 5
 6
 4

julia> map(x->x+3,a)
ERROR: map is not defined on sets
Stacktrace:
 [1] error(s::String)
   @ Base .\error.jl:33
 [2] map(f::Function, #unused#::Set{Int64})
   @ Base .\abstractarray.jl:2326
 [3] top-level scope
   @ REPL[26]:1
```

mcabbott:
Note that many functions like `sortslices(PeriodicArray(rand(3,3)); dims=1)` do work, via `similar`. In fact I think `map` will work without being explicitly overloaded, if you define this method:
```
Base.similar(a::PeriodicArray, ::Type{T}, dims::Tuple{Vararg{Int64, N}}) where {T, N} = PeriodicArray(similar(a.data, T, dims))
```
I think the one with a splat calls this. The docs mention `Base.Dims` which is `Tuple{Vararg{Int64, N}} where N` but maybe this could be much more explicit? Possibly these ones may be necc too, maybe only if you want to wrap things with offset indexing?
```
 [80] similar(a::AbstractArray, ::Type{T}, dims::Tuple{Vararg{Int64, N}}) where {T, N}
     @ abstractarray.jl:827
 [81] similar(a::AbstractArray, ::Type{T}, dims::Tuple{Integer, Vararg{Integer}}) where T
     @ abstractarray.jl:825
 [82] similar(a::AbstractArray, ::Type{T}, dims::Tuple{Union{Integer, Base.OneTo}, Vararg{Union{Integer, Base.OneTo}}}) where T
     @ abstractarray.jl:824
```

maartenvd:
But would you then rather not change zygote to correctly call broadcast where it should? Consider the example of Set, where map errors while broadcast works just fine. I really don't like having to define methods outside the official recommended interface, just to work around something that zygote arguably does wrong

mcabbott:
I do think this is the documented interface, although I agree many implementations don't quite follow it (including [some which I wrote](https://github.com/mcabbott/AxisKeys.jl/blob/648ba1e58d7ba97a038c42d6e3cb332251754ff9/src/functions.jl#L335)). Making Zygote more forgiving is OK. 

I can't see a way to push a Set into this bit of Zygote. Supposedly it does support Dict, for which neither broadcast nor map works. I think even Tuple doesn't reach this code, only arrays.

One reason people write `map` when there's one argument / shapes match is that broadcasting is pretty complex. E.g. for tuples this is very simple unrolled thing, instead of a giant tower of functions. But I think there's no chance this actually matters for performance here.

mcabbott (approved):

PR #1306, opened 2022-09-20
Make Zygote's logo a real PNG

Saransh-cpp:
Looks much better now! (like Flux's logo)

I couldn't push to a new branch in `Zygote` (access issues) for the docs preview to work, but here is how this looks (in the sidebar) -

## Before
image
image

## After
image
image


mcabbott:
Thanks!

Is there an easy way to make the black text more visible in dark mode, here and in Flux? Maybe the side-bar should be a lighter shade (and darker, in the light theme)? 

(And should we remove the text Zygote below the logo, seems a bit repetitive?)

Saransh-cpp:
image

image

Much better now!

mcabbott (approved):
That looks perfect! And sure, Flux too, why not?

PR #1307, opened 2022-09-20
Wct/fix repeats euclidean

willtebbutt:
Fixes an issue found in KernelFunctions: https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/474

Handles a problem that occurs when there are repeated inputs.

### DIFF
--- a/src/lib/distances.jl
+++ b/src/lib/distances.jl
@@ -80,7 +80,7 @@ end
 
 @adjoint function pairwise(::Euclidean, X::AbstractMatrix; dims=2)
 
-  _conditional(d, δ) = d > δ ? sqrt(d) : zero(δ)
+  _conditional(d, δ) = d > δ ? sqrt(d) : zero(d)
 
   function _pairwise_euclidean(X)
     δ = eps(eltype(X))^2
--- a/src/lib/distances.jl
+++ b/src/lib/distances.jl
@@ -65,7 +65,6 @@ end
 end
 
 @adjoint function pairwise(::Euclidean, X::AbstractMatrix, Y::AbstractMatrix; dims=2)
-  @show "in this other one"
   # Modify the forwards-pass slightly to ensure stability on the reverse.
   function _pairwise_euclidean(X, Y)
     δ = eps(promote_type(eltype(X), eltype(Y)))^2
--- a/src/lib/distances.jl
+++ b/src/lib/distances.jl
@@ -65,6 +65,7 @@ end
 end
 
 @adjoint function pairwise(::Euclidean, X::AbstractMatrix, Y::AbstractMatrix; dims=2)
+
   # Modify the forwards-pass slightly to ensure stability on the reverse.
   function _pairwise_euclidean(X, Y)
     δ = eps(promote_type(eltype(X), eltype(Y)))^2
--- a/src/lib/distances.jl
+++ b/src/lib/distances.jl
@@ -65,7 +65,7 @@ end
 end
 
 @adjoint function pairwise(::Euclidean, X::AbstractMatrix, Y::AbstractMatrix; dims=2)
-
+  @show "in this other one"
   # Modify the forwards-pass slightly to ensure stability on the reverse.
   function _pairwise_euclidean(X, Y)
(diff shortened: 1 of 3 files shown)


willtebbutt:
@DhairyaLGandhi @oxinabox @mzgubic looks like this is good to go. If someone has time for a review, would be greatly appreciated

src/lib/distances.jl
```diff
@@ -79,11 +79,16 @@ end
...
 
 @adjoint function pairwise(::Euclidean, X::AbstractMatrix; dims=2)
-  D, back = pullback(X -> pairwise(SqEuclidean(), X; dims = dims), X)
-  D .= sqrt.(D)
-  return D, function(Δ)
-    Δ = Δ ./ (2 .* max.(D, eps(eltype(D))))
-    Δ[diagind(Δ)] .= 0
-    return (nothing, first(back(Δ)))
+
+  _conditional(d, δ) = d > δ ? sqrt(d) : zero(δ)
```
mzgubic, line 83:
```suggestion
  _conditional(d, δ) = d > δ ? sqrt(d) : zero(d)
```
Shouldn't matter in this case as `d` and `δ` are the same type, but still to be as defensive as possible in case this code gets copied/reused/taken as an example.

src/lib/distances.jl
```diff
@@ -79,11 +79,16 @@ end
 end
 
 @adjoint function pairwise(::Euclidean, X::AbstractMatrix; dims=2)
-  D, back = pullback(X -> pairwise(SqEuclidean(), X; dims = dims), X)
-  D .= sqrt.(D)
-  return D, function(Δ)
-    Δ = Δ ./ (2 .* max.(D, eps(eltype(D))))
-    Δ[diagind(Δ)] .= 0
```
mzgubic, line 86: I wonder whether this worked in higher order AD

mzgubic (approved):
Just to make sure I understand: the issue was that with repeated inputs the distances were meant to be zero, but were not numerically zero, and that error propagated?

src/lib/distances.jl
willtebbutt: I can't imagine that it did, because mutation involved in `Δ[diagind(Δ)] .= 0`

willtebbutt:
> Just to make sure I understand: the issue was that with repeated inputs the distances were meant to be zero, but were not numerically zero, and that error propagated?

Almost -- the issue is that the distances were indeed zero, so we're calling `sqrt(0)` somewhere. The derivative of `sqrt` isn't defined at zero. This PR is a way to adopt the sub-gradient convention (or whatever it's called) such that the derivative of `sqrt` is treated as being `zero` for inputs very close to zero. This is essentially what the binary method of `pairwise` does, but implemented slightly differently.

willtebbutt:
Would like to merge, but I seem to have lost write access to Zygote -- could someone please re-instate it?

darsnack:
That's an inadvertent mistake, you should be able to merge now!

src/lib/distances.jl
```diff
@@ -79,11 +79,16 @@ end
...
+
+  function _pairwise_euclidean(X)
+    δ = eps(eltype(X))^2
+    D2 = pairwise(SqEuclidean(), X; dims=dims)
+    return _conditional.(D2, δ)
   end
+  D, back = pullback(_pairwise_euclidean, X)
+
+  _pairwise_pullback(Δ) = (nothing, back(Δ)...)
+  return D, _pairwise_pullback
```
devmotion, line 93: I assume this could be simplified to
```suggestion
@adjoint function pairwise(dist::Euclidean, X::AbstractMatrix; dims=2)

  _conditional(d, δ) = d > δ ? sqrt(d) : zero(d)

  function _pairwise_euclidean(sqdist::SqEuclidean, X)
    δ = eps(eltype(X))^2
    D2 = pairwise(sqdist, X; dims=dims)
    return _conditional.(D2, δ)
  end

  return pullback(_pairwise_euclidean, SqEuclidean(dist.thresh), X)
```
which would avoid the additional function + splatting and be closer to the primal as it would use the same `thresh` threshold.

PR #1310, opened 2022-09-21
Unify and simplify adjoints for `pairwise(::Euclidean, ...)`

devmotion:
Implements the suggestions in https://github.com/FluxML/Zygote.jl/pull/1307#discussion_r976398636 and unifies both `pairwise` adjoints.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.6.48"
+version = "0.6.49"
 
 [deps]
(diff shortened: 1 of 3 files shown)


willtebbutt:
Could you add a test for `pairwise(Euclidean(), X, X; dims=2)` for when the cols are repeats of one another? i.e. a test which is analogous to the one that I added in my last PR for `pairwise(Euclidean(), X; dims=2)`? Primarily to ensure that we avoid regression in this case

willtebbutt:
Yeah, I think it is. I wanted the case in which `X == Y` _and_ the columns of `X` are repeated, which I think is what you've tested here.

willtebbutt (approved):

PR #1316, opened 2022-09-27
Fix and update `CompatHelper` CI

Saransh-cpp:
See the failing CI jobs here - https://github.com/FluxML/Zygote.jl/actions/workflows/CompatHelper.yml

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -3,16 +3,37 @@ on:
   schedule:
     - cron: 0 0 * * *
   workflow_dispatch:
+
+permissions:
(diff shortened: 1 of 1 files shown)


devmotion (approved):
I guess the update to CompatHelper 3 would be sufficient but I don't think it hurts to update the action with the defaults suggested by CompatHelper.

devmotion:
It is working correctly again: https://github.com/FluxML/Zygote.jl/actions/runs/3138102092/jobs/5097045848 Thanks a lot, @Saransh-cpp!

Saransh-cpp:
Great! No worries! :)

Issue #1276, opened 2022-08-01
`cispi` has poor performance


roflmaostc:
Hi!

`cispi` seems to have type infereability issues:
```julia
f1(x) = sum(abs2, cispi.(x))
f2(x) = sum(abs2, cis.(x))

@time Zygote.gradient(f1, ones((1024,1024)))
# 1.121823 seconds (18.87 M allocations: 648.002 MiB, 31.55% gc time)

@time Zygote.gradient(f2, ones((1024,1024)))
#   0.066483 seconds (40 allocations: 168.001 MiB, 30.13% gc time)
```

Zygote uses ChainRules, doesn't it?
There is this [rule for `sincospi`](https://github.com/JuliaDiff/ChainRules.jl/pull/497/files) implemented which should be called by `cispi`.

Best,

Felix

ToucheSir:
It should use the rule, but I suspect something else is causing broadcasting to take a slow path here (you can guess because that path requires a couple of allocations per element to store pullbacks, and x1 million is not far off the reported amount)

ToucheSir:
Ah, this is the same issue as https://github.com/FluxML/Zygote.jl/issues/961#issuecomment-871084267 because `cispi` returns a complex number.

roflmaostc:
But `cis`has good performance?

And the linked issue was only for CUDA

ToucheSir:
I would not say it has good performance. Even though the memory allocations are better, they and overall runtime are strictly worse than a function with a more "expensive" forward pass.
```julia
manyexp(x) = exp(x) * exp(x) + exp(x)
x = ones(1024, 1024)

julia> @btime cis.($x);
  10.829 ms (2 allocations: 16.00 MiB)

julia> @btime manyexp.($x);
  17.025 ms (2 allocations: 8.00 MiB)

julia> @btime gradient(x -> sum(abs2, cis.(x)), $x);
  78.728 ms (38 allocations: 160.00 MiB)

julia> @btime gradient(x -> sum(abs2, manyexp.(x)), $x);
  22.156 ms (31 allocations: 40.00 MiB)
```
It just so happens that the pullback for `cis` is simple enough to not capture any values and thus be inlined even on the slow path. `cispi`'s pullback is not so fortunate (I think because of the `reverse` and splat?) and so has to be materialized for every array element.

> And the linked issue was only for CUDA

The linked comment discusses a general-purpose solution that works for the CPU path as well. I'm not sure what level of effort would be required to implement it.

[...]

Issue #1121, opened 2021-11-18
gradient returns nothing for `sum(abs2, x)` with a complex CuArray


LexaLutyi:
`gradient` returns `nothing` for `CuArrays{ComplexF32}`, but works fine with `Arrays{ComplexF32}`:

```
julia> a = CUDA.rand(ComplexF32, 2)
2-element CuArray{ComplexF32, 1, CUDA.Mem.DeviceBuffer}:
  0.040781975f0 + 0.31357694f0im
 0.0050712824f0 + 0.464033f0im

julia> gradient(t -> sum(abs2, t), a)
(nothing,)

julia> a = rand(ComplexF32, 2)
2-element Vector{ComplexF32}:
 0.6999197f0 + 0.343145f0im
 0.4877541f0 + 0.47177994f0im

julia> gradient(t -> sum(abs2, t), a)
(ComplexF32[1.3998394f0 + 0.68629f0im, 0.9755082f0 + 0.9435599f0im],)
```

`real` and `imag` behave similarly to `abs2`

DhairyaLGandhi:
Definitely a bug! Could you check on older Zygote releases, say v0.6.3?

LexaLutyi:
> 
> 
> Definitely a bug! Could you check on older Zygote releases, say v0.6.3?

Yes, on version v0.6.3 it works correctly

```
julia> a = CUDA.rand(ComplexF32, 2)
2-element CuArray{ComplexF32, 1, CUDA.Mem.DeviceBuffer}:
 0.12610757f0 + 0.47263396f0im
  0.4873352f0 + 0.42900836f0im

julia> gradient(t -> sum(abs2, t), a)
(ComplexF32[0.25221515f0 + 0.9452679f0im, 0.9746704f0 + 0.8580167f0im],)

julia> b = Array(a)
2-element Vector{ComplexF32}:
 0.12610757f0 + 0.47263396f0im
  0.4873352f0 + 0.42900836f0im

julia> gradient(t -> sum(abs2, t), b)
(ComplexF32[0.25221515f0 + 0.9452679f0im, 0.9746704f0 + 0.8580167f0im],)
```

DhairyaLGandhi:
I wonder if it's related to the projection related issues too.

LexaLutyi:
The problem occurs when using Flux, where Zygote is v0.6.30. If you install Zygote without Flux, then Zygote is v0.6.12 and everything works correctly.

[...]

PR #1324, opened 2022-10-25
Adding complex broadcasting for gradients on the GPU

ptiede:
This is a first attempt to add support for taking gradients of complex numbers when broadcasting and on the GPU. This targets issues #961 #1121 #1215. 

A nice side effect of this pull request is that complex broadcasting doesn't have to take the slow route anymore when on the CPU, and fixes the performance issues in #1276 

On the current Zygote.jl release, I get:
```julia
f1(x) = sum(abs2, cispi.(x))
f2(x) = sum(abs2, cis.(x))

@btime Zygote.gradient($f1, $ones((1024,1024)))
# 926.936 ms (18874448 allocations: 640.00 MiB)

@btime Zygote.gradient($f2, $(ones((1024,1024))))
# 39.655 ms (38 allocations: 160.00 MiB)
```

With this pull-request I get
```julia
@btime Zygote.gradient($f1, $(ones((1024,1024))))
#  19.668 ms (31 allocations: 72.00 MiB)

@btime Zygote.gradient($f2, $(ones((1024,1024))))
#  15.781 ms (31 allocations: 72.00 MiB)
```

### Approach
To fix these issues, I changed how `broadcast_forward` and `dual_function` work. This was inspired by @mcabbott [comment](https://github.com/FluxML/Zygote.jl/issues/961#issuecomment-871084267) but with some changes to ensure there are no dynamic dispatches or type instabilities.  Specifically, I had to change the `dual` function since
```julia
if any(a isa Complex for a in args...)
   ...
else 
  ...
end
``` 
was leading to some type instability warnings on the GPU and some other strange issues. 

On top of the change to `dual` another change is how `broadcast_forward` works. I had to make four separate functions depending on the output and the arguments broadcast. I am not sure if there is a better way to do this, but it currently works and passes all tests on my machine. One concern I had was what to do for complex->complex functions. For this, I just followed what was listed in https://juliadiff.org/ChainRulesCore.jl/stable/maths/complex.html, but maybe we don't want to follow that?

### Testing
In terms of testing, I have added some small tests to `cuda.jl` to ensure that nothing is not returned and that the gradient on the GPU and CPU are the same. Since I also changed broadcast_forward on the CPU (always taking the fast path) I believe there is already sufficient testing done there. 

### PR Checklist

- [x] Tests are added


src/lib/broadcast.jl
```diff
@@ -232,23 +235,44 @@ end
...
+
+# We do this because it ensures type stability so it compiles nicely on the gpu
+dual(x, i, N) = x
+dual(x::Bool, i, ::Val{N}) where {N} = x
+dual(x::Real, i, ::Val{N}) where {N} = Dual(x, ntuple(j-> i==j, Val(N)))
+# For complex since ForwardDiff.jl doesn't play nicely with complex numbers we
+# construct a Complex dual number and tag the real and imaginary parts separately
+function dual(x::Complex, i, ::Val{N}) where {N}
+    re_dual = Dual(real(x), ntuple(j->i==j, Val(2N)))
+    im_dual = Dual(imag(x), ntuple(j->(N+i)==j, Val(2N)))
```
ToucheSir, line 247:
```suggestion
    re_dual = Dual(real(x), ntuple(==(i), 2N))
    im_dual = Dual(imag(x), ntuple(==(N+i), 2N))
```
Some local testing suggests that constant propagation should be able to figure this out without rewrapping with `Val`. If it doesn't though, feel free to disregard.

src/lib/broadcast.jl
```diff
@@ -232,23 +235,44 @@ end
...
+    return Complex(re_dual, im_dual)
+end
 
 function dual_function(f::F) where F
-  function (args::Vararg{Any,N}) where N
-    ds = map(args, ntuple(identity,Val(N))) do x, i
-      dual(x, ntuple(j -> i==j, Val(N)))
+    function (args::Vararg{Any,N}) where N
+      ds = map(args, ntuple(identity,Val(N))) do x, i
+        tmp = dual(x, i, Val(N))
```
ToucheSir, line 254:
```suggestion
      ds = map(args, ntuple(identity, N)) do x, i
        tmp = dual(x, i, N)
```
Same as above. I believe all subsequent uses of `Val(N)` might be redundant as well.

ToucheSir (commented):
Mostly leaving this here to say the test failures are expected, but a couple suggestions while I'm at it:

src/lib/broadcast.jl
ptiede: So it looks like I can get rid of almost every Val. However, if I get rid of this one the following test fails
https://github.com/FluxML/Zygote.jl/blob/6742644da91ad22e3ab7081c4d8b77eeeafe4e79/test/cuda.jl#L20-L28
due to a type instability. 

If I instead do this
```suggestion julia
      ds = map(args, ntuple(identity, N) do x, i
        tmp = dual(x, i, Val(N))
```
then everything works on 1.6 and 1.8 on my machine, so I think this Val needs to remain.

src/lib/broadcast.jl
```diff
@@ -232,23 +235,44 @@ end
 import ForwardDiff
 using ForwardDiff: Dual
 
-dual(x, p) = x
-dual(x::Real, p) = Dual(x, p)
-dual(x::Bool, p) = x
+
+# We do this because it ensures type stability so it compiles nicely on the gpu
+dual(x, i, N) = x
+dual(x::Bool, i, ::Val{N}) where {N} = x
```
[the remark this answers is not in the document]
ptiede: Done!

src/lib/broadcast.jl
ptiede: Done!

test/cuda.jl
```diff
@@ -140,3 +140,21 @@ end
   @test_skip gradient((x,y) -> sum(vcat(x,y)), 1f0, r, 2f0, r)[2] isa CUDA.CuArray{Float32}
 end
 
+
+@testset "CUDA complex broadcasting" begin
+    # Issue 961 and 1121 and 1215
+    x = rand(Float32, 50)
+    y = complex(rand(Float32, 50))
```
mcabbott, line 147: Why define `x` here at all?

Also, this `y` has zero imaginary part. `rand(ComplexF64, 50)` would be a stronger test.
```
julia> complex(rand(Float32, 50))
50-element Vector{ComplexF32}:
  0.89825445f0 + 0.0f0im
  0.40070343f0 + 0.0f0im
  0.29411656f0 + 0.0f0im
  0.44503874f0 + 0.0f0im
```

mcabbott (commented):
This looks good, thanks for tacking it! 

I meant to take a closer read but haven't yet, sorry.

>  I believe there is already sufficient testing done there.

Sadly I would not assume this. There may be very few tests of complex broadcasting, not sure (maybe I missed a section). It might be worth trying to come up with some evil test cases, including e.g. fused broadcasts where only parts are complex.

test/cuda.jl
ptiede: Oops! That `x` was for a test I was doing on my machine. I think overall that the testing could be a bit better though so I've added another test that uses both real and complex arguments. I probably need to add some additional tests.

mcabbott: Cool. I think `x.^2 .*y .+ y` uses only functions which have special rules, and ought to work without this PR. I think even broadcasting trivial functions like `add(x,y) = x+y` will change the path it takes. But messy examples (e.g. with trig, conj/real/imag, in all sorts of ways)  are much more likely to expose mistakes like a `conj` missing somewhere.

mcabbott: Trying to invent some functions, did not try them on GPU:
```
r3 = Float32.(inv.(2:4))
c3 = ComplexF32.(inv.(5:7) .+ im ./ (8:10))

@test gradient(r -> sum(abs2, log.(1 .+ im .* r)./2), r3)[1] ≈ [0.2077734, 0.15268978, 0.11885023]
@test gradient(c -> sum(abs2, imag.(sqrt.(c .+ im))), c3)[1] ≈ [-0.4124833f0 + 0.49228126f0im, -0.4258298f0 + 0.49446818f0im, -0.43560573f0 + 0.49583605f0im]
@test gradient((r,c) -> sum(abs2, @. sin(conj(c)/r' - im) - imag(c + tanh(r/c'))), r3, c3)[2] ≈ [2.9423256f0 + 63.7845f0im, -2.7483354f0 + 55.08628f0im, -9.976982f0 + 48.902283f0im]
```
But locally, with this branch, I expected them to use the new code... but adding printing doesn't seem to work?
```
(jl_S8DfLf) pkg> st Zygote
Status `/private/var/folders/yq/4p2zwd614y59gszh7y9ypyhh0000gn/T/jl_S8DfLf/Project.toml`
  [e88e6eb3] Zygote v0.6.49 `https://github.com/ptiede/Zygote.jl#pt-complexbroadcast`

julia> @eval Zygote function dual(x::Complex, i, N)  # from PR, with printing
            @show x
            re_dual = Dual(real(x), ntuple(==(i), 2N))
            im_dual = Dual(imag(x), ntuple(==(N+i), 2N))
            return Complex(re_dual, im_dual)
        end;

julia> Zygote.refresh()

julia> @test gradient(r -> sum(abs2, log.(1 .+ im .* r)./2), r3)[1] ≈ [0.2077734, 0.15268978, 0.11885023]
Test Passed
```

ptiede: So I looked into this and this occurred because I hadn't added a Complex method for `_dual_safearg`. When I added this some issues started to appear. One of them was because the partials for the real and complex parts had different lengths. 

However, that is not the big issue. The big issue is that certain functions seem to be causing some type instabilities during the evaluation of the dual numbers. For instance,

```julia
x = rand(Complex{Float32}, 100)
f(x) = sum(abs2, log.(y))
@code_warntype Zygote.dual_function(f).(x)

MethodInstance for (::var"##dotfunction#314#7")(::Vector{ComplexF32})
  from (::var"##dotfunction#314#7")(x1) in Main
Arguments
  #self#::Core.Const(var"##dotfunction#314#7"())
  x1::Vector{ComplexF32}
Body::Union{Vector{ForwardDiff.Dual{Float32, Float32, 2}}, Vector{ForwardDiff.Dual{Float32, V, 2} where V}, Vector{ForwardDiff.Dual{Float32, Float64, 2}}}
1 ─ %1 = Zygote.dual_function::Core.Const(Zygote.dual_function)
│   %2 = (%1)(Main.f)::Core.Const(Zygote.var"#944#946"{typeof(f)}(f))
│   %3 = Base.broadcasted(%2, x1)::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1}, Nothing, Zygote.var"#944#946"{typeof(f)}, Tuple{Vector{ComplexF32}}}
│   %4 = Base.materialize(%3)::Union{Vector{ForwardDiff.Dual{Float32, Float32, 2}}, Vector{ForwardDiff.Dual{Float32, V, 2} where V}, Vector{ForwardDiff.Dual{Float32, Float64, 2}}}
└──      return %4```
```

Has a problem where the broadcast can't seem to figure out that eltype of the partial field in `Dual` should be a `Float32`. What is really annoying is that this problem does not occur for `Float64` where I get
```julia
x64 = Complex{Float64}.(x)
@code_warntype Zygote.dual_function(f)(x64)

MethodInstance for (::var"##dotfunction#313#6")(::Vector{ComplexF64})
  from (::var"##dotfunction#313#6")(x1) in Main
Arguments
  #self#::Core.Const(var"##dotfunction#313#6"())
  x1::Vector{ComplexF64}
Body::Vector{ForwardDiff.Dual{Float64, Float64, 2}}
1 ─ %1 = Zygote.dual_function::Core.Const(Zygote.dual_function)
│   %2 = (%1)(Main.f)::Core.Const(Zygote.var"#944#946"{typeof(f)}(f))
│   %3 = Base.broadcasted(%2, x1)::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1}, Nothing, Zygote.var"#944#946"{typeof(f)}, Tuple{Vector{ComplexF64}}}
│   %4 = Base.materialize(%3)::Vector{ForwardDiff.Dual{Float64, Float64, 2}}
└──      return %4
```


```

Comment (mcabbott) [reply test/cuda.jl]: Sometimes closing over a type goes badly, since `typeof(Float32) == DataType`. Could that be happening somewhere? I don't see where though... 

On a quick look I thought there were places where this closes over `N` and creates `Val(N)` inside another function; assigning the `Val(N)` to a variable outside might be better. But may not matter.

Comment (ptiede) [reply test/cuda.jl]: Ok looking into this more. It appears the `log` with `Complex{Dual{Float32}}` arguments is type unstable.
My guess is that this occurs because there isn't using the specific forward rule for a complex number for log, or likely any common functions.

Comment (mcabbott) [reply test/cuda.jl]: That is weird, `@code_warntype log(Dual(1f0, 1f0) + im)` is bad. Inside `Base.ssqs`, it looks like `ldexp(Dual(1f0, 2f0), 3)` makes a Float64 dual, by a method from ForwardDiff.

Comment (ptiede) [reply test/cuda.jl]: Ok sounds good! I'll skip log for now and make tests for other functions.

Comment (ptiede) [reply test/cuda.jl]: Alright I was able to add the  last test,
```julia
@test gradient((r,c) -> sum(abs2, @. sin(conj(c)/r' - im) - imag(c + tanh(r/c'))), r3, c3)[2] ≈ [2.9423256f0 + 63.7845f0im, -2.7483354f0 + 55.08628f0im, -9.976982f0 + 48.902283f0im]
```
and everything passes! The other two tests suggested both run into the `ldexp` problem with Float32. I have opened up an issue https://github.com/JuliaDiff/ForwardDiff.jl/issues/604 detailing the problem. The good news is that when I fix the problem locally all the tests pass!

ptiede: Here are a couple of updates on my end. First, I just realized I was running the previous test on the CPU. When I run it on the GPU, I get a scalar indexing error. The stack trace is 
```julia
julia>     @test gradcheck_gpu((r,c) -> sum(abs2, @. sin(conj(c)/r' - im) - imag(c + tanh(r/c'))), r3, c3)
Error During Test at /home/ptiede/.julia/dev/Zygote/test/cuda.jl:186
  Test threw exception
  Expression: gradcheck_gpu(((r, c)->begin
            sum(abs2, #= /home/ptiede/.julia/dev/Zygote/test/cuda.jl:186 =# @__dot__(sin(conj(c) / r' - im) - imag(c + tanh(r / c'))))
        end), r3, c3)
  Scalar indexing is disallowed.
  Invocation of getindex resulted in scalar indexing of a GPU array.
  This is typically caused by calling an iterating implementation of a method.
  Such implementations *do not* execute on the GPU, but very slowly on the CPU,
  and therefore are only permitted from the REPL for prototyping purposes.
  If you did intend to index this array, annotate the caller with @allowscalar.
  Stacktrace:
    [1] error(s::String)
      @ Base ./error.jl:35
    [2] assertscalar(op::String)
      @ GPUArraysCore ~/.julia/packages/GPUArraysCore/lojQM/src/GPUArraysCore.jl:87
    [3] getindex(::CuArray{ComplexF32, 2, CUDA.Mem.DeviceBuffer}, ::Int64, ::Int64)
      @ GPUArrays ~/.julia/packages/GPUArrays/fqD8z/src/host/indexing.jl:9
    [4] getindex
      @ ~/.julia/juliaup/julia-1.8.2+0.x64/share/julia/stdlib/v1.8/LinearAlgebra/src/adjtrans.jl:180 [inlined]
    [5] _unsafe_getindex_rs
      @ ./reshapedarray.jl:250 [inlined]
    [6] _unsafe_getindex
      @ ./reshapedarray.jl:247 [inlined]
    [7] getindex
      @ ./reshapedarray.jl:235 [inlined]
    [8] iterate
      @ ./abstractarray.jl:1167 [inlined]
    [9] iterate
      @ ./abstractarray.jl:1165 [inlined]
   [10] iterate
      @ ./generator.jl:44 [inlined]
   [11] _collect(c::Base.ReshapedArray{ComplexF32, 1, LinearAlgebra.Adjoint{ComplexF32, CuArray{ComplexF32, 2, CUDA.Mem.DeviceBuffer}}, Tuple{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64}}}, itr::Base.Generator{Base.ReshapedArray{ComplexF32, 1, LinearAlgebra.Adjoint{ComplexF32, CuArray{ComplexF32, 2, CUDA.Mem.DeviceBuffer}}, Tuple{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64}}}, ChainRulesCore.ProjectTo{Float32, NamedTuple{(), Tuple{}}}}, #unused#::Base.EltypeUnknown, isz::Base.HasShape{1})
      @ Base ./array.jl:807
   [12] collect_similar
      @ ./array.jl:716 [inlined]
   [13] map
      @ ./abstractarray.jl:2933 [inlined]
   [14] (::ChainRulesCore.ProjectTo{AbstractArray, NamedTuple{(:element, :axes), Tuple{ChainRulesCore.ProjectTo{Float32, NamedTuple{(), Tuple{}}}, Tuple{Base.OneTo{Int64}}}}})(dx::LinearAlgebra.Adjoint{ComplexF32, CuArray{ComplexF32, 2, CUDA.Mem.DeviceBuffer}})
      @ ChainRulesCore ~/.julia/packages/ChainRulesCore/C73ay/src/projection.jl:236
   [15] ProjectTo
      @ ~/.julia/packages/ChainRulesCore/C73ay/src/projection.jl:414 [inlined]
   [16] _project
      @ ~/.julia/dev/Zygote/src/compiler/chainrules.jl:184 [inlined]
   [17] unbroadcast(x::LinearAlgebra.Adjoint{Float32, CuArray{Float32, 1, CUDA.Mem.DeviceBuffer}}, x̄::CuArray{ComplexF32, 2, CUDA.Mem.DeviceBuffer})
      @ Zygote ~/.julia/dev/Zygote/src/lib/broadcast.jl:58
   [18] (::Zygote.var"#857#858"{CuArray{ComplexF32, 1, CUDA.Mem.DeviceBuffer}, LinearAlgebra.Adjoint{Float32, CuArray{Float32, 1, CUDA.Mem.DeviceBuffer}}, CuArray{ComplexF32, 2, CUDA.Mem.DeviceBuffer}})(Δ::CuArray{ComplexF32, 2, CUDA.Mem.DeviceBuffer})
      @ Zygote ~/.julia/dev/Zygote/src/lib/broadcast.jl:97
   [19] (::Zygote.var"#3669#back#859"{Zygote.var"#857#858"{CuArray{ComplexF32, 1, CUDA.Mem.DeviceBuffer}, LinearAlgebra.Adjoint{Float32, CuArray{Float32, 1, CUDA.Mem.DeviceBuffer}}, CuArray{ComplexF32, 2, CUDA.Mem.DeviceBuffer}}})(Δ::CuArray{ComplexF32, 2, CUDA.Mem.DeviceBuffer})
      @ Zygote ~/.julia/packages/ZygoteRules/AIbCs/src/adjoint.jl:67
   [20] Pullback
      @ ./none:0 [inlined]
   [21] (::typeof(∂(#13)))(Δ::Float32)
   [22] (::Zygote.var"#60#61"{typeof(∂(#13))})(Δ::Float32)
      @ Zygote ~/.julia/dev/Zygote/src/compiler/interface.jl:45
   [23] gradient(::Function, ::CuArray{Float32, 1, CUDA.Mem.DeviceBuffer}, ::Vararg{Any})
      @ Zygote ~/.julia/dev/Zygote/src/compiler/interface.jl:97
   [24] gradcheck_gpu(::Function, ::CuArray{Float32, 1, CUDA.Mem.DeviceBuffer}, ::Vararg{Any})
      @ Main ~/.julia/dev/Zygote/test/cuda.jl:9
   [25] top-level scope
```
From the look of the stack trace, this isn't due to this pull request. In fact, if I change the function definition to 
```julia
sin(conj(c)/$(transpose(r)) - im) - imag(c + tanh(r/c')))
```
then everything is fine, so my guess is that this is some funkiness related to the pullback of an adjoint of a real vector. I'll take a look into this, but I am not sure if that's part of this pull request.

Second, I have added some additional tests to ensure we hit every one of the `_broadcast_forward` branches.

ptiede:
@mcabbott mostly good news. The ldexp type instability was fixed in https://github.com/JuliaDiff/DiffRules.jl/pull/89.
However, now I am getting a really annoying issue that I wonder if you have some wisdom for. 
I am getting a `Reason: unsupported dynamic function invocation (call to exponent)`
```julia
r3 = Float32.(inv.(2:4))
f(r) = sum(abs2, log.(1 .+ im .* r)./2)
Zygote.gradient(f, r3)
```
but **only on Julia 1.6**. For Julia 1.7 and 1.8 everything works fine. 

Digging into this issue a bit more, on 1.6 when I call `Zygote.broadcast_forward` I get the following stack-trace:
```julia
ERROR: InvalidIRError: compiling kernel #broadcast_kernel#17(CUDA.CuKernelContext, CuDeviceVector{ForwardDiff.Dual{Nothing, Float32, 1}, 1}, Base.Broadcast.Broadcasted{CUDA.CuArrayStyle{1}, Tuple{Base.OneTo{Int64}}, Zygote.var"#938#939"{typeof(f)}, Tuple{Base.Broadcast.Extruded{CuDeviceVector{Float32, 1}, Tuple{Bool}, Tuple{Int64}}}}, Int64) resulted in invalid LLVM IR
Reason: unsupported dynamic function invocation (call to _broadcast_getindex_evalf(f::Tf, args::Vararg{Any, N}) where {Tf, N} in Base.Broadcast at broadcast.jl:648)
Stacktrace:
 [1] _broadcast_getindex
   @ ./broadcast.jl:621
 [2] getindex
   @ ./broadcast.jl:575
 [3] broadcast_kernel
   @ ~/.julia/packages/GPUArrays/fqD8z/src/host/broadcast.jl:57
Reason: unsupported dynamic function invocation (call to convert)
Stacktrace:
 [1] setindex!
   @ ~/.julia/packages/CUDA/DfvRa/src/device/array.jl:194
 [2] setindex!
   @ ~/.julia/packages/CUDA/DfvRa/src/device/array.jl:207
 [3] broadcast_kernel
   @ ~/.julia/packages/GPUArrays/fqD8z/src/host/broadcast.jl:57
Hint: catch this exception as `err` and call `code_typed(err; interactive = true)` to introspect the erronous code with Cthulhu.jl
Stacktrace:
  [1] check_ir(job::GPUCompiler.CompilerJob{GPUCompiler.PTXCompilerTarget, CUDA.CUDACompilerParams, GPUCompiler.FunctionSpec{GPUArrays.var"#broadcast_kernel#17", Tuple{CUDA.CuKernelContext, CuDeviceVector{ForwardDiff.Dual{Nothing, Float32, 1}, 1}, Base.Broadcast.Broadcasted{CUDA.CuArrayStyle{1}, Tuple{Base.OneTo{Int64}}, Zygote.var"#938#939"{typeof(f)}, Tuple{Base.Broadcast.Extruded{CuDeviceVector{Float32, 1}, Tuple{Bool}, Tuple{Int64}}}}, Int64}}}, args::LLVM.Module)
    @ GPUCompiler ~/.julia/packages/GPUCompiler/Fu1YT/src/validation.jl:141
  [2] macro expansion
    @ ~/.julia/packages/GPUCompiler/Fu1YT/src/driver.jl:418 [inlined]
  [3] macro expansion
    @ ~/.julia/packages/TimerOutputs/LHjFw/src/TimerOutput.jl:253 [inlined]
  [4] macro expansion
    @ ~/.julia/packages/GPUCompiler/Fu1YT/src/driver.jl:416 [inlined]
  [5] emit_asm(job::GPUCompiler.CompilerJob, ir::LLVM.Module; strip::Bool, validate::Bool, format::LLVM.API.LLVMCodeGenFileType)
    @ GPUCompiler ~/.julia/packages/GPUCompiler/Fu1YT/src/utils.jl:68
  [6] cufunction_compile(job::GPUCompiler.CompilerJob, ctx::LLVM.Context)
    @ CUDA ~/.julia/packages/CUDA/DfvRa/src/compiler/execution.jl:354
  [7] #222
    @ ~/.julia/packages/CUDA/DfvRa/src/compiler/execution.jl:347 [inlined]
  [8] JuliaContext(f::CUDA.var"#222#223"{GPUCompiler.CompilerJob{GPUCompiler.PTXCompilerTarget, CUDA.CUDACompilerParams, GPUCompiler.FunctionSpec{GPUArrays.var"#broadcast_kernel#17", Tuple{CUDA.CuKernelContext, CuDeviceVector{ForwardDiff.Dual{Nothing, Float32, 1}, 1}, Base.Broadcast.Broadcasted{CUDA.CuArrayStyle{1}, Tuple{Base.OneTo{Int64}}, Zygote.var"#938#939"{typeof(f)}, Tuple{Base.Broadcast.Extruded{CuDeviceVector{Float32, 1}, Tuple{Bool}, Tuple{Int64}}}}, Int64}}}})
    @ GPUCompiler ~/.julia/packages/GPUCompiler/Fu1YT/src/driver.jl:76
  [9] cufunction_compile(job::GPUCompiler.CompilerJob)
    @ CUDA ~/.julia/packages/CUDA/DfvRa/src/compiler/execution.jl:346
 [10] cached_compilation(cache::Dict{UInt64, Any}, job::GPUCompiler.CompilerJob, compiler::typeof(CUDA.cufunction_compile), linker::typeof(CUDA.cufunction_link))
    @ GPUCompiler ~/.julia/packages/GPUCompiler/Fu1YT/src/cache.jl:90
 [11] cufunction(f::GPUArrays.var"#broadcast_kernel#17", tt::Type{Tuple{CUDA.CuKernelContext, CuDeviceVector{ForwardDiff.Dual{Nothing, Float32, 1}, 1}, Base.Broadcast.Broadcasted{CUDA.CuArrayStyle{1}, Tuple{Base.OneTo{Int64}}, Zygote.var"#938#939"{typeof(f)}, Tuple{Base.Broadcast.Extruded{CuDeviceVector{Float32, 1}, Tuple{Bool}, Tuple{Int64}}}}, Int64}}; name::Nothing, kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ CUDA ~/.julia/packages/CUDA/DfvRa/src/compiler/execution.jl:299
 [12] cufunction
    @ ~/.julia/packages/CUDA/DfvRa/src/compiler/execution.jl:293 [inlined]
 [13] macro expansion
    @ ~/.julia/packages/CUDA/DfvRa/src/compiler/execution.jl:102 [inlined]
 [14] #launch_heuristic#246
    @ ~/.julia/packages/CUDA/DfvRa/src/gpuarrays.jl:17 [inlined]
 [15] _copyto!
    @ ~/.julia/packages/GPUArrays/fqD8z/src/host/broadcast.jl:63 [inlined]
 [16] copyto!
    @ ~/.julia/packages/GPUArrays/fqD8z/src/host/broadcast.jl:46 [inlined]
 [17] copy
    @ ~/.julia/packages/GPUArrays/fqD8z/src/host/broadcast.jl:37 [inlined]
 [18] materialize(bc::Base.Broadcast.Broadcasted{CUDA.CuArrayStyle{1}, Nothing, Zygote.var"#938#939"{typeof(f)}, Tuple{CuArray{Float32, 1, CUDA.Mem.DeviceBuffer}}})
    @ Base.Broadcast ./broadcast.jl:883
 [19] broadcast_forward(f::Function, args::CuArray{Float32, 1, CUDA.Mem.DeviceBuffer})
    @ Zygote ~/.julia/dev/Zygote/src/lib/broadcast.jl:271
 [20] top-level scope
    @ REPL[20]:1
 [21] top-level scope
    @ ~/.julia/packages/CUDA/DfvRa/src/initialization.jl:52
```

It looks like I can reduce this further by doing the following
```julia
julia> rd3 = first.(Zygote.dualize(r3)) # CuArray{ForwardDiff.Dual{Nothing, Float32, 1}, 1, CUDA.Mem.DeviceBuffer}
julia> log.(1im .* rd3) 
ERROR: InvalidIRError: compiling kernel #broadcast_kernel#17(CUDA.CuKernelContext, CuDeviceVector{Complex{ForwardDiff.Dual{Nothing, Float32, 1}}, 1}, Base.Broadcast.Broadcasted{CUDA.CuArrayStyle{1}, Tuple{Base.OneTo{Int64}}, typeof(log), Tuple{Base.Broadcast.Broadcasted{CUDA.CuArrayStyle{1}, Nothing, typeof(*), Tuple{Complex{Int64}, Base.Broadcast.Extruded{CuDeviceVector{ForwardDiff.Dual{Nothing, Float32, 1}, 1}, Tuple{Bool}, Tuple{Int64}}}}}}, Int64) resulted in invalid LLVM IR
Reason: unsupported dynamic function invocation (call to exponent)
Stacktrace:
 [1] ssqs
   @ ./complex.jl:474
 [2] log
   @ ./complex.jl:594
 [3] _broadcast_getindex_evalf
   @ ./broadcast.jl:648
 [4] _broadcast_getindex
   @ ./broadcast.jl:621
 [5] getindex
   @ ./broadcast.jl:575
 [6] broadcast_kernel
   @ ~/.julia/packages/GPUArrays/fqD8z/src/host/broadcast.jl:57
...
```
which suggests the problem is in `Base.ssqs`. This looks like an issue outside the scope of this pull-request so I am not too sure what we want to do here.

mcabbott:
Does more inlining help at all, e.g. `@inline function dual_function`?

ptiede:
Sadly no :( The MWE also shouldn't have an inlining issue right?

ptiede:
> Does more inlining help at all, e.g. `@inline function dual_function`?

Ok figured it out! So analyzing `Base.ssqs` with Cthulhu it looks like what was happening was sometimes the code would venture into the following call:
```julia
 • %282  = call #exponent(::ForwardDiff.Dual{Nothing, Float32, 1})::Union{}
```

which errors because `exponent(::Dual)` is never defined. This is what was causing the dynamic call. If I define the following method
```julia
Base.exponent(x::ForwardDiff.Dual{<:Real}) = Base.exponent(ForwardDiff.value(x))
```
everything works and we pass the tests on Julia 1.6. I believe this makes sense since exponent : Real -> Int so we only really care about the value of the function. The only thing I don't really understand is why this didn't cause an issue on 1.7/1.8.

ptiede:
Alright the dual exponent issue has been fixed. When a new version of ForwardDiff is released when a new version of ForwardDiff.jl get released the 1.6 tests should pass.

devmotion:
DynamicPPL test failures are caused by https://github.com/JuliaDiff/ForwardDiff.jl/issues/606.

ptiede:
@mcabbott I think this is finally ready to review again. All the tests are passing, and I have added some additional tests to ensure that every branch is getting hit.

CarloLucibello (approved):

ptiede:
Is this ready to merge?

ptiede:
Just a bump to see if this is ready to be merged or it there are some outstanding items that I still need to fix.

CarloLucibello:
Thanks! I'll tag a new release shortly

CarloLucibello:
@ptiede which of the issues mentioned in the OP should be closed?

ptiede:
This should fix 961, 1121, 1215, 1276, i.e. all of them since they were all the same problem in disguise.

CarloLucibello:
Do you think we need some extra tests or the ones in this PR are enough?

ptiede:
I think the tests should cover all of those cases

https://github.com/FluxML/Zygote.jl/blob/616bf6cfb7d7cea2b755a31f36274e9fff18f678/test/cuda.jl#L191-L192

should cover 1276 intrinsically because the type instability that was causing the slowdown is fixed. 

https://github.com/FluxML/Zygote.jl/blob/616bf6cfb7d7cea2b755a31f36274e9fff18f678/test/cuda.jl#L175

should cover the abs2 bug. But coming up with tests was tricky so it is possible that I missed something.

PR #1281, opened 2022-08-04
Add precompilation via SnoopPrecompile

ToucheSir:
Inspired by https://discourse.julialang.org/t/ann-new-package-snoopprecompile/84778, I threw together a quick POC using the existing (but unused) `precompile` function. Results may be found in [this post](https://discourse.julialang.org/t/ann-new-package-snoopprecompile/84778/21).

There's probably more to be done to claw back the lost load time, but I'm not well-versed in this area so any advice is appreciated.

PR #1332, opened 2022-11-28
revert SnoopCompile precompilation

CarloLucibello:
This reverts #1281 since it seems to be causing https://github.com/SciML/DiffEqFlux.jl/issues/783

We are not precompiling anything too odd, so likely the problem is in DiffEqFlux or somewhere else, but until we have a better understanding I think it is better to revert. 

cc @ToucheSir @ChrisRackauckas


axsk:
For me `using Lux` also does not terminate precompilation. Thought it might be related

CarloLucibello:
ok I'm going to just merge this

PR #1335, opened 2022-12-01
pullbacks for dict constructors

### LINKED ISSUE #1293: Gradient fails for Dict constructed with a generator or a vector of pairs
When taking the gradient of a function that uses a dict constructor using a generator or a list, it fails because there is a try/catch block somewhere.

e.g. for a generator
```julia
julia> using Zygote

julia> function f(x)
              a = Dict(c => x for c in 1:3)
              return a[1]
          end

julia> Zygote.gradient(f, 2.0)
ERROR: Compiling Tuple{Type{Dict}, Base.Generator{UnitRange{Int64}, var"#3#4"{Float64}}}: try/catch is not supported.
```

It works however if I slurp the vector
```julia
julia> function f(x)
              a = Dict([c => x for c in 1:3]...)
              return a[1]
          end

julia> Zygote.gradient(f, 2.0)
(1.0,)
```

CarloLucibello:
Trying to fix #1293
With respect to master the case 
```julia
gradient(2) do c
  d = Dict([i => i*c for i=1:3])
  return d[1] 
end
```
is now fixed, but I haven't cracked the  generator case  yet
```julia
gradient(2) do c
  d = Dict(i => i*c for i=1:3)
  return d[1] 
end
```
I've tried by composing pullbacks but so far it's been unsuccessful.

### DIFF
--- a/src/lib/base.jl
+++ b/src/lib/base.jl
@@ -231,3 +231,19 @@ end
     fallback_Fix2(y) = f(y, x)
     return _pullback(__context__, fallback_Fix2, y)
 end
+
+function ChainRulesCore.rrule(::typeof(Dict), xs::Pair...)
+  function Dict_pullback(Δ)
+    return (NoTangent(), ((first=ZeroTangent(), second=get(Δ, x[1], ZeroTangent())) for x in xs)...)
+  end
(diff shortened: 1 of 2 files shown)


CarloLucibello:
closing as outdated

Issue #796, opened 2020-09-22
Wrong gradient through hypergeometric function pFq


bertschi:
Zygote gives wrong gradients for

```{julia}
using HypergeometricFunctions
using Zygote
import ForwardDiff

f(z) = pFq([0.1], [0.2], z)
## Parameters don't matter ... gradient also wrong for other values

gradient(f, 100.2)
## Gives (46.42500000000001,)
ForwardDiff.gradient(f ∘ first, [100.2])
## Gives [9.993752431780506e42] -- Note: ForwardDiff.gradient only defined for vector functions
```
A quick numeric check suggests that ForwardDiff is correct. Unfortunately the example is certainly far from minimal. Would be happy for suggestions on how to track down this issue.

mcabbott:
```
julia> gradient(f, 0.75)[1]
0.84375

julia> ForwardDiff.derivative(f, 0.75)
1.0029644434748202

julia> @which  pFq([0.1], [0.2], 100.2)
pFq(α::AbstractArray{T,1} where T, β::AbstractArray{T,1} where T, z; kwds...) in HypergeometricFunctions at /Users/me/.julia/packages/HypergeometricFunctions/vfGlc/src/generalized.jl:8
```

This leads to this little function, seems too simple for much to go wrong? Maybe it would be worth replacing the for-loops with simple statements (and possibly changing the names, rather than updating the same variable) to see if anything changes?

https://github.com/JuliaMath/HypergeometricFunctions.jl/blob/5866fe52e6d2cf61b311936becb6def057ddf22a/src/specialfunctions.jl#L456

bertschi:
Thanks, I finally boiled it down to a simpler example. It appears that variables within the while loop get conflated in Zygote's backward pass ... the for loops appear not to be a problem.

```{julia}
function ggFoo(z::Float64)
    x = 1.0
    y = 1.0 + z
    j = 1
    while abs(x - y) > 1e-6
        ## x, y = y, (x + y) / 2
        ## Same issue with explicit temporary
        tmp = (x + y) / 2
        x = y
        y = tmp
        j += 1
        println([j, x, y])
    end
    return y
end
```

```{julia}
julia> ggFoo(2.0)
[1.0, 1.0, 3.0]
[2.0, 3.0, 2.0]
[3.0, 2.0, 2.5]
...
[22.0, 2.333333969116211, 2.3333330154418945]
2.3333330154418945

julia> gradient(ggFoo, 2.0)
[...]

PR #1337, opened 2022-12-05
Test/while

bertschi:
New test for issue #796.
Issue has been fixed in master and test is currently passing.

### DIFF
--- a/test/gradcheck.jl
+++ b/test/gradcheck.jl
@@ -2050,18 +2050,14 @@ end
     function foo(z::Float64)
         x = 1.0
         y = 1.0 + z
-        j = 1
         while abs(x - y) > 1e-6
-            tmp = (x + y) / 2
-            x = y
-            y = tmp
-            j += 1
+            y, x = (x + y) / 2, y
         end
         return y
     end
 
-    @test gradcheck(foo, 0.0)
-    @test gradcheck(foo, 2.0)
-    @test gradcheck(foo, -2.0)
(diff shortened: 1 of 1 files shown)


test/gradcheck.jl
```diff
@@ -2045,3 +2045,19 @@ end
   h(x) = sum(abs2, g(x))
   @test gradient(h, x)[1] isa typeof(x)
 end
+
+@testset "Zygote issue 796" begin
```
ToucheSir, line 2049:
```suggestion
@testset "Zygote #796" begin
```

ToucheSir (approved):
LGTM, one small style change for your consideration.

test/gradcheck.jl
```diff
@@ -2045,3 +2045,19 @@ end
...
+        while abs(x - y) > 1e-6
+            y, x = (x + y) / 2, y
+        end
+        return y
+    end
+
+    @test gradcheck(foo ∘ first, [0.0])
+    @test gradcheck(foo ∘ first, [2.0])
+    @test gradcheck(foo ∘ first, [-1e-5])
+    @test gradient(foo, 1024.0)[1] ≈ 2//3
```
MariusDrulea, line 2062: I would also add several tests such that Zygote matches ForwardDiff and ReverseDiff results; of course, this assumes both ADs give correct results.

bertschi: Did not include those as all other test sets in this file use finite differences.

PR #1340, opened 2022-12-18
Free CuArrays in the reverse pass

mcabbott:
This adds:
* A flag to `Context` to indicate that the pullback will never be called twice -- set to true for `gradient`, false for `jacobian`
* Modifications to many rules, esp. for broadcasting, so that `y=f(x)` in the forward pass has `finalize(y)` in the reverse. This increases the largest size of Flux model which can run on a given GPU.

Applying such modifications everywhere led to many errors, some from rules like `y = x .+ false` which return `y === x` under Zygote. So they now require a separate macro `@adjoint_final`. 

At present this modification is applied to all CR `rrule`s. This is probably unsafe and we should revert  2524163c8a5bd4aab9101c964d6d8d0676b501e8 . Unclear how best to opt-in within ChainRules. Xref https://github.com/JuliaDiff/ChainRulesCore.jl/pull/592 about the idea of a flag, but not entirely sure that's the right approach. 

Explicit finalising won't work well with thunks. Which doesn't matter at all yet, but might after #966.

It also does not work with second derivatives, hence is disabled. Other uses of the context flag (like testing `only_once(cfg)` & then over-writing some array) probably also need to be disabled.

Needs https://github.com/FluxML/ZygoteRules.jl/pull/23 so CI will fail. Locally, one failure, one failure to fail:
```
Global Params: Error During Test at /Users/me/.julia/dev/Zygote/test/features.jl:399
  KeyError: key :(Main.global_param) not found
  Stacktrace:
    [1] getindex(d::IdDict{Any, Any}, key::Any)
      @ Base ./iddict.jl:108
    [2] macro expansion
      @ ~/.julia/dev/Zygote/test/features.jl:404 [inlined]

Compiler: Error During Test at /Users/me/.julia/dev/Zygote/test/compiler.jl:35
 Unexpected Pass
 Expression: trace_contains(bt, :badly, "compiler.jl", 24)
 Got correct result, please change to @test if no longer broken.
```

### DIFF
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -196,17 +196,26 @@ _project(x::AbstractArray, dx::Tuple) = _project(x, reshape(collect(dx), axes(x)
 (project::ProjectTo{AbstractArray})(dx::Tangent) = dx
 
 """
-  ZBack{F}(back) <: Function
+  ZBack{Y,F}(y, back) <: Function
 
 Wrapper for a ChainRules pullback `back`, that causes it to follow Zygote conventions.
 (A functor here is used rather than a closure to avoid boxing issues);
+Now captures the forward result to call `finalize(y)` when done, if `only_once` says this is safe.
 """
-struct ZBack{F} <: Function
+struct ZBack{Y,F} <: Function
+  fwd::Y
   back::F
 end
-@inline (s::ZBack)(dy) = wrap_chainrules_output(s.back(wrap_chainrules_input(dy)))
+@inline (s::ZBack{Nothing})(dy) = wrap_chainrules_output(s.back(wrap_chainrules_input(dy)))
+@inline function (s::ZBack)(dy)
+  ∇s = wrap_chainrules_output(s.back(wrap_chainrules_input(dy)))
+  maybe_final(y)
+  ∇s
+end
+
 # `nothing->nothing` can be deleted after https://github.com/FluxML/Zygote.jl/issues/603
 # though it might be worth keeping as a performance optimization (benchmarking pending)
+@inline (s::ZBack{Nothing})(::Nothing) = nothing
 @inline (s::ZBack)(::Nothing) = nothing
 
 """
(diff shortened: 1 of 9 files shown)


chengchingwen:
Could we combine this with https://github.com/JuliaDiff/ChainRulesCore.jl/pull/592 ?

mcabbott:
It's possible. I think that means having two distinct structs, ZygoteRuleConfig and ZygoteOnceRuleConfig or something. 

At present, BTW, most of these `maybe_final`s seem not to be called & I'm not sure why.

src/compiler/chainrules.jl
```diff
@@ -195,17 +196,26 @@ _project(x::AbstractArray, dx::Tuple) = _project(x, reshape(collect(dx), axes(x)
...
-struct ZBack{F} <: Function
+struct ZBack{Y,F} <: Function
+  fwd::Y
   back::F
 end
-@inline (s::ZBack)(dy) = wrap_chainrules_output(s.back(wrap_chainrules_input(dy)))
+@inline (s::ZBack{Nothing})(dy) = wrap_chainrules_output(s.back(wrap_chainrules_input(dy)))
+@inline function (s::ZBack)(dy)
+  ∇s = wrap_chainrules_output(s.back(wrap_chainrules_input(dy)))
+  maybe_final(y)
```
CarloLucibello, line 212: wrong argument on the line?

chengchingwen:
> It's possible. I think that means having two distinct structs, ZygoteRuleConfig and ZygoteOnceRuleConfig or something.

Or introduce another type parameter like `ZygoteRuleConfig{once} where once`?

> At present, BTW, most of these maybe_finals seem not to be called & I'm not sure why.

Do you mean the `finalize` is not called, or it is called but the memory is not freed?

src/compiler/chainrules.jl
mcabbott: Oh right, should be
```suggestion
  maybe_final(s. fwd)
```

chengchingwen:
> But I don't think that fits CR's mechanism; the current struct is <: RuleConfig{Union{HasReverseMode,NoForwardsMode}} and the new ones would need different supertypes.

Couldn't it be done like `struct ZygoteRuleConfig{P<:PullbackCapability} <: RuleConfig{Union{HasReverseMode,NoForwardsMode,P}}`?

mcabbott:
Oh right, that ought to work.

Current status is that some arrays are freed too early (e.g. with Metalhead's ResNet, at addact(relu)) but it's hard to isolate. Still happens if I disable all thunks. In Zygote's tests, some failures due to too-early `fill!(x, NaN)` (included here as a test), perhaps related.

Issue #1349, opened 2023-01-07
Zygote.Buffer can fail for higher order AD


marius311:
For higher order AD where you can have Arrays{<:Float} on the foward pass but Arrays{<:Dual} on the backwards pass, Zygote.Buffer gives an error because its not allocated wide enough on the forward pass:

```julia
function buffer_identity(x)
    # toy function which uses a Buffer
    buf = Zygote.Buffer(x, length(x))
    buf[:] = x
    copy(buf)
end

x = rand(2)

Zygote.hessian(A -> sum(A .* buffer_identity(x)), 1.)
```

```julia
ERROR: MethodError: no method matching Float64(::ForwardDiff.Dual{ForwardDiff.Tag{Zygote.var"#104#105"{var"#9#10"}, Float64}, Float64, 1})
Stacktrace:
  [1] convert(#unused#::Type{Float64}, x::ForwardDiff.Dual{ForwardDiff.Tag{Zygote.var"#104#105"{var"#9#10"}, Float64}, Float64, 1})
    @ Base ./number.jl:7
  [2] setindex!(A::Vector{Float64}, x::ForwardDiff.Dual{ForwardDiff.Tag{Zygote.var"#104#105"{var"#9#10"}, Float64}, Float64, 1}, i1::Int64)
    @ Base ./array.jl:966
  [3] setindex!
    @ ./subarray.jl:347 [inlined]
  [4] copyto_unaliased!
    @ ./abstractarray.jl:1038 [inlined]
  [5] copyto!(dest::SubArray{Float64, 1, Vector{Float64}, Tuple{Base.Slice{Base.OneTo{Int64}}}, true}, src::Vector{ForwardDiff.Dual{ForwardDiff.Tag{Zygote.var"#104#105"{var"#9#10"}, Float64}, Float64, 1}})
    @ Base ./abstractarray.jl:1018
  [6] copyto!
    @ ./broadcast.jl:954 [inlined]
  [7] copyto!
    @ ./broadcast.jl:913 [inlined]
  [8] materialize!
    @ ./broadcast.jl:871 [inlined]
[...]

PR #1350, opened 2023-01-07
fix forward-over-reverse with Buffer in some cases

marius311:
Fixes https://github.com/FluxML/Zygote.jl/issues/1349

What's happening is that the Buffer is Array{<:Float} on the forward pass but the accumulator needs to be able to hold Duals on the backwards pass. This makes it so the first time we accumulate onto the Buffer, it will do a promotion if necessary and allocate an appropriately wide accumulator. 

I think there's probably functions you can concoct where the first accumulation is Float and the second is Dual, so this strategy is not foolproof. But it probably still covers alot of cases, and its better than not working at all. For a future PR, one could maybe imagine a continually widening strategy where you reallocate a wider buffer and copy in already accumulated values. 

I also switched `grad_mut` to use `get!` for consistency with the Buffer-specific versions and since its faster.

### DIFF
--- a/src/lib/buffer.jl
+++ b/src/lib/buffer.jl
@@ -1,4 +1,4 @@
-grad_mut(cx::Context, b::Buffer{T}, ::Type{S}=Union{}) where {T, S} =
+grad_mut(cx::Context, b::Buffer, ::Type=Union{}) =
   _get!(() -> fill!(similar(b.data, Any), nothing), cache(cx), b)
 grad_mut(cx::Context, b::Buffer{T}, ::Type{S}=Union{}) where {T<:Number, S} =
   _get!(() -> fill!(similar(b.data, float(promote_type(T, S))), 0), cache(cx), b)
--- a/src/lib/buffer.jl
+++ b/src/lib/buffer.jl
@@ -1,11 +1,13 @@
-grad_mut(b::Buffer) = fill!(similar(b.data, Any), nothing)
-grad_mut(b::Buffer{T}) where T<:Number = fill!(similar(b.data, float(T)), 0)
+grad_mut(cx::Context, b::Buffer{T}, ::Type{S}=Union{}) where {T, S} =
+  get!(() -> fill!(similar(b.data, Any), nothing), cache(cx), b)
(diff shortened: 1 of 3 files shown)


marius311:
Turns out `Base.get!` isnt differentiable so it was breaking reverse over reverse and hence a Flux test. I put that part back the way it was.

src/lib/buffer.jl
```diff
@@ -1,11 +1,13 @@
-grad_mut(b::Buffer) = fill!(similar(b.data, Any), nothing)
-grad_mut(b::Buffer{T}) where T<:Number = fill!(similar(b.data, float(T)), 0)
+grad_mut(cx::Context, b::Buffer{T}, ::Type{S}=Union{}) where {T, S} =
+  _get!(() -> fill!(similar(b.data, Any), nothing), cache(cx), b)
```
ToucheSir, line 2:
```suggestion
grad_mut(cx::Context, b::Buffer, ::Type) =
  _get!(() -> fill!(similar(b.data, Any), nothing), cache(cx), b)
```
Unless the forced specializations are helping somewhere.

src/lib/buffer.jl
```diff
@@ -1,11 +1,13 @@
...
+grad_mut(cx::Context, b::Buffer{T}, ::Type{S}=Union{}) where {T<:Number, S} =
+  _get!(() -> fill!(similar(b.data, float(promote_type(T,S))), 0), cache(cx), b)
```
ToucheSir, line 4:
```suggestion
grad_mut(cx::Context, b::Buffer{T}, ::Type{S}=Union{}) where {T<:Number, S} =
  _get!(() -> fill!(similar(b.data, float(promote_type(T, S))), 0), cache(cx), b)
```

ToucheSir (commented):
Basically LGTM, thanks!

src/lib/buffer.jl
marius311: They're not, was just code symmetry, will do this (although it still needs `=Union{}` arg default)

ToucheSir (approved):

PR #1353, opened 2023-01-10
Fix issue #1352 (Buffer regression)

devmotion:
I checked locally and it seems this change fixes #1352. I'll add the MWE as a test, provided that existing tests still pass.

Issue #1352, opened 2023-01-10
`Buffer` Regression in >v0.6.50 


frankschae:
We spotted that some of the SciMLSensitivity.jl tests that make use of `Zygote.Buffer` are failing on the recent Zygote release v0.6.52 (see https://github.com/SciML/SciMLSensitivity.jl/pull/766).  The tests are passing on earlier versions (https://github.com/SciML/SciMLSensitivity.jl/pull/765).

Here is an MWE
```julia
using Zygote

u₀ = [0.75, 0.5]
p = [-1.5, 0.05, 0.2, 0.01]

function g_nondiag!(du, u, p, t)
    du[1, 1] = p[3] * u[1] + p[4] * u[2]
    du[1, 2] = p[3] * u[1] + p[4] * u[2]
    du[2, 1] = p[4] * u[1] + p[3] * u[2]
    du[2, 2] = p[4] * u[1] + p[3] * u[2]
    nothing
end

v = randn(2)
A = zeros(2, 2)
_dy, back = Zygote.pullback(u₀, p) do u, p
    out_ = Zygote.Buffer(similar(A))
    g_nondiag!(out_, u, p, 1.0)
    copy(out_[:, 1])
end
tmp1, tmp2 = back(v)
```
On v0.6.52, a `MethodError` is thrown
```
ERROR: MethodError: no method matching float(::Type{Any})
Closest candidates are:
  float(::Type{Union{Missing, T}}) where T at missing.jl:112
  float(::Any) at float.jl:269
  float(::Complex{<:AbstractFloat}) at complex.jl:1086
  ...
Stacktrace:
 [1] float(#unused#::Type{Any})
   @ Base ./missing.jl:113
 [2] (::Zygote.var"#808#809"{Vector{Float64}, Float64, Zygote.Buffer{Float64, Matrix{Float64}}})()
   @ Zygote ~/.julia/packages/Zygote/arfUC/src/lib/buffer.jl:4
 [3] _get!
   @ ~/.julia/packages/Zygote/arfUC/src/lib/lib.jl:274 [inlined]
 [4] grad_mut(cx::Zygote.Context{false}, b::Zygote.Buffer{Float64, Matrix{Float64}}, ::Type{Vector{Float64}})
   @ Zygote ~/.julia/packages/Zygote/arfUC/src/lib/buffer.jl:3
[...]

PR #1355, opened 2023-01-11
full fix to #1352

marius311:
I did screw up #1350 slightly sorry, thanks for catching that in #1352 @frankschae, but @devmotion's #1353 is only a partial fix because the allocator ends up eltype `Any`. Here's a full fix. I also reduced the test (this test also would have caught the error in my initial PR).

### DIFF
--- a/test/gradcheck.jl
+++ b/test/gradcheck.jl
@@ -1569,46 +1569,13 @@ using Zygote: Buffer
   @test ∇W2 == W2
   @test ∇x == 6 .* x
 
-  @testset "incorrect promotion (#1352)" begin
-    u = [0.75, 0.5]
-    p = [-1.5, 0.05, 0.2, 0.01]
+  # reduced mwe of #1352
+  @test Zygote.gradient([0,0]) do x
+      buf = Zygote.Buffer(similar(x))
+      buf[:] = x
+      sum(copy(buf[1:2]))
+  end == ([1,1],)
 
-    # in-place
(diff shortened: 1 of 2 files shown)


test/gradcheck.jl
```diff
@@ -1553,46 +1553,13 @@ using Zygote: Buffer
...
-      du22 = p[4] * u[1] + p[3] * u[2]
-      return [du11 du12
-              du21 du22]
-    end
-    du1, back = Zygote.pullback(u, p) do u, p
-      du = g1352(u, p, 1.0)
-      return du[:, 1]
-    end
+  # reduced mwe of #1352
+  @test Zygote.gradient([0,0]) do x
```
devmotion, line 1557: Maybe use some more complicated function and some non-zero values? It happened to me before (eg in DistributionsAD) that too "nice" values and examples did not catch (all) problems.

marius311: Sure, any suggestions? The test is mainly of the pullback of `getindex(x::Buffer, ::UnitRange)` which was not getting tested before (only `getindex(x::Buffer, ::Int)` was).

ToucheSir: Gentle ping on this. If there are no good ideas then I think this is ready?

marius311: Fine by me to go ahead as-is or for someone to tweak the test. Just rebased on master for convenience.

ToucheSir (approved):

Issue #1342, opened 2022-12-22
Not able to @opt_out rules with RuleConfig


tansongchen:
Let's say I have a type `WeirdNumber <: Number` is so weird that I don't want its derivative of power function (`^`) be calculated by `rrule` of `^` and `literal_pow`, and should instead go through and differentiate its definition. Since Zygote defines AD-specific rules for `literal_pow` with `RuleConfig` in its code base, I had to also `@opt_out` this rule. However, the following MWE didn't work:

```julia
using ChainRulesCore: @opt_out, RuleConfig
import Base: ^

struct WeirdNumber <: Number
    a::Float64
end

^(x::WeirdNumber, p::Int) = WeirdNumber(x.a ^ p)

@opt_out rrule(::typeof(Base.literal_pow), ::typeof(^), x::WeirdNumber, ::Val{p}) where {p}
@opt_out rrule(::RuleConfig, ::typeof(Base.literal_pow), ::typeof(^), x::WeirdNumber,
               ::Val{p}) where {p}

using Zygote

fun(x::WeirdNumber) = (x^4).a

gradient(fun, WeirdNumber(2.))
```

Error:

```
ERROR: type Nothing has no field method
Stacktrace:
  [1] getproperty(x::Nothing, f::Symbol)
    @ Base ./Base.jl:38
  [2] has_chain_rrule(T::Type)
    @ Zygote ~/.julia/packages/Zygote/SmJK6/src/compiler/chainrules.jl:21
  [3] #s2948#1074
    @ ~/.julia/packages/Zygote/SmJK6/src/compiler/interface2.jl:20 [inlined]
  [4] var"#s2948#1074"(::Any, ctx::Any, f::Any, args::Any)
    @ Zygote ./none:0
  [5] (::Core.GeneratedFunctionStub)(::Any, ::Vararg{Any})
    @ Core ./boot.jl:582
  [6] _pullback
    @ ~/Applications/project/TaylorDiff.jl/.vscode/opt_lit_pow.jl:16 [inlined]
  [7] _pullback(ctx::Zygote.Context{false}, f::typeof(fun), args::WeirdNumber)
    @ Zygote ~/.julia/packages/Zygote/SmJK6/src/compiler/interface2.jl:0
  [8] pullback(f::Function, cx::Zygote.Context{false}, args::WeirdNumber)
    @ Zygote ~/.julia/packages/Zygote/SmJK6/src/compiler/interface.jl:44
  [9] pullback
[...]

PR #1358, opened 2023-01-13
MethodError if configured rrule is ambiguous

oxinabox:
This resolves https://github.com/FluxML/Zygote.jl/issues/1342
which is a special case of ambiguity where the ambiguity is for a configured rule and it's configured opt-out.
But it applies without an opt-out also.

This in particular implements [Sol0](https://github.com/FluxML/Zygote.jl/issues/1342#issuecomment-1363438666)
Which is to give an useful `MethodError`  rather than a confusing error about no field, as follows:

Before
```
julia> pullback(h_ambig, 1, 2)
ERROR: type Nothing has no field method
Stacktrace:
 [1] getproperty(x::Nothing, f::Symbol)
   @ Base ./Base.jl:33
 [2] has_chain_rrule(T::Type)
   @ Zygote ~/.julia/packages/Zygote/AS0Go/src/compiler/chainrules.jl:21
 [3] #s2941#1099
   @ ~/.julia/packages/Zygote/AS0Go/src/compiler/interface2.jl:20 [inlined]
 [4] var"#s2941#1099"(::Any, ctx::Any, f::Any, args::Any)
   @ Zygote ./none:0
 [5] (::Core.GeneratedFunctionStub)(::Any, ::Vararg{Any, N} where N)
   @ Core ./boot.jl:571
 [6] pullback(::Function, ::Zygote.Context{false}, ::Int64, ::Vararg{Int64, N} where N)
   @ Zygote ~/.julia/packages/Zygote/AS0Go/src/compiler/interface.jl:44
 [7] pullback(::Function, ::Int64, ::Int64)
   @ Zygote ~/.julia/packages/Zygote/AS0Go/src/compiler/interface.jl:42
 [8] top-level scope
   @ REPL[24]:1
```

Now:
```
julia> pullback(h_ambig, 1, 2)
ERROR: MethodError: rrule(::ZygoteRuleConfig{Zygote.Context{false}}, ::typeof(h_ambig), ::Int64, ::Int64) is ambiguous. Candidates:
  rrule(::RuleConfig, ::typeof(h_ambig), x::Int64, y::Int64) in Main at /home/oxinabox/JuliaEnvs/Zygote.jl/test/chainrules.jl:291
  rrule(::ZygoteRuleConfig, ::typeof(h_ambig), x, y) in Main at /home/oxinabox/JuliaEnvs/Zygote.jl/test/chainrules.jl:290
Possible fix, define
  rrule(::ZygoteRuleConfig, ::typeof(h_ambig), ::Int64, ::Int64)
Stacktrace:
 [1] chain_rrule
   @ ~/JuliaEnvs/Zygote.jl/src/compiler/chainrules.jl:223 [inlined]
 [2] macro expansion
   @ ~/JuliaEnvs/Zygote.jl/src/compiler/interface2.jl:0 [inlined]
 [3] _pullback(::Zygote.Context{false}, ::typeof(h_ambig), ::Int64, ::Int64)
   @ Zygote ~/JuliaEnvs/Zygote.jl/src/compiler/interface2.jl:9
 [4] pullback(::Function, ::Zygote.Context{false}, ::Int64, ::Vararg{Int64})
   @ Zygote ~/JuliaEnvs/Zygote.jl/src/compiler/interface.jl:44
 [5] pullback(::Function, ::Int64, ::Int64)
   @ Zygote ~/JuliaEnvs/Zygote.jl/src/compiler/interface.jl:42
 [6] top-level scope
   @ ~/JuliaEnvs/Zygote.jl/test/chainrules.jl:293
   ```

This behavior is consistent with our behavour for non-configured rules as implemented in https://github.com/FluxML/Zygote.jl/issues/1342

[Sol1](https://github.com/FluxML/Zygote.jl/issues/1342#issuecomment-1363438666) (to just ignore rules) would be easily implemented for both configured and unconfigured cases  by changing the final `if` to `is_ambig || match(...)` rather than `!is_ambig && match(...)
but should be a separate PR.

### PR Checklist

- [x] Tests are added

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.6.54"
+version = "0.6.55"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -18,7 +18,10 @@ such that if a suitable rule is defined later, the generated function will recom
 function has_chain_rrule(T)
   config_T, arg_Ts = Iterators.peel(T.parameters)
   configured_rrule_m = meta(Tuple{typeof(rrule), config_T, arg_Ts...})
-  if _is_rrule_redispatcher(configured_rrule_m.method)
+  is_ambig = configured_rrule_m === nothing  # this means there was an ambiguity error, on configured_rrule
+
+
+  if !is_ambig && _is_rrule_redispatcher(configured_rrule_m.method)
     # The config is not being used:
     # it is being redispatched without config, so we need the method it redispatches to
     rrule_m = meta(Tuple{typeof(rrule), arg_Ts...})
@@ -32,6 +35,8 @@ function has_chain_rrule(T)
     # it will be identical, and if it doesn't we don't care what it is.
     no_rrule_m = meta(Tuple{typeof(ChainRulesCore.no_rrule), config_T, arg_Ts...})
   end
+
+  is_ambig |= rrule_m === nothing  # this means there was an ambiguity error on unconfigured rrule
 
   # To understand why we only need to check if the sigs match between no_rrule_m and rrule_m
   # in order to decide if to use, one must consider the following facts:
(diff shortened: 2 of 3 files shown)


ToucheSir (approved):
LGTM, merge when ready :)

Issue #455, opened 2020-01-13
Need an adjoint for constructor Tridiagonal{Float32,Array{Float32,1}}. Gradient is of type Array{Float64,2}


homocomputeris:
When I use the following custom layer in Flux
```julia
struct MyLayer
    dl
    d
    du
end

function (a::MyLayer)(x::AbstractArray)
    W = Tridiagonal(a.dl, a.d, a.du)
    return W*x
end
``` 
I get the error:
```
Need an adjoint for constructor Tridiagonal{Float32,Array{Float32,1}}. Gradient is of type Array{Float64,2}
```
How should this adjoint be implemented, if `methods(Tridiagonal)` doesn't even have any method for `Tridiagonal{T, Array{T,1}}`?

If I use something similar to `@adjoint Symmetric` 
```julia
@adjoint Tridiagonal(dl,d,du) = Tridiagonal(dl,d,du), Δ->(Tridiagonal(Δ),)
Zygote.refresh()
```
it raises
```
BoundsError: attempt to access (nothing, [-0.09057819660705582 -0.14655860074988672 0.0 0.0 0.0 0.0 0.0 0.0 0.0; -0.4234162265727124 -0.6851018459828749 -0.685101845982875 0.0 0.0 0.0 0.0 0.0 0.0; 0.0 -0.4533717767992162 -0.45337177679921625 -0.2801991676018467 0.0 0.0 0.0 0.0 0.0; 0.0 0.0 0.35530036306762675 0.2195877005909712 4.575095642703382e-17 0.0 0.0 0.0 0.0; 0.0 0.0 0.0 0.12382269235811338 2.5798378449739945e-17 -0.12382269235811334 0.0 0.0 0.0; 0.0 0.0 0.0 0.0 1.2258624776753338e-18 -0.005883687331056945 -0.009520006080827294 0.0 0.0; 0.0 0.0 0.0 0.0 0.0 0.03193392014112915 0.0516701681823718 0.05167016818237181 0.0; 0.0 0.0 0.0 0.0 0.0 0.0 0.05067591651533174 0.050675916515331744 0.031319438817527155; 0.0 0.0 0.0 0.0 0.0 0.0 0.0 -0.1908189747028651 -0.11793261206477705])
```

MikeInnes:
You're on the right track. If you pass through the gradient unmodified (i.e. don't convert it back to tridiagonal), it'll probably just work.

homocomputeris:
```julia
[...]

PR #1362, opened 2023-01-20
Add missing overloads for Tridiagonal

ChrisRackauckas:
Upstreaming from DiffEqFlux (https://github.com/SciML/DiffEqFlux.jl/pull/800)


ToucheSir:
Are there tests to be upstreamed as well? Seems like this could fix https://github.com/FluxML/Zygote.jl/issues/455 too.

ChrisRackauckas:
The tests are really just the downstream DiffEqFlux tests. There's no simple test right now that we can upstream here, it's deep in a deep code (comes from SplineLayer which pulls in DataInterpolations etc.). Someone can probably dig in and make a test out of that.

mcabbott:
This looks like what https://github.com/JuliaDiff/ChainRulesCore.jl/pull/446 is meant to fix. Can you check easily?

CarloLucibello:
closing as outdated

PR #791, opened 2020-09-15
Add PyCall custom @adjoint test

samuela:
Add a test for https://github.com/FluxML/Zygote.jl/issues/789. This test is actually currently failing on master.

PR #1363, opened 2023-01-24
test pycall adjoint

CarloLucibello:
supersedes #791

### DIFF
--- a/test/features.jl
+++ b/test/features.jl
@@ -681,8 +681,6 @@ end
   end == ([8 112; 36 2004],)
 end
 
-@testset "pycall" begin
-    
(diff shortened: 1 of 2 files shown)


test/features.jl
```diff
@@ -681,6 +681,16 @@ end
   end == ([8 112; 36 2004],)
 end
 
+@testset "pycall" begin
+    
+@testset "PyCall custom @adjoint" begin
```
CarloLucibello, line 686:
```suggestion   
@testset "PyCall custom @adjoint" begin
```

ToucheSir (approved):
Seems odd that Zygote would fail here of all places, but if the test runs quickly we might as well.

PR #1371, opened 2023-02-08
Don't differentiate getproperty on const module fields

ToucheSir:
Testing to see if this fixes type stability issues like the one reported in https://discourse.julialang.org/t/type-stability-with-flux-gradient-of-loss-function-requiring-parameters/94156.


ToucheSir:
Buildkite failure on 1.8 looks like a flaky test tolerance. Failure on 1.6 also looks unrelated, no issues on #1367 but showed up right after merging it???

CarloLucibello:
Can you relax the tolerance of that test?

CarloLucibello (approved):

PR #1376, opened 2023-02-16
Always call reverse in `map`

### LINKED ISSUE #1374: Incorrect gradient with comprehensions
Hi, I encountered and analyzed a bug in the pullback function of the `collect()` function, which is used in Julia to implement comprehensions. Here is a minimal workable example only for demonstration of the incorrect gradient calculation.

Imagine the following linear regression model
```julia
struct Affine
    W
    b
end

(m::Affine)(x) = m.W * x + m.b

m = Affine(zeros(3,3), zeros(3,1))
```
which we want to fit to the following data
```julia
x = [ 1.0,  2.0,  3.0]
y = [-1.0, -2.0, -3.0]
```
using a squared error loss function
```julia
l(y,ŷ) = sum(abs2.(y - ŷ))/2
```

For this we need the gradient of the loss function with respect to the model parameters
```julia
grads = gradient(m -> l(y,m(x)), m)

julia> grads[1].W
3×3 Matrix{Float64}:
 1.0  2.0  3.0
 2.0  4.0  6.0
 3.0  6.0  9.0

julia> grads[1].b
3×1 Matrix{Float64}:
 1.0
 2.0
 3.0
```

This can be easily verified. The gradient with respect to `m.W` is `(m(x) - y) * x'` and since `m(x)` is a zero vector in this example (all model parameters are initialized to zero here), this gradient equals `-y*x'`. The gradient with respect to `m.b` is `m(x) - y`, which reduces to `-y` here.

Now we implement the matrix multiplication in the regression model using a comprehension as follows
```julia
(m::Affine)(x) = [sum(x.*r) for r in eachrow(m.W)] + m.b
```

I guess that no one would come up with this idea, but as I wrote this example is only for demonstration. Now the model gradients are calculated as follows

```julia
grads = gradient(m -> l(y,m(x)), m)

julia> grads[1].W
3×3 Matrix{Float64}:
 3.0  6.0  9.0
 2.0  4.0  6.0
 1.0  2.0  3.0

julia> grads[1].b
3×1 Matrix{Float64}:
 1.0
 2.0
 3.0
```

The gradient with respect to `m.b` is correct, but for `m.W` it is not. The rows of the matrix are reversed.

Debugging the pullback functions showed, that the problem is in [*src/lib/array.jl#L204-L205*](https://github.com/FluxML/Zygote.jl/blob/aea34ce61cae22f804ddf8a6d46e769694cdf0e3/src/lib/array.jl#L204-L205). These are part of the generated functions `∇map()` and `∇pmap()`, which are called for instance at [*src/lib/array.jl#L233*](https://github.com/FluxML/Zygote.jl/blob/aea34ce61cae22f804ddf8a6d46e769694cdf0e3/src/lib/array.jl#L233) as part of the pullback function of the `collect()` function. The problem is caused by the fact, that in code line [*src/lib/array.jl#L204*](https://github.com/FluxML/Zygote.jl/blob/aea34ce61cae22f804ddf8a6d46e769694cdf0e3/src/lib/array.jl#L204) the `_tryreverse()` function is resolved to [*src/lib/array.jl#L168*](https://github.com/FluxML/Zygote.jl/blob/aea34ce61cae22f804ddf8a6d46e769694cdf0e3/src/lib/array.jl#L168), since the gradient `Δ` is a column matrix, but in the following code line [*src/lib/array.jl#L205*](https://github.com/FluxML/Zygote.jl/blob/aea34ce61cae22f804ddf8a6d46e769694cdf0e3/src/lib/array.jl#L205) the `_tryreverse()` function is resolved to [*src/lib/array.jl#L173*](https://github.com/FluxML/Zygote.jl/blob/aea34ce61cae22f804ddf8a6d46e769694cdf0e3/src/lib/array.jl#L173), since `Δf_and_args_zipped` is a vector. That means when applying the pullbacks to the gradient `Δ` in line [*src/lib/array.jl#L204*](https://github.com/FluxML/Zygote.jl/blob/aea34ce61cae22f804ddf8a6d46e769694cdf0e3/src/lib/array.jl#L204) the order is not reversed, but during unzipping in line [*src/lib/array.jl#L205*](https://github.com/FluxML/Zygote.jl/blob/aea34ce61cae22f804ddf8a6d46e769694cdf0e3/src/lib/array.jl#L205) the order is reversed.

One might ask, why the gradient `Δ` could be a matrix at all, if the result of `collect()` is a vector. But this can happen quite easily. Here for instance the result of the comprehension is added to `m.b`, which is a sum of a vector and a column matrix. This array sum is handled correctly in the forward pass and results in a column matrix. But in the backward pass the pullback function [*src/lib/array.jl#L655*](https://github.com/FluxML/Zygote.jl/blob/aea34ce61cae22f804ddf8a6d46e769694cdf0e3/src/lib/array.jl#L655) simply propagates the gradient `Δ`, which is of course a column matrix, to the pullback function of each addend without any type conversion. I'm not sure, if this is another bug or if gradient calculation must be consistent for column matrices instead of vectors.

From my point of view this is a *serious bug*, since comprehensions are the proposed pattern for implementing recurrent neural network layers in [*Flux.jl*](https://fluxml.ai/Flux.jl/stable/models/recurrence/). At least I encountered this bug within my custom RNN layer. Furthermore I guess not everybody is aware of the problem, that providing column matrices to a pullback function instead of vectors could potentially produce different results.

mcabbott:
Fixes #1374 by removing the special case that `_tryreverse` only reverses vectors & tuples.

Possibly it will now fail on other types. That's better than silently failing to reverse them.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -165,12 +165,14 @@ end
 # needed for stateful functions.
 # See https://github.com/FluxML/Flux.jl/issues/1209
 # Should be generalized to abstract array, but reverse takes a dims keyword there
+
+# That lead to https://github.com/FluxML/Zygote.jl/issues/1374
+# ... and post-1.6 dims is not required to reverse arrays.
 _tryreverse(m, backs, Δ) = backs, Δ
-function _tryreverse(m::typeof(map), backs, Δ::Union{AbstractVector, Tuple})
-  return reverse(backs), reverse(Δ)
-end
+_tryreverse(m::typeof(map), backs, Δ) = reverse(backs), reverse(Δ)
+
 _tryreverse(m, x) = x
-_tryreverse(m::typeof(map), x::Union{AbstractVector, Tuple}) = reverse(x)
+_tryreverse(m::typeof(map), x) = reverse(x)
 
 # With mismatched lengths, map stops early. With mismatched shapes, it makes a vector.
 # So we keep axes(x) to restore gradient dx to its full length & correct shape.
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -161,13 +161,9 @@ function unzip(tuples)
   _unzip(tuples, Val(N))
 end
 
-# Reverse iteration order when ∇map is applied to vector,
-# needed for stateful functions.
-# See https://github.com/FluxML/Flux.jl/issues/1209
-# Should be generalized to abstract array, but reverse takes a dims keyword there
-
-# That lead to https://github.com/FluxML/Zygote.jl/issues/1374
-# ... and post-1.6 dims is not required to reverse arrays.
+# Reverse iteration order in ∇map, for stateful functions.
+# This is also used by comprehensions, which do guarantee iteration order.
+# Not done for pmap, presumably because all is lost if you are relying on its order.
 _tryreverse(m, backs, Δ) = backs, Δ
 _tryreverse(m::typeof(map), backs, Δ) = reverse(backs), reverse(Δ)
 
--- a/test/gradcheck.jl
+++ b/test/gradcheck.jl
@@ -401,6 +401,22 @@ end
   @test gradient((x,y) -> sum(map(*,x,y)), (1,2,3,4,5), [1 2; 3 4]) == ((1,3,2,4,nothing), [1 3; 2 4])
 end
 
+@testset "map: issye 1374" begin
+  # The code to reverse iteration in map was very sloppy, could reverse fwd & not reverse, wtf.
+  # https://github.com/FluxML/Zygote.jl/issues/1374
+  struct Affine1374
+    W
+    b
+  end
+  (m::Affine1374)(x) = [sum(x.*r) for r in eachrow(m.W)] + m.b
+  m = Affine1374(zeros(3,3), zeros(3,1))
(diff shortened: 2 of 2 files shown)


src/lib/array.jl
```diff
@@ -165,12 +165,14 @@ end
 # needed for stateful functions.
 # See https://github.com/FluxML/Flux.jl/issues/1209
 # Should be generalized to abstract array, but reverse takes a dims keyword there
+
+# That lead to https://github.com/FluxML/Zygote.jl/issues/1374
+# ... and post-1.6 dims is not required to reverse arrays.
```
ToucheSir, line 170: How about amending the above comment instead? As noted it turned stale the moment we set min compat to 1.6.

ToucheSir (approved):
Downstream tests pass so I'm happy modulo small comment change. If we end up falling victim to Hyrum's law so be it, probably good to know than to be ignorant anyhow. One wonders how many correctness demons like this are lurking in the depths of Zygote's less well-trodden rules...

src/lib/array.jl
mcabbott: I've shrunk the comment. Not sure what the right policy is for `pmap`, but I haven't changed it.

PR #1385, opened 2023-03-05
Use `_pullback` inside rules instead of `pullback`

ToucheSir:
This will give us more flexibility to implement internal changes such as https://github.com/FluxML/Zygote.jl/issues/603 without changing the user-facing API.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -279,7 +279,10 @@ end
   enumerate(xs), back
 end
 
-@adjoint Iterators.Filter(f, x) = pullback(filter, f, collect(x))
+function _pullback(cx::AContext, ::Type{<:Iterators.Filter}, f, x)
+  res, back = _pullback(cx, filter, f, collect(x))
+  return res, back ∘ unthunk_tangent
+end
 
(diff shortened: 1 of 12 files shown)


src/lib/array.jl
```diff
@@ -279,7 +279,10 @@ end
   enumerate(xs), back
 end
 
-@adjoint Iterators.Filter(f, x) = pullback(filter, f, collect(x))
+function _pullback(cx::AContext, ::Type{<:Iterators.Filter}, f, x)
```
ToucheSir, line 282: Note on why these are `_pullback` instead of `@adjoint`: `@adjoint` tries to do some clever postprocessing of the pullback return which results in mismatched results (differential arg tuple off by one). Using `_pullback` is slightly more code and requires a bit more care, but it removes a few layers from the call stack and brings us closer to removing `@adjoint` completely.

oxinabox: Yes, when Miha and I were trying to get rid of Zygote's own types we likewise removed some uses of `@adjoint` and wrote the `_pullback` overloads directly

oxinabox (approved):
If tests pass them this LGTM

Issue #1377, opened 2023-02-24
Incorrect pullback of `AbstractFFTs.[i]rfft`?


trahflow:
Zygote defines an adjoint of `AbstractFFTs.[i]rfft` [here](https://github.com/FluxML/Zygote.jl/blame/master/src/lib/array.jl#L733-L739).
On the other hand, `AbstractFFTs` defines `rrule`s for `[i]rfft` and other functions itself (implemented by @devmotion ) [here](https://github.com/JuliaMath/AbstractFFTs.jl/blame/master/src/chainrules.jl#L63-L82)

I assume that Zygote prefers its own adjoint over the `ChainRulesCore.rrule` of `AbstractFFTs` (at least that is what I think I'm seeing).

Now the problem is that the two implementations differ (not just by the additional projection into the real domain, but by additional scaling factors) and give different resuls.
I haven't yet taken the time to figure out how the scaling factors in `AbstractFFTs` are motivated (unfortunately there's no comment that gives a hint), but finite differencing seems to agree with `AbstractFFTs` implementation and disagree with Zygote's.

Side question: Can I somehow tell Zygote to prefer the `ChainRulesCore.rrule` over its own `@adjoint`?

ToucheSir:
https://github.com/FluxML/Zygote.jl/issues/1158 I believe is the same issue by another name. https://github.com/FluxML/Zygote.jl/issues/899 looks related. I think https://github.com/FluxML/Zygote.jl/pull/1173 was meant to be a fix, but have no expertise to evaluate it.



> Side question: Can I somehow tell Zygote to prefer the `ChainRulesCore.rrule` over its own `@adjoint`?

Not without resorting to piracy, no. I think at this point we should just remove all FFT-related rules from Zygote and let the appropriate chainrules take over, but that would require someone with the requisite experience on how to do so.

CarloLucibello:
I think removing the Zygote's adjoint is safe, @trahflow do you mind filing a PR?

trahflow:
> https://github.com/FluxML/Zygote.jl/issues/1158 I believe is the same issue by another name

Sorry for the duplicate. I did a search but probably only for "irfft".

>  @trahflow would you mind filing a PR?

I will give it a shot. I won't get around doing it before the end of the week though.

trahflow:
[...]

Issue #83, opened 2019-03-01
TL;DR of Supported Operations for Docs


jessebett:
When writing the docs, it would be nice to follow jax's lead and include a TL;DR of what sorts of operations are supported. From their docs:

>  TLDR Do use
> 
>         Functional programming
>         Many of NumPy’s functions (help us add more!)
>         Some SciPy functions
>         Indexing and slicing of arrays like x = A[[5, 1, 7], :, 2:4]
>         Explicit array creation from lists like A = np.array([x, y])
> 
>     Don’t use
> 
>         Assignment into arrays like A[0, 0] = x
>         Implicit casting to arrays like np.sum([x, y]) (use np.sum(np.array([x, y]) instead)
>         A.dot(B) method syntax for functions of more than one argument (use np.dot(A, B) instead)
>         Side-effects like mutation of arguments or mutation of global variables
>         The out argument of NumPy functions
>         Dtype casting like np.float64(x) (use x.astype('float64') or x.astype(np.float64) instead).
> 
>     For jit functions, also don’t use
> 
>         Control flow based on dynamic values if x > 0: .... Control flow based on shapes is fine: if x.shape[0] > 2: ... and for subarr in array.
>         Slicing A[i:i+5] for dynamic index i (use lax.dynamic_slice instead) or boolean indexing A[bool_ind] for traced values bool_ind.

MikeInnes:
This should be pretty easy for us since the only thing you can't do is assignment into arrays (yet – #75). Some general docs and usage tips are definitely a good idea though.

baggepinnen:
> This should be pretty easy for us since the only thing you can't do is assignment into arrays (yet – #75). Some general docs and usage tips are definitely a good idea though.

How about try/catch blocks?
https://github.com/FluxML/Zygote.jl/issues/47#issuecomment-470053165

MikeInnes:
Yup, just remembered that when looking at this issue :) I'm hoping that's an easy fix though, whereas the mutation restriction will probably be part of a release.

CarloLucibello:
we have a limitations section in the docs

PR #1386, opened 2023-03-06
drop adjoints for [i,r,b]fft()

trahflow:
Following the discussion in https://github.com/FluxML/Zygote.jl/issues/1377 I'm dropping the adjoints for `fft()`, `ifft()`, `rfft()` etc.
since the rules for `rfft()` and `irfft()` were not correct and correct rules have since been added to `AbstractFFTs` in https://github.com/JuliaMath/AbstractFFTs.jl/pull/58.

Note that this does not yet fully fix the issue, since the adjoints for multiplication/left division by a real fft `AbstractFFTs.Plan` are still incorrect.
There is an in-progress PR at `AbstractFFTs` to provide ChainRules for these:
https://github.com/JuliaMath/AbstractFFTs.jl/pull/67

For now, the adjoints for multiplication/left divison by `AbstractFFTs.Plan` are kept, even though they lead to incorrect results for real forward and inverse ffts.

**Note though that if the `dims` argument is not provided to `fft()` etc., the rules provided by https://github.com/JuliaMath/AbstractFFTs.jl/pull/58 aren't actually hit, and the compilation will fail trying to differentiate a try/except block in `AbstractFFTs.plan_fft()`**
This is still better than a silently incorrect result as in `rfft()` and `irfft()`, but a regression for `fft()` etc. (without the `dims` argument) where the rules defined in Zygote were seemingly correct.
(which is why I made this a draft. Opinions on this welcome!)

### PR Checklist

- [x] Tests are added (or rather adapted/removed)
- ~[ ] Documentation, if applicable~

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -27,7 +27,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 
 [compat]
-AbstractFFTs = "0.5, 1.0"
+AbstractFFTs = "1.3.1"
 ChainRules = "1.44.1"
 ChainRulesCore = "1.9"
 ChainRulesTestUtils = "1"
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -671,12 +671,6 @@ AbstractFFTs.brfft(x::Fill, d, dims...) = AbstractFFTs.brfft(collect(x), d, dims
 
 # the adjoint jacobian of an FFT with respect to its input is the reverse FFT of the
 # gradient of its inputs, but with different normalization factor
-@adjoint function fft(xs)
-  return AbstractFFTs.fft(xs), function(Δ)
-    return (AbstractFFTs.bfft(Δ),)
-  end
-end
-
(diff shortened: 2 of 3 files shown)


gaurav-arya:
I've made #83 to handle the chain rules issue you noted!

trahflow:
Thanks!
I'll reset the relevant tests here once that is merged

Project.toml
```diff
@@ -27,7 +27,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 
 [compat]
-AbstractFFTs = "0.5, 1.0"
+AbstractFFTs = "1.3.1"
```
trahflow, line 30: should this be done?

ToucheSir: Bumping the compat version of a dependency to add a new min bound? That's perfectly fine as long as we're not suddenly breaking anything. We do it all the time for e.g. Flux -> NNlib.

trahflow:
I think the build error is unrelated (and will be fixed by #1387 )

ToucheSir (approved):
Tests pass so LGTM. Thanks!

Issue #1227, opened 2022-05-13
Zygote.gradient returns `nothing` instead of `NotImplemented


simonmandlik:
```julia
using ChainRulesCore, Zygote

f(x) = 2x
ChainRulesCore.rrule(::typeof(f), x) = 2x, d -> (NoTangent(), @not_implemented(""), )

Zygote.gradient(f, 1)
```

The last line returns `(nothing, )` from v0.6.38 onwards but used to return `ChainRulesCore.NotImplemented` in the previous versions.

Is this intended?

ToucheSir:
Yes: https://github.com/FluxML/Zygote.jl/pull/1205. As an alternative, you could try using `rrule_via_ad(ZygoteRuleConfig(), f, 1)`. That won't help if a `NotImplemented` is generated in the process of running AD, but it will short-circuit if it finds a `rrule`.

If anyone has any ideas on how to refactor the internals such that `rrule_via_ad` doesn't have to go through an unwrapping/re-wrapping step (i.e. AD uses ChainRules types all the way through), let me know and I'd be happy to discuss + assist in implementation.

SimonMandlik:
Ok, thanks!

ToucheSir:
Re-opening as a tracker for https://github.com/JuliaDiff/ChainRules.jl/pull/521#issuecomment-1445842030.

PR #1389, opened 2023-03-13
Allow ChainRules zero types internally

ToucheSir:
Following on https://github.com/FluxML/Zygote.jl/pull/1385, this allows us to propagate `ZeroTangent`, `NoTangent` and `NotImplemented` inside the boundary of `pullback`. Doing so should prevent lossy conversions back and forth, as well as inadvertent dropping of the erroring functionality `NotImplemented` brings. Conversions remain the same for the user-level API, which means this only partially addresses https://github.com/FluxML/Zygote.jl/issues/1227.

This PR can be seen as an incremental step towards #603. There are a couple of options for next steps:
- Allowing `Tangent`s to pass through without conversion
- Supporting thunks (#966)
- Switching all internal rules to use CR zeros instead of Nothing (or converting them to `rrule`s.

### PR Checklist

- [x] Tests are added
- [N/A] Documentation, if applicable

### DIFF
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -109,10 +109,6 @@ Convert `x` from the differentials types ChainRules uses to the format Zygote us
 @inline wrap_chainrules_output(x) = x
 @inline wrap_chainrules_output(x::AbstractThunk) = wrap_chainrules_output(unthunk(x))  # For now we are just not going to deal with thunks
 @inline wrap_chainrules_output(x::Tuple) = map(wrap_chainrules_output, x)
-# Zygote convention: even if many AbstractZero partials (i.e. multi-input function), make just 1 nothing.
-@inline wrap_chainrules_output(x::Tuple{Vararg{ChainRules.AbstractZero}}) = nothing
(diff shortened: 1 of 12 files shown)


ToucheSir:
Turns out just changing Zygote's own `@adjoint`s to support CR zeros is not enough because downstream package rules are also affected. https://github.com/FluxML/ZygoteRules.jl/pull/25 should help with that and is now a blocker for this PR.

src/compiler/chainrules.jl
```diff
@@ -350,3 +352,43 @@ z2d(dx::NamedTuple{L,S}, primal::AbstractDict) where {L,S<:Tuple{Vararg{Union{Nu
 end
 
 z2d(dx::Ref, primal) = z2d(dx[], primal)  # mutable structs
+
+
+"""
+    differential2zygote(dx)
+
+Convert input `dx` from ChainRules differential types to the Zygote format.
+This is similar to `wrap_chainrules_output(dx)`, but converts zero types.
```
oxinabox, line 361:
```suggestion
This is similar to `wrap_chainrules_output(dx)`, but converts zero types,
and recursively converts Tangents.
```

CarloLucibello:
closing as outdated

PR #1390, opened 2023-03-13
move require usage to extensions on 1.9+

KristofferC:
The require usage in this package was profiled to spend a significant time in Requires code. This moves the require usage to extensions to avoid that.

Before:

```julia
julia> using Tracker, Colors, Distances

julia> @time using Zygote # runs the `@require` code
  0.942142 seconds (2.32 M allocations: 160.158 MiB, 5.37% gc time, 59.86% compilation time: <1% of which was recompilation)
```

After:

```julia
julia> using Tracker, Colors, Distances

julia> @time using Zygote # loads the extensions
  0.358596 seconds (1.02 M allocations: 72.098 MiB, 7.65% gc time, 1.44% compilation time)
```
  
Note that I didn't run the full tests locally because they took a significant time.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -26,12 +26,24 @@ SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 
+[weakdeps]
+Colors = "5ae59095-9a9b-59fe-a467-6f913c188581"
+Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
+Tracker= "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
+
+[extensions]
+ZygoteColorsExt = "Colors"
+ZygoteDistancesExt = "Distances"
+ZygoteTrackerExt = "Tracker"
+
 [compat]
 AbstractFFTs = "1.3.1"
 ChainRules = "1.44.1"
 ChainRulesCore = "1.9"
 ChainRulesTestUtils = "1"
+Colors = "0.12"
 DiffRules = "1.4"
+Distances = "0.10"
 FillArrays = "0.8, 0.9, 0.10, 0.11, 0.12, 0.13"
 ForwardDiff = "0.10"
 GPUArrays = "8.4.2"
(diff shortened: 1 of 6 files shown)


Project.toml
```diff
@@ -43,6 +55,7 @@ NaNMath = "0.3, 1"
 Requires = "1.1"
 SnoopPrecompile = "1.0.3"
 SpecialFunctions = "1.6, 2"
+Tracker = "0.2"
```
CarloLucibello, line 58: It seems that these 3 packages should be added to [extras] as well to make CI happy

KristofferC: Yep, I added that.

Project.toml
```diff
@@ -26,12 +26,24 @@ SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 
+[weakdeps]
+Colors = "5ae59095-9a9b-59fe-a467-6f913c188581"
+Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
```
devmotion, line 31: I wonder if we should move forward with https://github.com/JuliaStats/Distances.jl/pull/246 instead? But I guess there's also no harm in keeping this code in an extension for a few releases.

KristofferC: Doesn't matter from my p.o.v.

KristofferC:
I don't think there are any tests for Tracker btw.

Project.toml
ToucheSir: Using `ZygoteRuleConfig` for `rrules` defined here would prevent any overwrites. That said, it would be much nicer if we didn't maintain 2 separate copies of these rules. The question there is whether Distances.jl would be willing to use Requires or hard deps to support Julia <1.9.

ext/ZygoteTrackerExt.jl
```diff
@@ -0,0 +1,19 @@
...
+
+if isdefined(Base, :get_extension)
+    using Zygote
+    using Tracker: Tracker, TrackedArray, TrackedReal
+else
+    using ..Zygote
+    using ..Tracker: Tracker, TrackedArray, TrackedReal
+end
+
+using Tracker: TrackedArray, TrackedReal
```
devmotion, line 11: This seems redundant:
```suggestion
```

Project.toml
devmotion: BTW there's another difference, I just re-read my comment in the Distances PR and apparently the rules there differ from the currently implemented rules in Zygote:

> The implementation in this PR is different from the code in https://github.com/FluxML/Zygote.jl/pull/923 since tests fail with the rules in the Zygote PR (e.g., due to type inference issues).
The rules in this PR are also all explicit and do not call back into the AD system, in contrast to the current implementation in Zygote.

Issue #1393, opened 2023-03-14
Always calling `reverse` on `map` breaks stuff


simsurace:
#1376 broke this for example (this used to work on v0.6.55):
https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/issues/355

Could the change be reverted (or maybe there is an obvious fix) and a new patch be released?

mcabbott:
Is there a minimal example?

When calling `map` with a stateful function, pre-1376 behaviour was just to blindly apply the pullbacks in the wrong order, except on a vector, which sometimes gave silently wrong answers. Present behaviour is to error if `reverse` doesn't work.

willtebbutt:
I believe it boils down to

```julia
julia> using LinearAlgebra

julia> X = UpperTriangular(randn(3, 3))
3×3 UpperTriangular{Float64, Matrix{Float64}}:
 -0.211524  -1.40637   -0.315358
   ⋅         0.933569  -0.797453
   ⋅          ⋅        -0.339659

julia> reverse(X)
ERROR: ArgumentError: cannot set index in the lower triangular part (3, 2) of an UpperTriangular matrix to a nonzero value (-1.4063656013863817)
Stacktrace:
  [1] setindex!
    @ ~/.julia/juliaup/julia-1.8.5+0.x64.apple.darwin14/share/julia/stdlib/v1.8/LinearAlgebra/src/triangular.jl:233 [inlined]
  [2] _setindex!
    @ ./abstractarray.jl:1374 [inlined]
  [3] setindex!
    @ ./abstractarray.jl:1344 [inlined]
  [4] _reverse!(A::UpperTriangular{Float64, Matrix{Float64}}, dims::Tuple{Int64, Int64})
    @ Base ./arraymath.jl:89
  [5] _reverse!
    @ ./arraymath.jl:71 [inlined]
  [6] reverse!(A::UpperTriangular{Float64, Matrix{Float64}}; dims::Function)
    @ Base ./arraymath.jl:70
  [7] _reverse(A::UpperTriangular{Float64, Matrix{Float64}}, dims::Function)
    @ Base ./arraymath.jl:60
  [8] #reverse#247
    @ ./arraymath.jl:59 [inlined]
  [9] reverse(A::UpperTriangular{Float64, Matrix{Float64}})
    @ Base ./arraymath.jl:59
 [10] top-level scope
    @ REPL[3]:1
```

[...]

PR #1396, opened 2023-03-14
Fix `reverse` failure

simsurace:
Addresses part of #1393 by adding an internal function `_reverse` that now has some specializations for known cases in which the fallback `reverse` fails. This preserves the general behavior that was introduced by #1376 while fixing the known bugs it introduced.


src/lib/array.jl
```diff
@@ -165,10 +167,21 @@ end
...
+_tryreverse(m::typeof(map), x) = _reverse(x)
+
+# Fallback
+_reverse(x) = reverse(x)
+
+# Known cases in the standard library on which `reverse` errors (issue #1393)
+_reverse(x::LowerTriangular) = UpperTriangular(reverse(parent(x)))
+_reverse(x::UpperTriangular) = LowerTriangular(reverse(parent(x)))
+_reverse(x::UnitLowerTriangular) = UnitUpperTriangular(reverse(parent(x)))
+_reverse(x::UnitUpperTriangular) = UnitLowerTriangular(reverse(parent(x)))
```
devmotion, line 182: Should they maybe call `_reverse` instead of `reverse`?
```suggestion
_reverse(x::LowerTriangular) = UpperTriangular(_reverse(parent(x)))
_reverse(x::UpperTriangular) = LowerTriangular(_reverse(parent(x)))
_reverse(x::UnitLowerTriangular) = UnitUpperTriangular(_reverse(parent(x)))
_reverse(x::UnitUpperTriangular) = UnitLowerTriangular(_reverse(parent(x)))
```

simsurace:
@mcabbott

src/lib/array.jl
```diff
@@ -165,10 +167,21 @@ end
...
+_reverse(x::Hermitian) = Hermitian(reverse(collect(x)))
+_reverse(x::Symmetric) = Symmetric(reverse(collect(x)))
```
devmotion, line 184: I guess we could avoid `collect` here, I think the following could work:
```suggestion
_reverse(x::Hermitian) = Hermitian(_reverse(x.data), x.uplo == 'U' ? :L : :U)
_reverse(x::Symmetric) = Symmetric(_reverse(x.data), x.uplo == 'U' ? :L : :U)
```

src/lib/array.jl
simsurace: I think for this selection of types it does not matter, but when extending this one may run into infinite recursion (I can't think of an example right now).

src/lib/array.jl
simsurace: Good point, thanks!

simsurace: BTW. this deviates from `reverse`, which makes a copy, but I think it makes sense here.

devmotion: This will make a copy as well, won't it? If `x.data isa Array`, then `_reverse(x.data) = reverse(x.data)` which will be a copy of `x.data`.

simsurace: Yeah that's right. I confused myself there. Never mind.

src/lib/array.jl
devmotion: As long as all implementations of `_reverse` guarantee that the recurse only into an array/structure whose number of nested levels is smaller, it seems infinite recursion is not possible? And the definitions here and below seem to satisfy this property: the array on the LHS is a wrapper of the array on the RHS that `_reverse` is called with, so the number of levels is reduced by 1.

simsurace: Yeah you are right, one just needs to be careful when extending this in the future.

test/lib/array.jl
```diff
@@ -65,3 +65,14 @@ end
     end
     @test gradient(f_comprehension, w)[1] == ones(5)
 end
+
+@testset "_reverse" begin
+    m = [1 2 3; 4 5 6; 7 8 9]
+    @testset "$wrapper" for wrapper in [
+        Hermitian, Symmetric, LowerDiagonal, UpperDiagonal,
+        UnitLowerDiagonal, UnitUpperDiagonal,
```
devmotion, line 73:
```suggestion
        Hermitian, Symmetric, LowerTriangular, UpperTriangular,
        UnitLowerTriangular, UnitUpperTriangular,
```

mcabbott:
Would be good to have some tests of an end use, i.e. `gradient` or perhaps `pullback` calls. These can be kept even if e.g. the whole implementation is deleted in favour of https://github.com/JuliaDiff/ChainRules.jl/pull/671, or copied over to other AD systems.

simsurace:
> Would be good to have some tests of an end use, i.e. `gradient` or perhaps `pullback` calls. These can be kept even if e.g. the whole implementation is deleted in favour of [JuliaDiff/ChainRules.jl#671](https://github.com/JuliaDiff/ChainRules.jl/pull/671), or copied over to other AD systems.

I added the MWE from #1393. Any others that come to mind?

test/lib/array.jl
```diff
@@ -1,4 +1,6 @@
 using ChainRulesTestUtils
+using LinearAlgebra: Diagonal, Hermitian, LowerTriangular, UpperTriangular
+using LinearAlgebra: UnitLowerTriangular, UnitUpperTriangular
 using Zygote: ZygoteRuleConfig, _pullback
```
simsurace, line 4:
```suggestion
using Zygote: ZygoteRuleConfig, _pullback, _reverse
```

devmotion (approved):

PR #1397, opened 2023-03-15
New bugfix release

devmotion:
Version bump to release the bugfix in #1396.


CarloLucibello (approved):

simsurace:
@devmotion TagBot failed: https://github.com/FluxML/Zygote.jl/actions/runs/4425485995

devmotion:
TagBot is a bit unreliable in my experience, I've had many failed runs lately. Fortunately it's not needed for registration of new releases but only "nice to have".

simsurace:
I think it still needs to be tagged then (I don't have any permissions).

PR #1400, opened 2023-03-19
Improve literal function and module detection

ToucheSir:
Previously, instrumentation would be thrown off by storing a function in a variable/SSA value. Likewise, module references like `Main.Base` would be ignored because they're lowered as GlobalRefs.

This continues some of the work started in https://github.com/FluxML/Zygote.jl/pull/1371 and should fix a couple of edge cases.

### PR Checklist

- [x] Tests are added
- [N/A] Documentation, if applicable

### DIFF
--- a/src/compiler/reverse.jl
+++ b/src/compiler/reverse.jl
@@ -28,6 +28,18 @@ unwrapquote(x::QuoteNode) = x.value
 
 is_getproperty(ex) = iscall(ex, Base, :getproperty)
 
+# Allows us to resolve constants which have been stored in Variables.
+# e.g. `%1 = 1; %2 = %1``, or `%1 = identity; %1(...)`.
+trylookup(ir::IR, @nospecialize(v)) = v
+trylookup(ir::IR, v::Variable) = haskey(ir, v) ? trylookup(ir, ir[v].expr) : v
+# Only resolve GlobalRefs to traverse module hierarchies
+function trylookup(ir::IR, ref::GlobalRef)
+  isconst(ref.mod, ref.name) || return ref
+  val = getproperty(ref.mod, ref.name)
+  return val isa Module ? val : ref
+end
+
+
 # The initial premise of literal_getproperty was in some ways inherently flawed, because for
 # getproperty it was intended that _pullback falls back to literal_getproperty, but we actually
 # want the opposite to happen, since Zygote should fall back to recursing into the getproperty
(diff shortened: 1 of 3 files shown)


ToucheSir:
Buildkite error is spurious. Dealing with PyCall on CI is such a mess...

src/compiler/reverse.jl
```diff
@@ -28,23 +28,38 @@ unwrapquote(x::QuoteNode) = x.value
...
+# Allows us to resolve constants which have been stored in Variables.
+# e.g. `%1 = 1; %2 = %1``, or `%1 = identity; %1(...)`.
+trylookup(ir::IR, @nospecialize(v)) = v
+trylookup(ir::IR, v::Variable) = haskey(ir, v) ? trylookup(ir, ir[v].expr) : v
+# Only resolve GlobalRefs to traverse module hierarchies
+function trylookup(ir::IR, ref::GlobalRef)
+  isconst(ref.mod, ref.name) || return ref
+  val = getproperty(ref.mod, ref.name)
+  return val isa Module ? val : ref
+end
```
MasonProtter, line 40: Because this is probably going to be type unstable, you *may* want to write this as something like
```julia
function trylookup(ir::IR, v) # maybe @nospecialize(v) ?
    while true
        if v isa Variable
            v = ir[v].expr
        elseif v isa GlobalRef
           isconst(ref.mod, ref.name) || return ref
           val = getproperty(ref.mod, ref.name)
           return val isa Module ? val : ref
        else
           return v
        end
    end
end
```

ToucheSir: Thanks Mason. I remember thinking about a loop and rejecting it, but I can't think of why because the snippet you posted looks just fine. Will try incorporating it when I return to this PR.

Pangoraw:
I looked into the current literal detection when I discovered that some of my `throw` call were not ignored as they should and found this PR!

It seems to me that the `trylookup` can be generalized to checks to any `GlobalRef` for example in the following function:

https://github.com/FluxML/Zygote.jl/blob/6d983d5a4e710e814d21d05d4e6d4e5040f0f962/src/compiler/reverse.jl#L185-L193

which could be replaced with:

```julia
ignored_f(f) = f in (Base.not_int, 
                     Core.trinsics, :not_int, 
                     Core.:(===), 
                     Core.apply_type, 
                     Core.typeof, 
                     Core.throw, 
                     Base.kwerr, 
                     Core.kwfunc, 
                     Core.isdefined) 
```

This comes from the fact that if the symbols are unresolved, the `GlobalRef` to a call will be with respect to the calling module and these checks will fail. Example where `throw` is unresolved and placed as `GlobalRef(Main,:throw)` in the lowered IR:

```julia
julia> f(x) = x == 0 ? throw("invalid") : x

julia> @code_lowered f(1)
CodeInfo(
1 ─ %1 = x == 0
└──      goto #3 if not %1
2 ─ %3 = Main.throw("invalid")
└──      return %3
3 ─      return x
)

julia> (@code_lowered f(1)).code[3].args[1]
:(Main.throw)

julia> (@code_lowered f(1)).code[3].args[1] == GlobalRef(Core, :throw)
false
```

Since we are already calling `isconst(::GlobalRef)` for any called `GlobalRef`, Zygote already has the side-effect of resolving the bindings even if they are not used which the compiler will do anyway later (see https://github.com/JuliaLang/julia/issues/44604). Therefore, checking for strict equality (`isconst(ref) && getproperty(ref.mod, ref.name) === Base.getproperty` for example in `isliteral_getproperty`) could be beneficial.

ToucheSir:
Interesting, I could've sworn I had a PR which tried to normalize source modules for this exact reason but I can't find it now. I'm also not sure why the original code used GlobalRefs instead of just checking strict equality. Did changing to use the latter break any tests for you?

Pangoraw:
I have had problems with StackOverflows in the `_pullback(::typeof(literal_...))` (I added a fallback to `_pullback(typeof(...))`) and the following test failures in forward tests where `_pushforward(::typeof(literal_indexed_literal))` appears in the stacktrace:

```  Test threw exception
  Expression: D((x->begin
                D((y->begin
                                x = y
                            end), x) * x
            end), 1) == 1
  MethodError: no method matching *(::Nothing, ::Int64)
  
  Closest candidates are:
    *(::Any, ::Any, ::Any, ::Any...)
     @ Base operators.jl:578
    *(::T, ::T) where T<:Union{Int128, Int16, Int32, Int64, Int8, UInt128, UInt16, UInt32, UInt64, UInt8}
     @ Base int.jl:88
    *(::StridedArray{P}, ::Real) where P<:Dates.Period
     @ Dates ~/.julia/juliaup/julia-1.9.1+0.x64.linux.gnu/share/julia/stdlib/v1.9/Dates/src/deprecated.jl:44
```

ToucheSir:
Yes, I think I ran into similar issues at one point. This part of the AD transform is very finicky and difficult to debug...

CarloLucibello:
closing as outdated

PR #1412, opened 2023-04-09
Replace "may" with "might" to avoid ambiguity

jeffreyesun:
As written, "Non-mutating functions may also use mutation under the hood" might imply that non-mutating functions are _allowed to_ use mutation under the hood

[Please delete this text and describe your change here.
For bugfixes, please detail the bug and include a test case which your patch fixes.
If you are adding a new feature, please clearly describe the design, its rationale, the possible alternatives considered.
It is easiest to merge new features when there is clear precedent in other systems; we need to know we're taking
the right direction since it can be hard to change later.]

### PR Checklist

- [ ] Tests are added

### DIFF
--- a/docs/src/limitations.md
+++ b/docs/src/limitations.md
@@ -48,7 +48,7 @@ We got an error message and a long stacktrace. The error informs us that our cod
 
 !!! warning
 
-    Non-mutating functions may also use mutation under the hood. This can be done for performance reasons or code re-use.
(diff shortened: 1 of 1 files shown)


ToucheSir (approved):
I wonder if that note shouldn't be reworded to be even clearer (e.g. spell out what "non-mutating" means and how this includes functions the user may not even be aware of deep in the bowels of some (standard) library. Until that happens though, this seems like a reasonable change.

PR #1418, opened 2023-04-29
Updated PyCall module to Pythoncall 

tej2612:
PyCall module has been changed and updated to PythonCall along with the related methods.
It provides flexibility of conversion. PythonCall supports far more combinations of types of T and x and can be extended to support more types.
PythonCall uses a separate Conda environment for each Julia environment/project/package and installs Python (and other Python packages) into that. This means that different Julia projects can maintain an isolated set of Python dependencies (including the Python version itself).
PythonCall supports Julia 1.6.1+ and Python 3.7+. PyCall requires numpy to be installed, PythonCall doesn't (it provides the same fast array access through the buffer protocol and array interface).

### DIFF
--- a/src/forward/number.jl
+++ b/src/forward/number.jl
@@ -38,7 +38,7 @@ for f in [>, <, ==, ===, !=, in]
   @eval @tangent $f(a, b) = $f(a, b), (_, _) -> false
 end
 
-@tangent convert(T::Type{<:Real}, x::Real) = convert(T, x), (_, ẋ) -> convert(T, ẋ)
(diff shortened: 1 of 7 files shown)


darsnack (changes requested):
PyCall is only used in one place in the test suite. Zygote doesn't actually depend on it or use it. This PR should only change that one place in the test suite instead of find-replacing all instances of `Base.convert` with `pyconvert`.

ToucheSir:
This was fixed by https://github.com/FluxML/Zygote.jl/pull/1485.

PR #1419, opened 2023-04-30
Allow `f` to return a Tuple in `withgradient(f, args...)`

mcabbott:
Maybe we should allow this:
```julia
julia> withgradient([1,2,4]) do x
          z = 1 ./ x
          sum(z), z
       end
(val = (1.75, [1.0, 0.5, 0.25]), grad = ([-1.0, -0.25, -0.0625],))
```
 i.e. return `(val=f(x), grad=gradient(first∘f, x))` when `f` returns a Tuple: Currently an error.

### DIFF
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -119,13 +119,15 @@ julia> ∇ == gradient(/, 1, 2)
 true
 ```
 
-If `f` returns a Tuple or NamedTuple, then it calculates
-`gradient(first∘f, args...)` but returns the whole `f(args...)`:
+Allows you to capture auxillary outputs, in addition to the scalar
+used by `gradient`. To do this, `f` must return a Tuple or NamedTuple.
+Then it calculates `grad = gradient(first∘f, args...)
+but returns the whole `val = f(args...)`:
 
 ```jldoctest; setup=:(using Zygote)
 julia> withgradient([1,2,4]) do x
           z = 1 ./ x
-          sum(z), z
+          sum(z), z  # here z is an auxillary output
        end
 (val = (1.75, [1.0, 0.5, 0.25]), grad = ([-1.0, -0.25, -0.0625],))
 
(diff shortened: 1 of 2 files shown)


ToucheSir:
I like the goal but I'm not sure how I feel about the semantics here. Would be nice to get a second opinion on this.

mcabbott:
Maybe this needs an "RFC" label. I agree it's a little weird & potentially surprising. 

Motivation is things like this: https://stackoverflow.com/questions/75921783/accuracy-and-gradient-update-not-within-the-same-training-loop  where it seems cleaner than pushing to some storage, from inside gradient, and hiding that from AD.

darsnack:
We discussed this PR during the community call.

The basic question was: is there another interpretation for the additional elements in an output tuple that isn't auxiliary data? We could consider it from the perspective of `pullback` where multiple sensitivities could be passed in. The final decision was that this interpretation does not meet the definition of a `gradient`, so in the rare case that we do allow multiple sensitivities, we wouldn't call that function `gradient`.

The decision for this PR is to merge it as-is. The only outstanding question was whether `gradient` should mirror the changes to `withgradient` (similar to `jax.grad` and `jax.value_and_grad` linked above).

src/compiler/interface.jl
```diff
@@ -115,12 +115,32 @@ as a named tuple.
 julia> y, ∇ = withgradient(/, 1, 2)
 (val = 0.5, grad = (0.5, -0.25))
 
-julia> ∇ == gradient(/, 1, 2)  # explicit mode
+julia> ∇ == gradient(/, 1, 2)
 true
+```
+
+If `f` returns a Tuple or NamedTuple, then it calculates
+`gradient(first∘f, args...)` but returns the whole `f(args...)`:
```
CarloLucibello, line 123: maybe explain why we allow this, like "Sometimes it is convenient to return some auxiliary outputs"

ToucheSir (approved):

Issue #1431, opened 2023-05-31
@adjoint for logabsdet duplicates ChainRules rrule


ElOceanografo:
In [this issue](https://github.com/JuliaDiff/ChainRules.jl/issues/719) (trying to sort out how to do reverse-mode AD for the `logabsdet` of sparse matrices), it was pointed out that [Zygote's `@adjoint ` for `logabsdet`](https://github.com/FluxML/Zygote.jl/blob/e2d7839a476fc1cda407500d9a2ca125bd6e2aa5/src/lib/array.jl#L379) duplicates and overrides [an `rrule` in ChainRules](https://github.com/JuliaDiff/ChainRules.jl/blob/11c230cdf0f37a4f42de909d6c1f8500d1a80d69/src/rulesets/LinearAlgebra/dense.jl#L170). Would it make sense to delete that rule here and let ChainRules handle it?

ToucheSir:
I'd say go for it. If tests pass it should be an easy merge.

PR #1432, opened 2023-06-01
Remove @adjoint for logabsdet

ElOceanografo:
This PR removes the `@adjoint` for `logabsdet` from Zygote. There is an equivalent `rrule` defined in ChainRules, so this should be redundant. Merging will close https://github.com/FluxML/Zygote.jl/issues/1431 and will help resolve https://github.com/JuliaDiff/ChainRules.jl/issues/719.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -376,8 +376,6 @@ function _pullback(cx::AContext, ::typeof(kron), a::AbstractMatrix, b::AbstractM
   return res, back ∘ unthunk_tangent
 end
 
-@adjoint logabsdet(xs::AbstractMatrix) = logabsdet(xs), Δ -> (Δ[1] * inv(xs)',)
-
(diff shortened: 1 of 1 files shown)


ToucheSir:
LGTM, but I'm not sure if that DynamicPPL reverse CI failure is real or not. @devmotion or someone else, have you seen this before?

oxinabox:
The DynamicPPL errors are all about `length(::typeof(identity))` which seems like it can't be related.
but DynampicPPL does call `logabsdet` a ton via Bijectors.jl so I would like it if @devmotion  or @torfjelde  could check.

devmotion (approved):
Good to go, the DynamicPPL failures can be reproduced in a clean environment with the latest release of Zygote (AFAICT they are caused by a confusion between Bijectors.stack (which is already removed in the latest release of Bijectors) and Base.stack in 1.9).

PR #1434, opened 2023-06-02
Fix broken example in documentation

LilithHafner:
I pasted this into a repl and it errored "UndefVarError: `x` not defined". The new version runs fine.


ToucheSir (approved):
Thanks!

Issue #1439, opened 2023-07-10
Confusing failure with broadcasted functions with type instability


DomCRose:
When a function is broadcasted which is type unstable with Dual type inputs, there is a good chance the element type of the resulting output will be abstract, leading to a failure of the logic at https://github.com/FluxML/Zygote.jl/blob/2f4937096ee1db4b5a67c1c31fe3ebeab1c96c8c/src/lib/broadcast.jl#L284 (thanks to @ToucheSir who helped debug this). This can then cause an error much later on than the origin of the Dual, after Duals leak into the pullback definition and e.g. the gradient of the output is pulled onto the gradient of the input which assumed a non-Dual eltype, making it confusing to debug. Perhaps even worse, in some cases it causes the gradient to fail silently, either returning nothing or Duals for the gradient.

A MWE of silent failure on 1.9.0, in a temporary environment with only Zygote:
```julia
using Zygote
f(x) = x > 1.0 ? 1.0 : x^2
g(x) = sum(f.(x))
gradient(g, collect(0.5:0.25:1.5)) # (nothing,)
```
In contrast with the expected behaviour of:
```julia
using Zygote
f(x) = x > 1.0 ? one(x) : x^2
g(x) = sum(f.(x))
gradient(g, collect(0.5:0.25:1.5)) # ([1.0, 1.5, 2.0, 0.0, 0.0],)
```

A MWE of error, using repeat with the inner keyword as an example which doesn't allow the Dual to leak:
```julia
using Zygote
f(x) = x > 1.0 ? 1.0 : x^2
g(x) = sum(repeat(x, inner=2) .* f.(repeat(x, inner=2)))
gradient(g, collect(0.5:0.25:1.5))
```
results in:
```julia
ERROR: MethodError: no method matching Float64(::ForwardDiff.Dual{Nothing, Float64, 1})

Closest candidates are:
  (::Type{T})(::Real, ::RoundingMode) where T<:AbstractFloat
   @ Base rounding.jl:207
  (::Type{T})(::T) where T<:Number
   @ Core boot.jl:792
  (::Type{T})(::AbstractChar) where T<:Union{AbstractChar, Number}
   @ Base char.jl:50
  ...

Stacktrace:
 [1] convert(#unused#::Type{Float64}, x::ForwardDiff.Dual{Nothing, Float64, 1})
   @ Base .\number.jl:7
[...]

PR #1441, opened 2023-07-13
Fix broadcasts which are type unstable with Dual numbers

DomCRose:
Closes https://github.com/FluxML/Zygote.jl/issues/1439.

This is acheived by moving a dispatch from the element type of the output of broadcasting, to each individual element within the pullback, along with ensuring non-concrete element type arrays take the same path as concrete Dual arrays. This should hopefully compile away when the eltype is concrete, and indeed some simple benchmarks show no loss of performance, but I've hardly been exhaustive.

I've also added a few tests covering various real / complex input / output combinations, and a specific case that produced errors before rather than silently failing.

While its nice that this works, it could be worth adding a note to the documentation about the performance of broadcasting which has a type stable forward pass but becomes type unstable on Dual inputs, and perhaps that likewise such Dual input stability is required for the code to work on the GPU. But I'm not sure where that could go.

### PR Checklist

- [x] Tests are added

### DIFF
--- a/src/lib/broadcast.jl
+++ b/src/lib/broadcast.jl
@@ -282,25 +282,40 @@ end
   out = dual_function(f).(args...)
   T = eltype(out)
   if !isconcretetype(T) || T <: Union{Dual, Complex{<:Dual}}
-    return _broadcast_forward(out, args...)
+    if any(eltype(a) <: Complex for a in args)
+      return _broadcast_forward_complex(out, args...)
+    else
+      return _broadcast_forward(out, args...)
+    end
   else
     return (out, _ -> nothing)
   end
 end
 
+# Real input
 @inline _extract_value(x) = value(x)
 @inline _extract_value(x::Complex) = Complex(value(real(x)), value(imag(x)))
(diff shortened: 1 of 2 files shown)


mcabbott:
Thanks!

Cc @ptiede who most recently worked on all this, in #1324

ptiede:
I'm traveling right now, but I should be able to look at this in a couple of days. I think all the changes make sense and I'm sorry for missing this in the original PR. 
With regards to splitting the function, that's fine, but we shouldn't ruin the GPU complex number support. I have needed that in some of my projects, and it solved a bunch of previous issues with complex number GPU support.

DomCRose:
Great, thanks. No rush.

> With regards to splitting the function, that's fine, but we shouldn't ruin the GPU complex number support. I have needed that in some of my projects, and it solved a bunch of previous issues with complex number GPU support.

Certainly. I don't believe this breaks anything, I reverted my attempt to merge the functions further when I realized it broke GPU compilation (in a real -> real case actually, not even complex).

PR #1452, opened 2023-09-01
Mark AbstractFFT test broken

ToucheSir:
Having to look through every CI job to see if the failures are known is laborious. The Test stdlib gives us a way to automate this, so let's use it.


gaurav-arya:
FYI that there isn't a clear timeline for when AbstractFFTs will get its next release, so if you'd like to fix the Zygote CI without waiting on it I'd recommend just merging this, we can always change it back later

ToucheSir:
Thanks for the heads up. I think this PR is good to go then.

CarloLucibello (approved):

Issue #1450, opened 2023-08-30
ERROR: try/catch is not supported error when computing gradient of gradients on CUDA array


FerreolS:
Hi,
I'm trying to implement a gradient penalty with Lux. It is fine on CPU but raise a "try/catch" error on GPU (CUDA). It is seems to be linked to [the try catch here]( https://github.com/FluxML/Zygote.jl/blob/d4562e330d588cb986604bb4f1942bf9fca8ecc5/src/compiler/interface2.jl#L27C10-L27C10 ) but I'm not able to figure out what could be the problem.

```julia
using Zygote, Lux,LuxCUDA,Random,CUDA, cuDNN
function mwe(dev)
	rng = Random.default_rng()
	D = Dense(5,1, relu)
	ps, st = Lux.setup(rng, D) |> dev
	x = rand(5, 2) |> dev
	g(ps) = sum(abs2,only(gradient(x -> sum(first(D(x,ps,st))),x)))
	gradient(x->g(x),ps)
end
```

```julia
(@v1.9) pkg> status Lux, Zygote, LuxCUDA
Status `~/.julia/environments/v1.9/Project.toml`
  [b2108857] Lux v0.5.3
  [d0bbae9a] LuxCUDA v0.3.0
  [e88e6eb3] Zygote v0.6.63

julia> mwe(cpu_device())

((layer_1 = (weight = Float32[0.7818906 3.0653906 … 1.399884 -0.26843658; 0.0059568123 0.023353593 … 0.010664978 -0.0020450768; 0.47049373 1.8445637 … 0.8423642 -0.16152865], bias = Float32[0.0; 0.0; 0.0;;]), layer_2 = (weight = Float32[5.04391 -0.035778362 2.7771542], bias = nothing)),)

julia> mwe(gpu_device())
ERROR: try/catch is not supported.
Refer to the Zygote documentation for fixes.
https://fluxml.ai/Zygote.jl/latest/limitations

Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] instrument(ir::IRTools.Inner.IR)
    @ Zygote ~/.julia/packages/Zygote/4rucm/src/compiler/reverse.jl:128
  [3] #Primal#31
    @ ~/.julia/packages/Zygote/4rucm/src/compiler/reverse.jl:227 [inlined]
  [4] Primal
    @ ~/.julia/packages/Zygote/4rucm/src/compiler/reverse.jl:226 [inlined]
[...]

PR #1328, opened 2022-11-10
Excise getindex adjoint

ToucheSir:
We have a better rule in Chainrules now. 

Draft for now because I suspect downstream CI will have something to say about this. If it works though, would fix #820 and #1327.

### PR Checklist

- [ ] Tests are added (do we want any of the tests from #944?)

PR #1453, opened 2023-09-01
Remove redundant sum() rules

ToucheSir:
The pullback is non-differentiable, which messes with nested AD (https://github.com/FluxML/Zygote.jl/issues/1450). It's also not clear to me why this rule still exists when ChainRules has a seemingly GPU-compatible one. Let's see what CI says.

### PR Checklist

- ~~Tests are added~~
- ~~Documentation, if applicable~~


mcabbott:
I thought this existed in order to opt-out of the Zygote rule for sum which makes a FillArray. 
```
julia> gradient(sum, [2.0, 3.0])
(Fill(1.0, 2),)
```
We could delete that too, it saves one copy sometimes but rarely matters in real code, and causes problems.

ToucheSir:
Deleting that rule fixes all but one testsuite, https://github.com/FluxML/Zygote.jl/blob/master/test/lib/array.jl#L53. Not sure how best to fix it. Perhaps we could generalize https://github.com/FluxML/Zygote.jl/blob/612961353e4a81f9861fbca9db714e86f30ad0a3/src/lib/array.jl#L340-L342 to work on all `Integer`s and convert it to a `rrule(::ZygoteRuleConfig, ...)` for future-proofing at the same time?

mcabbott:
We could certainly delete the rule for bool arrays, as there's one here:

https://github.com/JuliaDiff/ChainRules.jl/blob/ba52ec89ddd97a07e79cc35a9fa39019915d203b/src/rulesets/Base/nondiff.jl#L80

IDK what the issue with that Dict test is. 

(Considering integers to be differentiable was a mistake, IMO, but a breaking change to fix that, here or in CR.)

ToucheSir:
> IDK what the issue with that Dict test is.

The old rule was arguably wrong, because it was passing through the gradient for the summed value without doing any form of projection. If this were a scalar function, asking to differentiate wrt an integer argument would return a float gradient. So in my mind the test is actually capturing incorrect and inconsistent behaviour of the current rule. If we agree on that, I'll just tweak the test and we'll be back on green CI (minus known AbstractFFT failures).

mcabbott:
Sorry I didn't look closely, but if the change is just that now you get a Dict of Floats not Ints, then that seems totally fine, we just adjust the test.

ToucheSir:
The one remaining failure:
```julia
sum, prod, cumsum: Test Failed at /var/lib/buildkite-agent/builds/gpuci-1/julialang/zygote-dot-jl/test/gradcheck.jl:117
  Expression: gradient(sum, [true, false, true]) == (nothing,)
 Evaluated: nothing == (nothing,)
```
Which comes from the `isnothing` ternary on https://github.com/FluxML/Zygote.jl/blob/e0d3d8b1a785ec291f0a41da3f12cad51d80eb6b/src/compiler/interface.jl#L98

@mcabbott do you recall why we're collapsing to `nothing` here? I can't recall how we're supposed to handle `nothing` vs `(nothing,)` vs `(nothing, ..., nothing)` when returned from the pullback.

mcabbott:
My memory is that Zygote is eager to collapse any tuple of nothings to just nothing, but doesn't always manage to do so. I think at least `withgradient` and perhaps `gradient` try to restore them & always make a tuple. But I may have forgotten things.

ToucheSir:
It looks like `gradient` is not trying to make a tuple when it goes get singular `nothing`. Should we make it do so? A version of this problem (more aggressive collapsing of zeros after moving to CR rules) is also causing the last two (non-unbreaking) test failures in https://github.com/FluxML/Zygote.jl/pull/1328.

mcabbott:
Want to rush this in & pretend it was part of breaking 0.7.0?

test/lib/array.jl
```diff
@@ -50,8 +50,8 @@ end
 @testset "dictionary comprehension" begin
     d = Dict(1 => 5, 2 => 6)
     g = gradient(d -> sum([v^2 for (_,v) in d]), d)[1]
-    @test g isa Dict{Int, Int}
-    @test g == Dict(1 => 10, 2 => 12)
+    @test g isa Dict{Int, Float64}
+    @test g == Dict(1 => 10.0, 2 => 12.0)
```
mcabbott, line 55: To trigger CI:
```suggestion
```

ToucheSir:
I totally forgot this one existed and was not looking forward to creating another PR to tackle https://discourse.julialang.org/t/second-order-gradient-with-lux-zygote-cuda-enzyme/124301. Take it away!

CC @pevnak for posterity, this likely would've addressed the issue you had with Zygote over Zygote.

mcabbott:
Done, I think. Molly.jl error is InitError: Exception[GLFW.GLFWError(65550, "X11: The DISPLAY environment variable is missing"), ErrorException("glfwInit failed")], unrelated.

test/gradcheck.jl
```diff
@@ -114,7 +114,7 @@ end
   @test gradtest(X -> sum(sum(x -> x^2, X; dims=1)), randn(10)) # issue #681
 
   # Non-differentiable sum of booleans
-  @test gradient(sum, [true, false, true]) == (nothing,)
+  @test_skip gradient(sum, [true, false, true]) == (nothing,)  # fine locally, fails on buidkite?
```
mcabbott, line 117:
```suggestion
  @test gradient(sum, [true, false, true]) == (nothing,)
```

mcabbott (approved):

PR #1458, opened 2023-09-26
Extend kron support

willtebbutt:
The current `kron` rules do not handle `kron(::AbstractMatrix, ::AbstractVector)` or `kron(::AbstractVector, ::AbstractMatrix)`. This PR adds support for both. Fortunately, the implementation strategy is precisely the same as the one currently taken.

A method of `_pullback` for `kron` is removed in favour of a single method of `_pullback` which can handle all four combinations of argument types.

Fixes https://github.com/JuliaGaussianProcesses/TemporalGPs.jl/pull/115

### PR Checklist

- [x] Tests are added

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.6.64"
+version = "0.6.65"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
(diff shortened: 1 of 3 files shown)


willtebbutt:
Has it caused you problems previously?

simsurace:
> I can't wait for us to be rid of this rule

Maybe in the not-so-distant future:
https://github.com/JuliaDiff/ChainRules.jl/pull/741

ToucheSir:
> Has it caused you problems previously?

Not so much. It was just frustrating to have CI broken for so long (in Zygote and Tracker) because of changes in Base and to have to patch the issue in the AD itself instead of in ChainRules. I'm glad to see progress is being made on a proper rule :)

PR #1462, opened 2023-10-11
Handle `Expr(:boundscheck)`

ToucheSir:
Fixes https://github.com/FluxML/Zygote.jl/issues/1438. We follow the same route as [Diffractor](https://github.com/JuliaDiff/Diffractor.jl/pull/177)and [Umlaut](https://github.com/dfdx/Umlaut.jl/pull/35)/Yota by always turning on bounds checking for correctness.

### PR Checklist

- [x] Tests are added
- [N/A] Documentation, if applicable

PR #1463, opened 2023-10-11
Fix `accum` genfunc error on nightly

ToucheSir:
Somewhere between 1.10 and 1.11, world age handling and permitted operations in `@generated` functions became more strict. This makes string interpolation fail, apparently.

Alongside #1462, this should fix the remaining visible errors on nightly CI.

### PR Checklist

- [ ] Tests are added


ToucheSir:
This must've been fixed on a later nightly, because I don't see it in CI logs any more.

PR #1465, opened 2023-10-15
Handle unreachable blocks in the adjoint CFG

### LINKED ISSUE #1118: return within at-inbounds yields invalid IR
Reduced from https://github.com/JuliaLang/julia/issues/43023:
```julia
julia> import Zygote

julia> f() = @inbounds return 0.0

julia> y, back = Zygote.pullback(f)
(0.0, Zygote.var"#55#56"{typeof(∂(f))}(∂(f)))

julia> ci = @code_lowered back.back(1.0)
CodeInfo(
1 ─      $(Expr(:meta, :inline))
│   %2 = Base.getfield(#self#, :t)
│   %3 = Base.getindex(%2, 1)
│   %4 = %3 !== 0x01
└──      goto #4 if not %4
2 ─      goto #3
3 ─      phi_1_1 = nothing
└──      goto #1
4 ─ %9 = Zygote.tuple(nothing)
└──      return %9
)

julia> ci.slotnames
3-element Vector{Symbol}:
 Symbol("#self#")
 :Δ
 :phi_1_1

julia> ci.slotflags
2-element Vector{UInt8}:
 0x00
 0x00
```
Julia determines the number of slots based on `length(ci.slotflags)`, so there will be no slot for `phi_1_1` and trying to access it will result in an error:
```julia
julia> @code_typed back(1.0)
ERROR: BoundsError: attempt to access 2-element Vector{Any} at index [3]
```
This is with Julia 1.6.

Actually calling `back(1.0)` is also likely to lead to a segfault during codegen, although in this reduced example, it seems to go ok.

The CodeInfo is produced by Zygote, so it is likely not Julia that is to blame here. It might be that the root cause is in IRTools, but drilling down into Zygote to hunt this down further is better done by someone knowing his way around the code.

### LINKED ISSUE #1380: `Zygote.withgradient` crashes Julia REPL
crosspost from: https://github.com/JuliaLang/julia/issues/48817

MWE:

```julia
using SpecialFunctions, Zygote
function maternKernel(
    nu,
    d
)
    if d > 0
        x = sqrt(2* nu*d)
        return 2^(1-nu)/gamma(nu) * x^nu * besselk(nu,x)
    else
        return 1 # variance
    end
    throw(ArgumentError("the squared Norm should never be negative!"))
end
Zygote.withgradient(x->maternKernel(1.,x), 0.)
```

Pangoraw:
The error described in #1118 and #1380, stopped happening in Julia 1.10-beta (maybe because of https://github.com/JuliaLang/julia/pull/50943) but the gradient is now wrong.

```julia
julia> f(x) = @inbounds return x

julia> Zygote.gradient(f,1.)
(nothing,)
```

This PR adds unreachable branches at the end of blocks in the adjoint when those blocks are unreachable in the primal which fixes the issue in both 1.9 and 1.10 because it avoids implicit branches.

I also though of removing unreachable blocks altogether since some operations are invalid for these blocks in IRTools (see [`dominators`](https://github.com/FluxML/IRTools.jl/blob/6fcac25fbda04cdd6de5a163ae8f9b8c6193bc8a/src/passes/passes.jl#L62) for example) but it seemed to do a lot more work than this fix.

Fixes #1380
Fixes #1118

> [!note]
> It depends on https://github.com/FluxML/IRTools.jl/pull/115.


oxinabox (approved):
this seems good to me, based on the new test. But I am not that familar with this bit of the code

Pangoraw:
I add to had the `Core.throw` call because having an `unreachable` in a reachable path throws a compiler error. And it is possible to manually construct a pullback which reaches these adjoint blocks whose primal blocks are unreachable.

```julia
julia> using Zygote

julia> function f(x, cond)
           if cond
               return x
           else
               return 2x
           end
           return 3x
       end

julia> back = Zygote.Pullback{Tuple{typeof(f), Float64, Bool}, Any}(([Returns([1,2,3])], [Returns([1,2,3])], 0x03))
∂(f)

julia> back(1.)
ERROR: "unreachable"
Stacktrace:
 [1] f
   @ ./REPL[2]:7 [inlined]
 [2] (::Zygote.Pullback{Tuple{typeof(f), Float64, Bool}, Any})(Δ::Float64)
   @ Zygote ~/Projects/Zygote.jl/src/compiler/interface2.jl:0
 [3] top-level scope
   @ REPL[5]:1
```

ToucheSir:
My understanding is that this may incur an allocation for the string "unreachable" in some methods. Is this reliably optimized out in practice? If not, it might make sense to define a custom immutable error type. That type could also track more information such as file and line number to help with troubleshooting.

Pangoraw:
The string `"unreachable"` string literal is actually allocated once by the parser and permanently rooted.

```julia
julia> using Zygote

julia> f(x) = @inbounds return x;

julia> invalid_pull = Zygote.Pullback{Tuple{typeof(f), Float64}, Any}(0x02,);

julia> m = only(methods(Zygote.adjointcfg));
       root = m.roots[findfirst(==("unreachable"), m.roots)];

julia> unreachable_string = try invalid_pull(1.); catch exc; exc end
"unreachable"

julia> unreachable_string === root
true
```

One problem is that it may taint the effects of the pullback compared to the case without unreachable blocks (which we could have if we removed unreachable blocks from the primal):

```julia
julia> Core.Compiler.infer_effects(invalid_pull, (Float64,))
(!c,!e,!n,!t,!s,!m,+i)′

julia> g(x) = return x;

julia> _, pull = Zygote.pullback(g, 1.);
       Core.Compiler.infer_effects(pull.back, (Float64,))
(+c,+e,+n,+t,+s,+m,+i)
```

ToucheSir:
A quick sanity check before I merge: does this pass locally on nightly for you?

Pangoraw:
The added testset does, but the entire test suite fails with CUDA and a missing `make_seed`  attribute just like in CI.

ToucheSir:
Could you try temporarily commenting out the CUDA test imports and the test block in https://github.com/FluxML/Zygote.jl/blob/49a11849157222e9ae5c995f5886a9e19b8505aa/test/runtests.jl#L8-L15? Either locally or on this branch to run CI would be fine.

Pangoraw:
I had to cherry-pick #1462 but there is one `@optout` and two grad checks tests failing. I don't know if those are related to this pr.

PR #1466, opened 2023-10-24
Test changes from https://github.com/FluxML/ZygoteRules.jl/pull/26

ToucheSir:
We have both Zygote's own CI and downstream tests here, so it'll be much easier to see what the impact is.

### PR Checklist

- [ ] Tests are added

### DIFF
--- a/src/Zygote.jl
+++ b/src/Zygote.jl
@@ -6,6 +6,11 @@ using LinearAlgebra: copytri!, AbstractTriangular
 import ZygoteRules: @adjoint, @adjoint!, AContext, adjoint, _pullback, pullback,
   literal_getproperty, literal_getfield, unthunk_tangent
 
+ZygoteRules.gradtuple0(::Tuple{Vararg{Nothing}}) = nothing
+ZygoteRules.gradtuple1(::Tuple{Vararg{Nothing}}) = nothing
+ZygoteRules.gradtuple2(::Tuple{Vararg{Nothing}}) = nothing
+ZygoteRules.gradtuple3(::Tuple{Vararg{Nothing}}) = nothing
+
 using ChainRulesCore
 using ChainRules: ChainRules, rrule, unthunk, canonicalize
 using IRTools


devmotion:
Seems the Zygote test failures only point to a few tests that have to be updated since they rely on the current non-collapsing behaviour.

ToucheSir:
Most if not all of those should be fixed by https://github.com/FluxML/Zygote.jl/pull/1328, but it hasn't been merged yet. Do you have any insight into the DynamicPPL error? If it doesn't look related, the ZygoteRules PR should be good to go.

devmotion:
DynamicPPL can be ignored - AbstractMCMC 4.5 contains a breaking change which was accidentally released in a non-breaking release (https://github.com/TuringLang/AbstractMCMC.jl/pull/132#issue-1959740311) but hopefully it will be yanked soon (https://github.com/JuliaRegistries/General/pull/94031).

devmotion:
I went through the CI logs once more and I think https://github.com/FluxML/Zygote.jl/actions/runs/6631670834/job/18015683033?pr=1466#step:6:818 might be problematic: There `(nothing,)` seems to be correct - at least I assumed the return value of `gradient` should always be a `Tuple` with one element for each of the function arguments. Otherwise splatting, indexing, etc. could lead to nasty surprises. I still think that the other two failing tests are actually improved by the ZygoteRules change because tuples of `nothing`s that are returned by pullbacks are simplified.

devmotion:
I guess my assumption about the return type of `gradient` was wrong (but this is not documented anywhere, is it?): https://github.com/FluxML/Zygote.jl/blob/f7551273080569316f5127cc6a537b62ef50f592/src/compiler/interface.jl#L98

ToucheSir:
Yeah, so this goes all the way back to the early early days of Zygote and I don't think the behaviour nor the reasoning for it have ever been documented: https://github.com/FluxML/Zygote.jl/blame/087b55467afb4ee312ad1325de02bdf91e07e790/src/compiler/interface.jl#L60. We should make a more formal guarantee now, but of what? It would be good to know where other AD libraries fall on this, AbstractDifferentiation.jl and its backends come to mind. Also cc @mcabbott since I think he last had to deal with this part of the codebase.

devmotion:
Yeah, I think such conventions should be documented more clearly and followed more consistently. IMO it would be reasonable to demand that `gradient` always returns a tuple with gradients for each input argument (as done e.g. in AbstractDifferentiation). Also for the pullback functions, I guess it would be helpful for developers to document that `nothing`s should always be collapsed.

PR #1467, opened 2023-10-25
Fix tests for ZygoteRule change

devmotion:
Fixes of the tests for https://github.com/FluxML/ZygoteRules.jl/pull/26. See https://github.com/FluxML/Zygote.jl/pull/1466.

### DIFF
--- a/test/gradcheck.jl
+++ b/test/gradcheck.jl
@@ -114,7 +114,7 @@ end
   @test gradtest(X -> sum(sum(x -> x^2, X; dims=1)), randn(10)) # issue #681
 
   # Non-differentiable sum of booleans
-  @test gradient(sum, [true, false, true]) == (nothing,)
+  @test gradient(sum, [true, false, true]) === nothing
   @test gradient(x->sum(x .== 0.0), [1.2, 0.2, 0.0, -1.1, 100.0]) == (nothing,)
 
   # https://github.com/FluxML/Zygote.jl/issues/314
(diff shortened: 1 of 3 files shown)


test/gradcheck.jl
```diff
@@ -114,7 +114,7 @@ end
   @test gradtest(X -> sum(sum(x -> x^2, X; dims=1)), randn(10)) # issue #681
 
   # Non-differentiable sum of booleans
-  @test gradient(sum, [true, false, true]) == (nothing,)
+  @test gradient(sum, [true, false, true]) === nothing
```
devmotion, line 117:
```suggestion
  @test gradient(sum, [true, false, true]) === (nothing,)
```

test/structures.jl
```diff
@@ -64,5 +64,5 @@ end
   end
 
   m, b = Zygote._pullback(Zygote.Context(), nameof, M)
-  @test b(m) == (nothing, nothing)
+  @test b(m) === nothing
```
devmotion, line 67:
```suggestion
  @test b(m) == (nothing, nothing)
```

devmotion:
Seems #1328 "fixed" all but one test error.

oxinabox (approved):

devmotion:
Flux test failure indicates that the latest Zygote release fixed a test that is currently marked as broken. Should be fixed by https://github.com/FluxML/Flux.jl/pull/2349.

ToucheSir:
That was my bad, I thought it was merged too and tagged a release! I agree with your idea in https://github.com/FluxML/Zygote.jl/pull/1466#issuecomment-1780649701, just need a quick sanity check from someone else before we start merging and updating the Zygote interface too.

PR #1474, opened 2023-12-05
Support try/catch on the happy (nothrow) path

Pangoraw:
This enables differentation support to try/catch when no exception is thrown on the try block.

The changes are twofolds:
 - Stop throwing when encountering `:enter` and `:leave` expressions, add an dedicated error message in the adjoint of the catch block instead.
 - Make sure that alphas in the first block are stored as stacks if it is a try block since they are not garanteed to run (an exception can be thrown before).

Here is an example of the behavior:

```julia
julia> using Zygote

julia> function foo(f)
         y = 1
         try
           y = f()
         catch
           y
         end
         y
       end
foo (generic function with 1 method)

julia> gradient(x -> foo(() -> x), 1) # 1
(1.0,)

julia> _, pull = pullback(foo, () -> 0//0)
(1, Zygote.var"#76#77"{Zygote.Pullback{Tuple{typeof(foo), var"#9#10"}, Any}}(∂(foo)))

julia> pull(1.)
ERROR: Can't differentiate function execution in catch block at #= REPL[4]:4 =#.
Stacktrace:
...
```

https://github.com/FluxML/IRTools.jl/pull/117 is a companion change that fixes some of the error that can currently happen with try/catch blocks in IRTools.jl.

Supersedes https://github.com/FluxML/Zygote.jl/pull/466.


ToucheSir:
I'm not really sure how to evaluate any risks from this change other than reading through the old PR. @oxinabox as the reviewer on that, would you be able to take a look here?

docs/src/limitations.md
```diff
@@ -82,27 +82,30 @@ julia> gradient(rand(3)) do y
...
-Any expressions involving `try`/`catch` statements is not supported.
-```julia
-function tryme(x)
-  try
-    2 * x
-  catch e
-    throw(e)
-  end
-end
+Exceptions containing try-catch statements can be differentiated if the catch block is not reached (no error are thrown).
```
oxinabox, line 85:
```suggestion
Code containting try-catch blocks can be differentiated as long as no exception is actually thrown.
```

src/compiler/reverse.jl
```diff
@@ -279,7 +274,7 @@ xaccum(ir, xs...) = push!(ir, xcall(Zygote, :accum, xs...))
 
 function passthrough_expr(ex::Expr)
     # Metadata we want to preserve
-    isexpr(ex, GlobalRef, :call, :isdefined, :inbounds, :meta, :loopinfo) && return true
+    isexpr(ex, GlobalRef, :call, :isdefined, :inbounds, :meta, :loopinfo, :enter, :leave, :catch) && return true
```
oxinabox, line 277: This should work on older version of julia that do not have `enter` and `leave`.
But the feature will not

src/compiler/reverse.jl
```diff
@@ -124,11 +124,6 @@ function instrument(ir::IR)
     ex = st.expr
     if isexpr(ex, :foreigncall, :isdefined)
       continue
-    elseif isexpr(ex, :enter, :leave)
-      error("""try/catch is not supported.
-            Refer to the Zygote documentation for fixes.
-            https://fluxml.ai/Zygote.jl/latest/limitations
-            """)
```
oxinabox, line 131: I believer we still need this on versions of julia that do not have `enter` and `leave`?
so maybe we keep this and add a `&& VERSION<v"1.10"` (is that the right bounds?)

src/compiler/reverse.jl
```diff
@@ -321,6 +321,18 @@ function adjoint(pr::Primal)
...
+    # This is corresponds to a catch blocks which technically
+    # has predecessors but they are not modelled in the IRTools CFG.
+    # We put an error message at the beginning of said block.
+    if has_leave && isempty(predecessors(b)) && b.id != 1
+        _, f_stmt = first(b)
+        li = pr.ir.lines[f_stmt.line]
+        li = LineNumberNode(Int(li.line), li.file)
+        pushfirst!(rb, stmt(xcall(Base, :error,
+                                  "Can't differentiate function execution in catch block at $(li).")))
+    end
```
oxinabox, line 334: I don't think throwing the error in the catch block is right.
I think we need to put it at the `throw` location.
Because I assume we also do not support `finally` blocks?

test/compiler.jl
```diff
@@ -245,3 +245,46 @@ end
     @test_nowarn g = back(1.)
     @test only(g) ∈ (1., 2.)
 end
+
+function f_try_catch(x,y)
```
oxinabox, line 249:
```suggestion
function throws_and_catches_if_x_negative(x,y)
```

oxinabox (commented):
Nice work.
See my comments

src/compiler/reverse.jl
Pangoraw: Zygote has compat for Julia 1.6 which already has `enter` and `leave` to represent exception handlers so it should be fine.

src/compiler/reverse.jl
Pangoraw: Try finally is implemented in term of `enter` and `leave` which can roughly be mapped to the following pseudo code:

```julia
try
  foo()
  origin = 1
catch
  origin = 2
end

if origin == 2
  rethrow()
end
```

I will improve the error message to include try/finally as another construct which is not supported when an error is thrown in the block.

src/compiler/reverse.jl
oxinabox: oh nice, i thought they were way newer

Pangoraw:
Tests are failing because an update to IRTools is needed (https://github.com/FluxML/IRTools.jl/pull/120).

test/compiler.jl
```diff
@@ -259,14 +259,49 @@ function f_try_catch(x,y)
...
+    x
+end
+
+if VERSION >= v"1.8"
+    # try/catch/else is invalid syntax prior to v1.8
+    eval(Meta.parse("""
+        function try_catch_else(cond, x)
+        end
+    """))
+end
```
oxinabox, line 282: this code seems unused and incomplete

oxinabox (approved):

ToucheSir:
@oxinabox is this still good to merge?

oxinabox:
I see no reason why it wouldn't be.

PR #1485, opened 2023-12-31
Swap PyCall to PythonCall

ToucheSir:
The former was giving us setup-related headaches on CI, and we can use any Python FFI for the one test which needs it.

As a bonus, stops loading CUDA.jl on GHA so we can get rid of errors on nightly. Hopefully with this we'll be back to green CI.

### PR Checklist

- [x] Tests are added


ToucheSir:
Looks like fixing these issues exposed actual test failures on nightly. We should address those and consider having better isolation for our testsets so that earlier errors don't prevent subsequent tests from running, but those can be done later. For know, I can confirm that the reverse CI failures are unrelated and this is good to go.

darsnack (approved):

ToucheSir:
We were mainly having problems with Buildkite, which uses a completely different system. PythonCall is so much easier to work with in general that it was worth the switchover regardless.

Issue #1484, opened 2023-12-27
Method ambiguities reported by Aqua


prbzrg:
In Julia v1.10, I used Aqua to detect method ambiguities in my package https://github.com/impICNF/ContinuousNormalizingFlows.jl/issues/356 and some of its report were related to this package:

```julia
Ambiguity #44
broadcasted(f, gs::Zygote.Grads, a::Union{AbstractArray{<:T}, T} where T<:Number) @ Zygote ~/.julia/packages/Zygote/WOy6z/src/compiler/interface.jl:381
broadcasted(::Base.Broadcast.AndAnd, a, b) @ Base.Broadcast broadcast.jl:196

Possible fix, define
  broadcasted(::Base.Broadcast.AndAnd, ::Zygote.Grads, ::Union{AbstractArray{<:T}, T} where T<:Number)

Ambiguity #45
broadcasted(f, gs::Zygote.Grads, gss::Union{Zygote.Grads, AbstractDict}...) @ Zygote ~/.julia/packages/Zygote/WOy6z/src/compiler/interface.jl:378
broadcasted(style::Base.Broadcast.BroadcastStyle, f::F, args...) where F @ Base.Broadcast broadcast.jl:1349

Possible fix, define
  broadcasted(::Base.Broadcast.BroadcastStyle, ::Zygote.Grads, ::Vararg{Union{Zygote.Grads, AbstractDict}})

Ambiguity #46
broadcasted(f, a::Union{AbstractArray{<:T}, T} where T<:Number, gs::Zygote.Grads) @ Zygote ~/.julia/packages/Zygote/WOy6z/src/compiler/interface.jl:380
broadcasted(style::Base.Broadcast.BroadcastStyle, f::F, args...) where F @ Base.Broadcast broadcast.jl:1349

Possible fix, define
  broadcasted(::Base.Broadcast.BroadcastStyle, ::F, ::Zygote.Grads) where F<:(Union{AbstractArray{<:T}, T} where T<:Number)

Ambiguity #47
broadcasted(f, gs::Zygote.Grads, a::Union{AbstractArray{<:T}, T} where T<:Number) @ Zygote ~/.julia/packages/Zygote/WOy6z/src/compiler/interface.jl:381
broadcasted(::Base.Broadcast.OrOr, a, b) @ Base.Broadcast broadcast.jl:203

Possible fix, define
[...]

PR #1486, opened 2023-12-31
Remove superfluous methods on `Grads`

ToucheSir:
As far as I can tell, these were never used, tested or documented. Moreover, they don't make sense semantically since `Grads` behaves like a Dict rather than an ordinal-indexed collection like an Array. Meanwhile, their continued existence is causing issues like https://github.com/FluxML/Zygote.jl/issues/1484.

### PR Checklist

- [ ] ~~Tests are added~~

### DIFF
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -375,40 +375,6 @@ function Base.copy(gs::Grads)
   merge!(gs_new, gs)
 end
 
-broadcasted(f, gs::Grads, gss::ADictOrGrads...) = map(f, gs, gss...)
-
-broadcasted(f, a::Numeric, gs::Grads) = map(x -> f(a, x), gs)
-broadcasted(f, gs::Grads, a::Numeric) = map(x -> f(x, a), gs)
-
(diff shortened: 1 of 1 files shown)


ToucheSir:
After blaming the failed test code, it appears the src file blame is messed up on GH and the [actual PR](https://github.com/FluxML/Zygote.jl/pull/902) which introduced this _does_ have tests + proper motivation for it. Now I'm not sure what we should do with this functionality. Deprecating it wouldn't get rid of the ambiguities, but it would set us up for removing it along with possibly more of the implicit grads machinery in the future (if we ever get around to that). Adding the suggested methods in #1484 would work, but feels rather unsatisfying and a little hacky. Maybe we could avoid ambiguities by restricting the list of broadcastable functions? 

I wish https://docs.julialang.org/en/v1/manual/interfaces/#man-interfaces-broadcasting would be a little clearer on how to implement the interface, especially since this is Base's own internal overloads conflicting with us here. WDYT @CarloLucibello @darsnack?

PR #1491, opened 2024-01-03
Preserve 0d arrays in `Zygote.accum`

lxvm:
`Zygote.accum` mirrored an [old Julia bug](https://github.com/JuliaLang/julia/issues/28866) that got fixed by Julia 1.6, which is the version supported by Zygote. This pr changes the behavior to match Julia's

### PR Checklist

- [x] Tests are added


mcabbott:
Do you have an example of user-facing code which triggers this?
```
julia> gradient(x -> first(x) + sum(x), fill(1.0))[1]
0-dimensional Array{Float64, 0}:
2.0
```

PR #1493, opened 2024-01-03
Remove `@adjoint`s for `sort` and `filter`

### LINKED ISSUE #1492: `sort(x; rev=true)` is not supported
Currently there is no support for `sort(::AbstractVector; rev::Bool)`. This is similar to what happened in #586, and would just need to add the keyword to the `@adjoint` definition. Alternatively, it is possible to just remove the definition and rely on the one that is defined by `ChainRules`, which automatically handles both options. I can open up a PR for this, if someone could let me know which solution is preferred.

```julia-repl
julia> using Zygote
julia> jacobian(x -> sort(x), [1,2,3])
([1 0 0; 0 1 0; 0 0 1],)
julia> jacobian(x -> sort(x; rev=true), [1,2,3])
ERROR: MethodError: no method matching adjoint(::Zygote.Context{false}, ::typeof(sort), ::Vector{Int64}; rev::Bool)
```

lkdvos:
This removes the definitions of `sort` and `filter`, as these are covered by ChainRules.
This also fixes #1492 , as the ChainRules `rrule` for `sort` includes all keywords.

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -251,20 +251,6 @@ end
 
 @adjoint iterate(r::UnitRange, i...) = iterate(r, i...), _ -> nothing
 
-@adjoint function sort(x::AbstractArray; by=identity)
-  p = sortperm(x, by=by)
-  return x[p], x̄ -> (x̄[invperm(p)],)
-end
-
-@adjoint function filter(f, x::AbstractVector)
-    t = map(f, x)
-    x[t], Δ -> begin
-        dx = _zero(x, eltype(Δ))
-        dx[t] .= Δ
-        (nothing, dx)
-    end
-end
-
 # Iterators
 
 @adjoint function enumerate(xs)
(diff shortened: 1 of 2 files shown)


devmotion:
> This also fixes #1492 , as the ChainRules `rrule` for `sort` includes all keywords.

I think it would be good to add a test for #1492.

darsnack:
The Molly.jl failure seems real, but the downstream test set is complex/foreign enough that I can't decipher where `sort` or `filter` are used. @jgreener64 can you confirm that this change is causing the failures and where?

jgreener64:
That test does fail sometimes due to stochasticity, I should improve it. The Molly CI run for the latest commit seems fine though, so this change doesn't seem to break anything.

lkdvos:
I think in that case this is ready for review + merge, unless I need to add something else?

darsnack (approved):

PR #1494, opened 2024-01-08
Add rule for `push!(::IdSet)`

ToucheSir:
This is required on 1.11 because the underlying implementation changed in https://github.com/JuliaLang/julia/commit/cb01a3b4521abbca1756f64537082561e815dc5e from using an IdDict (which Zygote can diff through) to a direct `ccall`.

### PR Checklist

- [x] Tests are added

### DIFF
--- a/src/lib/base.jl
+++ b/src/lib/base.jl
@@ -17,6 +17,19 @@ end
   end
 end
 
+# IdSet (needed for nested AD with implicit params)
+
+grad_mut(::IdSet) = IdSet()
+
+function _pullback(cx::AContext, ::typeof(push!), s::IdSet, @nospecialize(x))
+  res = push!(s, x)
+  function idset_push!_pullback(_)
+    Δ = pop!(grad_mut(cx, s), x, nothing)
+    (nothing, Δ, nothing)
+  end
+  return res, idset_push!_pullback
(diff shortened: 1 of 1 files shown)


ToucheSir:
NNlib reverse CI will be fixed by https://github.com/FluxML/NNlib.jl/pull/559. DynamicPPL tests still appear unrelated. Remaining failures (not errors) don't use implicit params, and may be related to ChainRules since they're testing those rules rather than Zygote ones.

ToucheSir:
Revisiting this because it squashes one error on nightly that is making all PRs fail. It does expose another error related to try-catch, but I imagine https://github.com/FluxML/Zygote.jl/pull/1474 would help with that.

CarloLucibello:
Add test?

CarloLucibello:
or no test is needed, it fixes an error on nightly?

ToucheSir:
The latter. The codepath was already tested, but it was broken ever since 1.11 became the nightly version: https://github.com/FluxML/Zygote.jl/blob/2517e67ecd233c21a89cd28e8d8e3c2d7214b5df/test/interface.jl#L265-L267

CarloLucibello (approved):

PR #1511, opened 2024-05-06
Adjust to upcoming julia#54341

Keno:
Corresponding IRTools changes are in https://github.com/FluxML/IRTools.jl/pull/125.


Keno:
I merged the IRTools change, but since approval is required here and I don't want to abuse my superuser powers, since I haven't worked on Flux in a long time, I'll let someone else take this one. @ToucheSir maybe?

CarloLucibello (approved):

Keno:
Could we get a new release of IRTools/Zygote that includes these changes for PkgEval purposes? I'd like to make sure that Zygote's downstreams are ok also before merging the base change.

PR #1524, opened 2024-09-16
Prefer numeric zero over ZeroTangent for numeric arrays

BioTurboNick:
If Zygote produces an `Array{Union{Nothing, <: Number}}`, when passed through `wrap_chainrules_input`, produces an `Array{Any}` because the `nothing` objects are mapped to `ZeroTangent` objects. However, ChainRules uses the `zero` of the type whenever possible so that the array maintains a concrete element type. This PR provides such a method.

Related to https://github.com/JuliaDiff/ChainRules.jl/issues/807

Note: It's possible that this case was only possible due to an upstream issue. Currently, I suspect that when Zygote provides an array with Tangents with NamedTuple components and ZeroTangents, the components of the NamedTuple are extracted and the ZeroTangent is mapped to `nothing` instead of performing the `zero(T)` optimization. But I wasn't able to track down the responsible code yet.

Also added tests to cover the related methods.

### PR Checklist

- [x] Tests are added
- [x] Documentation, if applicable **Doesn't seem to be documented in any way**

### DIFF
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -162,6 +162,7 @@ end
 # For arrays, whitelist the safe ones, but always look inside Any[]:
 @inline wrap_chainrules_input(dxs::AbstractArray{<:Number}) = dxs
 @inline wrap_chainrules_input(dxs::AbstractArray{<:AbstractArray{<:Number}}) = dxs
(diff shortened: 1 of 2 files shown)


BioTurboNick:
I believe the failing test is unrelated.

ToucheSir:
Do you mind force pushing (good change to rebase and squash too)? Just so we can double-check the GPU tests are unaffected since they failed to run properly last time.

ToucheSir (approved):
LGTM, thanks!

PR #1525, opened 2024-09-17
Support `enumerate` in conditional generators/comprehensions

BioTurboNick:
Trying to differentiate over a comprehension or generator over `enumerate` with a conditional errors because the tangent input will contain `nothing` instead of `Tuple{Nothing, T}`, and the mapped `last` to extract the `T` value fails.

This PR adds a `back` method for this case which maps `nothing` to `nothing`, and otherwise taking `last`.

### PR Checklist

- [x] Tests are added
- [x] Documentation, if applicable - **does not appear to be applicable**

### DIFF
--- a/src/lib/array.jl
+++ b/src/lib/array.jl
@@ -260,6 +260,7 @@ end
 @adjoint function enumerate(xs)
   back(::AbstractArray{Nothing}) = nothing
   back(dy::NamedTuple{(:itr,)}) = tuple(dy.itr)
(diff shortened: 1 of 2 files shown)


BioTurboNick:
Nightly failure appears unrelated

BioTurboNick:
@Keno - would you be able to review this and my other PR here? Saw you committed the last version pushed to the registry.

ToucheSir (approved):
LGTM, thanks!

PR #1548, opened 2025-01-04
Improve the limitations page

mcabbott:
This is far from perfect, but aims to:
* discourage anyone from using Buffer, as it doesn't really work well, and
* say more what it means to just avoid mutation. Maybe that could be longer still.


ToucheSir (approved):
The perfect is the enemy of the good, and given how much TLC the docs need this LGTM.

PR #1549, opened 2025-01-04
Drop support for Julia 1.6

mcabbott:
No very strong reason, could wait until someone has a bug which is simpler to solve without fixing 1.6

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.6' # Replace this with the minimum Julia version that your package supports.
+          - '1.10' # Replace this with the minimum Julia version that your package supports.
           - '1'   # automatically expands to the latest stable 1.x release of Julia
           - 'nightly'
         os:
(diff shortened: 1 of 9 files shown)


CarloLucibello:
Maybe the are some `VERSION` branches in the code that can be eliminated

CarloLucibello (approved):

.github/workflows/ci.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.6' # Replace this with the minimum Julia version that your package supports.
+          - '1.10' # Replace this with the minimum Julia version that your package supports.
```
devmotion, line 20: You could update setup-julia to v2 and use
```suggestion
          - 'min'
```

.github/workflows/ci.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.6' # Replace this with the minimum Julia version that your package supports.
```
ToucheSir, line 20: While we're modifying CI config, could we bump the versions of actions and switch over to the more useful https://github.com/julia-actions/cache?

pxl-th:
Should we merge this PR? I can then add a paralellized testing which would be much easier with Julia 1.9+.

PR #1551, opened 2025-01-16
Unthunk tangents (if any) before returning gradient

pxl-th:
Fixes: https://github.com/FluxML/Flux.jl/pull/2574#issuecomment-2595049570

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -57,7 +57,7 @@ Requires = "1.1"
 SpecialFunctions = "1.6, 2"
 Statistics = "1"
 Tracker = "0.2"
-ZygoteRules = "=0.2.5"
+ZygoteRules = "0.2.7"
 julia = "1.6"
 
 [extras]
--- a/Project.toml
+++ b/Project.toml
@@ -57,7 +57,7 @@ Requires = "1.1"
 SpecialFunctions = "1.6, 2"
 Statistics = "1"
 Tracker = "0.2"
-ZygoteRules = "0.2.5"
+ZygoteRules = "=0.2.5"
 julia = "1.6"
 
 [extras]
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -1,6 +1,6 @@
 # ToDo: Move some of this to ZygoteRules, or move unthunk_tangent for Tuple and NamedTuple from
 # Zygote rules here?
-function unthunk_tangent end
+# function unthunk_tangent end
 @inline unthunk_tangent(x::AbstractThunk) = wrap_chainrules_output(unthunk(x))
 @inline unthunk_tangent(x::NTuple{N,<:Number}) where N = x
 @inline unthunk_tangent(x::AbstractArray{<:Number,N}) where N = x
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -1,11 +1,10 @@
 # ToDo: Move some of this to ZygoteRules, or move unthunk_tangent for Tuple and NamedTuple from
 # Zygote rules here?
-# function unthunk_tangent end
-@inline unthunk_tangent(x::AbstractThunk) = wrap_chainrules_output(unthunk(x))
-@inline unthunk_tangent(x::NTuple{N,<:Number}) where N = x
-@inline unthunk_tangent(x::AbstractArray{<:Number,N}) where N = x
-@inline unthunk_tangent(x::AbstractArray) = map(unthunk_tangent, x)
-unthunk_tangent(d::IdDict) = IdDict([unthunk_tangent(k) => unthunk_tangent(v) for (k, v) in d])
+@inline ZygoteRules.unthunk_tangent(x::AbstractThunk) = wrap_chainrules_output(unthunk(x))
+@inline ZygoteRules.unthunk_tangent(x::NTuple{N,<:Number}) where N = x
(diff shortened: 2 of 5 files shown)


CarloLucibello:
Can you come up with a test?

pxl-th:
Hm... While it does fix the issue with `Flux.Dense` it still returns thunks for plain struct:
```julia
using Zygote

struct Dense
    w::Matrix{Float32}
end

(d::Dense)(x) = d.w * x

function main()
    layers = [Dense(rand(Float32, 3, 3))]
    # layers = [Dense(3, 3)]
    x = ones(Float32, 3)
    g = gradient(layers -> sum(layers[1](x)), layers)[1]
    @show g
end
main()
```
Chaing `@_adjoint_keepthunks` to `@adjoint` "fixes" it, but probably is not the right way, since it will unthunk every `getfield`.
So maybe needs unthunking at some other stage.
I'd be fine with unthunking at the very end, but maybe someone has better ideas...

CC @mcabbott @ToucheSir

ToucheSir:
Can you remind me what if anything we lose by unthunking at the top level (before `gradient` returns)? I think the problem is that `unthunk_tangent` was not taught about `Tuple`s and `NamedTuple`s in https://github.com/FluxML/Zygote.jl/pull/966/files#diff-e0bc7da8f1a33a59f5ecfa67257c04038f0b4915b3f74bdf39780818fd0010a2R3-R8.

mcabbott:
I think what's happening is that `m.W*x` returns `dm1 = (W=Thunk(), b=nothing)` while `.+m.b` returns `dm2 = (W=nothing, b=Thunk())`, and these are added by `accum(dm1, dm2)`. That's altering `accum` works here, but seems like the wrong place -- it will un-thunk things which aren't returned, and probably won't work for structs where only one field is used.

I agree that making `unthunk_tangent` recurse not just into arrays but also NamedTuples/Tuples seems like the right path.

It might be worth having `accum(dx::AbstractThunk, ::Nothing) = dx` etc, or otherwise making sure that this returns `dx` instead of making a new thunk? I can't prove this matters for performance but it seems worth trying a little to avoid making thunks of thunks of thunks.

pxl-th:
Hm... it was taught, `unthunk_tangent` is imported from ZygoteRules and there it has definition for NamedTuple:
https://github.com/FluxML/ZygoteRules.jl/blob/f9bf0e367fa259c5aa68f0e14ccbf2125d734bd6/src/adjoint.jl#L39

ToucheSir:
I don't think it is imported or used, per the comment in the code diff I linked.

pxl-th:
It is improted [here](https://github.com/FluxML/Zygote.jl/blob/4bfc545384effff9686beb92620e85ebee66233f/src/Zygote.jl#L8).
And if defining `unthunk_tangent` for `NamedTuple` in Zygote it complains that it iverwrites the one from ZygoteRules:

```
WARNING: Method definition unthunk_tangent(Any) in module ZygoteRules at /home/pxlth/.julia/packages/ZygoteRules/M4xmc/src/adjoint.jl:37 overwritten in module Zygote at /home/pxlth/.julia/dev/Zygote/src/compiler/chainrules.jl:4.
```

ToucheSir:
That's strange, because the code in #966 is definitely defining an `unthunk_tangent` function from scratch. I wonder if that's because it's in one of the submodules.

pxl-th:
I added tests, but I'm a bit out of ideas, so I made it unthunk before returning gradients.
Other approaches I've tried broke laziness...

pxl-th:
I've tested with Flux and all tests pass. Maybe this is fine for now?

ToucheSir:
The current approach LGTM, but perhaps it would make sense to have the new overloads for `unthunk_tangent` in ZygoteRules.jl? If nobody has any strong feelings about that though, happy to approve.

pxl-th:
Agree, moved them: https://github.com/FluxML/ZygoteRules.jl/pull/28

pxl-th:
> The current approach LGTM, but perhaps it would make sense to have the new overloads for unthunk_tangent in ZygoteRules.jl?

Based on https://github.com/FluxML/ZygoteRules.jl/pull/28#issuecomment-2601956822 and other comments, maybe it'd be better to move on with this PR and have a separate PR that will resolve this.

pxl-th:
> It also turns out that defining function unthunk_tangent end twice is a no-op? Can you remove that line

Already did: https://github.com/FluxML/Zygote.jl/pull/1551/files#diff-e0bc7da8f1a33a59f5ecfa67257c04038f0b4915b3f74bdf39780818fd0010a2R3

Project.toml
```diff
@@ -57,7 +57,7 @@ Requires = "1.1"
 SpecialFunctions = "1.6, 2"
 Statistics = "1"
 Tracker = "0.2"
-ZygoteRules = "0.2.5"
+ZygoteRules = "=0.2.5"
```
ToucheSir, line 60: This bound needs to be updated after https://github.com/FluxML/ZygoteRules.jl/pull/31.

ToucheSir (approved):
In hindsight, #966 should not have been merged with the `unthunk_tangent` changes because it was accidentally committing type piracy. But undoing that is going to take some work, and it's not clear if ZygoteRules.jl will still exist in its current form by the time someone gets around to doing said work because of how complex the phasing needs to be.

PR #1552, opened 2025-01-17
Improve catch block identification

### LINKED ISSUE #1546: try/catch error message regression
We used to have a nice error message for differentiating through try/catch block, saying "Can't differentiate function execution in catch block". On julia v1.11, we have a cryptic "unreachable" error  instead:

```julia
julia> function try_catch_finally(cond, x)
    try
        x = 2x
        cond && throw(DomainError())
    catch
        x = 2x
    finally
        x = 3x
    end
    x
end

julia> res, pull = pullback(try_catch_finally, true, 1.)
(12.0, Zygote.var"#78#79"{Zygote.Pullback{Tuple{typeof(try_catch_finally), Bool, Float64}, Any}}(∂(try_catch_finally)))

julia> pull(1)
ERROR: unreachable
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] try_catch_finally
   @ ./REPL[12]:8 [inlined]
 [3] (::Zygote.Pullback{Tuple{typeof(try_catch_finally), Bool, Float64}, Any})(Δ::Int64)
 [4] (::Zygote.var"#78#79"{Zygote.Pullback{Tuple{typeof(try_catch_finally), Bool, Float64}, Any}})(Δ::Int64)
   @ Zygote ~/.julia/dev/Zygote/src/compiler/interface.jl:91
 [5] top-level scope
   @ REPL[16]:1
```

Pangoraw:
Fixes #1546.

### DIFF
--- a/src/compiler/reverse.jl
+++ b/src/compiler/reverse.jl
@@ -299,6 +299,7 @@ end
 
 function adjoint(pr::Primal)
   ir, sigs = adjointcfg(pr)
+  catch_blocks = falses(length(blocks(pr.ir)))
   for b in reverse(blocks(pr.ir))
     rb = block(ir, b.id)
     grads = Dict()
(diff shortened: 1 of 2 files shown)


ToucheSir (approved):
Thanks!

PR #1560, opened 2025-03-06
Cleanup & remove some deps

### LINKED ISSUE #22: Seems to be #265ing
Maybe there is a catch that is not getting updated when a function is redefined?

# Code:
```
using Zygote;
g(x) = 2x;
@show g'(10);
g(x) = 3x;
@show g'(10);
```
Expected output is **2**, then **3**
Actual output is **2**, then **2** again.

## Demo:

```
julia> using Zygote;

julia> g(x) = 2x;

julia> @show g'(10);
(g')(10) = 2

julia> g(x) = 3x;

julia> @show g'(10);
(g')(10) = 2
```

pxl-th:
- Get rid of `Requires`.
- Remove GPUArrays dep as it is not used anywhere.
- Move Atom to extension.
- Minor cleanups after dropping 1.6 support.
- Remove `Zygote.refresh()` as it seems it is no longer required (closes #22):
```julia
julia> using Zygote

julia> g(x) = 2x;

julia> @show g'(10);
(g')(10) = 2.0

julia> g(x) = 3x;

julia> @show g'(10);
(g')(10) = 3.0
```


ToucheSir (approved):
https://github.com/FluxML/Zygote.jl/pull/1272 is why the GPUArrays compat bound was put in place. As long is the downgrade issue there is no longer a problem, I think we can remove it. Otherwise this LGTM.

pxl-th:
> #1272 is why the GPUArrays compat bound was put in place. As long is the downgrade issue there is no longer a problem, I think we can remove it. Otherwise this LGTM.

Shouldn't be a problem. All GPU backends now use `GPUArraysCore@0.2` and that itself will require newer GPUArrays package.

PR #1562, opened 2025-03-13
Add AccThunk to avoid prematurely unthunking thunks

pxl-th:
- Add `AccThunk` type (as suggested in https://github.com/FluxML/Zygote.jl/pull/1555) that holds thunks to-be-accumulated without unthunking prematurely.
- Add reduced MWE from https://github.com/FluxML/Flux.jl/issues/2585 as test, that leads to infinite compilation without `AccThunk`.

### PR Checklist

- [x] Tests are added

### DIFF
--- a/src/Zygote.jl
+++ b/src/Zygote.jl
@@ -13,6 +13,7 @@ using IRTools
 using MacroTools
 using MacroTools: @forward
 
+import ChainRules
 import Distributed: pmap, CachingPool, workers
 export Params, withgradient, gradient, withjacobian, jacobian, hessian, diaghessian, pullback, pushforward, @code_adjoint
 export rrule_via_ad
--- a/src/lib/lib.jl
+++ b/src/lib/lib.jl
@@ -1,5 +1,13 @@
-using Base: RefValue
-using Base: ismutabletype
+using Base: RefValue, ismutabletype, tail
+
+struct AccThunk{T} <: AbstractThunk
+    thunks::Vector{T}
+end
+
+_maybe_thunks(x) = x
+_maybe_thunks(x::AccThunk) = x.thunks
+
+ChainRules.unthunk(acc::AccThunk) = reduce(accum, unthunk.(acc.thunks))
 
 # Interfaces
 
@@ -7,9 +15,8 @@ accum() = nothing
 accum(x) = x
 
 accum(x, y) =
-  x === nothing ? y :
-  y === nothing ? x :
-  x + y
+  x === nothing ?
+    y : (y === nothing ? x : x + y)
 
 accum(x, y, zs...) = accum(accum(x, y), zs...)
 
@@ -35,9 +42,9 @@ accum(x::ChainRulesCore.Tangent, y::NamedTuple) = accum(wrap_chainrules_output(x
 accum(x::Nothing, y::AbstractThunk) = y
 accum(x::AbstractThunk, y::Nothing) = x
 
-accum(x, y::AbstractThunk) = accum(x, unthunk(y))
-accum(x::AbstractThunk, y) = accum(unthunk(x), y)
-accum(x::AbstractThunk, y::AbstractThunk) = accum(unthunk(x), unthunk(y))
+accum(x::AbstractThunk, y::AbstractThunk) = AccThunk(vcat(_maybe_thunks(x), _maybe_thunks(y)))
+accum(x::AbstractThunk, y) = AccThunk(vcat(_maybe_thunks(x), [y]))
+accum(x, y::AbstractThunk) = AccThunk(vcat([x], _maybe_thunks(y)))
 
 # Core functions
 @_adjoint_keepthunks deepcopy(x) = deepcopy(x), ȳ -> (ȳ,)
@@ -55,7 +62,7 @@ accum_param(::Context{false}, _, Δ) = Δ
   isbitstype(x) && return :(Δ)
   quote
     if haskey(cache(cx), x)
-      cache(cx)[x] = accum(cache(cx)[x],Δ)
+      cache(cx)[x] = accum(cache(cx)[x], Δ)
       return
     else
       return Δ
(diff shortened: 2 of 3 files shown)


pxl-th:
Test failures are unrelated.

pxl-th:
> Coming from [#1555 (comment)](https://github.com/FluxML/Zygote.jl/pull/1555#issuecomment-2611733494), I think we need at least one test that shows we aren't introducing new type instabilities when multiple paths try to accumulate different (un)thunked types.

IIUC you mean the case when `AccThunk` holds thunks of different type and if that leads to more type-instability -> worse performance?

ToucheSir:
Yes. I'd expect any overhead from type instability to be a non-issue for GaussianSplatting.jl, because each operation is much larger and any AD bookkeeping overhead is relatively small. But for problems more sensitive to AD overhead, it would be good to have some numbers.

pxl-th:
@ToucheSir do you have such a test in mind?

ToucheSir:
I was thinking something which involves smaller input sizes. e.g. this regression-esque example:
```julia
using Zygote, LinearAlgebra

x = rand(10, 100)
w = rand(2, 10)

f(w, x) = sum(w * x) + norm(w)

@benchmark gradient(f, $w, $x)
```

pxl-th:
On your example this PR is slower:

- This PR:
```julia
BenchmarkTools.Trial: 10000 samples with 6 evaluations per sample.
 Range (min … max):  5.714 μs …  1.974 ms  ┊ GC (min … max):  0.00% … 99.07%
 Time  (median):     8.062 μs              ┊ GC (median):     0.00%
 Time  (mean ± σ):   9.420 μs ± 39.797 μs  ┊ GC (mean ± σ):  13.65% ±  3.90%
     ▂██▇▃             ▃▂                                     
  ▁▂▅█████▇▅▄▃▃▂▂▃▂▄▅▆█████▇▆▄▃▃▂▂▂▁▁▁▁▁▁▁▂▂▂▂▃▄▃▃▄▃▃▃▂▂▂▂▂▁ ▃
  5.71 μs        Histogram: frequency by time        12.8 μs <
 Memory estimate: 13.47 KiB, allocs estimate: 47.
```
- `master`:
```julia
BenchmarkTools.Trial: 10000 samples with 8 evaluations per sample.
 Range (min … max):  3.047 μs … 603.206 μs  ┊ GC (min … max):  0.00% … 98.22%
 Time  (median):     3.696 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   4.684 μs ±  11.762 μs  ┊ GC (mean ± σ):  10.44% ±  5.14%
   ▃▆▇▇██▇▆▄▃▂▂▁▁                   ▁▁▁▁▁▁▁▁▁         ▁▁▁▂▁   ▂
  ▆█████████████████▇▆█▇▆▆▅▆▅▅▃▃▂▃▅▇█████████▇▇▆▆▆▇▇▇████████ █
  3.05 μs      Histogram: log(frequency) by time      9.14 μs <
 Memory estimate: 11.91 KiB, allocs estimate: 15.
```

But the times are so small it is not obvious if it's actually 2x slower on other examples.

Trying out reduced MWE that I've added as a test:

```julia
function main()
    W = ones(Float32, 10, 10)
    x = [ones(Float32, 10) for i in 1:512]
    gs = gradient(W) do W
        sum((W * xi)[1] for xi in x)
    end
    return
end
@benchmark main()
```

- This PR:
```julia
BenchmarkTools.Trial: 438 samples with 1 evaluation per sample.
 Range (min … max):   9.155 ms … 21.202 ms  ┊ GC (min … max):  0.00% … 11.38%
 Time  (median):     11.008 ms              ┊ GC (median):    14.24%
 Time  (mean ± σ):   11.387 ms ±  1.351 ms  ┊ GC (mean ± σ):  13.36% ±  3.84%
           ▅█▅▇▄                                               
  ▃▃▂▂▃▄▄▅▆██████▅▆▅▅▆▄▄▄▄▃▂▅▁▃▂▂▃▃▂▁▃▁▃▁▂▁▁▂▂▁▁▁▁▂▁▁▁▁▁▁▁▁▁▃ ▃
  9.15 ms         Histogram: frequency by time        17.9 ms <
 Memory estimate: 31.41 MiB, allocs estimate: 49627.
```
- `master`:
```julia
BenchmarkTools.Trial: 395 samples with 1 evaluation per sample.
 Range (min … max):  10.143 ms … 23.214 ms  ┊ GC (min … max):  0.00% … 29.74%
 Time  (median):     12.530 ms              ┊ GC (median):     0.00%
 Time  (mean ± σ):   12.623 ms ±  2.049 ms  ┊ GC (mean ± σ):  10.17% ± 10.01%
    ▄█  ▁                                                      
  ▃████▅█▅▆▆▅▅▄▃▃▃▃▃▃▃▅▅▇▆▅▆▆█▅▃▅▃▅▄▅▅▄▅▃▄▃▃▃▃▃▃▄▂▃▃▁▂▁▂▂▁▂▃▃ ▃
  10.1 ms         Histogram: frequency by time        17.6 ms <
 Memory estimate: 11.82 MiB, allocs estimate: 64480.
```

ToucheSir:
My main worry would be that this introduces allocations into workloads which were low to zero allocation before. The loop example is perhaps the best case scenario for this change, because it involves a large number of accumulated thunks which all have the same type.

My other worry is that this makes debugging Zygote code quite a bit harder. Currently Cthulhu.jl works well because a lot of functions can be statically inferred. But with this change, many functions which were type stable under differentiation will no longer be, so that tool will not be available. The change probably makes sense on balance, but losing access to such an important debugging tool (really the _only_ debugging tool, because the Debugger.jl does not work well enough) makes me uncomfortable.

Anyhow, I'll defer to @mcabbott or @CarloLucibello for a second opinion on this.

pxl-th:
> The change probably makes sense on balance, but losing access to such an important debugging tool (really the only debugging tool, because the Debugger.jl does not work well enough) makes me uncomfortable.

We can make this opt-in then and disable thunks by default.

PR #1564, opened 2025-03-27
chore: add unthunk_tangent rule for Ref

DhairyaLGandhi:
Sometimes internal types can "leak" out after the back pass when working with adjoints of types. See https://github.com/SciML/ModelingToolkit.jl/actions/runs/13965380427/job/39094467215#step:6:5266 for an example where an `InplaceableThunk` is returned. This is because of a missing dispatch for `unthunk_tangent(::RefValue)` which this PR adds. I don't have a very simple case yet where this fails for a test, but it is seen in the wild with v0.7.x in SciML

### PR Checklist

- [ ] Tests are added

### DIFF
--- a/src/compiler/chainrules.jl
+++ b/src/compiler/chainrules.jl
@@ -4,6 +4,7 @@
 @inline ZygoteRules.unthunk_tangent(x::NTuple{N,<:Number}) where N = x
 @inline ZygoteRules.unthunk_tangent(x::AbstractArray{<:Number,N}) where N = x
 @inline ZygoteRules.unthunk_tangent(x::AbstractArray) = map(unthunk_tangent, x)
+@inline ZygoteRules.unthunk_tangent(r::Base.RefValue) = r[] = unthunk_tangent(r[])
 ZygoteRules.unthunk_tangent(d::IdDict) = IdDict([unthunk_tangent(k) => unthunk_tangent(v) for (k, v) in d])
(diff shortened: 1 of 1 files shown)


src/compiler/chainrules.jl
```diff
@@ -4,6 +4,7 @@
 @inline ZygoteRules.unthunk_tangent(x::NTuple{N,<:Number}) where N = x
 @inline ZygoteRules.unthunk_tangent(x::AbstractArray{<:Number,N}) where N = x
 @inline ZygoteRules.unthunk_tangent(x::AbstractArray) = map(unthunk_tangent, x)
+@inline ZygoteRules.unthunk_tangent(r::Base.RefValue) = r[] = unthunk_tangent(r[])
```
CarloLucibello, line 7: Having second thoughts. Why is this mutating and not creating a new Ref?

DhairyaLGandhi:
Can this also get a release

CarloLucibello:
can you answer the comment above?

src/compiler/chainrules.jl
DhairyaLGandhi: This is the same ref used in the gradient accumulation, creating a new ref would break the provenance that the ref was holding on. In case the ref needs to be accessed downstream from where the unthunking happens. It keeps a single source of truth.

CarloLucibello:
tagged, thanks!

PR #1569, opened 2025-04-29
Import `typesof` from Base instead of InteractiveUtils

serenity4:
`typesof` was defined in Base, and simply used by InteractiveUtils. https://github.com/JuliaLang/julia/pull/57909 removed the `using` statement from InteractiveUtils, breaking the import.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Zygote"
 uuid = "e88e6eb3-aa80-5325-afca-941959d7151f"
-version = "0.7.6"
+version = "0.7.7"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
--- a/src/compiler/interface.jl
+++ b/src/compiler/interface.jl
@@ -1,5 +1,5 @@
 using InteractiveUtils
-using InteractiveUtils: typesof
+using Base: typesof
 using Core: Typeof
 import Base: copy!, IdSet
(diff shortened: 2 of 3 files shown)


mcabbott (approved):

mcabbott:
This change will be in Julia 1.13.  Tests still fail on CI, but locally it can be loaded & works: 
```julia
julia> using Zygote
Info Given Zygote was explicitly requested, output will be shown live 
WARNING: Constructor for type "Variable" was extended in `Zygote` without explicit qualification or import.
  NOTE: Assumed "Variable" refers to `IRTools.Variable`. This behavior is deprecated and may differ in future versions.
  NOTE: This behavior may have differed in Julia versions prior to 1.12.
  Hint: If you intended to create a new generic function of the same name, use `function Variable end`.
  Hint: To silence the warning, qualify `Variable` as `IRTools.Variable` in the method signature or explicitly `import IRTools: Variable`.

julia> Zygote.gradient(x -> sum(abs2, x), [1,2,3,4.])
([2.0, 4.0, 6.0, 8.0],)

julia> VERSION
v"1.13.0-DEV.461"
```

serenity4:
Yes, as far as I can retrace it `typesof` was always defined in Base, never in InteractiveUtils, so this should really be a no-op until 1.13.

mcabbott:
Thanks for fixing this.

I added a commit to silence the warning above, just because this was easier than making a fresh PR.

PR #1574, opened 2025-05-25
Accumulate mutable structs with `Ref`

DhairyaLGandhi:
In several downstream packages in SciML, ref https://github.com/SciML/ModelingToolkit.jl/pull/3585, https://github.com/SciML/SciMLSensitivity.jl/issues/1212 and others, accumulation into a mutable Tangent throws errors of the form.

```julia
julia> gs = gradient(new_p) do new_p
           new_params = SciMLStructures.replace(SciMLStructures.Tunable(), mtkparams, new_p)
           new_prob = remake(prob, p = new_params)
           new_sol = solve(new_prob, Tsit5())
           sum(new_sol)
       end
ERROR: MethodError: no method matching +(::Base.RefValue{…}, ::@NamedTuple{…})
The function `+` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  +(::Any, ::Any, ::Any, ::Any...)
   @ Base operators.jl:596
  +(::MutableArithmetics.Zero, ::Any)
   @ MutableArithmetics ~/.julia/packages/MutableArithmetics/tNSBd/src/rewrite.jl:64
  +(::Any, ::MutableArithmetics.Zero)
   @ MutableArithmetics ~/.julia/packages/MutableArithmetics/tNSBd/src/rewrite.jl:65
  ...

Stacktrace:
  [1] accum(x::Base.RefValue{…}, y::@NamedTuple{…})
    @ Zygote ~/.julia/packages/Zygote/wfLOG/src/lib/lib.jl:9
  [2] accum(::Base.RefValue{…}, ::@NamedTuple{…}, ::Nothing, ::Vararg{…})
    @ Zygote ~/.julia/packages/Zygote/wfLOG/src/lib/lib.jl:14
  [3] maybe_eager_initialize_problem
    @ ~/.julia/packages/SciMLBase/z4OYD/src/remake.jl:1211 [inlined]
  [4] (::Zygote.Pullback{Tuple{…}, Any})(Δ::Tuple{Thunk{…}, @NamedTuple{…}})
    @ Zygote ~/.julia/packages/Zygote/wfLOG/src/compiler/interface2.jl:0
  [5] #remake#753
    @ ~/.julia/packages/SciMLBase/z4OYD/src/remake.jl:266 [inlined]
  [6] (::Zygote.Pullback{Tuple{…}, Any})(Δ::Base.RefValue{Any})
    @ Zygote ~/.julia/packages/Zygote/wfLOG/src/compiler/interface2.jl:0
[...]

PR #1576, opened 2025-06-06
chore: handle `accum` when no grads are present

DhairyaLGandhi:
Handles the case when there are no gradients to accumulate, as a follow on to #1574 

cc @ChrisRackauckas 

### PR Checklist

- [ ] Tests are added


ChrisRackauckas:
@oxinabox

oxinabox (approved):
This is obviously correct

PR #1583, opened 2025-08-09
Add Downgrade.yml and Fix Minimum Versions

ChrisRackauckas:
I think a lot of downgrade CI in SciML is pointing to minimum version issues in Zygote, so I'm trying to now track that down here in isolation.


ChrisRackauckas:
So it seems this did most of it, with the lower bound of MacroTools.jl being very crucial. But CUDA.jl seems to have some issues of its own to track down.

PR #1605, opened 2026-03-15
Fix Zygote for Julia 1.12

murrellb:
## Summary

  This attempts to fix a set of Zygote issues that arose in Julia 1.12. Caveat: I have no idea what I'm doing.

  The main breakages were:
  - regressions around Julia 1.12 global access lowering
  - failures in generated pullback code for keyword/default-wrapper methods
  - forward-mode failures on new 1.12 internal IR intrinsics

  These regressions showed up in small MWEs and in downstream code paths such as ChainStorm on Julia 1.12, including GPU-backed training-style workloads.

  ## What Was Broken

  ### 1. Global access on Julia 1.12

  Julia 1.12 changed the lowered/internal representation of module global access. Zygote was still assuming the older form, which caused incorrect differentiation results for code that reads globals or module bindings.

  ### 2. Keyword/default-wrapper pullback generation

  On Julia 1.12, some synthesized keyword/default-wrapper methods could lead Zygote to emit malformed lowered `:isdefined` expressions in generated pullback `CodeInfo`.

  That invalid IR could then fail during code generation, including the crash path hit by a reduced ChainStorm reproducer.

  ### 3. Forward-mode interaction with new 1.12 internals

  Julia 1.12 introduced internal IR operations that should be treated as non-differentiable metadata from Zygote's perspective. Without that handling, forward-mode paths could fail or hard-error.

  ## Fixes

  ### Reverse mode / globals

  - instrument `Core.getglobal` in reverse-mode compilation
  - add an adjoint for `getglobal`

  This restores correct behavior for gradients that depend on module/global access.

  ### Pullback generation / kw wrappers

  - sanitize invalid lowered `Expr(:isdefined, ...)` nodes in generated pullback `CodeInfo` before returning it

  This fixes the Julia 1.12 kw/default-wrapper crash path without changing user-facing semantics.

  ### Forward mode / 1.12 internals

  - treat new Julia 1.12 metadata-style internal operations as non-differentiable
  - guard those additions so Julia 1.11 remains compatible
  - stop hard-failing on 1.12 `foreigncall` IR in the forward compiler path

  ## Tests

  Adds/updates tests for:
  - Julia 1.12 global access regressions
  - Julia 1.12 forward-mode regressions
  - a minimal keyword-wrapper/logging reproducer for the pullback-generation failure
  - Flux-facing MWEs (`Flux.normalise` / `Flux.LayerNorm`) that exercise the same fixed paths

  ## Validation

  Verified on Julia 1.12:
  - targeted `compiler`, `forward`, and `flux` test items pass
  - reduced ChainStorm GPU backward paths now run on GPU 0
  - a fixed-batch ChainStorm benchmark runs successfully on both Julia 1.11 and 1.12

  The remaining known Julia 1.12 issue is an inference regression for `usesmod`: the gradient value is correct, but inference is wider than on Julia 1.11.


devmotion:
>  Caveat: I have no idea what I'm doing.

Does this mean the changes are AI generated?

murrellb:
> > Caveat: I have no idea what I'm doing.
> 
> Does this mean the changes are AI generated?

The LLMs are doing most of the coding here, yes, but that is fairly normal now? I add that caveat to *additionally* flag that I have no particular Zygote expertise, except for a few large DL projects that fail on Julia 1.12 with Zygote master.

In any case, I found a few more tests marked as broken on 1.12 and there is a bit more work to do, so I'm going to close this for now so I don't trigger this rather large CI.

devmotion:
> The LLMs are doing most of the coding here, yes, but that is fairly normal now?

I'd like to know whether a PR is implemented without AI, with AI assistance or written by AI (or something in-between) because it's a primer for whether I should invest time in reviewing a PR or not and to what level I should review it and provide suggestions.

murrellb:
That's fair. This was my motivation for including the caveat, but more about me than the LLMs. To be clear though: I don't think anyone should be reviewing this yet (I marked it as draft, etc).

Issue #1664, opened 2026-07-27
Zygote throws error during precompilation in Julia nightly


ronisbr:
My tests are failing when using Zygote in Julia nightly because it throws the following error in precompilation:

```
Internal error: during type inference of
_pullback(Zygote.Context{false}, typeof(Zygote.pow), Int64, Int64)
```

The full log is available here:

https://github.com/JuliaSpace/ReferenceFrameRotations.jl/actions/runs/30275856779

Issue #147, opened 2019-04-12
#mutate: push!() fails to differentiate


NiklasGustafsson:
The following code raises an exception when using the #mutate branch:

```
using Pkg; Pkg.activate(".");
pkg"add Zygote#mutate"

using Zygote

function f(w::Array{Float64,2}, x::Array{Float64,2})

    result = []

    for i = 1:size(x,2)
        h = w[:,i] .+ x[:,i]
        push!(result, h)
    end

    cat = hcat(result...)
    return sum(cat)
end

s = rand(5,100)
x = rand(5,100)

v = f(s,x)

l = Zygote.gradient((w)-> f(w, x), s)
```

MikeInnes:
Now working on the latest commit.

overshiki:
Hi
I just tested the above piece of code in Zygote v0.6.40, it does not work. Is differentiation of `push!` currently a support feature?

ToucheSir:
It is not. Note the use of an experimental branch above. If you must have a mutable vector, `Buffer` or hiding the mutation behind a `rrule` are the only options.

overshiki:
Hi @ToucheSir 
I see. 
Cool, thanks!

PR #1665, opened 2026-08-03
Support Julia 1.13 and 1.14

CarloLucibello:
Makes Zygote build and test cleanly on **Julia 1.13** (rc1) and **1.14** (nightly).

### Julia 1.14 — precompile crash (#1664)
The root cause is in IRTools, not Zygote: `update!` sets `CodeInfo.edges` to a `Vector{Core.MethodInstance}`, which Julia 1.14's `compute_edges!` now type-asserts against `Union{Nothing, SimpleVector, Vector{Any}}` (the edges format is heterogeneous, so the container must be `Vector{Any}`). Filed and to be fixed upstream in **FluxML/IRTools.jl#147**; this PR just raises the compat floor to `IRTools = "0.4.20"` rather than carrying a workaround.

> ⚠️ Depends on the IRTools release containing the #147 fix — CI on 1.14/nightly will stay red until `IRTools 0.4.20` is registered.

### Julia 1.13 — `@spawn` differentiation
The `ccall`/`:foreigncall` target representation changed from `QuoteNode(:name)` to the tuple expression `(:name,)`, so `passthrough_expr` stopped recognising `jl_set_task_threadpoolid` and the `threads` tests failed with *"Can't differentiate foreigncall"*. Added `foreigncall_name` to accept both forms.

### Test guards (genuine Julia bugs, not Zygote)
- **utils.jl** — `ForwardDiff.derivative(x -> ForwardDiff.derivative(y->y^4, x), 7)` (pure ForwardDiff, no Zygote) miscompiles to an **illegal instruction** on Julia ≥1.13; works on 1.12, and works with a `Float64` seed. A SIGILL is uncatchable and takes down the test worker, so the integer case is `@test_skip`ped on ≥1.13 and a `Float64` variant added to keep coverage. Worth an upstream JuliaLang report.
- **gradcheck_p2.jl** — the near-degenerate `sqrt(::Symmetric)` finite-difference check (already `broken` on x86_64 since 1.12) also fails on aarch64 from 1.13; extended the guard.

### Verification
- Full suite green on **1.13.0-rc1** (only expected `broken`).
- #1664 fix confirmed on **1.14.0-DEV** in an isolated env (dev Zygote + IRTools with the #147 fix) across the no-edge, `materialize`, and rrule-edge paths.

Note: version guards use `v"1.13-"` so they also match the current `1.13.0-rc1` prerelease (`v"1.13.0-rc1" >= v"1.13"` is `false`).


CarloLucibello:
nightly CI failes to update the registry (needed for IRTools 0.4.20)

